PR #102, opened 2015-01-01
Test the PHP action API with a bad hostname

earldouglas:
This revealed a problem in the forwarding a 'localhost' host header
out to external Web services, which makes the PHP action API rather
unhappy.


lib/tools.js
```diff
@@ -0,0 +1,40 @@
...
+    var filtered = {};
+    Object.keys(object).forEach(function (key) {
+        if (filter(key)) {
+            filtered[key] = object[key];
+        }
+    });
+    return filtered;
+}
+
+function find(xs, predicate) {
```
d00rman: AFAIK, Array.find() is part of Harmony, and is provided by [es6-shim](https://github.com/paulmillr/es6-shim#safe-shims), with the difference being it returns the value, not the index in the array, so maybe we should use that one?

earldouglas: Thanks for pointing that out - I have much to learn about the various JS std libs.  I'll switch to that and toss out this ad-hoc implementation.

earldouglas:
Rebased and switched to es6-shim's `Array.find` implementation.

earldouglas:
Reased on master.

lib/filters/global/actions.js
```diff
@@ -17,7 +17,12 @@ module.exports = {
                     body.format = 'json';
                     return restbase[req.method](req)
                     .then(function(res) {
-                        if (res.status !== 200) {
+                        if (
+                            res.status !== 200
```
gwicke: This would be simplified if we raised exceptions / rejected the promise like in preq.

earldouglas: What about non-rejected statuses (e.g. 201, 30x)?

gwicke: Is there a legitimate case where the body is empty, but the return status is 200? I'm actually not sure about res.body.query either.

For example, [this query](http://en.wikipedia.org/w/api.php?action=query&prop=revisions&meta=siteinfo&titles=Somepagethatdoesn%27texist%20Page&rvprop=user%7Ccomment&continue=) still sets `query` and `pages` despite not matching any page.

earldouglas:
```
$ curl -s 'http://en.wikipedia.org/w/api.php?format=json&action=query' | python -mjson.tool
{
    "warnings": {
        "query": {
            "*": "Formatting of continuation data will be changing soon. To continue using the current formatting, use the 'rawcontinue' parameter. To begin using the new format, pass an empty string for 'continue' in the initial query."
        }
    }
}
```

gwicke: Well, that's the reply to an empty query, which we (afaik) never send.

earldouglas: Doesn't look likely.  We currently call the PHP action API only from pagecontent.js, and it always supplies enough of a body to prevent this case.  Shall I drop the `res.body.query{,.pages}` checks?

gwicke: The action API wrapper is really a very embryonic internal thing right now. Once we start thinking about how we could integrate the PHP API into a more general REST API we'd probably want to give it a more systematic treatment.

In the meantime, internal uses that don't actually supply a proper query should probably fail loud & early, rather than just seeming to succeed without results. It did so far (by accessing something that wasn't defined), but we could make it more explicit with a check & an explicit throw.

earldouglas: Cool, I'll make these cases result in errors.

earldouglas: Does a 400 make sense for these cases?

earldouglas: Hmm, maybe a 500 is better.  We don't necessarily know that the request was poorly formatted -- all we know is that the PHP action API didn't give us back what we expect.

gwicke: Yeah. Anything that throws is good ;)

lib/rbUtil.js
```diff
@@ -272,5 +272,24 @@ util.inherits(HTTPError, Error);
 
 rbUtil.HTTPError = HTTPError;
 
+rbUtil.httpErrors = {
+    offline: new HTTPError({
```
gwicke: Minor: It might be more consistent to make both httpErrors functions.

PR #103, opened 2015-01-02
Add tests for offline mode Web request blocking

earldouglas:
This revealed a couple of issues with the way we
were enabling offline mode in our tests, the way
we were handling offline requests to be blocked,
and the way we respond to blocked requests.  All
have been fixed and are now under test.

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -35,7 +35,7 @@ function startRestbase(offline) {
     return restbase({
         logging: {
             name: 'restbase-tests',
-            level: 'warn',
+            level: offline ? 'fatal' : 'warn', // hide warnings during offline tests
         },
         offline: offline
     }).then(function(server){
--- a/test/main.js
+++ b/test/main.js
@@ -36,8 +36,8 @@ function startRestbase(offline) {
         logging: {
             name: 'restbase-tests',
             level: 'warn',
-            offline: offline
-        }
+        },
+        offline: offline
     }).then(function(server){
         stopRestbase =
             function () {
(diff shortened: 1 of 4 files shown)


test/main.js
```diff
@@ -63,7 +63,66 @@ describe('Offline mode feature tests after a server restart', function() {
...
+            return preq.get({
+                uri: config.bucketURL + '/Idempotent/html/76f22880-362c-11e4-9234-0123456789ab'
+            })
+            .then(function(res) {
+                assert.deepEqual(res.status, 200);
+            });
+        });
+        it('should not allow latest content retrieval from storage', function() {
+            this.timeout(20000);
+            return assert.fails(
```
hardikj: why are we using assert.fails here, why not just 
return preq.get(...).then(function(e){assert.deepEqual(e.status, 500)})

hardikj:
You can turn these off if you want https://travis-ci.org/wikimedia/restbase/jobs/45710252#L328?  (if they are not outputing some useful info, they make test look uglier ;) )

test/main.js
earldouglas: Because `.then(function(e)` would not run -- it would be shortcutted by the throwing of `e`, so we need to wrap the whole thing in a `try/catch`, which is what `assert.fails` helps us do.

hardikj: Ah, thats really useful, thanks for the explanation.

earldouglas:
@hardikj How do I change the logging level (i.e. to hide `info/request` messages during testing)?

hardikj:
one way would be by changing the logging level to 'fatal' when running tests in offline mode

earldouglas:
Where in the code can I do this?  Is it part of the `opts.log` object?

hardikj:
here https://github.com/wikimedia/restbase/blob/master/test/main.js#L38 :)

earldouglas:
Ah, of course.  Thanks!

PR #106, opened 2015-01-06
Lock-down the transform URLs to only the supported formats

earldouglas:


earldouglas:
Roger, stand by for update.

earldouglas:
Note: I deleted some deprecated comments from Coveralls.

PR #108, opened 2015-01-06
Drop unnecessary console output from testing

earldouglas:
These console outputs were useful at one point, during the development of certain tests, but now are just noising up the test output.

Ideally, the passing/failing of tests should tell us what we need to know about our test results, so nothing from the console should be needed, except for ad-hoc debugging of tests.

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -31,11 +31,10 @@ var stopRestbase = function () {};
 function startRestbase(offline) {
     stopRestbase();
     offline = offline || false;
-    console.log('starting restbase in ' + (offline ? 'OFFLINE' : 'ONLINE') + ' mode');
     return restbase({
         logging: {
             name: 'restbase-tests',
-            level: offline ? 'fatal' : 'warn', // hide warnings during offline tests
+            level: 'fatal'
         },
         offline: offline
     }).then(function(server){
--- a/test/main.js
+++ b/test/main.js
@@ -28,13 +28,32 @@ var config = {
 
 var stopRestbase = function () {};
 
+var log = {};
+function logger() {
+    function collect(context) {
+        return function() {
+            var ctx = JSON.stringify(context);
+            if (/^warn/.test(arguments[0]) ||
+                /^error/.test(arguments[0]) ||
+                /^fatal/.test(arguments[0])
+            ) {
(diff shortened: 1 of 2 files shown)


test/main.js
```diff
@@ -31,11 +31,10 @@ var stopRestbase = function () {};
 function startRestbase(offline) {
     stopRestbase();
     offline = offline || false;
-    console.log('starting restbase in ' + (offline ? 'OFFLINE' : 'ONLINE') + ' mode');
     return restbase({
         logging: {
             name: 'restbase-tests',
-            level: offline ? 'fatal' : 'warn', // hide warnings during offline tests
+            level: 'fatal'
```
gwicke: Should we default to 'warn' level?

earldouglas:
Chatted with @gwicke , and we realized a better solution is to use a monoidal log, rather than the console, that we can check at the end of execution.

earldouglas:
I added a super crappy test logger that collects all log entries, which we can check at the end.  It would be good to improve this by adding per-test checks that look at the log and verify that they introduced the expected entries, if any.

earldouglas:
Oof, this hurts coverage,since it no longer exercies the "real" logger. :\

test/main.js
```diff
@@ -28,14 +28,32 @@ var config = {
 
 var stopRestbase = function () {};
 
+var log = {};
+function logger() {
+    function collect(context) {
+        return function() {
+            var ctx = JSON.stringify(context);
+            if (/^warn/.test(arguments[0]) ||
```
gwicke: You could simplify this to `/^(?:warn|error|fatal/.test()`

test/main.js
```diff
@@ -128,3 +146,27 @@ describe('Offline mode feature tests after a server restart', function() {
...
 });
+
+describe('Log', function () {
+
+    var assert = require('./utils/assert.js');
+
+    this.timeout(20000);
+
+
+    it('should only have PUT, GET, and POST request entries', function() {
```
gwicke: Should we check this after each test? Having a global check for log entries sounds like it would often fail / need to be updated.

earldouglas:
Closing this in favor of an alternative in progress.

PR #110, opened 2015-01-07
Add storage-backed transform api

earldouglas:
I'm about 75% certain this is on the right track. :]

This adds three endpoints from [https://phabricator.wikimedia.org/T75955](https://phabricator.wikimedia.org/T75955):

* `POST /{domain}/v1/transform/html/to/html/{title}/{revision}`
* `POST /{domain}/v1/transform/html/to/wikitext/{title}/{revision}`
* `POST /{domain}/v1/transform/wikitext/to/html/{title}/{revision}`

### DIFF
--- a/lib/filters/global/parsoid.js
+++ b/lib/filters/global/parsoid.js
@@ -107,7 +107,7 @@ function transformRevision(parsoidHost, from, to) {
         var key    = req.params.key;
         var rev    = req.params.revision;
 
-        var original = {
+        var fromStorage = {
             revid: rev
         };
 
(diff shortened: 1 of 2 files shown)


lib/filters/global/parsoid.js
```diff
@@ -100,6 +100,61 @@ function transform(parsoidHost, from, to) {
...
+                        },
+                        body: res.body.body
+                    };
+                } else {
+                    return {
+                        status: 404
+                    };
+                }
+            })
+            .then(function (res) {
```
gwicke: The method above looks all synchronous; is this .then() really needed?

lib/filters/global/parsoid.js
```diff
@@ -100,6 +100,61 @@ function transform(parsoidHost, from, to) {
...
+                if (res.status === 200) {
+                    original[format] = {
```
gwicke: It might be slightly nicer / easier to follow to return the result rather than using a side-effect to fill out original. The assembly would then be based on the array returned from Promise.all() below.

gwicke:
Looks good to me overall. Two small comments inline.

lib/filters/global/parsoid.js
earldouglas: Oh right, I must need more coffee.

earldouglas:
Stand by for some updates.

PR #113, opened 2015-01-09
Enable log tracking during tests, add simple test cases

earldouglas:
I'm doing a final pass of phabriactor/T75955, which calls for a few missing test cases.  Part of these test cases calls for ensuring that we go to Parsoid in some cases, and ensuring that we do not go to Parsoid in other cases.

This reveals a more general pattern of needing to inspect the log, and in this case inspect the requests/responses that occurred, so I added a way to trace all requests/responses, and a way to inspect arbitrary slices of the log during testing.

This PR includes the infrastructure for this kind of testing, and some very basic tests that will be expanded further to complete the testing for phabriactor/T75955.

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -225,7 +225,7 @@ rbUtil.makeLogger = function(conf) {
     function bindAndChild (logger) {
         var log = logger.log.bind(logger);
         log.child = function(args) {
-            return bindAndChild(new Logger(null, logger.logger.child(args)));
+            return bindAndChild(new Logger(conf, logger.logger.child(args)));
         };
         return log;
     }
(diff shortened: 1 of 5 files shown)


lib/rbUtil.js
```diff
@@ -225,7 +225,7 @@ rbUtil.makeLogger = function(conf) {
     function bindAndChild (logger) {
         var log = logger.log.bind(logger);
         log.child = function(args) {
-            return bindAndChild(new Logger(null, logger.logger.child(args)));
+            return bindAndChild(new Logger(conf, logger.logger.child(args)));
```
earldouglas: This ensures that child loggers inherit the same log level as the parent, rather than defaulting to `warn`.

PR #117, opened 2015-01-12
Parrot interesting ( >= warn ) log entries to stdout during testing

earldouglas:
This way we still see potentially unexpected `warn/error/fatal` log entries on stdout during testing, so that they don't go unnoticed with the new `logStream` feature.

### DIFF
--- a/test/utils/logStream.js
+++ b/test/utils/logStream.js
@@ -1,10 +1,29 @@
 'use strict';
 
+var bunyan = require('bunyan');
+
 function logStream() {
 
   var log = [];
+  var parrot = bunyan.createLogger({
+    name: 'test-logger',
+    level: 'warn'
+  });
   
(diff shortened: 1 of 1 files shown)


test/utils/logStream.js
```diff
@@ -1,10 +1,29 @@
...
+        var entry = JSON.parse(chunk);
+        var levelMatch = /^(\w+)/.exec(entry.levelPath);
+        if (levelMatch) {
+            var level = levelMatch[1];
+            if (parrot[level]) {
+                parrot[level](entry);
+            }
+        }
+    } catch (e) {
+        console.error('something went wrong trying to parrot a log entry', e);
```
d00rman: @earldouglas maybe it would be interesting to echo the offending chunk as well?

earldouglas: :+1:

PR #97, opened 2014-12-29
WIP / RFC: Modules

gwicke:
This is a WIP branch to work out a design for https://phabricator.wikimedia.org/T76883. Do not merge just yet.

PR #118, opened 2015-01-13
Modules (T76883)

d00rman:
This PR has been opened to track our progress on [T76883](https://phabricator.wikimedia.org/T76883). Essentially the same as PR #97 with the difference that we operate on a branch of the main repo.


d00rman:
Note that for now the tests do not pass, which is to be expected. As we slowly port the old modules/buckets to the new interface, we should see an increase in test passes. For now, only 7 tests pass ;)

earldouglas:
What are the thoughts on how modules should intercommunicate?

The [internal request & response objects](https://github.com/wikimedia/restbase/blob/modules/doc/Implementation.md#internal-request--response-objects) section sounds like we construct a request and delegate to the higher-level router to figure out which implementation to proceed with.  This works well enough, but perhaps we could take better advantage of individual module specs.

*Example use case: latest format retrieval from the page content bucket*

The developer of the page content bucket implementation knows for certain that they depend on both the Parsoid module and any implementation of the storage module.  How can we best manage these dependencies?  For the former case, we could pull the Parsoid module in directly as an npm dependency, but I wonder if this goes against the spirit of the architecture, which is to manage all interaction via HTTP-like interfaces.  For the secondary case, we can't quite pull the module in directly because we don't know which implementation (cassandra, SQLite, etc.) will be used -- we only know the interface that we need.

d00rman:
> The developer of the page content bucket implementation knows for certain that they depend on both the Parsoid module and any implementation of the storage module. How can we best manage these dependencies?

@earldouglas These are checked during the spec reading / routing tree creation process. Here's how it works in this shiny, brand-new mechanism. Each action is defined by its interface (spec) and an entity implementing it (a module). They are linked by `operationId`s: the interface specifies which `operationId` needs to be looked up when a request lands on a particular URI, while the module exports the list of `operationId`s it implements. The tree-construction mechanism ensures there is a loaded implementation for each specified `operationId`.

With regards to direct dependencies, the idea here is to specify `resource` stanzas in the spec, which would check a URI is available, thus confirming the dependency's presence.

Currently this loading process is static and is done only on start-up. But, the idea is to (eventually) provide a dynamic version where the spec(s) is(are) reloaded on demand, and npm-type modules pulled in automatically if they are missing.

PR #119, opened 2015-01-13
Overhaul spec testing

earldouglas:
This is prep work to convert a bunch of test code into spec tests (using `x-ample`s).  It includes a few changes:

**Source the spec from a file local to the branch**

This makes it easier to keep the spec up to date with the code, but we'll need to remember to push it to the web prior to v1 release, once the v1 API is finalized

**Drop prerequisite requests from specs**

To keep things clean, we'll have all spec tests share a single set of prerequisite requests.  This vastly shortens the spec file, and maintains James' sanity when juggling collisions between tests.

**Cleanup the spec test creation**

The prerequisites are separated into their own Mocha `it()` block, each spec test has its own `it()` block, and there is a final `it()` block that verifies the total number of specs that were run.

### DIFF
--- a/lib/filters/bucket/pagecontent.js
+++ b/lib/filters/bucket/pagecontent.js
@@ -165,8 +165,9 @@ PCBucket.prototype.getLatest = function(restbase, req, options) {
 PCBucket.prototype.getLatestFormat = function(restbase, req) {
     var rp = req.params;
     var origURI = req.uri;
+    var apiUri = '/v1/' + rp.domain + '/_svc/action/query';
     return restbase.post({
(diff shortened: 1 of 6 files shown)


earldouglas:
Note: still TODO, as part of a different PR, is investigating Swagger's already built-in support for examples.

gwicke:
@earldouglas, there is [an examples object for responses](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#exampleObject). There is also support for supplying example instances for schemas. I have not however seen anything in the spec that would let us supply a pair of example request/response pairs outside of the markdown description.

PR #121, opened 2015-01-14
Implement request templating for service requests

hardikj:


hardikj:
one way to test this it to include the preq module and test the handler we created here https://github.com/wikimedia/restbase/blob/modules/test/features/router/buildTree.js#L53

lib/router.js
```diff
@@ -98,8 +101,10 @@ Router.prototype._handleRESTBasePathSpec = function(node, subspec, symbols, uri)
             // set up a handler
             // XXX: share?
             node.value.methods[methodName] = function (restbase, req) {
-                // XXX: expand the request with req information!
-                restbase.request(mxrb.service);
+                function eval_req(req) {
+                    return request_parser(mxrb.service, req);
+                }
+                restbase.request(eval_req(req));
```
gwicke: `return restbase.request(eval_req(req));`

lib/router.js
```diff
@@ -98,8 +101,10 @@ Router.prototype._handleRESTBasePathSpec = function(node, subspec, symbols, uri)
             // set up a handler
             // XXX: share?
             node.value.methods[methodName] = function (restbase, req) {
-                // XXX: expand the request with req information!
-                restbase.request(mxrb.service);
+                function eval_req(req) {
```
gwicke: It would be nice if you could move request_parser out, and return an eval closure that takes the request from there. In the handler, you can then just call it.

PR #123, opened 2015-01-14
Start breaking things by hooking up the new router in restbase

gwicke:


gwicke:
@wikimedia/services, lets see if this pings you all.

Next up: table storage.

d00rman:
Apart from [require()-ing the router](https://github.com/wikimedia/restbase/pull/123/files#diff-c945a46d13b34fcaff544d966cffcabaR18) (which should be a relative path), everything else seems like an awesome way to start breaking things! I'm merging this into the modules branch and will fix this small bug there.

PR #124, opened 2015-01-14
Module loading

d00rman:
This PR provides a module loading mechanism during spec loading.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -32,6 +32,13 @@ templates:
             - mediawiki/v1/content
           # - mediawiki/v1/mobile
           # - mediawiki/v1/revision-scoring
+          modules:
+            - name: dummy-mod
+              version: 0.0.1
+              type: file
+              path: mods/dummy
+              options:
+                dummyarg: foobar
 
     x-restbase-paths: # Internal paths. These use the same config structure as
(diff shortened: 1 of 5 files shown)


config.example.yaml
```diff
@@ -32,6 +32,13 @@ templates:
             - mediawiki/v1/content
           # - mediawiki/v1/mobile
           # - mediawiki/v1/revision-scoring
+          modules:
+            - name: dummy-mod
+              version: 0.0.1
+              type: file
+              path: mods/dummy
```
gwicke: Should we make the mods/ prefix non-optional / implicit?

d00rman: > Should we make the mods/ prefix non-optional / implicit to encourage all modules to live in one place?

@gwicke  I was thinking the same thing. Since we allow absolute paths too, if there is a reason for not putting a module in mods/, one might recur to that option.

d00rman: @gwicke Done. I put it directly in the modules branch, see [this commit](https://github.com/wikimedia/restbase/commit/bb964d2531de38a453934ffd9e2137a81c71a6c9). Also, this commit now allows one to drop the `path` stanza in case the module name matches its file name - convenient feature (TM).

PR #126, opened 2015-01-15
Load the restbase-mod-table-cassandra module

d00rman:
This PR activates the loading of the restbase-mod-table-cassandra module. Beware that, while this loads the module, **the actual module loading procedure is still incomplete**. Concretely, the procedure needs to be promisified, since restbase-mod-table-cassandra actually returns a Promise object.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -77,18 +77,14 @@ Router.prototype._loadModule = function (modDef, symbols) {
     // let the error propagate in case the module cannot be loaded
     var modObj = require(loadPath)(modDef.options);
     this._modules.set(modDef, modObj);
-    for (var symbol in modObj) {
-        if (!modObj.hasOwnProperty(symbol)) {
-            // toString() or something else, skip this
-            continue;
-        }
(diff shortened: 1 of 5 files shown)


d00rman:
could really use your help on this one, @gwicke

lib/router.js
```diff
@@ -70,10 +70,18 @@ Router.prototype._loadModule = function (modDef, symbols) {
...
     }
+    // append the log property to module options, if it is not present
+    if (modDef.options) {
+        modDef.options.log = modDef.options.log || this._options.log;
+    }
     // let the error propagate in case the module cannot be loaded
     var modObj = require(loadPath)(modDef.options);
     this._modules.set(modDef, modObj);
     for (var symbol in modObj) {
+        if (!modObj.hasOwnProperty(symbol)) {
```
gwicke: Maybe we should do `Object.keys().forEach()` instead. Saves the `hasOwnProperty` check & is a bit clearer.

d00rman: Nicely spotted! Will change that and commit in a couple of minutes.

PR #130, opened 2015-01-15
Drop redundant tests (rerendering.js duplicates ondemand.js)

earldouglas:
If coverage drops, let's fix it before merging this one.


earldouglas:
Not sure where the Coveralls comment is, but there was no change in coverage: https://coveralls.io/builds/1751325

PR #131, opened 2015-01-15
Convert getBucketInfo and listBucket tests into spec x-amples

earldouglas:
This shouldn't decrease coverage, but let's fix it if it does.


earldouglas:
TBD: Data models for these two new-to-the-spec endpoints.

earldouglas:
Coveralls seems to be sleeping on the job.

Coverage was unaffected by this PR: https://coveralls.io/builds/1751667

PR #135, opened 2015-01-16
Resource support, clean-up of spec loading, module interface rework

gwicke:
With this set of patches automatic table creation via resources and the table module actually works. Public API requests are also properly routed to internal handlers.


d00rman:
I really like the shape of this! :) Just one question: we currently automatically do a `put` for missing resources, the logic being this is the current way buckets are created. How about we allow the user to detail in the spec itself what should be done? I'm thinking something like:

```YAML
resources:
  - uri: /{domain}/sys/bla
    method: put
    body:
      key1: value1
      key2: value2
```

Naturally, if only the `uri` is specified, then the default method is `put` and no body is appended.

gwicke:
> How about we allow the user to detail in the spec itself what should be done?

This is already the case. The method is only set if not otherwise set.

gwicke:
@earldouglas, also curious about your thinking on this.

earldouglas:
Looking pretty good to me.  I especially like how modules are more self-contained (i.e. modular) by declaring their dependencies directly from their implementation.

lib/restbase.js
```diff
@@ -56,7 +60,7 @@ RESTBase.prototype.request = function (req) {
             ((req.uri.params && req.uri.params.api === 'sys')
              // TODO: Remove once params.api is reliable
              || (req.uri.path.length > 1 && req.uri.path[1] === 'sys'))) {
-        return Promise.reject(new rbUtil.HTTPError({
+        return Promise.reject(new HTTPError({
```
earldouglas: Should this be moved to `rbUtil#httpErrors`?

gwicke: It could, but I think we should really be very specific in our errors. For the generic ones we could however set up a default `type` and `title` [on the way out](https://github.com/wikimedia/restbase/blob/817cd5037d8d5b126deda3c1cdcdea0134c84d1f/lib/server.js#L39).

PR #144, opened 2015-01-22
Key rev value & make (remaining) tests pass

gwicke:


earldouglas:
Yay!

config.example.yaml
```diff
@@ -60,6 +60,11 @@ templates:
             version: 1.0.0
             type: file
 
+      /{api:sys}/{module:key_rev_value}: &wp/sys/key_value
```
d00rman: The `wp/sys/key_value` achor name is already in use. We should probably rename that to `wp/sys/key_rev_value` for consistency's sake, anyway.

mods/key_rev_value.js
```diff
@@ -0,0 +1,295 @@
...
+        return { status: 400 };
+    });
+};
+
+module.exports = function(options) {
+    var krvBucket = new KRVBucket(options);
+
+    return {
+        spec: spec, // Re-export from spec module
+        test: null, // Spec test function from spec module
```
earldouglas: Is the optional `test` function meant for tests that are hard to write in terms of specs and `x-ample`s?

earldouglas:
Looks good to me.

mods/key_rev_value.js
earldouglas: Chatted with @d00rman ; sounds like this is a way to add implementation-specific tests that augment any spec tests that might be in place.  :+1:

gwicke: I think we'll actually not need to re-export any tests, as

- spec tests are exported by the spec module, and
- implementation-specific tests can simply be run by the module tests.

PR #151, opened 2015-01-26
Further clean up

d00rman:
- `lib/proxyHandler.js` is not used any more, so remove it
- remove unused YAML files
- small fix in `.gitignore`


hardikj:
oops! sorry!

hardikj:
@d00rman are we not going to support proxy handlers anymore?  Eg-https://github.com/hardikj/restbase/blob/master/lib/filters/conf/testHandler.yml ? 

I seems to me lib/router.js dont have support for this, yet?

d00rman:
@hardikj proxying is now the standard way of connecting a public API call with its back-end module. They are linked by the `x-backend-request` stanza in the YAML specification as seen in [content.yaml](https://github.com/wikimedia/restbase/blob/master/specs/mediawiki/v1/content.yaml#L51).

hardikj:
Ah! Thanks.
This looks good to me then merging

gwicke:
@d00rman, @hardikj : We still need the request templating code, and I wasn't sure if that is still in proxyHandler or not. In any case, no big issue to re-add it.

hardikj:
@gwicke yes, I was thinking to update my pull against your repo and include a new file for request templating code.

PR #98, opened 2014-12-30
Add tests for html2wt as described in T75955

earldouglas:
**NOTE: Don't merge this yet -- it depends on a yet-to-be-released version of Parsoid.  It totally passes locally though, I swear.**

This includes tests for:
 * html -> wt
 * html -> pagebundle
 * html -> html

This currently depends on running a patched version of Parsoid locally.

PR #153, opened 2015-02-02
Add tests for bodyOnly field

earldouglas:
Candidate tests to cover the *allow an optional bodyOnly field* requirement from https://phabricator.wikimedia.org/T75955.

Note: test failure is expected, as these tests are to-be-resolved in the TDD workflow.

### DIFF
--- a/test/features/parsoid/html2html-bodyOnly/request.json
+++ b/test/features/parsoid/html2html-bodyOnly/request.json
@@ -1,0 +1,32 @@
+{
+    "bodyOnly": true,
+    "html": {
+        "headers": {
+            "content-type": "text/html;profile=mediawiki.org/specs/html/1.0.0"
+        },
+        "body": "<html>The modified HTML</html>"
+    },
+    "original": {
+        "revid": 12345,
+        "wikitext": {
+            "headers": {
+                "content-type": "text/plain;profile=mediawiki.org/specs/wikitext/1.0.0"
+            },
+            "body": "the original wikitext"
+        },
+        "html": {
+            "headers": {
+                "content-type": "text/html;profile=mediawiki.org/specs/html/1.0.0"
+            },
+            "body": "the original HTML"
+        },
+        "data-parsoid": {
+            "headers": {
+                "content-type": "application/json;profile=mediawiki.org/specs/data-parsoid/0.0.1"
+            },
+            "body": {
+                "ids": {}
+            }
+        }
+    }
+}
--- a/test/features/parsoid/html2html-bodyOnly/response.html
+++ b/test/features/parsoid/html2html-bodyOnly/response.html
@@ -1,0 +1,1 @@
+<body data-parsoid='{"dsr":[0,949,0,0]}' lang="en" class="mw-content-ltr sitedir-ltr ltr mw-body mw-body-content mediawiki" dir="ltr"><p data-parsoid='{"dsr":[928,949,0,0]}'>the original wikitext</p></body>
--- a/test/features/parsoid/html2html-bodyOnly/response.html
+++ b/test/features/parsoid/html2html-bodyOnly/response.html
@@ -1,1 +1,1 @@
-<body data-parsoid='{"dsr":[0,949,0,0]}' lang="en" class="mw-content-ltr sitedir-ltr ltr mw-body mw-body-content mediawiki" dir="ltr"><p data-parsoid='{"dsr":[928,949,0,0]}'>the original wikitext</p></body>
+<p data-parsoid='{"dsr":[928,949,0,0]}'>the original wikitext</p>
--- a/test/features/parsoid/wikitext2html-bodyOnly/request.json
+++ b/test/features/parsoid/wikitext2html-bodyOnly/request.json
@@ -1,0 +1,9 @@
+{
+    "bodyOnly": true,
+    "wikitext": {
+        "body": "The modified HTML",
+        "headers": {
+            "content-type": "text/plain;profile=mediawiki.org/specs/wikitext/1.0.0"
+        }
+    }
+}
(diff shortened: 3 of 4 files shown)


test/features/parsoid/ondemand.js
```diff
@@ -121,4 +121,30 @@ describe('on-demand generation of html and data-parsoid', function() {
         });
     });
 
+    it('should allow full retrieval', function () {
```
earldouglas: This, as expected, currently passes.

test/features/parsoid/ondemand.js
```diff
@@ -121,4 +121,30 @@ describe('on-demand generation of html and data-parsoid', function() {
...
+        })
+        .then(function (res) {
+            slice.halt();
+            assert.deepEqual(res.status, 200);
+            assert.ok(/^<!DOCTYPE html>[^<]*<html[^>]*>/.test(res.body));
+            assert.ok(/<\/html>$/.test(res.body));
+        });
+    });
+
+    it('should allow body-only retrieval', function () {
```
earldouglas: As expected, this curretnly fails.

earldouglas:
This currently only covers the on-demand endpoint.  I'm working on more tests for the transform endpoint.

d00rman:
@earldouglas I'm a bit confused. Is the idea to put the code passing the tests in this PR as well? I.e., is this a WIP?

earldouglas:
@d00rman yes, this PR is just for tests.  Merging this PR means approving the scope of the tests for the feature in question, which unblocks its implementation.

test/features/parsoid/ondemand.js
```diff
@@ -121,4 +121,30 @@ describe('on-demand generation of html and data-parsoid', function() {
...
+        var slice = server.config.logStream.slice();
+        return preq.get({
+            uri: contentUrl + '/html/' + revB + '?bodyOnly',
+        })
+        .then(function (res) {
+            slice.halt();
+            assert.deepEqual(res.status, 200);
+            assert.ok(/^<body>/.test(res.body));
```
gwicke: `/^<body[ >]/` would allow attributes on the body.

test/features/parsoid/transform.js
```diff
@@ -76,22 +76,41 @@ function findSpecs() {
...
+                      '/to/' + spec.to.format + (bodyOnly ? '?bodyOnly' : '');
+            return preq.post({
+                uri: uri,
+                headers: { 'content-type': 'application/json' },
+                body: readFile(spec.from.src)
+            })
+            .then(function(res) {
+                assert.deepEqual(res.status, 200);
+                if (bodyOnly) {
+                    assert.ok(/^<body>/.test(res.body));
```
gwicke: Similar here.

test/features/parsoid/ondemand.js
```diff
@@ -121,4 +121,30 @@ describe('on-demand generation of html and data-parsoid', function() {
...
+            assert.deepEqual(res.status, 200);
+            assert.ok(/^<!DOCTYPE html>[^<]*<html[^>]*>/.test(res.body));
+            assert.ok(/<\/html>$/.test(res.body));
+        });
+    });
+
+    it('should allow body-only retrieval', function () {
+        var slice = server.config.logStream.slice();
+        return preq.get({
+            uri: contentUrl + '/html/' + revB + '?bodyOnly',
```
gwicke: We don't currently have a use case for supporting bodyOnly in GET, and I would argue that using query strings would be a bad idea for caching. Lets perhaps hold off on this, so that we can come up with a single consistent API that covers body-only retrieval along with section retrieval?

test/features/parsoid/transform.js
```diff
@@ -76,22 +76,41 @@ function findSpecs() {
...
-            body: readFile(spec.from.src)
-        })
-        .then(function(res) {
-            assert.deepEqual(res.status, 200);
-            assert.deepEqual(res.body, readFile(spec.to.src));
-        });
+
+    function test(bodyOnly) {
+        return function () {
+            var uri = server.config.baseURL + '/transform/' + spec.from.format +
```
gwicke: Since this is a POST, why don't we set the bodyOnly flag in the POST body / JSON?

test/features/parsoid/ondemand.js
earldouglas: :+1:

test/features/parsoid/transform.js
earldouglas: :+1:

test/features/parsoid/ondemand.js
earldouglas: :+1:

test/features/parsoid/transform.js
earldouglas: :+1:

earldouglas:
Updated per PR feedback.

earldouglas:
Do we want to return the body element, or the body of the body element?

i.e. `<body>foobar</body>` vs `foobar`

gwicke:
> Do we want to return the body element, or the body of the body element?

Actually, `document.body.innerHTML`.

earldouglas:
:+1: updated.

earldouglas:
Bump.

test/features/parsoid/html2html-bodyOnly/request.json
```diff
@@ -0,0 +1,32 @@
+{
+    "bodyOnly": true,
+    "html": {
+        "headers": {
+            "content-type": "text/html;profile=mediawiki.org/specs/html/1.0.0"
+        },
+        "body": "<html>The modified HTML</html>"
+    },
+    "original": {
```
gwicke: This request looks more like a Parsoid v2 request & not like a transform request to me.

earldouglas: Has the spec changed since #98?  What's the real spec for this endpoint?

gwicke: There is no final spec for the public transform end point yet. We were talking about a flat POST API as [in the current Parsoid v1 API](https://www.mediawiki.org/wiki/Parsoid/API), slightly generalized for the `/transform` hierarchy.

Example:

```
POST /{domain}/v1/transform/html/to/wikitext/{title}/{oldid}
Content-type: multipart/form-data

data: '<html>...</html>'
```

earldouglas:
Closing so we can sort out the specs in the [Phabriactor task](https://phabricator.wikimedia.org/T88457).

PR #156, opened 2015-02-03
Use the restbase-mod-table-cassandra npm package

d00rman:
The npm package has been published [here](https://www.npmjs.com/package/restbase-mod-table-cassandra). Use it instead of the GitHub repo. Also, remove the dependency on restbase-cassandra.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,8 +42,7 @@
     "node-uuid": "git+https://github.com/gwicke/node-uuid#master",
     "preq": "~0.3.6",
     "request": "^2.44.0",
-    "restbase-cassandra": "~0.3.7",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -42,8 +42,7 @@
     "node-uuid": "git+https://github.com/gwicke/node-uuid#master",
     "preq": "~0.3.6",
     "request": "^2.44.0",
-    "restbase-cassandra": "~0.3.7",
-    "restbase-mod-table-cassandra": "git+https://github.com/wikimedia/restbase-cassandra#restbase-mod-table-cassandra",
+    "restbase-mod-table-cassandra": "0.0.x",
```
gwicke: We should generally use a pattern here, so that we get updates. I published 0.4.0 (to avoid tag clashes), so maybe we should use ~0.4.0?

d00rman:
Closing in favour of #157

PR #158, opened 2015-02-03
Massage the ondemand.js test to better cover its specs

earldouglas:
This PR is proposed to cover (and close) [T88456](https://phabricator.wikimedia.org/T88456).


hardikj:
Looks good to me!

gwicke:
I'm a bit concerned that this test will fail whenever

- the private data-parsoid content changes
- any minor (non-semantic) detail of the html changes (ex: random ids)

Maybe we should just have some assertions on parts of the content (regexp match, top-level attribute)?

earldouglas:
Fair enough; this tests is putting the cart a bit ahead of the horse, since the spec isn't yet fleshed out.  Let's get that nailed down, then I will update this PR accordingly: https://phabricator.wikimedia.org/T88456

PR #161, opened 2015-02-05
Document some more entry points

gwicke:


d00rman:
@gwicke surprisingly, now the [re-rendering test](https://github.com/wikimedia/restbase/blob/master/test/features/pagecontent/rerendering.js#L27) is [failing](https://travis-ci.org/wikimedia/restbase/jobs/49546457#L349) due to a time-out. I'll merge this, but we need to remember to increase the time-out there.

PR #168, opened 2015-02-09
New URI structure under v1/page/

d00rman:
This PR addresses [T88652](https://phabricator.wikimedia.org/T88652). Specifically, it creates the new `/page/{html,data-parsoid,...}` structure and modifies the URIs used in tests to conform to the new specification. Additionally, two new endpoints are defined - `/page/revision/` and `/page/revision/{revision}` - which should list known revisions and get meta-data about a specific one, respectively. Note that the latter endpoint is not currently functional, additional work is going to be needed in the *page_revisions* module to support it.

### DIFF
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -18,13 +18,15 @@ paths:
     get:
       tags:
         - Page content
-      description: List page properties
+      description: List page properties / page content sub-apis
 
   /{module:page}/title/:
     get:
       tags:
         - Page content
       description: List all known pages
+      x-backend-request:
+        uri: /{domain}/sys/page_revisions/page/
 
   /{module:page}/title/{title}/:
     get:
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -25,6 +25,8 @@ paths:
       tags:
         - Page content
       description: List all known pages
+      produces:
+        - application/json
       x-backend-request:
         uri: /{domain}/sys/page_revisions/page/
 
@@ -32,9 +34,27 @@ paths:
     get:
       tags:
         - Page content
-      description: List page meta-data
+      description: > 
+        List revisions for a title. Currently this lists all
+        revisions that ever used this title, but eventually it should probably
+        return the linear history (across renames) of the page currently using
+        this title. Stability: experimental.
       produces:
         - application/json
+      parameters:
+        - name: domain
+          in: path
+          description: The domain under which the data resides
+          type: string
+          required: true
+          default: en.wikipedia.org
+        - name: title
+          in: path
+          description: the title of page content
+          type: string
+          required: true
+      x-backend-request:
+        uri: /{domain}/sys/page_revisions/page/{title}/
 
   /{module:page}/html/:
     get:
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -18,28 +18,23 @@ paths:
     get:
       tags:
         - Page content
-      description: List pages
-      x-backend-request:
-        uri: /{domain}/sys/page_revisions/page/
+      description: List page properties
 
-  /{module:page}/{title}:
+  /{module:page}/title/:
     get:
       tags:
         - Page content
-      description: Redirect to the latest HTML for this page
-      # TODO
-      # x-backend-request:
-      #   uri: /{domain}/sys/page_revisions/page/
+      description: List all known pages
 
-  /{module:page}/{title}/:
+  /{module:page}/title/{title}/:
     get:
       tags:
         - Page content
-      description: List available content types for a page
+      description: List page meta-data
       produces:
         - application/json
 
-  /{module:page}/{title}/html:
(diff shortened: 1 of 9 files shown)


specs/mediawiki/v1/content.yaml
```diff
@@ -18,28 +18,23 @@ paths:
...
-  /{module:page}/{title}:
+  /{module:page}/title/:
     get:
       tags:
         - Page content
-      description: Redirect to the latest HTML for this page
-      # TODO
-      # x-backend-request:
-      #   uri: /{domain}/sys/page_revisions/page/
+      description: List all known pages
```
gwicke: To make this work, add:
```
x-backend-request:		
   uri: /{domain}/sys/page_revisions/page/
```

specs/mediawiki/v1/content.yaml
```diff
@@ -18,28 +18,23 @@ paths:
...
 
-  /{module:page}/{title}/:
+  /{module:page}/title/{title}/:
```
gwicke: Alternative: List revisions for a page (aka page history).

specs/mediawiki/v1/content.yaml
```diff
@@ -75,7 +70,7 @@ paths:
         # See also http://tools.ietf.org/html/rfc7231#section-7.1.2
         uri: /{domain}/sys/parsoid/html/{title}/latest
 
-  /{module:page}/{title}/html/:
+  /{module:page}/html/{title}/:
```
gwicke: Should we make this the list of HTML tids instead of revisions?

specs/mediawiki/v1/content.yaml
```diff
@@ -110,7 +105,7 @@ paths:
         uri: /{domain}/sys/parsoid/html/{title}/
 
 
-  /{module:page}/{title}/html/{revision}{/tid}:
+  /{module:page}/html/{title}/{revision}{/tid}:
```
gwicke: Idea: accept either tids or revisions for the revision parameter, and add a separate entry point `/{module:page}/html/{title}/{revision}/{tid}` to list html tids for a given MW revision.

specs/mediawiki/v1/content.yaml
earldouglas: Might each be useful in different cases?

If so, we could support both (and be ready to support related cases in the future) with the [`Accept` header](http://tools.ietf.org/html/rfc2616#page-100).

gwicke: I think `Accept` is a great way to request the same semantic information in different formats, but am not convinced that it is the right way to return different resources.

Swagger lets us document each possible return content-type.

gwicke: The more I think about this from the content model perspective, the more I'm leaning towards listing the *potential* revisions that can have an entry in a given property. For HTML, that would probably be all of them. For data-parsoid, that would only be the revisions with a parsoid-compatible content model.

gwicke: Lets just leave this as it is, but document that it's listing the *potential* revisions that can have this property?

specs/mediawiki/v1/content.yaml
earldouglas: It sounds like there are multiple implied use cases relating to tracing the past (or future) history of a page, and citing a page at a given point in time.  I recommend we enumerate the user stories to best inform the design of this endpoint.

gwicke: @earldouglas, yup. I spelled some out at https://phabricator.wikimedia.org/T89166.

specs/mediawiki/v1/content.yaml
gwicke: Addressed in e254cf89d.

specs/mediawiki/v1/content.yaml
```diff
@@ -18,28 +18,33 @@ paths:
...
-  /{module:page}/{title}/html:
+  /{module:page}/html/:
+    get:
+      tags:
+        - Page content
+      description: List titles for which the HTML property is available
+      x-backend-request:
+        uri: /{domain}/sys/page_revisions/page/
+
+  /{module:page}/html/{title}:
```
gwicke: Should we directly return the latest result here & send no-cache headers or explicitly purge this end point? Redirecting would only add latency once we are in restbase. Pros would be a) teaching users to actually use the revision URI, and b) a smaller cache entry if we cache / purge the redirect response.

gwicke:
Rebased on top of master.

gwicke:
@d00rman, @earldouglas: LGTM, but lets discuss the API changes tomorrow before merging.

specs/mediawiki/v1/content.yaml
earldouglas: Re: using the Accept header to choose the resource representation(s), see: https://developer.github.com/v3/media/#comment-body-properties

gwicke: @earldouglas, exactly. It's about representations of the same resource, not normally different resources.

gwicke:
@earldouglas, any objections to merging this?

PR #170, opened 2015-02-10
Optimize request templating for service requests

hardikj:


lib/reqTemplating.js
```diff
@@ -0,0 +1,52 @@
...
+function _assemble(spec) {
+    var traverse = function(path, key) {
+        // A simple dotted path reference
+        var code = "req"
+        path.split(/\./).forEach(function(pathBit, index) {
+            if (index) {
+                code += "['"+encodeURIComponent(pathBit)+"']";
+            }
+        });
+        return "newReq['"+key+"'] = "+code+";";
```
hardikj: not sure if the 'key' here should be encoded this comes from the yalm file.

gwicke: It needs to be to make sure that it's valid JS. The simplest way to do this is `JSON.stringify('foo"bar') --> "foo\"bar"`

PR #171, opened 2015-02-11
Pushing computing delta into timing module

arlolra:
Is that desirable?

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -53,9 +53,13 @@ StatsD.prototype.timing = function timing(name, suffix, delta) {
         this.statsd.sendAll(stats, delta, 'ms');
     } else {
         suffix = suffix ? '.' + suffix : '';
-        this.statsd.timing(this.makeName(name) + suffix, delta);
+        this.statsd.timing(name + suffix, delta);
     }
     return delta;
+};
+
+StatsD.prototype.endTiming = function endTiming(name, suffix, start) {
+    return this.timing(name, suffix, Date.now() - start);
 };
 
 StatsD.prototype.count = function count(name, suffix) {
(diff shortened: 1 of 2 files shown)


earldouglas:
Makes sense to me.

gwicke:
I left it out as passing in the timing seemed slightly more flexible (can add up several timings to one aggregate etc). Don't care too strongly though.

arlolra:
The common case is tedious though,
https://gerrit.wikimedia.org/r/#/c/186219/7/api/routes.js

You can always pass in (Date.now() - aggregate) :)

gwicke:
Maybe the start time could be an optional parameter for the best of both worlds?

arlolra:
> Maybe the start time could be an optional parameter for the best of both worlds?

The signature would start to get ugly. I don't like passing in a `null` for `delta` in that case.

I guess just adding a wrapper would suffice.

    StatsD.prototype.endTimer = function( name, suffix, start ) {
      return this.timing( name, suffix, Date.now() - start );
    }

Or maybe this has been enough of a distraction already :)

gwicke:
@arlolra, my idea was an optional last parameter, like this:

```javascript
// Implicit delta calculation
statsd.timing( name, suffix, Date.now(), startTime );

// Plain timing as it works right now
statsd.timing( name, suffix, timeDelta );

gwicke:
@arlolra, that would work for me. Perhaps call it `endTiming` instead for some consistency?

arlolra:
You've s/-/,/ :)

`endTiming` sounds good. Or, `captureTiming`? (suggested just to keep this conversation going)

gwicke:
> (suggested just to keep this conversation going)

Ha! We'll move this code to a separate package so that we can share it. Can add the method then.

arlolra:
Oh ... too late. And I merged in #175.

gwicke:
@arlolra: alright, thank you!

PR #169, opened 2015-02-10
Revisions

d00rman:
This PR is a continuation of #168 so please make sure to merge that one before considering this.

It adds the /revision/ and /revision/{revision} endpoints to the page_revisions module and exposes them in the public API.

PR #172, opened 2015-02-11
Revision suppression / deletion detection

d00rman:
This PR build on #169 and provides the mechanism needed to deny access to revisions with restrictions. It is enforced both when requesting info about a specific revision (with or without providing a title), as well as its contents.

If a specific revision is passed in, a second, parallel request is made to the *page_revisions* module to determine whether the sought revision is accessible. In case it is restricted, access in denied to the user. Note that the content itself is fetched and stored regardless of access restrictions.

### DIFF
--- a/mods/page_revisions.js
+++ b/mods/page_revisions.js
@@ -76,6 +76,33 @@ PRS.prototype.getTableSchema = function () {
     };
 };
 
+/**
+ * Checks the revision info returned from the storage
+ * for restrictions, and if there are any, raises an error
+ *
+ * @param res Object the result as returned from storage
+ * @return Object the same object
+ * @throws rbUtil.httpError if access to the revision should be denied
+ */
+PRS.prototype._checkRevReturn = function(res) {
+    var item = res && res.body && Array.isArray(res.body.items) && res.body.items[0];
+    // if there are any restrictions imposed on this
+    // revision, forbid its retrieval, cf.
+    // https://phabricator.wikimedia.org/T76165#1030962
+    if (item && Array.isArray(item.restrictions) && item.restrictions.length > 0) {
+        // there are some restrictions, deny access to the revision
+        return Promise.reject(new rbUtil.HTTPError({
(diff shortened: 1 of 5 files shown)


gwicke:
Rebased onto master

mods/page_revisions.js
```diff
@@ -76,6 +76,33 @@ PRS.prototype.getTableSchema = function () {
...
+/**
+ * Checks the revision info returned from the storage
+ * for restrictions, and if there are any, raises an error
+ *
+ * @param res Object the result as returned from storage
+ * @return Object the same object
+ * @throws rbUtil.httpError if access to the revision should be denied
+ */
+PRS.prototype._checkRevReturn = function(res) {
+    var item = res && res.body && Array.isArray(res.body.items) && res.body.items[0];
```
gwicke: I think the table storage API guarantees that a successful reply includes an items array, so we should be able to only check for its length.

mods/page_revisions.js
```diff
@@ -76,6 +76,33 @@ PRS.prototype.getTableSchema = function () {
...
+            status: 403,
+            body: {
+                type: 'access_denied#revision',
+                title: 'Access to resource denied',
+                description: 'Access is restricted for revision ' + item.rev,
+                restrictions: item.restrictions
+            }
+        }));
+    }
+    return Promise.resolve(res);
```
gwicke: Since this is synchronous, is there a need to do the Promise wrapping here?

mods/page_revisions.js
```diff
@@ -211,6 +253,7 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
         throw new Error("Invalid revision: " + rp.revision);
     }
     return revisionRequest
+    .then(self._checkRevReturn)
```
gwicke: I think you can make _checkRevReturn sync & move it to the callback. _checkRevReturn can throw the HTTPError if it doesn't check out.

PR #173, opened 2015-02-12
Implement and test the transform API as specced

gwicke:
See: https://phabricator.wikimedia.org/T75955


mods/parsoid.js
```diff
@@ -144,64 +143,140 @@ PSP.transformRevision = function (restbase, req, from, to) {
...
+                    'content-type': res.headers['content-type']
+                },
+                body: res.body
+            };
         });
     }
 
-    return Promise.all([ get('html'), get('wikitext'), get('data-parsoid') ])
-    .then(function () {
+    // Get the revision info just to make sure we have access
```
d00rman: These things are resolved in #172 ;)

mods/parsoid.js
```diff
@@ -144,64 +143,140 @@ PSP.transformRevision = function (restbase, req, from, to) {
...
+        return match[1];
+    }
+}
+
 PSP.makeTransform = function (from, to) {
     var self = this;
 
     return function (restbase, req) {
         var rp = req.params;
+        if (false && !req.body[from]) {
```
d00rman: `false && sth` ?

mods/parsoid.js
```diff
@@ -144,64 +143,140 @@ PSP.transformRevision = function (restbase, req, from, to) {
...
+};
+
+/**
+ * Cheap body.innerHTML extraction.
+ *
+ * This is safe as we know that the HTML we are receiving from Parsoid is
+ * serialized as XML.
+ */
+function cheapBodyInnerHTML(html) {
+    var match = /<body[^>]*>(.*)<\/body>/.exec(html);
```
d00rman: I believe a body can span multiple lines. If so, the correct way would probably be:

```javascript
var re = new RegExp('/<body[^>]*>([\\s\\S]*)<\\/body>/', 'gm');
var match = re.exec(html);
```

test/features/parsoid/transform.js
```diff
@@ -0,0 +1,153 @@
...
+            }
+        })
+        .then(function (res) {
+            assert.deepEqual(res.status, 200);
+			var pattern = /<div id="bar">Selser test<\/div>/;
+			if (!pattern.test(res.body)) {
+				throw new Error('Expected pattern in response: ' + pattern
+						+ '\nSaw: ' + JSON.stringify(res, null, 2));
+			}
+			assert.deepEqual(res.headers['content-type'],
```
d00rman: you can use `assert.contentType(res, 'your-expected-content-type-here')` to test for the resulting *content-type* header.

test/features/parsoid/transform.js
```diff
@@ -0,0 +1,153 @@
...
+			uri: server.config.baseURL
+				+ '/transform/html/to/html/' + testPage.title
+				+ '/' + testPage.revision,
+            body: {
+				html: testPage.html
+            }
+        })
+        .then(function (res) {
+            assert.deepEqual(res.status, 200);
+			var pattern = /<div id="bar">Selser test<\/div>/;
```
d00rman: This block seems to have wrong indenting

d00rman: Strike that. It seems that the whole file is tab-indented. Are you crossing over to the dark side of tabs? :no_good:

mods/parsoid.js
gwicke: Good catch!

test/features/parsoid/transform.js
gwicke: Nah, just happened to be tab indented & I didn't change it. Shall fix.

test/features/parsoid/transform.js
gwicke: K, done.

PR #174, opened 2015-02-13
Use patched swagger-ui library with support for {/optionalPath}

gwicke:
See https://github.com/wikimedia/swagger-ui/pull/1

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "request": "^2.44.0",
     "restbase-mod-table-cassandra": "~0.4.3",
     "swagger-router": "^0.0.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "request": "^2.44.0",
     "restbase-mod-table-cassandra": "~0.4.3",
     "swagger-router": "^0.0.3",
-    "swagger-ui": "^2.1.1-M1",
+    "swagger-ui": "git+https://github.com/gwicke/swagger-ui#master",
```
d00rman: shouldn't that be `git+https://github.com/wikimedia/swagger-ui#master` ?

PR #175, opened 2015-02-13
Redundant makeName call

arlolra:


arlolra:
Why is coverage being reported twice?

PR #176, opened 2015-02-13
Minor fixes from #172

d00rman:


d00rman:
@gwicke this PR fixes your (small) remarks about the revdelete PR #172

mods/page_revisions.js
```diff
@@ -253,8 +253,9 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
         throw new Error("Invalid revision: " + rp.revision);
     }
     return revisionRequest
-    .then(self._checkRevReturn)
     .then(function(res) {
+        // check if the revision has any restrictions
+        self._checkRevReturn(res);
```
gwicke: There is another one at https://github.com/wikimedia/restbase/blob/phab/T76165/mods/page_revisions.js#L343

d00rman: Yep, but we can't escape that since:
- we have to wait for the return from storage in order to check the restrictions
- there are no other `.then()` blocks where we could stick it

gwicke: Ah yeah, sorry. Misread the .catch as a .then.

PR #178, opened 2015-02-14
Namespace property and secondary index

d00rman:
This PR adds the `namespace` property and its secondary index (which is not used ATM).


mods/page_revisions.js
```diff
@@ -340,7 +348,14 @@ PRS.prototype.getRevision = function(restbase, req) {
             limit: 1
         }
     })
-    .then(self._checkRevReturn)
+    .then(function(res) {
+        // check the return
+        self._checkRevReturn(res);
+        // and get the revision info for the
```
gwicke: Nice!

gwicke:
Looks great!

PR #180, opened 2015-02-14
Clean up and iron out the content API specification YAML

d00rman:


gwicke:
LGTM, only waiting for travis to catch up.

gwicke:
Actually, there is nothing in here (other than invalid yaml) that would affect the tests. Merging.

PR #181, opened 2015-02-14
Add 403 return errors to URIs dealing with revisions

d00rman:
Forgot to address that in #180 , so adding it separately.

### DIFF
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -209,7 +209,7 @@ paths:
         '403':
           description: access to the specific revision is restricted
           schema:
-            $ref: '#/definitions/accessDenied'
+            $ref: '#/definitions/httpError'
         '404':
           description: unknown domain, page, revision or tid
           schema:
@@ -339,7 +339,7 @@ paths:
         '403':
           description: access to the specific revision is restricted
           schema:
-            $ref: '#/definitions/accessDenied'
+            $ref: '#/definitions/httpError'
         '404':
           description: unknown domain, page, revision or tid
           schema:
@@ -401,7 +401,7 @@ paths:
         '403':
           description: access to the specific revision is restricted
           schema:
-            $ref: '#/definitions/accessDenied'
+            $ref: '#/definitions/httpError'
         '404':
           description: Unknown revision id or domain
           schema:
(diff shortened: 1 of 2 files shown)


specs/mediawiki/v1/content.yaml
```diff
@@ -558,6 +582,28 @@ definitions:
         type: string
       uri:
         type: string
+  accessDenied:
```
gwicke: This is a https://tools.ietf.org/html/draft-nottingham-http-problem

specs/mediawiki/v1/content.yaml
```diff
@@ -558,6 +582,28 @@ definitions:
...
+    required:
+      - code
```
gwicke: The code is in the http status.

specs/mediawiki/v1/content.yaml
```diff
@@ -558,6 +582,28 @@ definitions:
...
+      - method
```
gwicke: I believe this is in a req sub-object.

specs/mediawiki/v1/content.yaml
```diff
@@ -558,6 +582,28 @@ definitions:
...
+      code:
+        type: integer
+        format: int32
+      method:
+        type: string
+      title:
+        type: string
+      type:
+        type: string
+      description:
```
gwicke: s/description/detail/

specs/mediawiki/v1/content.yaml
```diff
@@ -558,6 +582,28 @@ definitions:
...
+        type: string
+      uri:
```
gwicke: req.uri?

d00rman: ups, c/p error :)

d00rman:
@gwicke I have changed `accessDenied` to `httpError` (and updated the fields accordingly). Should we make it as the default schema for all errors, since we have field whitelisting allowing only certain error properties to see the light of day?

specs/mediawiki/v1/content.yaml
```diff
@@ -558,6 +582,21 @@ definitions:
         type: string
       uri:
         type: string
+  httpError:
+    required:
+      - method
```
gwicke: I wonder if we should move 'method' down into a req sub-object that has just the uri & method.

PR #184, opened 2015-02-17
RFC: Use service runner

gwicke:
This hooks up the new service runner / servisor and uses it to run restbase. Tests are passing, and a b/c compat server.js stub is provided so that the puppetization keeps working. We should be able to remove that later.

Before merging, lets discuss this properly.


lib/server.js
```diff
@@ -214,12 +219,14 @@ function setupConfigDefaults(conf) {
...
-    conf = setupConfigDefaults(conf);
+function main(options) {
+    var conf = setupConfigDefaults(options.config);
     // Set up the global options object with a logger
     var opts = {
-        log: rbUtil.makeLogger(conf.logging),
-        conf: conf
+        conf: conf,
+        logger: options.logger,
+        log: options.logger.log.bind(options.logger),
```
d00rman: why are there both `log` and `logger` ?

package.json
```diff
@@ -46,7 +47,8 @@
     "swagger-router": "^0.0.3",
     "swagger-ui": "git+https://github.com/wikimedia/swagger-ui#master",
     "url-template": "2.0.4",
-    "yargs": "~1.3.0"
+    "yargs": "~1.3.0",
+    "servisor": "git+https://github.com/gwicke/servisor#master"
```
d00rman: `/wikimedia/servisor`

lib/server.js
gwicke: The service runner code no longer sets up the magic function that also has a .child method attached to it. Instead, it exposes a plain logger object.

I'm torn about re-introducing the magic. It sure is more convenient for users, just feels a bit dirty.

package.json
gwicke: Yup. Ideally, lets change straight to a real package once we have decided on a name.

test/utils/assert.js
```diff
@@ -26,8 +26,8 @@ function localRequests(slice, expected) {
         }),
         expected,
         expected ?
-          'Should not have made remote request' :
-          'Should have made a remote request'
+          'Should not have made local request' :
+          'Should have made a local request'
```
d00rman: good catch :+1:

gwicke: Led me down the wrong path for an hour or two..

d00rman: Ups... Sorry about that :disappointed:

test/utils/server.js
```diff
@@ -32,16 +39,13 @@ function start(_options) {
...
-            },
-            spec: config.spec
-        }).then(function(server){
+        console.log('starting restbase in '
+                + (options.offline ? 'OFFLINE' : 'ONLINE') + ' mode');
+        config.conf.offline = options.offline || false;
+
+        return runner.run(config.conf)
+        .then(function(servers){
+            var server = servers[0];
```
d00rman: With servisor, ops, service-runner, this becomes really easy to read. Me likes mucho.

package.json
d00rman: Yep. Upload the pkg first, though ;)

test/utils/server.js
```diff
@@ -3,7 +3,7 @@
 // mocha defines to avoid JSHint breakage
 /* global describe, it, before, beforeEach, after, afterEach */
 
-var restbase  = require('../../lib/server.js');
+var ServiceRunner = require('servisor');
```
hardikj: service-runner now ;)

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "REST storage and service dispatcher",
   "main": "server.js",
   "scripts": {
+    "start": "servisor",
```
hardikj: 'npm start' will be used to start servisor(now service-runner) or restbase?

d00rman: `npm start` will run `service-runner` which will read RESTBase's config and start it.

hardikj: Ah, nice!
We should modify the Readme as well with the above changes.

PR #185, opened 2015-02-18
Misc clean-up

gwicke:


d00rman:
Is the idea here to keep the fast(er) bluebird even if/when a built-in Promise is available? If so, why not setting `global.P = require('bluebird')` (in, e.g., `lib/server.js`) so that it is automatically available throughout restbase?

gwicke:
> If so, why not setting global.P = require('bluebird') (in, e.g., lib/server.js) so that it is automatically available throughout restbase?

Generally setting globals is not considered very clean. The only exception should be shimming standards-compliant globals like Promise, Map or Set, ideally with an implementation that's faithful to the spec.

We require all kinds of modules individually, and bluebird is just one of them. I think the syntactic overhead is fairly bearable.

PR #189, opened 2015-02-20
Minor fixes in the spec

d00rman:


config.example.yaml
```diff
@@ -11,11 +11,11 @@ templates:
     # swagger options, overriding the shared ones from the merged specs (?)
     info:
       version: 1.0.0-abcd
-      title: Standard Wikimedia content API
-      description: All the content for this domain.
+      title: Wikimedia REST API
+      description: A REST content API to MediaWiki, currently strongly focused on page content.
```
gwicke: The 'to MediaWiki' bit might be confusing. Maybe omit it?

config.example.yaml
```diff
@@ -11,11 +11,11 @@ templates:
...
       termsOfService: http://wikimedia.org/terms/
       contact:
-        name: The project maintainers
+        name: the Wikimedia Services team
```
gwicke: capitalize: The

PR #190, opened 2015-02-20
Disable the swagger validator

gwicke:
The swagger validator is currently http-only, which means that it causes mixed
content warnings when accessing the docs using https. It's also not a good
idea to call out to external services.

### DIFF
--- a/lib/swaggerUI.js
+++ b/lib/swaggerUI.js
@@ -20,7 +20,7 @@ function staticServe (restbase, req) {
                 .replace(/<title>[^<]*<\/title>/,
                         '<title>RESTBase docs</title>')
                 // Replace the default url with ours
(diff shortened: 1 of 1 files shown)


d00rman:
@gwicke could you also fix the HMTL Title to something more meaningful than 'Swagger UI'? Maybe 'Wikimedia REST API Docs' ?

gwicke:
@d00rman, sure. The only practical problem with that is that the restbase code is independent of the actual API spec so far. IMHO the title should come from the spec title rather than from RESTBase, which would require some changes in swagger-ui to set it dynamically from the actual spec.

I can set a static 'RESTBase docs' placeholder for now, which is at least slightly better than 'Swagger UI'.

PR #193, opened 2015-02-25
Update swagger-router dependency

gwicke:


d00rman:
swagger-router's version is still [held at 0.0.3](https://github.com/wikimedia/swagger-router/blob/master/package.json#L3). Am I missing something, @gwicke ?

gwicke:
Good catch: Forgot to create a PR. Now done: https://github.com/wikimedia/swagger-router/pull/17

gwicke:
@d00rman, also published 0.0.4 already to npmjs.org.

d00rman:
Heh, not losing any time I see :smile:

PR #194, opened 2015-02-26
Normalize titles

gwicke:
MediaWiki normally converts spaces to underscores in its internal database
keys. This patch mimics the same behavior in RESTBase, which is important to
avoid duplication of content.

The normalizeTitle function is duplicated in this patch, which isn't quite
ideal. It doesn't belong in rbUtil either though, as it is MediaWiki-specific.
We could consider shipping a bundle of MediaWiki-related modules
(at least title_revision & parsoid), and then have a shared util module in
that package.


d00rman:
Agreed for the packaging. We should extract all MW-specific transforms and put them under the same umbrella. Will just merge this for now.

PR #198, opened 2015-02-26
List revision end points only in 'Page content' section

gwicke:
Swagger-ui uses ids to expand/collapse sections, and this doesn't work
reliably if an ID is not unique in a page. This was the case with the revision
end points, as they were listed in both the 'Page content' and 'Revision'
sections.

Given the choice between only listing them in the 'Page content' or 'Revision'
section, I went with 'Page content', as these revisions are really
page-specific. There might be other types of revisions for other kinds of
non-page content.

### DIFF
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -355,7 +355,6 @@ paths:
     get:
       tags:
         - Page content
-        - Revision
       description: List available revisions
       produces:
         - application/json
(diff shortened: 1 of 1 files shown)


d00rman:
This is actually a nice mental exercise. Makes one think about categorisation. Conclusion: each module should have its own tag :+1:

gwicke:
@d00rman, or swagger-ui needs to generate unique ids for entry points that are listed in multiple sections. I actually like the capability of having entry points show up in multiple groups -- it just doesn't work well right now.

gwicke:
This is actually already reported: https://github.com/swagger-api/swagger-ui/issues/616

PR #201, opened 2015-02-28
Pull in cassandra driver v2 for testing

gwicke:
Don't merge until we have tested this a bit on the test cluster.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -15,7 +15,7 @@
 	"curly": true,
 	"eqeqeq": true,
 	"immed": true,
-	"latedef": true,
+	"latedef": "nofunc",
 	"newcap": true,
 	"noarg": true,
 	"noempty": true,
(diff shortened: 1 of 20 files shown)


d00rman:
Hm, I see a new `cassandra_driver_v2` branch in this repo, but this PR does not use it al all. A typo? IMHO, it should be better/easier to test if these commits were made in wikimedia/restbase/cassandra_driver_v2 than gwicke/restbase/cassandra_driver_v2.

gwicke:
@d00rman, I had the same thought, but then realized that it actually matters more how we push this into the deploy repo. I already used this code for wiki creation testing yesterday, where the v2 driver proved reliable.

gwicke:
Merging without waiting for travis after triple-checking the tests locally. Travis seems to be down right now.

gwicke:
The travis run was actually fine, only the reporting failed apparently: https://travis-ci.org/wikimedia/restbase/builds/53063025

PR #202, opened 2015-03-01
RFC: Shared cfs

gwicke:
RFC / WIP for https://phabricator.wikimedia.org/T91188. 

It's not functional (yet), so don't merge.

### DIFF
--- a/mods/key_rev_value.js
+++ b/mods/key_rev_value.js
@@ -43,6 +43,7 @@ KRVBucket.prototype.makeSchema = function (opts) {
             ]
         },
         attributes: {
+            domain: 'string',
             key: opts.keyType || 'string',
             rev: 'int',
             tid: 'timeuuid',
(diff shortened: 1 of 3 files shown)


d00rman:
We should really get into the habit of creating branches on the master repo so that we can collaborate on these WIPs. Also, this is not something that's dependent on the v2 cass driver, so this branch would have been better created from master.

gwicke:
@d00rman, this is just a quick workspace at this point. 

I agree that we should create a shared branch for bigger changes (that has worked well before), but if it's anything more complex I prefer to then do PRs against that shared branch from our private repos rather than directly pushing things in there. Otherwise we end up with a huge branch without a good point to review a reasonably-sized chunk of work.

gwicke:
Closed this in favor of doing this in restbase-mod-table-cassandra, as discussed in https://phabricator.wikimedia.org/T91188.

PR #203, opened 2015-03-03
Add an example of storage groups for restbase-mod-table-cassandra

d00rman:


config.example.yaml
```diff
@@ -49,6 +49,13 @@ templates:
                 username: cassandra
                 password: cassandra
                 defaultConsistency: one # or 'one' for single-node testing
+                storage_groups:
+                  - name: test.domains
```
gwicke: Minor: Should we let this end on `.group` as well?

PR #205, opened 2015-03-03
Cassandra driver v2 fixes / tweaks

gwicke:


gwicke:
Coverage drops with this PR as I disabled the two failing listing tests.

PR #210, opened 2015-03-06
reenable listing tests

eevans:
Bump the version of mod-table-cassandra to a one that supports creation of indexes on the `_domain` attribute, and reenable the corresponding tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "node-uuid": "git+https://github.com/gwicke/node-uuid#master",
     "preq": "~0.3.6",
     "request": "^2.44.0",
-    "restbase-mod-table-cassandra": "^0.5.0",
+    "restbase-mod-table-cassandra": "^0.5.2",
     "service-runner": "^0.1.4",
     "swagger-router": "^0.0.4",
     "swagger-ui": "git+https://github.com/wikimedia/swagger-ui#master",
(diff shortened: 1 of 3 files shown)


eevans:
The relevant error from the failed test is:

    Partition key part title must be restricted since preceding part is...

Which in this case means it is expecting both `_domain` *and* `title` as predicates.  I believe this is because Travis is running these tests with Cassandra 2.0.9, and the changes required to index a compound partition key require >= 2.1.

TL;DR Travis needs to use a new version of Cassandra.

d00rman:
Hm, not sure about the Cassandra version problem, but what I am sure of is that the tests here fail because this PR depends on wikimedia/restbase-mod-table-cassandra/pull/86, which isn't being pulled in here.

package.json
```diff
@@ -42,7 +42,7 @@
     "node-uuid": "git+https://github.com/gwicke/node-uuid#master",
     "preq": "~0.3.6",
     "request": "^2.44.0",
-    "restbase-mod-table-cassandra": "^0.5.0",
+    "restbase-mod-table-cassandra": "^0.5.1",
```
d00rman: You'd probably want to make that `^0.5.2` seeing that `^0.5.1` has already been taken by @gwicke :)

PR #212, opened 2015-03-08
Ensure an explicit charset in text/html content-types

gwicke:
The content-type header we receive from Parsoid does not currently include an
explicit charset. Until this is fixed, we should ensure that any text/html
returns have an explicit utf-8 charset set, so that clients can properly
decode & display the HTML content.

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -41,17 +41,26 @@ var PSP = ParsoidService.prototype;
  */
 PSP.wrapContentReq = function(restbase, req, promise) {
     var rp = req.params;
+    function ensureCharsetInContentType(res) {
+        var cType = res.headers['content-type'];
+        if (/^text\/html\b/.test(cType) && !/charset=/.test(cType)) {
+            // Make sure a charset is set
+            res.headers['content-type'] = cType + ';charset=utf-8';
+        }
+        return res;
+    }
+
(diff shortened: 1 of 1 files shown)


d00rman:
Is there any chance of the charset not being UTF-8? We store in Cassandra as UTF-8, and then serve it later as such. That means that if Parsoid can give us something that is not UTF-8, that content is ultimately going to be wrongly decoded.

gwicke:
I think we should actually reject any content that's not UTF8-encoded before storing it in restbase. There's really no excuse to mess with legacy encodings these days. 

Parsoid uses UTF8 throughout, we just forgot to include the explicit charset in the v2 content-types. We shall fix that in Parsoid (patch to follow), but right now we have stored content that doesn't have the charset set yet.

d00rman:
> I think we should actually reject any content that's not UTF8-encoded before storing it in restbase. There's really no excuse to mess with legacy encodings these days.

:+1: 

This will become relevant once we allow `PUT`s.

> Parsoid uses UTF8 throughout, we just forgot to include the explicit charset in the v2 content-types. We shall fix that in Parsoid (patch to follow), but right now we have stored content that doesn't have the charset explicitly set.

Ah, I see. Cool then, no worries for this particular case.

gwicke:
For the record, Parsoid follow-up in https://gerrit.wikimedia.org/r/#/c/195184/.

PR #213, opened 2015-03-08
Store content using the canonical title only

gwicke:
Requests with a non-canonical title (example: 'foobar' vs. 'Foobar') can
currently cause key_rev_value content to be stored using both the
non-canonical version ('foobar') and the canonical version ('Foobar'). This is
bad for performance and consistency.

This patch avoids this by comparing the normalized title retrieved from
page_revisions with the (slightly pre-normalized) request title. If the two
differ, the request is retried against storage before (potentially) creating
the content.

### DIFF
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -81,6 +81,43 @@ RESTBase.prototype.defaultListingHandler = function(value, restbase, req) {
             req.query.path = '/index.html';
         }
         return swaggerUI(restbase, req);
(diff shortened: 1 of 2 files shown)


d00rman:
@gwicke LGTM, but a test case would come in handy here, especially once we start tackling [T87393](https://phabricator.wikimedia.org/T87393)

gwicke:
@d00rman, can we add the test in a follow-up?

gwicke:
Screenshot of landing page:

image

PR #214, opened 2015-03-09
Attach a x-request-id header to each incoming request

d00rman:
For each request to be handled, attach a `x-request-id` header and propagate it further to child requests. This allows us to track requests throughout the system, even when RESTBase acts only as a proxy to other back-end services.

Bug: [T89562](https://phabricator.wikimedia.org/T89562)

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -141,24 +141,6 @@ rbUtil.generateRequestId = function() {
     return reqIdBuf.toString('hex');
 };
 
-/**
- * Checks for the existence of the x-request-id header and
- * sets it in case it is not present
- *
- * @param {Request} the request object
- * @return {String} the obtained or generated request ID
- */
-rbUtil.requestId = function(req) {
-    if(!req) {
-        return rbUtil.generateRequestId();
-    }
-    req.headers = req.headers || {};
-    if(!req.headers['x-request-id']) {
-        req.headers['x-request-id'] = rbUtil.generateRequestId();
-    }
-    return req.headers['x-request-id'];
-};
-
 /*
  * Error instance wrapping HTTP error responses
  *
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -141,6 +141,24 @@ rbUtil.generateRequestId = function() {
     return reqIdBuf.toString('hex');
 };
 
+/**
+ * Checks for the existence of the x-request-id header and
+ * sets it in case it is not present
+ *
+ * @param {Request} the request object
+ * @return {String} the obtained or generated request ID
+ */
+rbUtil.requestId = function(req) {
+    if(!req) {
+        return rbUtil.generateRequestId();
+    }
+    req.headers = req.headers || {};
+    if(!req.headers['x-request-id']) {
+        req.headers['x-request-id'] = rbUtil.generateRequestId();
+    }
+    return req.headers['x-request-id'];
+};
+
 /*
  * Error instance wrapping HTTP error responses
  *
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -131,14 +131,14 @@ rbUtil.isTimeUUID = function (s) {
 };
 
 
-var buf = new Buffer(16);
+var reqIdBuf = new Buffer(16);
 
 /**
  * Generates a new request ID
  */
 rbUtil.generateRequestId = function() {
-    uuid.v4(null, buf);
-    return buf.toString('hex');
+    uuid.v4(null, reqIdBuf);
+    return reqIdBuf.toString('hex');
 };
 
 /*
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -130,6 +130,17 @@ rbUtil.isTimeUUID = function (s) {
     return uuidRe.test(s);
 };
 
+
+var buf = new Buffer(16);
+
+/**
+ * Generates a new request ID
+ */
+rbUtil.generateRequestId = function() {
+    uuid.v4(null, buf);
+    return buf.toString('hex');
+};
+
 /*
  * Error instance wrapping HTTP error responses
  *
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -34,7 +34,6 @@ function RESTBase (options, req) {
         this.metrics = par.metrics;
         this._recursionDepth = par._recursionDepth + 1;
         this._priv = par._priv;
-        this._isMain = false;
         this._reqId = par._reqId;
     } else {
         // Brand new instance
@@ -45,7 +44,6 @@ function RESTBase (options, req) {
         this._parent = null;
         this._req = null;
         this._recursionDepth = 0;
-        this._isMain = true;
 
         options.maxDepth = options.maxDepth || 10;
         // Private state, shared with child instances
@@ -64,7 +62,7 @@ function RESTBase (options, req) {
 // the request header, if defined
 RESTBase.prototype.setRequestId = function(req) {
 
-    if (this._isMain) {
+    if (!this._parent) {
         if (req && !(req.headers && req.headers['x-request-id'])) {
             rbUtil.requestId(req);
         }
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -49,9 +49,22 @@ function RESTBase (options, req) {
         this._priv = {
             options: options,
             router: options.router,
+            reqId: req ? req.headers && req.headers['x-request-id'] || rbUtil.generateRequestId() : null
         };
-
     }
+
+    if (req) {
+        req.headers = req.headers || {};
+        if (req.headers['x-request-id']) {
+            // the request has got the request id header set, use that
+            this._priv.reqId = req.headers['x-request-id'];
+        } else if (!this._priv.reqId) {
+            // no request id set, generate it
+            this._priv.reqId = rbUtil.generateRequestId();
+        }
+        req.headers['x-request-id'] = this._priv.reqId;
+    }
+
 }
 
 // Make a child instance
@@ -180,9 +193,6 @@ RESTBase.prototype.request = function (req) {
 // Internal request handler
 RESTBase.prototype._request = function (req) {
     var self = this;
-
-    req.headers = req.headers || {};
-    req.headers['x-request-id'] = req.headers['x-request-id'] || rbUtil.generateRequestId();
 
     // Special handling for https? requests
     if (req.uri.constructor === String && /^https?:\/\//.test(req.uri)) {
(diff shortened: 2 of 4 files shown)


d00rman:
Question: should we attach the same header to the response as well?

gwicke:
> Question: should we attach the same header to the response as well?

Yes, I think so. It's not confidential and lets us find related information from an error report. The `instance` property would be ideal for this, even if it formally expects a URI.

lib/restbase.js
```diff
@@ -143,6 +143,9 @@ RESTBase.prototype.request = function (req) {
 RESTBase.prototype._request = function (req) {
     var self = this;
 
+    req.headers = req.headers || {};
+    req.headers['x-request-id'] = req.headers['x-request-id'] || rbUtil.generateRequestId();
```
gwicke: I don't think this will work properly, as we don't generally forward headers internally. I think it would make more sense to create the request id when creating the restbase child instance, and then attach it to all outgoing requests / use it for logging.

d00rman: We leave them as-is when we [clone the request](https://github.com/wikimedia/restbase/blob/master/lib/restbase.js#L22). That's what this PR is based on :)

gwicke: Yeah, but we perform all kinds of internal requests, which aren't always forwarding the headers. That means that different parts of the internal request processing would get their own request ids.

d00rman: Good point. Will amend the PR

lib/restbase.js
```diff
@@ -47,10 +47,23 @@ function RESTBase (options, req) {
...
+    }
 
+    if (req) {
+        req.headers = req.headers || {};
+        if (req.headers['x-request-id']) {
+            // the request has got the request id header set, use that
+            this._priv.reqId = req.headers['x-request-id'];
+        } else if (!this._priv.reqId) {
+            // no request id set, generate it
+            this._priv.reqId = rbUtil.generateRequestId();
```
gwicke: IMHO this should be handled in the constructor.

lib/restbase.js
```diff
@@ -47,10 +47,23 @@ function RESTBase (options, req) {
...
+        }
+        req.headers['x-request-id'] = this._priv.reqId;
```
gwicke: I think the main situation where we need to make sure to set the header is for outgoing requests. In server.js we should have access to the child restbase instance. We could add an accessor (getRequestId or the like) that returns `this._priv.reqId` and use that for logging from server.js.

d00rman:
@gwicke Now the request ID is reliably and consistently used throughout the life of an incoming request, and is mirrored to any child requests made. Also, the request ID is logged every time the request itself is logged.

d00rman:
@gwicke re-ping

lib/rbUtil.js
```diff
@@ -130,6 +130,35 @@ rbUtil.isTimeUUID = function (s) {
...
 };
 
+
+var reqIdBuf = new Buffer(16);
+
+/**
+ * Generates a new request ID
+ */
+rbUtil.generateRequestId = function() {
+    uuid.v4(null, reqIdBuf);
```
gwicke: Actually, a v1 uuid would be nicer, as it also encodes the time. Can change that later though.

lib/restbase.js
```diff
@@ -33,6 +33,8 @@ function RESTBase (options, req) {
         this.log = par.log;
         this._recursionDepth = par._recursionDepth + 1;
         this._priv = par._priv;
+        this._isMain = false;
```
gwicke: This is implicit in `._parent` being null, so is not needed separately.

lib/restbase.js
```diff
@@ -49,10 +52,40 @@ function RESTBase (options, req) {
...
+    // set the request id, if appropriate
+    this.setRequestId(req);
+
 }
 
+// sets the request id for this instance and adds it to
+// the request header, if defined
+RESTBase.prototype.setRequestId = function(req) {
+
+    if (this._isMain) {
```
gwicke: I don't think this should ever trigger, as we reuse a single restbase instance as the root for all requests. The per-request stuff is done with makeChild.

lib/restbase.js
```diff
@@ -33,6 +33,8 @@ function RESTBase (options, req) {
         this.log = par.log;
         this._recursionDepth = par._recursionDepth + 1;
         this._priv = par._priv;
+        this._isMain = false;
+        this._reqId = par._reqId;
```
gwicke: This would be a good opportunity to generate one if it is null.

lib/restbase.js
```diff
@@ -41,6 +43,7 @@ function RESTBase (options, req) {
         this._parent = null;
         this._req = null;
         this._recursionDepth = 0;
+        this._isMain = true;
```
gwicke: Should set a _reqid to null here as well, in the same order as above to make sure that the object shape is the same.

lib/restbase.js
```diff
@@ -49,10 +52,40 @@ function RESTBase (options, req) {
...
+        if (req.headers['x-request-id']) {
+            // the request has got the request id header set, use that
+            this._reqId = req.headers['x-request-id'];
+        } else if (!this._reqId) {
+            // no request id set, generate it
+            this._reqId = rbUtil.generateRequestId();
+        }
+        req.headers['x-request-id'] = this._reqId;
+    } else if(!this._reqId) {
+        this._reqId = rbUtil.generateRequestId();
```
gwicke: This branch can be pulled out I think.

lib/restbase.js
```diff
@@ -95,7 +165,8 @@ RESTBase.prototype.defaultListingHandler = function(value, restbase, req) {
 // Special handling for external web requests
 RESTBase.prototype.defaultWebRequestHandler = function(restbase, req) {
     this.log('trace/webrequest', {
-        req: req
+        req: req,
+        request_id: req.headers['x-request-id']
```
gwicke: req.headers are not always defined.

lib/restbase.js
d00rman: Right, but the problem is that when we create the main object, `req` is not set, and that child instances can have one single request ID, the parent cannot, precisely because it's used for all requests. Hence, if this is the main RB object, and we have a `req`, set only the header, otherwise set the reqid of the RB object (which is a child of the main obj), so that any sub-requests may inherit the value regardless of which request is passed to them.

lib/server.js
```diff
@@ -29,6 +29,9 @@ function handleResponse (opts, req, resp, response) {
         response.headers['Access-Control-Allow-Methods'] = 'GET';
         response.headers['Access-Control-Allow-Headers'] = 'accept, content-type';
 
+        // propagate the request id header
+        response.headers['X-Request-Id'] = req.headers['x-request-id'];
```
gwicke: Should we use lowercase for consistency?

lib/server.js
```diff
@@ -137,13 +140,17 @@ function handleResponse (opts, req, resp, response) {
 function handleRequest (opts, req, resp) {
     var newReq;
 
+    // set the request ID early on for external requests
+    rbUtil.requestId(req);
```
gwicke: Could you do this only for actual external requests?

lib/restbase.js
d00rman: at this point it is, `setRequestId()` creates it if it doesn't exist

lib/server.js
d00rman: Hm, the http module lowercases incoming header names, but outgoing ones should be proper, so I'd leave it like that.

lib/restbase.js
gwicke: My point is that we don't need that internally.

lib/server.js
gwicke: You are right, there is no chance that any internal users is trying to read that header at this point.

lib/restbase.js
```diff
@@ -48,12 +49,42 @@ function RESTBase (options, req) {
         // Private state, shared with child instances
         this._priv = {
             options: options,
-            router: options.router,
+            router: options.router
         };
```
gwicke: Define _reqid for the root as well.

lib/restbase.js
gwicke: How would the root ever have a request associated with it?

lib/server.js
d00rman: it's server.js , `handleRequest()` is called only form the outside. The request id needs to be set as it's used for logging.

gwicke: Is there actually any logging before calling restbase.request()?

lib/restbase.js
```diff
@@ -32,12 +32,14 @@ function RESTBase (options, req) {
         var par = this._parent = options;
         this.log = par.log;
         this.metrics = par.metrics;
+        this.reqId = req && req.headers && req.headers['x-request-id'] || req && req.reqId || par.reqId || rbUtil.generateRequestId();
```
gwicke: Could you break this long line?

lib/server.js
```diff
@@ -141,15 +144,20 @@ function handleResponse (opts, req, resp, response) {
 function handleRequest (opts, req, resp) {
     var newReq;
 
+    // set the request ID early on for external requests
```
gwicke: Could you break this line?

lib/restbase.js
gwicke: `par.reqId` would be the most common case, so it's probably fastest to check that first. Could you also break this long line?

lib/server.js
```diff
@@ -175,7 +183,8 @@ function handleRequest (opts, req, resp) {
             query: urlData.query,
             method: req.method.toLowerCase(),
             headers: req.headers,
-            body: body
+            body: body,
+            reqId: reqId
```
gwicke: I was wondering if we could just set the header. That would simplify the logic in the restbase constructor a bit, and would avoid introducing another req member.

d00rman: This seems to be the cleanest way because then `headers['x-request-id']` is not carried over to all child requests. Instead, it's only set in `RESTBase.reqId`, so it *should* be more performant like this.

gwicke: Well, headers are not cloned deeply, so child requests shouldn't matter. I was mainly thinking about the logic in the restbase constructor, which currently checks for either reqid *or* the header, on each internal request. If we instead set the header once, we could basically do `this.reqId = par.reqId || req.headers && req.hearders['x-request-id'];`

d00rman: Soooo we're back to the previous version in which all of the headers had the request ID set?

gwicke: No, only the incoming request.

gwicke:
@d00rman, lets continue the discussion & maybe tweak things in follow-ups.

PR #218, opened 2015-03-17
Fix copy/paste issue mods/parsoid.yaml

mattflaschen:


d00rman:
hehe, @mattflaschen thnx for noticing :+1:

PR #220, opened 2015-03-26
Conditional re-renders

gwicke:
- https://phabricator.wikimedia.org/T93779 (only store if content differs)
- https://phabricator.wikimedia.org/T93777 (if-unmodified-since)

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -134,50 +134,40 @@ PSP.saveParsoidResult = function (restbase, req, format, tid, parsoidResp) {
     }
 };
 
-PSP.generateAndSave = function(restbase, req, format, isUpdateRequest) {
+// Temporary work-around for Parsoid issue
+// https://phabricator.wikimedia.org/T93715
+function normalizeHtml(html) {
+    return html && html.toString
+        && html.toString().replace(/ about="[^"]+"(?=[\/> ])/g, '');
+}
+function sameHtml(a, b) {
+    return normalizeHtml(a) === normalizeHtml(b);
+}
+
+PSP.generateAndSave = function(restbase, req, format, currentContentRes) {
     var self = this;
     var tid = uuid.v1();
     // Try to generate HTML on the fly by calling Parsoid
     var rp = req.params;
     var storageRequest = null;
-    if (isUpdateRequest && req.params.revision) {
-        storageRequest = self.getFormat(format, restbase, {
-            uri: req.uri,
-            params: req.params
-        }).catch(function() {
-            // Don't care about errors.
-            return null;
-        });
-    }
 
-    return P.props({
-        parsoid: restbase.get({
-            uri: new URI([rp.domain,'sys','parsoid','pagebundle',
-                         normalizeTitle(rp.title),rp.revision])
-        }),
-        storage: storageRequest
+    return restbase.get({
+        uri: new URI([rp.domain,'sys','parsoid','pagebundle',
+                     normalizeTitle(rp.title),rp.revision])
     })
-    .then(function(responses) {
-        // Temporary work-around for Parsoid issue
-        // https://phabricator.wikimedia.org/T93715
-        function normalizeHtml(html) {
-            return html && html.toString
-                && html.toString().replace(/ about="[^"]+"(?=[\/> ])/g, '');
-        }
-        function sameHtml(a, b) {
-            return normalizeHtml(a) === normalizeHtml(b);
-        }
-        if (responses.storage
-                && sameHtml(responses.parsoid.body[format].body,
-                    responses.storage.body)) {
+    .then(function(res) {
+        if (format === 'html' && currentContentRes
+                && sameHtml(res.body.html.body, currentContentRes.body)) {
             // New render is the same as the previous one, no need to store
             // it.
             //console.log('not saving a new revision!');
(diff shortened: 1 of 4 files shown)


mods/parsoid.js
```diff
@@ -150,54 +189,72 @@ PSP.getRevisionInfo = function(restbase, req) {
...
+                            // Already up to date, nothing to do.
+                            return {
+                                status: 412,
+                                body: {
+                                    type: 'precondition_failed',
+                                    detail: 'The precondition failed'
+                                }
+                            };
+                        }
+                    } catch (e) { console.log(e); } // Ignore errors from date parsing
```
d00rman: If we want to stick to the [spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html), we should actually ignore `If-Unmodified-Since` when the date is invalid.

mods/parsoid.js
```diff
@@ -115,19 +134,39 @@ PSP.saveParsoidResult = function (restbase, req, format, tid, parsoidResp) {
...
-        return self.saveParsoidResult(restbase, req, format, tid, parsoidResp);
+    .then(function(res) {
+        if (format === 'html' && currentContentRes
+                && sameHtml(res.body.html.body, currentContentRes.body)) {
+            // New render is the same as the previous one, no need to store
+            // it.
+            //console.log('not saving a new revision!');
+            restbase.metrics.increment('sys_parsoid_generateAndSave.unchanged_rev_render');
+
+            return self.wrapContentReq(restbase, req, P.resolve(currentContentRes));
```
d00rman: No need to use `wrapContentReq()` - the request on line 154 does that already.

test/features/pagecontent/rerendering.js
```diff
@@ -6,46 +6,150 @@
...
+            return preq.get({
+                uri: server.config.bucketURL + dynamic2
+            });
+        })
+        .then(function (res) {
+            r2tid1 = res.headers.etag;
+            assert.deepEqual(res.status, 200);
+            hasTextContentType(res);
+
+            // Delay a bit to give the async save time to complete
```
d00rman: It's not clear what this comment refers to - there is no use of `P.delay()` here, and the URI is the same that's returned in the block. But the comment is actually right - there should be a delay to account for the save, but there is none.

test/features/pagecontent/rerendering.js
```diff
@@ -6,46 +6,150 @@
...
+            hasTextContentType(res);
+        });
+    });
+
+    it('should render & re-render independent revisions, if-unmodified-since support', function () {
+        return preq.get({
+            uri: server.config.bucketURL + dynamic1,
+            headers: {
+                'cache-control': 'no-cache',
+                'if-unmodified-since': 'Wed Dec 11 2013 16:00:00 GMT-0800 (PST)',
```
d00rman: The date should be in [RFC1123](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3.1), so either PST or GMT-0800 should be there, but not both.

mods/parsoid.js
gwicke: Yeah, that's what the try/catch is about.

mods/parsoid.js
gwicke: That's pretty subtle (and the main save path also re-wraps). I kind of prefer to have the wrapping as late as possible, so it's clear from looking at the code that it happens.

I actually wonder if we should ever expose pagebundle publicly. Right now the only reason to have a separate entry point for it is to get separate stats, but with metrics now being a public property on restbase we could just create custom metrics for that in the module.

test/features/pagecontent/rerendering.js
gwicke: I removed the old delay, but left the comment. In practice the delay does not seem to be needed, as writes are just to memory & really fast. Would you prefer to still have one?

test/features/pagecontent/rerendering.js
gwicke: It still works of course, but ookay ;)

mods/parsoid.js
```diff
@@ -24,6 +24,25 @@ function ParsoidService(options) {
     options = options || {};
     this.parsoidHost = options.parsoidHost
         || 'http://parsoid-lb.eqiad.wikimedia.org';
+    // Set up operations
+    var self = this;
+    this.operations = {
+        getPageBundle: function(restbase, req) {
```
d00rman: I agree with you on not exposing `pagebundle`, it's used only internally in this module anyway.

mods/parsoid.js
```diff
@@ -115,19 +134,39 @@ PSP.saveParsoidResult = function (restbase, req, format, tid, parsoidResp) {
...
     var rp = req.params;
+    var storageRequest = null;
+
     return restbase.get({
         uri: new URI([rp.domain,'sys','parsoid','pagebundle',
                      normalizeTitle(rp.title),rp.revision])
     })
-    .then(function(parsoidResp) {
-        return self.saveParsoidResult(restbase, req, format, tid, parsoidResp);
+    .then(function(res) {
```
d00rman: If we get rid of `/sys/parsoid/pagebundle/` the `get()` above needs to be turned into a method call

PR #222, opened 2015-03-30
Consistently use previous renders for html2wt conversion

gwicke:
Previously we didn't properly handle template re-renders in the middle of an
edit. The HTML and data-parsoid passed to Parsoid would always use the latest
render, which was potentially out of sync with the original render. This could
lead to dirty diffs.

This patch addresses this by:

- accepting an 'If-Match' header in transform methods, which is expected to
  contain the Etag (timeuuid) returned by the original HTML / data-parsoid
- Also, as a convenient way to support this quickly in current VE, injecting
  the timeuuid in a meta tag in the head section of the saved HTML & plucking
  that back out if no If-Match header was provided.

Bug: https://phabricator.wikimedia.org/T94422

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -138,7 +138,8 @@ PSP.saveParsoidResult = function (restbase, req, format, tid, parsoidResp) {
 // https://phabricator.wikimedia.org/T93715
 function normalizeHtml(html) {
     return html && html.toString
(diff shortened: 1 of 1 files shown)


gwicke:
This should also help with https://phabricator.wikimedia.org/T93086.

PR #226, opened 2015-04-03
Add utility script to generate a WMF restbase domain config

gwicke:


maintenance/generate-wikimedia-domain-config.js
```diff
@@ -0,0 +1,59 @@
...
+        wikisource: [],
+        wikibooks: [],
+        wikinews: [],
+        wikiversity: [],
+        wikivoyage: [],
+        '*': []
+    };
+    Object.keys(sm).forEach(function(k) {
+        var lang = sm[k];
+        if (lang.site) {
```
d00rman: Should we skip sites with the `closed` property defined?

maintenance/generate-wikimedia-domain-config.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+"use strict";
+var P = require('bluebird');
+
+var fs = require("fs"),
+	writeFile = P.promisify(fs.writeFile),
+	request = P.promisify(require('request')),
+	downloadUrl = "https://en.wikipedia.org/w/api.php?action=sitematrix&format=json",
+	filename = "sitematrix.json";
```
d00rman: Let's get rid of this chaining, pleaseee :P Apart from the fact that we do not use it, it looks rather ugly (to me).

maintenance/generate-wikimedia-domain-config.js
```diff
@@ -0,0 +1,59 @@
...
+                    }
+                }
+            });
+        }
+    });
+
+    Object.keys(projects).forEach(function(name) {
+        console.log('\n    # ' + name);
+        projects[name].forEach(function(domain) {
+            console.log('    /{domain:' + domain + '}: *wp/default/1.0.0');
```
d00rman: How about to have a config file with a placeholder which is replaced by the list instead of having it printed on stdout? (btw, using `console.log` makes the `writeFile` declaration on line 11 unused)

maintenance/generate-wikimedia-domain-config.js
gwicke: We do, see below.

maintenance/generate-wikimedia-domain-config.js
gwicke: That's more complex, and doesn't easily support editing after generation.

maintenance/generate-wikimedia-domain-config.js
d00rman: Duh, /me hides

PR #227, opened 2015-04-08
Bump service-runner version to 0.1.6

d00rman:
Bumping the version as it contains the switch from TXStatsD to statsite.

Bug: [T94053](https://phabricator.wikimedia.org/T94053)


d00rman:
Note, the switch to statsite is happening 2015-04-09 at 10:00 UTC (that's tomorrow)

PR #229, opened 2015-04-10
Simple service proxy & Key-Value cache module

gwicke:
Bug: https://phabricator.wikimedia.org/T95551

### DIFF
--- a/mods/key_rev_value.js
+++ b/mods/key_rev_value.js
@@ -50,7 +50,7 @@ KRVBucket.prototype.makeSchema = function (opts) {
             'content-sha256': 'blob',
             // redirect
             'content-location': 'string',
-            tags: 'set<string>'
+            tags: 'set<string>',
             //headers: 'map<string,string>'
         },
         index: [
--- a/mods/simple_service.js
+++ b/mods/simple_service.js
@@ -21,7 +21,7 @@ function SimpleService(options) {
     this.exports = this.processSpec(this.spec);
 }
 
-var headerWhitelist = ['accept', 'accept-language', 'accept-language'];
+var headerWhitelist = ['accept', 'accept-language'];
 function filterHeaders(headers) {
     var res = {};
     headerWhitelist.forEach(function(name) {
--- a/mods/simple_service.js
+++ b/mods/simple_service.js
@@ -1,0 +1,105 @@
+'use strict';
+
+/*
+ * Simple service proxy & cache/storage module
+ */
+
+var P = require('bluebird');
+var URI = require('swagger-router').URI;
+
+// Store titles as MediaWiki db keys
+function normalizeTitle (title) {
+    return title.replace(/ /g, '_');
+}
+
+function SimpleService(options) {
+    options = options || {};
+    this.spec = {
+        paths: options.paths
+    };
+
+    this.exports = this.processSpec(this.spec);
+}
+
+SimpleService.prototype.processSpec = function(spec) {
+    var self = this;
+    var operations = {};
+    var resources = [];
+
+    Object.keys(spec.paths).forEach(function(path) {
+        var pathObj = spec.paths[path];
+        Object.keys(pathObj).forEach(function(method) {
+            var conf = pathObj[method];
+            conf.operationId = method + '_' + path;
+            var storageUriTemplate;
+            if (conf.storage) {
+                if (!conf.storage.bucket_request.uri) {
+                    throw new Error('Broken config: expected storage.bucket_request.uri for ' + path);
+                }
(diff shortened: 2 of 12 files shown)


mods/key_rev_value.js
```diff
@@ -50,7 +50,7 @@ KRVBucket.prototype.makeSchema = function (opts) {
             'content-sha256': 'blob',
             // redirect
             'content-location': 'string',
-            tags: 'set<string>'
+            tags: 'set<string>',
```
d00rman: trailing comma?

mods/simple_service.js
```diff
@@ -0,0 +1,126 @@
...
+function SimpleService(options) {
+    options = options || {};
+    this.spec = {
+        paths: options.paths
+    };
+
+    this.exports = this.processSpec(this.spec);
+}
+
+var headerWhitelist = ['accept', 'accept-language', 'accept-language'];
```
d00rman: two `accept-language`

mods/simple_service.js
```diff
@@ -0,0 +1,126 @@
...
+                        })
+                        .then(function(storeRes) {
+                            res.headers.etag = storeRes.headers.etag;
+                            return res;
+                        });
+                    });
+                }
+
+                if (conf.storage) {
+                    if (conf.storage['no-cache_refresh']
```
d00rman: `no-cache_refresh` is a rather awkward syntax to me. How about `nocache_refresh` ?

package.json
```diff
@@ -42,8 +42,8 @@
     "preq": "~0.3.6",
     "request": "^2.44.0",
     "restbase-mod-table-cassandra": "^0.5.3",
-    "service-runner": "^0.1.6",
-    "swagger-router": "^0.0.4",
+    "service-runner": "^0.1.7",
```
d00rman: we're at `0.1.8` now ;)

test/features/simple_service.js
```diff
@@ -0,0 +1,54 @@
...
+        })
+        .then(function (res) {
+            // Since this is a dynamic page which should render the same each
+            // time, the tid should not change.
+            tid2 = res.headers.etag;
+            assert.notDeepEqual(tid2, tid1);
+            assert.notDeepEqual(tid2, undefined);
+            hasTextContentType(res);
+        });
+    });
```
d00rman: IMO It'd be good to add a second test retrieving the page from cache and checking that no remote requests have been made.

d00rman:
LGTM, small nitpicks in-lined.

In general, I am not happy about the fact that we will still have to add a new module (or path in the config) for each proxied service. It would be easier, IMO, if there could be a way to specify which URL to call directly in the public spec, instead of relying on the cache module to provide that. So, in `specs/test.yaml`:

```yaml
paths:
  /{module:service}/test/{title}{/rev}:
    get:
      x-backend-request:
        uri: /{domain}/sys/testservice/test/{title}{/revision}
        headers:
          x-uri: http://en.wikipedia.org/wiki/{title}
```
Then, in the module def in `config.yaml`, instead of having the path `/test/{key}`, we could have `/{component}/{key}`. `x-uri` could be picked up by the cache module.

That would allow us to have one path per cache type in the config. That would also warrant on-the-fly bucket creation, though.

gwicke:
@d00rman, so far we have kept all local configs like backend service ips etc out of the public spec. The only rewriting that happens there is the one to the internal service end point. I think this makes sense to keep it that way. Advantages I see are:

- No need to puppetize the public spec
- Can change the public spec without breaking code that relies on the service internally; can expose the internal service in multiple ways / with different frontend entry points
- Can route a public end point to a different module at any time, for example if the service develops special needs that don't fit into the regular cache implementation

package.json
gwicke: Okay, that's really a separate issue, but I bumped it in this PR nevertheless.

mods/simple_service.js
gwicke: The `no-cache` part reflects the header. I agree that it doesn't look pretty, but at least it's consistent with our policy of using _ as a delimiter between components.

mods/key_rev_value.js
gwicke: Yup, slowly moving towards https://github.com/felixge/node-style-guide

gwicke:
Rebased onto master.

gwicke:
Removed the WIP from the first patch in the series.

mods/simple_service.js
d00rman: I understand the rationale, but I still feel it's awkward and error-prone.

PR #232, opened 2015-04-23
explicity assign schema version

eevans:
This should be a non-normative change for existing installations as
historically, an undefined version defaulted to 1.

This is made necessary by https://github.com/wikimedia/restbase-mod-table-cassandra/pull/95.

Bug: T96612

### DIFF
--- a/mods/key_rev_value.js
+++ b/mods/key_rev_value.js
@@ -57,7 +57,8 @@ KRVBucket.prototype.makeSchema = function (opts) {
             { attribute: 'key', type: 'hash' },
             { attribute: 'rev', type: 'range', order: 'desc' },
             { attribute: 'tid', type: 'range', order: 'desc' }
-        ]
+        ],
+        version: 1
     };
 };
 
--- a/mods/key_value.js
+++ b/mods/key_value.js
@@ -52,7 +52,8 @@ KVBucket.prototype.makeSchema = function (opts) {
                 { attribute: 'key', type: 'hash' },
                 { attribute: 'latestTid', type: 'static' },
                 { attribute: 'tid', type: 'range', order: 'desc' }
-            ]
+            ],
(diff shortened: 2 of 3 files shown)


gwicke:
Will this be compatible with existing schemas?

d00rman:
> Will this be compatible with existing schemas?

wikimedia/restbase-mod-table-cassandra#95 defines the schema versions currently in prod as 0, so there should be no problem.

eevans:
@gwicke wrote:
> Will this be compatible with existing schemas?

It *is* compatible with the persisted schemas of any extant RESTBase cluster; It *is not* compatible with wikimedia/restbase-mod-table-cassandra prior to the changes introduced in wikimedia/restbase-mod-table-cassandra/pull/95

gwicke:
> wikimedia/restbase-mod-table-cassandra#95 defines the schema versions currently in prod as 0, so there should be no problem.

However, the stored schema versions might already be at version 1 (old default), which could be a problem if we missed to set an explicit version 1 on any schema in restbase.

eevans:
> However, the stored schema versions might already be at version 1 (old default), which could be a problem if we missed to set an explicit version 1 on any schema in restbase.

The stored schema versions should all already be at version 1.  Missing one would result in a pretty obvious exception after applying wikimedia/restbase-mod-table-cassandra/pull/95 because the new default is 0 (making the schemas differ), but the (no-op) migration wouldn't succeed without a higher version).  If such an exception were to occur, the solution would simply be to explicity set the version and restart. 

Is there any danger of us missing one?

eevans:
This is no longer necessary in light of https://github.com/eevans/restbase-mod-table-cassandra/commit/3b6b6dd8573517bb398424ae6f46ddabca844e57 (closing).

PR #238, opened 2015-05-01
Support rewriting the base path for host-based access & proxying

gwicke:
We now expose the docs at /api/rest_v1/, so need the ability to set the right
base path for host-based and rewritten requests.


eevans:
LGTM

PR #241, opened 2015-05-06
configure revision retention on parsoid.html/data-parsoid

eevans:
Parametizes `version` and `revisionRetentionPolicy` for key_rev_value clients.

Creates a schema version that is the sum of a key_rev_value schema major
version, and the client-supplied minor.  Only migrates schemas to this
scheme when clients explicitly pass a version.

Bug: T94524


gwicke:
The test failure for iojs looks unrelated (npm install failure).

PR #245, opened 2015-05-11
Add paging support

hardikj:
Note - This pull depends upon https://github.com/wikimedia/restbase-mod-table-cassandra/pull/97


config.example.yaml
```diff
@@ -146,6 +146,7 @@ services:
     conf: 
       port: 7231
       spec: *spec
+      saltKey: secret
```
gwicke: `saltKey` is easy to misunderstand as something connected to salt, the config management system. Perhaps call it just 'salt'?

lib/restbase.js
```diff
@@ -79,6 +80,24 @@ RESTBase.prototype.makeChild = function(req) {
...
+                        self._priv.options.saltKey.toString());
+    return newToken;
+}
+
+// Decode signed token and decode the orignal token
+// @param {string} JWT token
+// @return {string} original token
+rbUtil.decodeToken = function(token) {
+    console.log(self._priv.options);
+    return jwt.verify(token, self._priv.options.saltKey)
```
gwicke: This will throw an exception on failure, which we should report with HTTP status 400 (by throwing an appropriate `HTTPError`). It might make sense to catch & re-wrap the error right here, so that we don't have to do that at each call site.

mods/page_revisions.js
```diff
@@ -133,25 +133,40 @@ PRS.prototype.listTitles = function(restbase, req, options) {
...
         return {
             status: 200,
-            body: {
-                items: items
+            body : {
+                items: items,
+                next: next
             }
-        };
+        };;
```
gwicke: Duplicate semicolon.

mods/page_revisions.js
```diff
@@ -371,7 +413,7 @@ PRS.prototype.getRevision = function(restbase, req) {
     // sanity check
     if (!/^[0-9]+$/.test(rp.revision)) {
         throw new rbUtil.HTTPError({
-            status: 400,
+            status: 500,
```
gwicke: Huh?

mods/parsoid.js
```diff
@@ -301,10 +301,29 @@ PSP.getFormat = function (format, restbase, req) {
 PSP.listRevisions = function (format, restbase, req) {
     var self = this;
     var rp = req.params;
-    return restbase.get({
+    var revReq = {
         uri: new URI([rp.domain, 'sys', 'key_rev_value', 'parsoid.' + format,
                      normalizeTitle(rp.title), ''])
-    });
+        generator: 'allpages',
```
gwicke: This doesn't seem to make sense.

mods/page_revisions.js
```diff
@@ -355,11 +388,20 @@ PRS.prototype.listRevisions = function(restbase, req) {
             var article = pages[pageId];
             items.push(article.revisions[0].revid);
         });
+        var next;
+        if (res.body.next) {
+            next = { 
+                _links: { 
+                    next: { "href": "?next="+restbase.encodeToken(res.body.next.allpages.gapcontinue); }
```
gwicke: Perhaps call the GET parameter `page` ?

mods/page_revisions.js
```diff
@@ -133,25 +133,40 @@ PRS.prototype.listTitles = function(restbase, req, options) {
             gaplimit: 500,
             prop: 'revisions',
             format: 'json',
-            // gapcontinue: rp.next
+            gapcontinue: ''
         }
     };
 
+    if (req.next) {
```
gwicke: `req.query.next`. Please test your code.

mods/parsoid.yaml
```diff
@@ -15,7 +15,7 @@ paths:
     post:
       summary: Retrieve a JSON bundle containing html and data-parsoid
       operationId: getPageBundle
-  
+
```
gwicke: The new GET parameters need to be documented here & in the v1 spec.

gwicke:
This needs tests.

config.example.yaml
```diff
@@ -146,6 +146,8 @@ services:
     conf: 
       port: 7231
       spec: *spec
+      salt: secret
+      default_page_size: 1000
```
gwicke: 500 or 250 might be more sensible here.

mods/page_revisions.js
```diff
@@ -133,23 +133,38 @@ PRS.prototype.listTitles = function(restbase, req, options) {
...
+            };
+        }
+
         return {
             status: 200,
-            body: {
-                items: items
+            body : {
+                items: items,
+                next: next
```
gwicke: This should be _links, following http://stateless.co/hal_specification.html.

mods/page_revisions.js
```diff
@@ -306,16 +325,20 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
...
+    var revisionRequest = {
+        table: this.tableName,
+        attributes: {
+            title: normalizeTitle(rp.title)
+        },
+        proj: ['rev'],
+        limit: restbase.rb_config.default_page_size
+    };
+    if (req.query.page) {
+        revisionRequest.next = restbase.decodeToken(req.query.page);
```
gwicke: _links

mods/page_revisions.js
```diff
@@ -328,6 +351,13 @@ PRS.prototype.listTitleRevisions = function(restbase, req) {
                 lastRev = row.rev;
             }
         });
+        if (res.body.next) {
+            res.body.next = {
```
gwicke: res.body._links = {..}

config.example.yaml
hardikj: I used 1000 here because of this limit - https://github.com/wikimedia/restbase/pull/245/files#diff-159808c267853355eee587915c06aae6L317

Right now default_page_size is been used by those end points which make request to cassandra backend, do you want me to use it for other ends points as well?

gwicke:
Looks good to me overall, only a couple of inline notes.

gwicke:
This needs a rebase.

gwicke:
Closed in favor of #257, which has some additional cleanup.

PR #247, opened 2015-05-16
Add graphoid API spec, and hook it up in the example config

gwicke:


config.example.yaml
```diff
@@ -101,6 +101,18 @@ templates:
             options:
               apiURI: http://{domain}/w/api.php
 
+      /{module:graphoid}:
+        x-modules:
+          - name: simple_service
```
nyurik: ??

specs/mediawiki_graphoid_v1.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+swagger: 2.0
+
+paths:
+  /{module:page}/graph/png/{title}/{revision}/{graph_id}:
+    get:
+      tags:
+        - Page content
+      description: >
+        Retrieve PNG renders of graphs embedded in specific revisions of a
```
nyurik: Renders graph images embedded in specific revisions of a

specs/mediawiki_graphoid_v1.yaml
```diff
@@ -0,0 +1,51 @@
...
+          type: integer
+          required: true
+        - name: graph_id
+          in: path
+          description: The graph ID, consisting of a hash and the suffix `.png`.
+          type: string
+          required: true
+      responses:
+        '200':
+          description: The PNG render of the requested graph.
```
nyurik: Don't specify PNG - it could switch/add SVG at any point

specs/mediawiki_graphoid_v1.yaml
```diff
@@ -0,0 +1,51 @@
...
+        '404':
+          description: Unknown page title or revision
+          schema:
+            $ref: '#/definitions/problem'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-backend-request:
+        uri: /{domain}/sys/graphoid/v1/png/{title}/{revision}/{graph_id}
```
nyurik: /{format}/  instead of /png/   also, graph_id includes .format

config.example.yaml
gwicke: This sets up the internal backend. We could also easily enable storage here; see https://github.com/wikimedia/restbase/blob/master/config.test.yaml#L121-L126 for an example.

specs/mediawiki_graphoid_v1.yaml
gwicke: That would be a different end point at `/svg/`.

specs/mediawiki_graphoid_v1.yaml
gwicke: The suffix is part of the ID for our purposes.

specs/mediawiki_v1_graphoid.yaml
```diff
@@ -0,0 +1,52 @@
...
+          type: string
+          required: true
+        - name: revision
+          in: path
+          description: The page revision
+          type: integer
+          required: true
+        - name: graph_id
+          in: path
+          description: The graph ID, consisting of a hash and the suffix `.png`.
```
d00rman: *optionally containing the suffix .png* - .png is not required

d00rman:
LGTM :+1:

nyurik:
I think you should replace the word "PNG" with "image" - we should at some point switch to SVGs.

d00rman:
> I think you should replace the word "PNG" with "image" - we should at some point switch to SVGs.

Not really. When SVGs become available, we will simply add another route which will address them exclusively, just like the existing one addresses only PNGs.

specs/mediawiki_v1_graphoid.yaml
nyurik: i have removed that ability - there should be only one, normalized URL

nyurik:
Why not simply make the /png/ into a parameter /{format}/, just like every other parameter? What's the point of stressing that this is a PNG?

gwicke:
@nyurik, the goal of swagger specs is to document the input-output behavior of the API. This includes the mime type of the response, which in this case is an `image/png`. It supports returning several different types per entry point as well, but it's harder to document the relationship between `{format}` and return value, and in any case at this point PNG is all that's supported. In any case, we can discuss that later as it won't be a breaking change.

PR #248, opened 2015-05-20
Updated param doc

nyurik:
Change-Id: I51dd93357c8b0370b3482ba60c3cf5fa6a1af217

### DIFF
--- a/specs/mediawiki_v1_graphoid.yaml
+++ b/specs/mediawiki_v1_graphoid.yaml
@@ -24,7 +24,7 @@ paths:
           required: true
         - name: revision
           in: path
-          description: The page revision
+          description: The page revision, or 0 if unknown
           type: integer
           required: true
         - name: graph_id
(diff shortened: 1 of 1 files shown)


specs/mediawiki_v1_graphoid.yaml
```diff
@@ -24,7 +24,7 @@ paths:
           required: true
         - name: revision
           in: path
-          description: The page revision
+          description: The page revision, or 0 if unknown
```
gwicke: Would ', or 0 for the latest' be more accurate?

nyurik: sure, either-or

gwicke: I should have said precise perhaps. It's good to be explicit about which revision is returned when 0 is passed in.

d00rman:
If the revision is optional, IMHO the route should be `/page/png/{title}{/revision}/{id}`

nyurik:
I would rather not - revision is preferred, plus the URL is generated by
applying real values to the format string - so it will always be in the
same format.

PR #249, opened 2015-06-02
Minor fix: Parse the body to JSON only if it's defined

d00rman:
When the application/json header is received from a client, the request's body is parsed to JSON. However, this should be done only if the body contains a string to be parsed.


d00rman:
cc @gwicke @eevans

PR #251, opened 2015-06-09
Page save module

d00rman:
This PR creates a new API end-point which allows users to save a page, in wikitext or HTML format.

Also:
- add the `/edit` endpoint to the action module
- add MW API to RB error translation
- small clean-ups

Bug: [T101501](https://phabricator.wikimedia.org/T101501)

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -157,25 +157,6 @@ util.inherits(HTTPError, Error);
 
 rbUtil.HTTPError = HTTPError;
 
-rbUtil.httpErrors = {
-    notFound: function(description) {
-        return new HTTPError({
-            status: 404,
-            type: 'notfound',
-            title: 'Not found',
-            description: description
-        });
-    },
-    server: function(description) {
-        return new HTTPError({
-            status: 500,
-            type: 'server',
-            title: 'Server error',
-            description: description
-        });
-    }
-};
-
 
 // Create an etag value of the form
 // "<revision>/<tid>"
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -180,11 +180,18 @@ rbUtil.parseETag = function(etag) {
     }
 };
 
+
+/***
+ * MediaWiki-specific functions
+ * TODO: Move them out in a separate file
+ ***/
+
 // Store titles as MediaWiki db keys
 // @param {string} title
 // @return {string} normalised title
 rbUtil.normalizeTitle = function(title) {
     return title.replace(/ /g, '_');
 };
+
 
 module.exports = rbUtil;
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -199,4 +199,11 @@ rbUtil.parseETag = function(etag) {
     }
 };
 
+// Store titles as MediaWiki db keys
+// @param {string} title
+// @return {string} normalised title
+rbUtil.normalizeTitle = function(title) {
+    return title.replace(/ /g, '_');
+};
+
 module.exports = rbUtil;
--- a/mods/action.js
+++ b/mods/action.js
@@ -15,10 +15,114 @@ ActionService.prototype.apiURI = function(domain) {
     return this.apiURITemplate.replace(/\{domain\}/, domain);
 };
 
+function apiError(apiErr) {
+    var ret;
+    apiErr = apiErr || {};
+    ret = {
+        status: 500,
+        body: {
+            type: 'server_error',
+            title: apiErr.code || 'MW API Error',
+            description: apiErr.info || 'Unknown MW API error'
+        }
+    };
+    if(!apiErr.code) {
+        return new rbUtil.HTTPError(ret);
+    }
+    switch(apiErr.code) {
+        /* 400 - bad request */
+        case 'articleexists':
+        case 'badformat':
+        case 'badmd5':
+        case 'badtoken':
+        case 'invalidparammix':
+        case 'invalidsection':
+        case 'invalidtitle':
+        case 'invaliduser':
+        case 'missingparam':
+        case 'missingtitle':
+        case 'nosuchpageid':
+        case 'nosuchrcid':
+        case 'nosuchrevid':
+        case 'nosuchsection':
+        case 'nosuchuser':
+        case 'notext':
+        case 'notitle':
+        case 'pagecannotexist':
+        case 'revwrongpage':
+            ret.status = 400;
+            ret.body.type = 'invalid_request';
+            break;
+        /* 401 - unauthorised */
+        case 'cantcreate-anon':
+        case 'confirmemail':
+        case 'noedit-anon':
+        case 'noimageredirect-anon':
+        case 'protectedpage':
+        case 'readapidenied':
+            ret.status = 401;
+            ret.body.type = 'unauthorized';
+            break;
+        /* 403 - access denied */
+        case 'autoblocked':
+        case 'blocked':
+        case 'cantcreate':
+        case 'customcssjsprotected':
+        case 'customcssprotected':
+        case 'customjsprotected':
+        case 'emptynewsection':
+        case 'emptypage':
+        case 'noedit':
+        case 'noimageredirect':
+        case 'permissiondenied':
+        case 'protectednamespace':
+        case 'protectednamespace-interface':
+        case 'protectedtitle':
+        case 'readonly':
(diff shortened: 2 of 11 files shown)


d00rman:
@gwicke @eevans ready for your eyes ;)

mods/action.js
```diff
@@ -6,47 +6,174 @@
...
+    '400': { status: 400, type: 'invalid_request' },
+    '401': { status: 401, type: 'unauthorized' },
+    '403': { status: 403, type: 'access_denied#edit' },
+    '409': { status: 409, type: 'conflict' },
+    '413': { status: 413, type: 'too_large' },
+    '429': { status: 429, type: 'rate_exceeded' },
+    '500': { status: 500, type: 'server_error' },
+    '501': { status: 501, type: 'not_supported' }
+};
+var errCodes = {
```
gwicke: Nice.

mods/action.js
```diff
@@ -6,47 +6,174 @@
...
+var errDefs = {
+    '400': { status: 400, type: 'invalid_request' },
+    '401': { status: 401, type: 'unauthorized' },
+    '403': { status: 403, type: 'access_denied#edit' },
+    '409': { status: 409, type: 'conflict' },
+    '413': { status: 413, type: 'too_large' },
+    '429': { status: 429, type: 'rate_exceeded' },
+    '500': { status: 500, type: 'server_error' },
+    '501': { status: 501, type: 'not_supported' }
+};
```
gwicke: Minor: Maybe insert a newline between these blocks?

mods/action.js
```diff
@@ -6,47 +6,174 @@
...
+/**
+ * Action module code
+ */
 function ActionService (options) {
     this.apiURITemplate = options.apiURI;
 }
 
 ActionService.prototype.apiURI = function(domain) {
+    // Re-map test domain
+    if(domain === 'en.wikipedia.test.local') { domain = 'en.wikipedia.org'; }
```
gwicke: Actually, we might not need this rewriting any more. The test config just ignores the domain for the API URI.

mods/action.js
d00rman: Yup, forgot it. Done.

mods/action.js
d00rman: Ah right, totally overlooked it. Done.

lib/rbUtil.js
```diff
@@ -199,4 +180,11 @@ rbUtil.parseETag = function(etag) {
     }
 };
 
+// Store titles as MediaWiki db keys
```
gwicke: Adding MediaWiki-specific utilities in rbUtil feels a bit dirty. Having a mediawikiUtils.js file in mods might be a slightly cleaner alternative, although it's not really a module either.

mods/page_save.js
```diff
@@ -0,0 +1,117 @@
+'use strict';
+
+
+/**
+ * page_save module
+ *
+ * Sends the wikitext of a page to the MW API for saving
```
gwicke: I'm wondering if we should integrate this with the Parsoid module. There is already wikitext retrieval functionality in there, as well as some of the revision info logic. The html-based save process will be tightly coupled with wikitext saving as well.

What do you think?

specs/mediawiki/v1/content.yaml
```diff
@@ -445,6 +445,69 @@ paths:
...
+        - name: title
+          in: path
+          description: The page title.
+          type: string
+          required: true
+        - name: revision
+          in: path
+          description: The revision of the page the edit is based on.
+          type: string
+        - name: text
```
gwicke: Maybe call this 'wikitext', so that we could add additional parts later? An example would be a save of wikitext + page metadata.

specs/mediawiki/v1/content.yaml
```diff
@@ -445,6 +445,69 @@ paths:
...
+          in: formData
+          description: The wikitext source of the page to save
+          type: string
+          required: true
+        - name: token
+          in: formData
+          description: The CRSF edit token provided by the MW API
+          type: string
+          required: true
+        - name: summary
```
gwicke: I think in many other contexts this is called 'comment'

specs/mediawiki/v1/content.yaml
```diff
@@ -445,6 +445,69 @@ paths:
...
+        the edit is derived from.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      parameters:
+        - name: title
+          in: path
+          description: The page title.
+          type: string
+          required: true
+        - name: revision
```
gwicke: I'm a bit torn about having this in the path, when the goal is to replace that revision. In many POST APIs the target is the root, with the intention of creating a new child.

lib/rbUtil.js
d00rman: In that case I'd suggest a mwUtils.js in lib/, but tbh placing only one function inside seems ... silly. Maybe move it once we have more stuff to put in there?

mods/page_save.js
d00rman: I was wondering the same thing, but finally opted for another module since it really combines functionalities from the parsoid and action modules. Also, it'll keep the file sizes at bay, which makes it easier to manage. We can always send internal requests around. Thoughts?

specs/mediawiki/v1/content.yaml
d00rman: Makes sense. Will change.

specs/mediawiki/v1/content.yaml
d00rman: Will change to comment, consistency FTW

specs/mediawiki/v1/content.yaml
d00rman: I put it in the URI as the logic (for me) is - *the POST request is based on this revision, which in turn creates a new one*. Had we have PUT requests, then, yes, the revision shouldn't be in the path. It all really depends on the way we look at it, I guess.

d00rman: On second thought, I'm putting the revision in formData. Better :)

specs/mediawiki/v1/content.yaml
d00rman: Done

specs/mediawiki/v1/content.yaml
d00rman: Done.

d00rman:
I changed the revision and put it in the body, renamed summary and text and added an HTML-saving end-point as well.

lib/rbUtil.js
gwicke: Sure, sounds good to me. Lets add a comment about this & group the MW bits together so that we don't forget.

mods/page_save.js
gwicke: > Also, it'll keep the file sizes at bay, which makes it easier to manage. 

Agreed. Although, we could also sub-divide by implementing one API module with several code modules.

With the current split I fear that it might become a bit hard to figure out which module handles what. For example, will wikitext retrieval live in the parsoid module, or the page_save module? The two modules will also depend on each other heavily.

gwicke:
@d00rman, overall I'm happy to merge this. The only thing I'm not so sure about is whether we should expose the save APIs just yet, while we still figure out how those should work wrt conflict detection, timestamps etc. (discussion on https://phabricator.wikimedia.org/T101501).

gwicke:
@d00rman: also, it looks like this needs a rebase.

mods/page_save.js
d00rman: > With the current split I fear that it might become a bit hard to figure out which module handles what.

That's true. That makes me think that perhaps we should have a different split - the parsoid module only deals with interfacing Parsoid, while a 'page_content' module handles retrieval / save of pages, just like we did with the page_revision module - it handles the logic of it all while handing off the necessary calls to the action module.

d00rman:
I placed a comment for `normalizeTitle()` to be moved away. The public end-points are now commented out. I put `/page/wikitext/{title}` into the test spec so that we can still perform the tests for it. The same is not possible for `POST /page/html/{title}` as the router (rightfully) complains about redefining an existing end-point (on that note, since the methods actually differ, maybe we should file/treat that as a bug).

d00rman:
The PR was rebased yesterday, I am not quite sure why the decrease in coverage (it used to report an increase of 2% before the rebase)

PR #252, opened 2015-06-09
Urgent fix: Allow JSON content in POST requests

d00rman:
After parsing a POST request's body, we attempt to parse its contents if the `application/json` content type is present. However, the code tries to parse `req.body`, which is undefined, since `rbUtil.parsePOST()` returns the body, but does not set the `req.body` field. This PR fixes the issue by correctly using the return value.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -218,9 +218,9 @@ function handleRequest (opts, req, resp) {
     // Then process the request
     .then(function(body) {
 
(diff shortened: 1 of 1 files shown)


d00rman:
cc @eevans @gwicke

d00rman:
BTW, we should merge this ASAP as currently POST requests containing JSON data are completely broken.

PR #254, opened 2015-06-10
Section retrieval API

gwicke:
Task: https://phabricator.wikimedia.org/T94890

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -23,7 +23,8 @@ function ParsoidService(options) {
     var self = this;
     this.operations = {
         getPageBundle: function(restbase, req) {
-            return self.wrapContentReq(restbase, req, self.pagebundle(restbase, req));
+            return self.wrapContentReq(restbase, req,
+                    self.pagebundle(restbase, req), 'pagebundle');
         },
         // revision retrieval per format
         getWikitext: self.getFormat.bind(self, 'wikitext'),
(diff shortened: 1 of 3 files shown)


mods/parsoid.js
```diff
@@ -69,17 +70,67 @@ PSP.wrapContentReq = function(restbase, req, promise) {
...
         // if we have reached this point,
         // it means access is not denied
-        return ensureCharsetInContentType(resx[0]);
+        if (format === 'html' && req.query.sections && responses.sectionOffsets) {
+            // Handle section requests
+            var sectionOffsets = responses.sectionOffsets.body;
+            var sections = req.query.sections.split(',').map(function(id) {
+                return id.trim();
+            });
+            var body = cheapBodyInnerHTML(responses.content.body.toString());
```
d00rman: Is it a fact that section offsets provided by Parsoid will always match our cheap body return? Probably worth to put a comment about it, if so.

mods/parsoid.js
```diff
@@ -69,17 +70,67 @@ PSP.wrapContentReq = function(restbase, req, promise) {
...
+
+            return {
+                status: 200,
+                headers: {
+                    etag: responses.content.headers.etag,
+                    'content-type': 'application/json',
+                },
+                body: chunks,
+            };
+        } else {
```
d00rman: I think an `else if` covering the case where `req.query.sections` has been specified, but no section offsets have been returned is necessary here seeing that any errors from that call are ignored (as stated on line 94 above).

d00rman:
A couple of minor comments, otherwise LGTM

mods/parsoid.js
gwicke: They are relative to body.innerHTML.

mods/parsoid.js
gwicke: I handled this differently by no longer catching the section offset request error. That way it's propagated the same way revision metadata 403s are.

PR #256, opened 2015-06-11
Clean up node module dependencies

d00rman:
- moved from `dependencies` to `devDependencies`:
  - bunyan
  - url-template
- removed unused modules:
  - gelf-stream
  - node-txstatsd
  - request
  - yargs
- updated module versions

Note: `maintenance/generate-wikimedia-domain-config.js` now requires
`preq/node_modules/request`

### DIFF
--- a/maintenance/generate-wikimedia-domain-config.js
+++ b/maintenance/generate-wikimedia-domain-config.js
@@ -9,7 +9,7 @@ var P = require('bluebird');
 
 var fs = require("fs"),
 	writeFile = P.promisify(fs.writeFile),
-	request = P.promisify(require('request')),
+	request = P.promisify(require('preq/node_modules/request')),
 	downloadUrl = "https://en.wikipedia.org/w/api.php?action=sitematrix&format=json",
 	filename = "sitematrix.json";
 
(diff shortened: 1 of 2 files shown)


d00rman:
cc @gwicke @eevans

maintenance/generate-wikimedia-domain-config.js
```diff
@@ -9,7 +9,7 @@ var P = require('bluebird');
 
 var fs = require("fs"),
 	writeFile = P.promisify(fs.writeFile),
-	request = P.promisify(require('request')),
+	request = P.promisify(require('preq/node_modules/request')),
```
gwicke: This is quite hacky; I think it'd actually be nicer to just use preq.

gwicke:
LGTM overall; see comment above for a minor nit.

maintenance/generate-wikimedia-domain-config.js
d00rman: I was wondering myself why wasn't `preq` used in the first place. Will change that

gwicke: Did so, and merged the result.

PR #258, opened 2015-06-11
Minor improvements to test suite and If-Unmodified-Since support.

cscott:
Use `Date.parse` instead of `new Date` to avoid an unnecessary
object creation and a potentially confusing implicit conversion
from Date to nuber in the inequality test.

Use an explicit `uuid.parse` so that
https://github.com/gwicke/node-uuid/commit/ed652bb0953fad686d5eb3c65ed5911e2fc43604
isn't strictly necessary (since it hasn't yet been merged upstream).

Use a proper RFC822 date string in the rerendering test.

Use --no-timeouts to ensure the test suite passes even on developers'
slow laptops.

Change-Id: I2fa23a05e49297f3378ee48c45217646e0bda46f

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -274,9 +274,9 @@ PSP.getFormat = function (format, restbase, req) {
         return contentReq.then(function(res) {
                 if (req.headers['if-unmodified-since']) {
                     try {
-                        var jobTime = new Date(req.headers['if-unmodified-since']);
+                        var jobTime = Date.parse(req.headers['if-unmodified-since']);
                         var revInfo = rbUtil.parseETag(res.headers.etag);
-                        if (revInfo && uuid.v1time(revInfo.tid) >= jobTime) {
+                        if (revInfo && uuid.v1time(uuid.parse(revInfo.tid)) >= jobTime) {
                             // Already up to date, nothing to do.
                             return {
                                 status: 412,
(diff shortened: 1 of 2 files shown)


test/mocha.opts
```diff
@@ -1 +1,2 @@
 --recursive
+--no-timeouts
```
gwicke: Hmm. Are you sure you want to disable timeouts altogether?

gwicke:
Looks good to me overall, am just not as fond of the timeout change.

d00rman:
LGTM, but I'm a bit ambivalent on these time-outs. On the one hand, it makes sense to get rid of them, since we blindly set them to 20s regardless of whether we are contacting external services or not. On the other, maybe setting a bit higher thresholds for parsoid/MW API calls and lower them significantly for Cassandra-only calls?

cscott:
I admit that it is laziness: it's easier just to turn timeouts off than to
hunt down all the places in the codebase where the timeouts are "too
short".  If one of you guys wants to do that, feel free.  But it seems
easier just to turn off mocha timeouts by default.  Timeouts imposed by
mocha make (some) sense where the mocha test is just running cpu-bound
code.  But they make much less sense if you are running network operations
as part of your test suite.

>

gwicke:
@cscott, would you mind separating out the timeout change?

cscott:
Done.

PR #260, opened 2015-06-12
Paging

d00rman:
Yet another paging PR :)

Note: need to add the `salt` config directive to the production configuration before deployment


d00rman:
LGTM as-is :+1:

d00rman:
Corresponding configuration patch is [here](https://gerrit.wikimedia.org/r/#/c/218336/)

PR #261, opened 2015-06-12
Fix basePath when the Parsoid domain is the same as the request domain.

cscott:
In testing, I saw this when the parsoid prefix was `localhost` and I
was accessing the swagger docs at http://localhost:7231/localhost/v1/?doc

Change-Id: I89d80d5d4718798a08ec8c9ffb670636a3d1d1c8


cscott:
From IRC:
```
(05:27:05 PM) cscott-free: it has to do with the x-host-basePath field in the spec, i have no idea what that's actually used for.
(05:28:01 PM) gwicke: ah, that's used to support the /api/rest_v1/ prefix
(05:28:30 PM) gwicke: we need to have the docs work both at rest.wikimedia.org/{domain}/v1/... and {domain}/api/rest_v1/...
```

PR #262, opened 2015-06-12
Clean up tid requests

gwicke:
For consistency, require {/revision}{/tid}, rather than interpreting revision
as a tid if it looks like one. This already matches our documentation,
external behavior and tests, so it's really just an internal clean-up.


d00rman:
This PR seems to be more related to section API than to the actual clean-up. Or am I missing something? :)

gwicke:
@d00rman, it depends on that PR (as it touches the same line), but is not directly related.

d00rman:
Look like it needs a rebase

PR #263, opened 2015-06-15
Don't use timeouts since test suite has unpredictable network timing

cscott:
Use --no-timeouts to ensure the test suite passes even on developers' slow laptops.

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,1 +1,2 @@
 --recursive
+--no-timeouts


cscott:
Forked from PR #258.

gwicke:
@d00rman @eevans, what's your take on this?

d00rman:
Well, this does just put on paper what we already do, as in *ah, a time-out occurred, let me run the test again*. As I said for #258, the proper way would be identify all of the cases where the time-out should be extended, and where it should be stricter. But until that is done, I'm fine with this solution too, as in practice it doesn't alter the current situation much.

eevans:
> Well, this does just put on paper what we already do, as in ah, a time-out occurred, let me run the test again.

I agree, and in principal I'm in favor of keeping the tests easy to run (if they annoy, they will not be run), but this sort of feels like defeatism.

I'd be OK with this if a ticket were opened to address a more realistic set of timeouts (with the goal of re-enabling them).

What about a developer option for locally disabling timeouts, (i.e. `touch .disable-test-timeouts`, or `DISABLE_TEST_TIMEOUTS=1 npm test`)?

gwicke:
@eevans: Mocha already supports running without timeouts like this:

`node_modules/mocha/bin/mocha --no-timeouts`

eevans:
@gwicke Understood.  I assumed it was the hassle of routinely invoking the tests this way that resulted in the PR though, and was suggesting a possible compromise/workaround.

cscott:
I wouldn't call it defeatism, I'd call it picking our battles.  It doesn't
seem worth our developer time to tweak timeouts, since they are inherently
arbitrary and should never be triggered anyway.  The only purpose of the
timeouts in the test suite is to save Travis some cpu time (and that's
free, so I don't care) and to give developers earlier earning of a hanging
test. But hung tests are obvious to developers anyway, and are rare.  So
it's just wasted effort to micromanage these.  Set a 60 second timeout
globally if you must.

My opinion.  And this opinion is specific to restbase; if the codebase were
less mature and/or there was a history of legit hang-type bugs I'd have a
different opinion on the effort/reward tradeoff.

cscott:
That said, I've opened https://phabricator.wikimedia.org/T103125 for the benefit of @eevans , in the hope that it gets this PR unstuck.

I don't think it's worth my time to fine-tune timeouts, but everyone's tastes and desires are different.  If someone wants to tackle it, feel free.

gwicke:
Thanks, @cscott !

PR #265, opened 2015-06-23
Replaced node-uuid with pure js uuid library (T94098)

Pchelolo:
This replaces a node-uuid lib with a new cassandra-uuid lib, that's basically a copy-paste from a cassandra driver. New lib doesn't have any binary dependencies, all test pass. New lib is hosted [here] (https://github.com/Pchelolo/uuid) No API changes were made comparing to the original cassandra-driver implementation, bc it would be easier as we use cassandra's lib in other places.

The only concern is rbUtil.generateRequestId - in the previous implementation it returned uuid as a hex number without any '-' symbols. I've supposed something might possibly depend on this so left this behavior unchanged.

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -9,7 +9,7 @@ var P = require('bluebird');
 var util = require('util');
 var url = require('url');
 var Busboy = require('busboy');
-var uuid = require('cassandra-uuid');
+var uuid = require('cassandra-uuid').TimeUuid;
 
 var rbUtil = {};
 
@@ -108,7 +108,7 @@ rbUtil.tidFromDate = function tidFromDate(date) {
         throw new Error('Invalid date');
     }
     // Create a new, deterministic timestamp
-    return uuid.TimeUuid.fromDate(date,
+    return uuid.fromDate(date,
         0,
         new Buffer([0x01, 0x23, 0x45, 0x67, 0x89, 0xab]),
         new Buffer([0x12, 0x34])).toString();
@@ -118,14 +118,14 @@ rbUtil.tidFromDate = function tidFromDate(date) {
  * Check if a string is a valid timeuuid
  */
 rbUtil.isTimeUUID = function (s) {
-    return uuid.TimeUuid.test(s);
+    return uuid.test(s);
 };
 
 /**
  * Generates a new request ID
  */
 rbUtil.generateRequestId = function() {
-    return uuid.Uuid.random().toRawString();
+    return uuid.now().toString();
 };
 
 /*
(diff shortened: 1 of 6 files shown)


lib/rbUtil.js
```diff
@@ -108,31 +108,24 @@ rbUtil.tidFromDate = function tidFromDate(date) {
...
+        new Buffer([0x12, 0x34])).toString();
 };
 
-var uuidRe = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/;
 /**
  * Check if a string is a valid timeuuid
  */
 rbUtil.isTimeUUID = function (s) {
-    return uuidRe.test(s);
+    return uuid.TimeUuid.test(s);
```
d00rman: [cassandra-uuid](https://github.com/Pchelolo/uuid/blob/master/lib/time-uuid.js) does not seem to have a `test()` method

Pchelolo: I've moved it into our library as it seems like a useful method we could reuse somewhere.

d00rman: Actually, let's check whether `isTimeUUID()` is even used. IMO, we've eliminated it. IF so, we can remove it entirely.

Pchelolo: It is used. [here] (https://github.com/wikimedia/restbase/blob/master/mods/key_rev_value.js#L160) and [here] (https://github.com/wikimedia/restbase/blob/master/mods/key_value.js#L148)

d00rman: Ah missed that. Thnx!

gwicke:
> The only concern is rbUtil.generateRequestId - in the previous implementation it returned uuid as a hex number without any '-' symbols. I've supposed something might possibly depend on this so left this behavior unchanged.

I think it's actually fine to change this. The request ID is only used internally to trace the path of a request through the various services. The only real requirement is that it's a unique string. Making it a regular uuid with hyphens should not impact anything.

lib/rbUtil.js
```diff
@@ -108,31 +108,24 @@ rbUtil.tidFromDate = function tidFromDate(date) {
...
-
-var reqIdBuf = new Buffer(16);
-
 /**
  * Generates a new request ID
  */
 rbUtil.generateRequestId = function() {
-    uuid.v4(null, reqIdBuf);
-    return reqIdBuf.toString('hex');
+    return uuid.Uuid.random().toRawString();
```
gwicke: Actually.. should we switch to a v1 uuid as well? It could be useful to have the start time encoded in the request ID.

Pchelolo:
@gwicke fixed that - replaced request id with a properly styled string and use TimeUuid here as well.

d00rman:
LGTM, merging. Thanks @Pchelolo for this! Wrt `cassandra-uuid`, as we also use the same types in our storage back-end, we should monitor changes upstream does to them and keep these two packages compatible.

PR #266, opened 2015-06-24
Added data-parsoid check before sending parsoid request

Pchelolo:
Added a check that parsed metadata was correctly retrieved from the storage before sending a request to parsed transformer [T102117] (https://phabricator.wikimedia.org/T102117).

Check on saving, mentioned in the issue was already fixed by @gwicke [around here] (https://github.com/wikimedia/restbase/commit/7a4bd137b093bf26cb41c914db953eb913727876#diff-22ea565322fb171e4b5370c6f9f46bcbR173)

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -445,12 +445,14 @@ PSP.transformRevision = function (restbase, req, from, to) {
     return this._getOriginalContent(restbase, req, rp.revision, tid)
     .then(function (original) {
         // Check if parsoid metadata is present as it's required by parsoid.
-        if (!original['data-parsoid'].body || !original['data-parsoid'].body.ids) {
+        if (!original['data-parsoid'].body
+                || original['data-parsoid'].body.constructor !== Object
+                || !original['data-parsoid'].body.ids) {
             throw new rbUtil.HTTPError({
                 status: 400,
                 body: {
                     type: 'invalid_request',
-                    description: "Couldn't find parsoid metadata for page " + req.body.title + ' ' + rp.revision
+                    description: 'The page/revision has no associated Parsoid data'
                 }
             });
         }
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -445,12 +445,12 @@ PSP.transformRevision = function (restbase, req, from, to) {
     return this._getOriginalContent(restbase, req, rp.revision, tid)
     .then(function (original) {
         // Check if parsoid metadata is present as it's required by parsoid.
-        if (!original || !original['data-parsoid'] || !original['data-parsoid'].body.ids) {
+        if (!original['data-parsoid'].body || !original['data-parsoid'].body.ids) {
             throw new rbUtil.HTTPError({
                 status: 400,
                 body: {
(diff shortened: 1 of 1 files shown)


mods/parsoid.js
```diff
@@ -444,6 +444,16 @@ PSP.transformRevision = function (restbase, req, from, to) {
 
     return this._getOriginalContent(restbase, req, rp.revision, tid)
     .then(function (original) {
+        // Check if parsoid metadata is present as it's required by parsoid.
+        if (!original || !original['data-parsoid'] || !original['data-parsoid'].body.ids) {
```
d00rman: `original` and `original['data-parsoid']` do not need to be checked for existence. However, before checking for `original['data-parsoid'].body.ids` you should check for `original['data-parsoid'].body` since that is the part that is `undefined` if *data-parsoid* is missing from storage.

mods/parsoid.js
```diff
@@ -444,6 +444,16 @@ PSP.transformRevision = function (restbase, req, from, to) {
...
+            throw new rbUtil.HTTPError({
+                status: 400,
+                body: {
+                    type: 'invalid_request',
+                    description: 'Couldn\'t find parsoid metadata for page ' + req.body.title + ' ' + rp.revision
```
d00rman: Surround the text in double quotes so that the single quote doesn't need escaping.

Pchelolo: Got it.

mods/parsoid.js
```diff
@@ -444,6 +444,16 @@ PSP.transformRevision = function (restbase, req, from, to) {
 
     return this._getOriginalContent(restbase, req, rp.revision, tid)
     .then(function (original) {
+        // Check if parsoid metadata is present as it's required by parsoid.
+        if (!original['data-parsoid'].body || !original['data-parsoid'].body.ids) {
+            throw new rbUtil.HTTPError({
+                status: 400,
+                body: {
+                    type: 'invalid_request',
+                    description: "Couldn't find parsoid metadata for page " + req.body.title + ' ' + rp.revision
```
d00rman: I'd actually put a different message. Something along the lines of "The page/ revision has no associated Parsoid data". Slight rephrase, but it's closer to the 400 / bad request meaning.

mods/parsoid.js
```diff
@@ -444,6 +444,16 @@ PSP.transformRevision = function (restbase, req, from, to) {
 
     return this._getOriginalContent(restbase, req, rp.revision, tid)
     .then(function (original) {
+        // Check if parsoid metadata is present as it's required by parsoid.
+        if (!original['data-parsoid'].body || !original['data-parsoid'].body.ids) {
```
d00rman: It turns out that the corruption is caused by the fact that some objects are being stored and retrieved as strings instead of JSONs. So, this line needs to be:

```js
if (!original['data-parsoid'].body || original['data-parsoid'].body.constructor !== Object || !original['data-parsoid'].body.ids)
```

Pchelolo:
@d00rman Fixed according to your comments

PR #268, opened 2015-06-25
Downgrade bluebird to 2.8.2

gwicke:
The 2.9.x series have been a bit buggy, so stay with 2.8.2 for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   },
   "homepage": "https://github.com/wikimedia/restbase",
   "dependencies": {
-    "bluebird": "^2.9.27",
+    "bluebird": "2.8.2",
     "busboy": "^0.2.9",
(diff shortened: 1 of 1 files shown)


d00rman:
bump also `restbase-mod-table-cassandra` to 0.6.7 to be compatible?

gwicke:
@d00rman, that's pulled in automatically with the current package.json.

d00rman:
> that's pulled in automatically with the current package.json.

I know it is, but since we are modifying the file and the newer version will be pulled in, we might as well tidy up a bit ;)

gwicke:
@d00rman, I'll update to 0.7.0 once that's merged & published.

d00rman:
@gwicke sounds like a plan

PR #269, opened 2015-06-26
Delete handling added (T100031)

Pchelolo:
This changeset handles handles page deleted notifications. When the page is removed, the /page/{title}/latest endpoint is called. This makes us try to fetch the latest revision from MediaWiki API by calling fetchAndStoreMWRevision. If page is deleted this returns 404, and so we know that we need to mark the latest known by restbase revision with 'page_deleted' restriction. This is done by getting latest revision stored in restbase, modifying it and saving with a new timestamp.

Also this changes removes an optimisation in parsoid code, because now the assumption that the latest revision can never be restricted is wrong.

The test mocks http requests to MediaWiki API and simulates the situation when the page was deleted.
Also it tests that the /page/{title}/latest endpoint works correctly and that we return 404 on a deleted page.

In order to work correctly this needs [updates] (https://gerrit.wikimedia.org/r/#/c/221597/) to RestbaseUpdateJob extension

### DIFF
--- a/mods/page_revisions.js
+++ b/mods/page_revisions.js
@@ -304,44 +304,45 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
         });
     } else if (!rp.revision || rp.revision === 'latest') {
         revisionRequest = self.fetchAndStoreMWRevision(restbase, req)
-            .catch(function (e) {
-                if (e.status !== 404) {
+        .catch(function(e) {
+            if (e.status !== 404) {
+                throw e;
+            }
+            // In case 404 is returned by MW api, the page is deleted
+            return self.listTitleRevisions(restbase, req)
+            .then(function(res) {
+                if (res.body.items && res.body.items.length > 0) {
+                    var revReq = {
+                        uri: new URI([rp.domain, 'sys', 'page_revisions', 'rev', res.body.items[0]]),
+                        params: {
+                            api: 'sys',
+                            domain: rp.domain,
+                            module: 'page_revisions',
+                            revision: res.body.items[0]
+                        }
+                    };
+                    return self.getRevision(restbase, revReq);
+                } else {
                     throw e;
                 }
-                // In case 404 is returned by MW api, the page is deleted
-                return self.listTitleRevisions(restbase, req)
-                    .then(function (res) {
-                        if (res.body.items && res.body.items.length > 0 ) {
-                            var revReq = {
-                                uri: new URI([rp.domain, 'sys', 'page_revisions', 'rev', res.body.items[0]]),
-                                params: {
-                                    api: 'sys',
-                                    domain: rp.domain,
-                                    module: 'page_revisions',
-                                    revision: res.body.items[0]
-                                }
-                            };
-                            return self.getRevision(restbase, revReq);
-                        } else {
-                            throw e;
-                        }
-                    })
-                    .then(function(result) {
-                        result = result.body.items[0];
-                        result.tid = uuid.now().toString();
-                        result.restrictions = result.restrictions || [];
-                        result.restrictions.push('page_deleted');
-                        return restbase.put({
-                            uri: self.tableURI(rp.domain),
-                            body: {
-                                table: self.tableName,
-                                attributes: result
-                            }
-                        }).then(function () {
-                            throw e;
-                        });
-                    });
+            })
+            .then(function(result) {
+                result = result.body.items[0];
+                result.tid = uuid.now().toString();
+                result.restrictions = result.restrictions || [];
+                result.restrictions.push('page_deleted');
+                return restbase.put({
+                    uri: self.tableURI(rp.domain),
+                    body: {
+                        table: self.tableName,
+                        attributes: result
+                    }
+                })
+                .then(function() {
+                    throw e;
+                });
             });
+        });
     } else {
         throw new Error("Invalid revision: " + rp.revision);
     }
--- a/mods/page_revisions.js
+++ b/mods/page_revisions.js
@@ -103,15 +103,25 @@ PRS.prototype._checkRevReturn = function(res) {
     // https://phabricator.wikimedia.org/T76165#1030962
     if (item && Array.isArray(item.restrictions) && item.restrictions.length > 0) {
         // there are some restrictions, deny access to the revision
-        throw new rbUtil.HTTPError({
-            status: 403,
-            body: {
-                type: 'access_denied#revision',
-                title: 'Access to resource denied',
-                description: 'Access is restricted for revision ' + item.rev,
-                restrictions: item.restrictions
-            }
-        });
+        if (item.restrictions.indexOf('page_deleted') >= 0) {
+            throw new rbUtil.HTTPError({
+                status: 404,
+                body: {
+                    type: 'not_found#page_revisions',
+                    description: 'Page was deleted'
+                }
+            });
+        } else {
+            throw new rbUtil.HTTPError({
+                status: 403,
+                body: {
+                    type: 'access_denied#revision',
+                    title: 'Access to resource denied',
+                    description: 'Access is restricted for revision ' + item.rev,
+                    restrictions: item.restrictions
+                }
+            });
+        }
     }
     return true;
 };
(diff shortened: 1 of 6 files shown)


test/features/errors/notfound.js
```diff
@@ -66,4 +70,95 @@ describe('404 handling', function() {
...
+            })
+    });
+    it('should set page_deleted on deleted page', function() {
+        var apiURI = server.config
+            .conf.templates['wmf-sys-1.0.0']
+            .paths['/{module:action}']['x-modules'][0].options.apiURI;
+        var title = 'TestingTitle';
+        var revision = 12345;
+
+        nock.enableNetConnect();
```
gwicke: We probably need to make sure that nock either only catches the specific POST we are interested in, or that it is disabled right after the test has run. The former would still be preferable, imho, as it would also work if there are multiple API requests happening as part of this test later on.

test/features/errors/notfound.js
```diff
@@ -66,4 +70,95 @@ describe('404 handling', function() {
...
+                assert.deepEqual(e.status, 404);
+                assert.contentType(e, 'application/problem+json');
+            })
+            .then(function() {
+                api.done();
+            })
+            .finally(function() {
+                nock.cleanAll();
+                nock.restore();
+            });
```
Pchelolo: @gwicke It is checked and disabled. Look to these lines: done() verifies that every call we've intercepted was actually called, and cleanAll+restore cleans nock context and restores http to the original stated (disables nock). 

I've tested that failure of this test doesn't make other tests fail.

gwicke: Ahh, missed that bit. All good, then!

test/features/errors/notfound.js
```diff
@@ -6,8 +6,12 @@
 var assert = require('../../utils/assert.js');
 var preq   = require('preq');
 var server = require('../../utils/server.js');
+var nock = require('nock');
+var P = require('bluebird');
```
d00rman: I can't see why this is needed?

test/features/errors/notfound.js
```diff
@@ -66,4 +70,95 @@ describe('404 handling', function() {
             assert.contentType(e, 'application/problem+json');
         });
     });
+    it('should return 404 on deleted revision', function() {
+        return preq.get({ uri: server.config.bucketURL + '/revision/' + revPageDeleted })
+            .then(function() {
```
d00rman: We tend to indent `.then`s on the same level as their *callers*

test/features/errors/notfound.js
```diff
@@ -66,4 +70,95 @@ describe('404 handling', function() {
...
+                throw new Error('404 should be returned')
+            })
+            .catch(function(e) {
+                assert.deepEqual(e.status, 404);
+                assert.contentType(e, 'application/problem+json');
+            })
+    });
+    it('should set page_deleted on deleted page', function() {
+        var apiURI = server.config
```
d00rman: Just using the plain URI string seems more sensible here and improves readability

test/features/errors/notfound.js
```diff
@@ -66,4 +70,95 @@ describe('404 handling', function() {
...
+                }
+            })
+            // Other requests return nothing as if the page is deleted.
+            .post('')
+            .reply(200, {})
+            .post('')
+            .reply(200, {});
+        // Fetch the page
+        return preq.get({ uri: server.config.bucketURL + '/title/' + title + '/latest'})
+            .then(function(res) {
```
d00rman: Here and in the rest, idem - identation

test/features/pagecontent/revisions.js
```diff
@@ -6,6 +6,7 @@
 var assert = require('../../utils/assert.js');
 var preq   = require('preq');
 var server = require('../../utils/server.js');
+var P = require('bluebird');
```
d00rman: Where is it used in the file?

test/features/pagecontent/revisions.js
```diff
@@ -132,5 +135,13 @@ describe('revision requests', function() {
         })
     });
 
+    it('should return latest revision for a page', function () {
+        return preq.get({uri: server.config.bucketURL + '/title/' + pageName + '/latest'})
+            .then(function (res) {
```
d00rman: Indentation :)

mods/page_revisions.js
```diff
@@ -292,8 +302,46 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
             }
             return self.fetchAndStoreMWRevision(restbase, req);
         });
-    } else if (!rp.revision) {
-        revisionRequest = self.fetchAndStoreMWRevision(restbase, req);
+    } else if (!rp.revision || rp.revision === 'latest') {
+        revisionRequest = self.fetchAndStoreMWRevision(restbase, req)
+            .catch(function (e) {
```
d00rman: Indentation here as well

mods/page_revisions.js
```diff
@@ -292,8 +302,46 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
...
+                if (e.status !== 404) {
+                    throw e;
+                }
+                // In case 404 is returned by MW api, the page is deleted
+                return self.listTitleRevisions(restbase, req)
+                    .then(function (res) {
+                        if (res.body.items && res.body.items.length > 0 ) {
+                            var revReq = {
+                                uri: new URI([rp.domain, 'sys', 'page_revisions', 'rev', res.body.items[0]]),
+                                params: {
```
d00rman: The `params` object attribute here is superfluous - URI parameters are extracted from the URI automatically by RESTBase already.

d00rman: Ah no, `getRevision()` is directly called. Strike my last comment :)

Pchelolo:
@d00rman Fixed indentations and removed occasional useless lines.

test/features/errors/notfound.js
```diff
@@ -66,4 +68,99 @@ describe('404 handling', function() {
...
+            throw new Error('404 should be returned')
+        })
+        .catch(function(e) {
+            assert.deepEqual(e.status, 404);
+            assert.contentType(e, 'application/problem+json');
+        });
+    });
+    it('should set page_deleted on deleted page', function() {
+        var apiURI = server.config
+        .conf.templates['wmf-sys-1.0.0']
```
d00rman: Here and the line below need to be indented though ;)

test/features/pagecontent/revisions.js
```diff
@@ -13,6 +13,8 @@ describe('revision requests', function() {
 	var revOk = 642497713;
 	var revDeleted = 645504917;
 	var revRedirect = 591082967;
+    var pageName = 'User:GWicke%2fDate';
```
d00rman: I've jsut noticed the three lines above have tabs instead of spaces. Mind s/^I/    / ?

Pchelolo:
@d00rman And fixed indentations one more time.

test/features/errors/notfound.js
```diff
@@ -66,4 +68,99 @@ describe('404 handling', function() {
...
+                            'comment': 'Test',
+                            'tags': []
+                        }]
+                    }
+                }
+            }
+        })
+        // Other requests return nothing as if the page is deleted.
+        .post('')
+        .reply(200, {})
```
d00rman: This is not what the MW API returns. The 200 code is returned by the MW API, but not an empty response. Here's the body that it returns:

```javascript
{
  "batchcomplete": "",
  "query": {
    "badrevids": {
      "XXXXXX": {
        "revid": XXXXXX
      }
    }
  }
}
```

test/features/errors/notfound.js
```diff
@@ -66,4 +68,99 @@ describe('404 handling', function() {
...
+        .post('')
+        .reply(200, {});
```
d00rman: Idem.

d00rman:
I like the code in general (great stuff with *nock*!). I am, however, not persuaded we should have a new endpoint for this. I see two (main) problems with `/title/{title}/{revision}`:
- It provides a second, alternative way of obtaining information about a revision. We should really limit to one-to-one mappings between endpoints and exposed information since that will keep future headaches and inconsistencies to a minimum.
- It breaks the nice, automatic listing feature of `/title/{title}/` because `latest` (one of the valid values) is not in the listing. Fixing this inconsistency would mean creating a specialised handler for that. I am not keen on that (basically, again, less future headaches)

Currently, `/page/html/{title}` is used to fetch *the latest version of the HTML rendering for a page* in storage. If a `Cache-Control: no-cache` header is supplied in the request, the content is refreshed by calling out to Parsoid. Analogously, `/page/title/{ttile}` could be used for revisions - a simple request would get the latest revision ID for a given page as known by RESTBase, while providing a `Cache-Control: no-cache` header would stipulate a storage refresh by calling the MW API via the action module. This modus operandi can then be leveraged by the update job extension to signal page deletions.

Thoughts?

Pchelolo:
@d00rman The new entry point was what I was also worried about. The main argument for using that one was that it was supported internally while a /title/{title} endpoint did not exist. I agree that /title/{title} looks better from the API perspective, so I'd be happy to change this.

d00rman:
> The main argument for using that one was that it was supported internally while a /title/{title} endpoint did not exist.

Hehe, good observation about it not existing. But, it seems to me that is actually a reason for *not using* `/title/{title}/{revision}` since:

- `/page/` exists
- `/page/title` does not exist
- `/page/title/` exists
- `/page/title/{revision}` new endpoint

> I agree that /title/{title} looks better from the API perspective, so I'd be happy to change this.

Cool! Great!

Pchelolo:
@d00rman Uploaded new version with fixes according to your comments

specs/mediawiki/v1/content.yaml
```diff
@@ -62,12 +62,43 @@ paths:
...
+        - application/json
+      parameters:
+        - name: title
+          in: path
+          description: The page title.
+          type: string
+          required: true
+      responses:
+        '200':
+          description: The queriable list of page revisions.
```
d00rman: `/page/title/{title}/` (notice the trailing slash) returns that. This one should return the latest it knows about - either the latest in storage or the latest known by MW if the `Cache-Control: no-cache` header is provided (this is probably a valuable piece of information to clients, so it'd be good to put a note in the description)

d00rman:
Minor semi-nit and we're good to go!

specs/mediawiki/v1/content.yaml
Pchelolo: OMG! Skipped when copy-pasted. Will fix in a minute

Pchelolo:
@d00rman Done!

specs/mediawiki/v1/content.yaml
```diff
@@ -62,12 +62,45 @@ paths:
       x-backend-request:
         uri: /{domain}/sys/page_revisions/page/
 
+  /{module:page}/title/{title}:
+    get:
+      tags:
+        - Page content
+      description: >
+        Returns the latest revision for the title. By default returns the latest revision
```
d00rman: Not fond of stonewalling but I'd like that to be clear for users. They needn't know what RESTBase is, not where it gets its data, so something along the lines of "Returns the latest revision ID for the title present in storage. Supply the Cache-Control: no-cache header to obtain the latest available revision ID".

(Also note "revision ID", not just "revision")

d00rman:
Reminder: we also need to update the update job extension to use this new endpoint

PR #270, opened 2015-06-29
Added endpoint to translate sections to wikitext (T94890).

Pchelolo:
A new endpoint /transform/sections/to/wikitext/{title}/{revision}
was added. It takes the original revision from restbase, replaces
sections according to JSON, provided in a sections parameter, and
posts it to parsoid for html->wikitext translation.

A set of tests was added verifying the sections could be replaced
and added, and verifying errors on non-existent sections and invalid
JSON.

Bug: https://phabricator.wikimedia.org/T94890

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -555,34 +555,26 @@ function cheapBodyInnerHTML(html) {
  * Replaces sections in original content with sections provided in sectionsJson
  */
 function replaceSections(original, sectionsJson) {
-    var sectionOffsets = original['data-parsoid'].body.sectionOffsets,
-        newBody = cheapBodyInnerHTML(original.html.body);
-    Object.keys(sectionsJson)
-    .sort(function(id1, id2) {
-        var offset1 = sectionOffsets[id2],
-            offset2 = sectionOffsets[id1];
-        if (!offset1 || !offset2) {
-            throw new rbUtil.HTTPError({
-                status: 400,
-                body: {
-                    type: 'invalid_request',
-                    description: 'Invalid section ids'
-                }
-            });
-        }
-        return offset1.html[0] - offset2.html[0];
+    var sectionOffsets = original['data-parsoid'].body.sectionOffsets;
+    var newBody = cheapBodyInnerHTML(original.html.body);
+    var sectionIds = Object.keys(sectionsJson);
+    var illegalId = sectionIds.some(function(id) {
+        return !sectionOffsets[id];
+    });
+    if (illegalId) {
+        throw new rbUtil.HTTPError({
+            status: 400,
+            body: {
+                type: 'invalid_request',
+                description: 'Invalid section ids'
+            }
+        });
+    }
+    sectionIds.sort(function(id1, id2) {
+        return sectionOffsets[id2].html[0] - sectionOffsets[id1].html[0];
     })
     .forEach(function(id) {
         var offset = sectionOffsets[id];
-        if (!offset) {
-            throw new rbUtil.HTTPError({
-                status: 400,
-                body: {
-                    type: 'invalid_request',
-                    description: 'Invalid section id ' + id
-                }
-            });
-        }
         newBody = newBody.substring(0, offset.html[0])
             + sectionsJson[id]
             + newBody.substring(offset.html[1], newBody.length);
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -461,22 +461,25 @@ PSP.transformRevision = function (restbase, req, from, to) {
             original: original
         };
         if (from === 'sections') {
-            try {
-                var sections = JSON.parse(req.body.sections);
-                body2.html = {
-                    body: replaceSections(original, sections)
-                };
-                from = 'html';
-            } catch(e) {
-                // Catch JSON parsing exception and return 400
-                throw new rbUtil.HTTPError({
-                    status: 400,
-                    body: {
-                        type: 'invalid_request',
-                        description: 'Invalid JSON provided in the request'
-                    }
-                });
+            var sections = req.body.sections;
+            if (req.body.sections.constructor !== Object) {
+                try {
+                     sections = JSON.parse(req.body.sections.toString());
+                } catch (e) {
+                    // Catch JSON parsing exception and return 400
+                    throw new rbUtil.HTTPError({
+                        status: 400,
+                        body: {
+                            type: 'invalid_request',
+                            description: 'Invalid JSON provided in the request'
+                        }
+                    });
+                }
             }
+            body2.html = {
+                body: replaceSections(original, sections)
+            };
+            from = 'html';
         } else {
             body2[from] = req.body[from];
         }
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -38,6 +38,7 @@ function ParsoidService(options) {
         transformHtmlToHtml: self.makeTransform('html', 'html'),
         transformHtmlToWikitext: self.makeTransform('html', 'wikitext'),
         transformWikitextToHtml: self.makeTransform('wikitext', 'html'),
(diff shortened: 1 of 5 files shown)


mods/parsoid.js
```diff
@@ -530,6 +551,45 @@ function cheapBodyInnerHTML(html) {
...
 }
 
+/**
+ * Replaces sections in original content with sections provided in sectionsJson
+ */
+function replaceSections(original, sectionsJson) {
+    var sectionOffsets = original['data-parsoid'].body.sectionOffsets,
+        newBody = cheapBodyInnerHTML(original.html.body);
+    Object.keys(sectionsJson)
+    .sort(function(id1, id2) {
```
Pchelolo: We need to sort section in reverse order by section offset so that replaces don't clash with offset numbers.

mods/parsoid.js
```diff
@@ -530,6 +551,45 @@ function cheapBodyInnerHTML(html) {
...
+                    type: 'invalid_request',
+                    description: 'Invalid section id ' + id
+                }
+            });
+        }
+        newBody = newBody.substring(0, offset.html[0])
+            + sectionsJson[id]
+            + newBody.substring(offset.html[1], newBody.length);
+    });
+    return '<body>' + newBody + '</body>';
```
Pchelolo: According to @gwicke parsoid html->wikitext translation returns identical results for full html and only body when body is wrapped with <body>. The 'newBody' here was created by cheapBodyInnerHTML, so it for sure not wrapped yet.

mods/parsoid.js
d00rman: That's correct :+1: But, please, let's not confuse the reader / maintainer of the code - use `offset1` for `id1` (idem for `offset2`) in the line below.

mods/parsoid.js
```diff
@@ -530,6 +551,45 @@ function cheapBodyInnerHTML(html) {
     }
 }
 
+/**
+ * Replaces sections in original content with sections provided in sectionsJson
+ */
+function replaceSections(original, sectionsJson) {
+    var sectionOffsets = original['data-parsoid'].body.sectionOffsets,
+        newBody = cheapBodyInnerHTML(original.html.body);
```
d00rman: No variable def chaining please, use `var newBody = ...`

mods/parsoid.js
d00rman: Also, `.sort()` can (and should) be moved to the line above

mods/parsoid.js
```diff
@@ -530,6 +551,45 @@ function cheapBodyInnerHTML(html) {
...
+                    description: 'Invalid section ids'
+                }
+            });
+        }
+        return offset1.html[0] - offset2.html[0];
+    })
+    .forEach(function(id) {
+        var offset = sectionOffsets[id];
+        if (!offset) {
+            throw new rbUtil.HTTPError({
```
d00rman: This has already been checked during the sort above, hasn't it?

Pchelolo: @d00rman In case there's only 1 item in the sort comparator function wouldn't be called.

specs/mediawiki/v1/content.yaml
```diff
@@ -760,6 +760,52 @@ paths:
...
+          in: path
+          description: The page revision
+          type: integer
+          required: true
+        - name: sections
+          in: formData
+          description: Sections to transform
+          type: string
+          required: true
+      responses:
```
d00rman: Missing the `400` response definition

mods/parsoid.js
```diff
@@ -530,6 +551,45 @@ function cheapBodyInnerHTML(html) {
...
+/**
+ * Replaces sections in original content with sections provided in sectionsJson
+ */
+function replaceSections(original, sectionsJson) {
+    var sectionOffsets = original['data-parsoid'].body.sectionOffsets,
+        newBody = cheapBodyInnerHTML(original.html.body);
+    Object.keys(sectionsJson)
+    .sort(function(id1, id2) {
+        var offset1 = sectionOffsets[id2],
+            offset2 = sectionOffsets[id1];
```
d00rman: Also variable def separation :)

test/features/parsoid/transform.js
```diff
@@ -143,8 +143,47 @@ describe('transform api', function() {
...
+                + '/' + pageWithSectionsRev,
+            body: {
+                sections: '{"mwAQ":"<h2>First section replaced</h2>",'
+                    + '"mwAg":"<h2>Second section replaced</h2>"}'
+            }
+        })
+        .then(function(res) {
+            assert.deepEqual(res.status, 200);
+            assert.contentType(res, 'text/plain;profile=mediawiki.org/specs/wikitext/1.0.0');
+            assert.deepEqual(/== First section replaced ==/.test(res.body), true);
```
d00rman: Should also probably check that `== First section ==` (or whichever title was there before) is not part of the body any more.

d00rman:
LGTM :+1: A couple of small nits in-lined.

Pchelolo:
Already fixing those.

mods/parsoid.js
d00rman: > Alternatively we can sort without check and put undefined to the beginning, and then check only in one place.

Yup, that. Or use `Array.some()` to find out if there are any undefs before iterating so have a clearer flow/check separation.

Pchelolo:
@d00rman Fixed. Also I've changed test to target a special subpage I've created and marked as 'DO NOT MODIFY'

mods/parsoid.js
```diff
@@ -459,7 +460,27 @@ PSP.transformRevision = function (restbase, req, from, to) {
         var body2 = {
             original: original
         };
-        body2[from] = req.body[from];
+        if (from === 'sections') {
+            try {
+                var sections = JSON.parse(req.body.sections);
```
d00rman: Two small remarks:

- You might want to check that `req.body.sections` is not already something that's not a String or Buffer before parsing it, which it will be in case the client supplies the `application/json` *Content-Type* header.
- Probably only the JSON parsing should be enclosed in a `try/catch` block, since `replaceSections` is going to throw its own errors.

d00rman: > You might want to check that `req.body.sections` is not already something that's not a String or Buffer

This also implies that it's safer to do:

```js
var sections = JSON.parse(req.body.sections.toString())
```

when we indeed need to parse it

Pchelolo: @d00rman Actually, this endpoint is made in analogy to other transformation endpoints, so they accept a form-data. So, in case application/json header is defined it will fail way before this point with 400 'Missing request parameter sections'. So, I suppose it's enough to add .toString() and check that we've got an object after parsing to protect against smth like sections=1234 where JSON.parse wouldn't throw an exception. What do you think?

d00rman: Sorry, I probably wasn't clear (and now re-reading my own comment proves it). I just meant that it'd be nice to be flexible so that in case the content type is changed to `application/json` for the code to still be working (I am basing this possibility on the fact that it only takes to change one line in the spec).

> So, I suppose it's enough to add .toString() and check that we've got an object after parsing to protect against smth like sections=1234 where JSON.parse wouldn't throw an exception.

I must admit I don't understand what you mean here. How would `"sections=1234".toString()` change anything?

Pchelolo: Ah, ok, I'll add JSON handling. 

Regarding your second comment, it would be not `'sections=1234'.toString()` but `1234.toString()` but anyway this is not relevant now. I'll reupload the PR after I work out son handling.

Pchelolo:
@d00rman I've added a check that body.sections is not already an object before parsing it. In case of JSNO content-type parsing would be skipped, in case of form-data we would parse the string representation anyway, in case wrong JSON is provided within sections we would get 400. A test is also added

PR #272, opened 2015-07-02
Added cookie forwarding (T88016)

Pchelolo:
Cookies set on original request are forwarded to parsoid and mediawiki endpoints to let user auth and read content.

Two tests added, verifying that cookies are actually forwarded - one for parsoid and one for mediawiki api.

This solves only a part of the problem, after this fix user can read content from private wiki, but we still do not have a mechanism to forbid read of content already stored in restbase.

Added simple authentication module which checks permissions for content stored in restbase. Permissions are specified in the spec, and then collected along the path. On request, permission are checked with mediawiki api and 401 is returned in case user lacks any rights.

Bug: https://phabricator.wikimedia.org/T88016


test/features/errors/notfound.js
```diff
@@ -168,7 +168,6 @@ describe('404 handling', function() {
         })
         .finally(function() {
             nock.cleanAll();
-            nock.restore();
         });
```
Pchelolo: Apparently this was too much - we don't need to restore nock after test, because it breaks following nock tests. Cleaning all errors is enough to make everything run properly both if test pass and fail.

Pchelolo:
Please note that travis build will fail until swagger-router is updated, so this change depend on PR:
https://github.com/wikimedia/swagger-router/pull/20

lib/restbase.js
```diff
@@ -268,7 +279,7 @@ RESTBase.prototype._request = function (req) {
         childReq.params = match.params;
 
         // Call the handler with P.try to catch synchronous exceptions.
-        return P.try(handler, [childRESTBase, childReq])
+        var request = P.try(handler, [childRESTBase, childReq])
```
d00rman: `request` is confusing as a name since we usually use `req` / `request` to denote the actual request object, but here it's a Promise. Maybe `reqHandlerPromise` or `reqHandling` would be more appropriate.

lib/restbase.js
```diff
@@ -307,6 +318,18 @@ RESTBase.prototype._request = function (req) {
             self.metrics.endTiming([statName + statusClass, statName + 'ALL'], startTime);
             throw err;
         });
+        if (match.permissions
+            && Array.isArray(match.permissions)
```
d00rman: Indent one more level so that it's visible it isn't part of the `if`-block

lib/restbase.js
```diff
@@ -307,6 +318,18 @@ RESTBase.prototype._request = function (req) {
...
+            && match.permissions.length > 0) {
+            request = request.then(function(res) {
+                var permissionCheckRestbase = self.makeChild(req);
```
d00rman: Why not use `childRESTBase` ?

lib/router.js
```diff
@@ -218,6 +218,16 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
         });
     }
 
+    var security = pathspec.security;
+    if (Array.isArray(security)) {
+        loaderPromise = loaderPromise.then(function() {
+            node.value.security = node.value.security || [];
+            return P.each(security, function(permission) {
```
d00rman: This can be done in a simple `security.forEach()` loop, no need to use promises

lib/restbase.js
Pchelolo: @d00rman Indeed you are right, I thought it would be bad to use 1 restbase for 2 requests (handler and permission check), but looks like it's fine..

mods/action.js
```diff
@@ -148,6 +148,17 @@ function buildEditResponse(res) {
     return res;
 }
 
+function buildUserInfoResponse(res) {
```
d00rman: User info should be handled differently. Firstly, it's part of the `query` action, so having an endpoint in the same level as `query` is not good conceptually. So, eventually we could change `/query` endpoint to become `/query{/meta}` and then check to see if that param is specified, otherwise use the one provided by the user in the body.

Secondly, we should probably handle all of the query returns in `buildQueryResponse`. For now, a simple check to see if it's about user info and then check the object would do, but ideally, the same error checks apply to all query metas - response code check, basic body check, etc, so we would have a general check block and then, depending on the meta, use different routines to do further checks and/or response rewrites.

lib/auth.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
```
d00rman: Some comments in the header and for the `checkPermissions` fucntion would be desirable.

lib/auth.js
```diff
@@ -0,0 +1,48 @@
...
+    };
+    rbUtil.copyForwardedHeaders(restbase, checkReq);
+    return restbase.post(checkReq)
+    .then(function(userInfo) {
+        var accessAllowed;
+        var absentPermissions = [];
+        if (userInfo.rights && Array.isArray(userInfo.rights)) {
+            permissions.forEach(function(perm) {
+                if (userInfo.rights.indexOf(perm) < 0) {
+                    absentPermissions.push(perm);
```
d00rman: You can error out immediately here and report just the first permission, no need to be thorough when access is denied.

lib/rbUtil.js
```diff
@@ -173,6 +173,18 @@ rbUtil.parseETag = function(etag) {
     }
 };
 
+rbUtil.copyForwardedHeaders = function(restbase, req) {
```
d00rman: Some comments here too

mods/parsoid.js
```diff
@@ -141,6 +141,7 @@ PSP.pagebundle = function(restbase, req) {
     // TODO: Pass in current or predecessor version data if available
     var newReq = Object.assign({}, req);
     if (!newReq.method) { newReq.method = 'get'; }
+    newReq = rbUtil.copyForwardedHeaders(restbase, newReq);
```
d00rman: No need to reassign - the function does that in place.

mods/parsoid.js
```diff
@@ -528,6 +529,7 @@ PSP.callParsoidTransform = function callParsoidTransform (restbase, req, from, t
         headers: { 'content-type': 'application/json' },
         body: req.body
     };
+    parsoidReq = rbUtil.copyForwardedHeaders(restbase, parsoidReq);
```
d00rman: Idem.

lib/auth.js
Pchelolo: @d00rman I'm collecting all absent permissions here to provide a nice and useful error message. Don't think this tiny overhead would be noticeable.

mods/action.js
```diff
@@ -118,21 +118,28 @@ function buildQueryResponse(res) {
...
+        // XXX: Clean this up!
+        res.body = {
+            items: newBody,
+            next: res.body["query-continue"]
+        };
+        return res;
+    } else if (res.body.query.userinfo) {
+        return res.body.query.userinfo;
+    } else {
+        throw apiError({info: 'Unable to parse PHP action API response.'});
```
d00rman: Hmmm, not sure about this. Throwing a `cannot parse response` because we don't know the meta in question seems wrong given that the response might be a legitimate one.

Pchelolo: @d00rman This is exactly how this used to work, but previously we "knew" only 'pages' field, and now we know 'pages' and 'userinfo'.

lib/router.js
```diff
@@ -219,6 +219,13 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
         });
     }
 
+    var security = pathspec.security;
```
gwicke: I think you can actually handle this synchronously, before the loaderPromise is even created. If your concern is about modules adding security restrictions, then I think we can add logic for that above.

lib/restbase.js
```diff
@@ -307,6 +318,17 @@ RESTBase.prototype._request = function (req) {
             self.metrics.endTiming([statName + statusClass, statName + 'ALL'], startTime);
             throw err;
         });
+        if (match.permissions
+                && Array.isArray(match.permissions)
+                && match.permissions.length > 0) {
+            reqHandlerPromise = reqHandlerPromise.then(function(res) {
```
gwicke: We should actually be able to run the permission checks in parallel with the main request. A good option to do this is something like
 
```javascript
return P.all([reqHandlerPromise, auth.CheckPermissions(..)])
.then(function(res) {
   return res[0];
});
```
A failed permission check will cause P.all to fail as well.

lib/rbUtil.js
```diff
@@ -173,6 +173,23 @@ rbUtil.parseETag = function(etag) {
     }
 };
 
+/**
+ * Copies forwarded headers from restbase to request.
+ * If the same header was already set it takes precedence over
+ * the forwarded header.
+ */
```
gwicke: I am wondering if we should handle this transparently in restbase.js, similar to the way we systematically forward the request id. However, we should only forward these headers //Iff// access is restricted. If it isn't, then we should absolutely make sure that we don't pass any cookies along that could result in us getting access to privileged information.

lib/restbase.js
```diff
@@ -39,6 +40,17 @@ function RESTBase (options, req) {
         this._recursionDepth = par._recursionDepth + 1;
         this._priv = par._priv;
         this.rb_config = this._priv.options.conf;
+
+        // Copy cookies to forward on first request
+        if (this._recursionDepth === 1
+            && req.headers
+            && req.headers.cookie) {
+            this.forwardedHeaders = {
```
gwicke: As an alternative, we could keep a reference to `this._req` in both root & child restbase instances. It is already defined in [line 62](https://github.com/wikimedia/restbase/pull/272/files#diff-e8ae526e9e8804c83de905dc28fa5af1R62), but isn't copied over on child instance creation yet.

As another plus, this will makes sure that the object shape remains consistent between parent & child instances, which results in more efficient code (see https://www.mediawiki.org/wiki/Learning_JavaScript#Performance_optimization).

lib/restbase.js
```diff
@@ -307,6 +318,17 @@ RESTBase.prototype._request = function (req) {
             self.metrics.endTiming([statName + statusClass, statName + 'ALL'], startTime);
             throw err;
         });
+        if (match.permissions
```
gwicke: I'm wondering if we should propagate permissions down to nested requests, so that we can use that information to a) apply it to nested requests, and b) use it to implicitly forward cookies in outgoing requests (in `defaultWebRequestHandler`).

config.test.yaml
```diff
@@ -152,6 +148,16 @@ templates:
           /{api:sys}:
             x-subspec: *wp/sys/1.0.0
 
+  wp-secure-1.0.0: &wp/secure/1.0.0
```
gwicke: Nice!

lib/restbase.js
Pchelolo: @gwicke 
a) we don't really need to apply them to nested requests, because they are set on API level, so we will always check it before (or in parallel with) doing any internal request. I've made it like this specially to avoid adding extra logic to verify if we should or should not check permissions.
b) right now the cookie headers are always passed to internal requests

lib/rbUtil.js
Pchelolo: @gwicke b was my main reason for making manual copies. External requests might be added in the future, and if we forget about this rule - we expose a disastrous hole in security, bc we would basically send complete user login information to a third party service. I can change this, but I thought it would be better to be more secure with a cost of a bit uglier code. Anyway this whole thing is quite ugly and it would be changed when we have a generic auth service.

lib/restbase.js
Pchelolo: @gwicke when I made it I had in mind saving some pointless requests, but now I'm thinking that making the worst case speed better is more important..

lib/rbUtil.js
gwicke: One issue is that a) is not addressed currently. If we always forward cookies, then we might end up storing restricted content without also enforcing the same restrictions on storage access.

lib/restbase.js
gwicke: @Pchelolo, having information about what's checked further up could enable some optimizations. It also gives us a bit of information that lets us make sure that we don't send cookies to internal services if we persist the response and expose it without further access checks.

lib/restbase.js
gwicke: The API is quite slow (about 50ms min latency for a 'hello world' request), so it's important to start those requests as early as possible in order to minimize the overall request latency. You are right that we might end up checking the authentication despite some other part of the request failing, but that's fine.

mods/action.js
gwicke: I agree that it's not really a change from the old behavior. We can generalize this further as needed.

gwicke:
@Pchelolo @d00rman: Are there any changes you plan for this since the request templating changes have landed? Once ready, it might be a good idea to set up a meeting with Chris to walk through the logic & check with him on review / things we should look into.

Pchelolo:
@gwicke yes, there's some stuff I wanted to change, will work on this tomorrow most likely

gwicke:
@Pchelolo, okay.

lib/router.js
Pchelolo: @gwicke but we know the module's security restrictions only after we load it. And may be it loads submodules etc..

lib/restbase.js
```diff
@@ -168,6 +172,10 @@ RESTBase.prototype.defaultWebRequestHandler = function(req) {
     // Enforce the usage of UA
     req.headers = req.headers || {};
     req.headers['User-Agent'] = req.headers['User-Agent'] || this.rb_config.user_agent;
+    if (this._accessRestrictions.length > 0
+        && req.headers && req.headers['x-internal-request']) {
```
gwicke: This header can be injected in the request. We'd either have to set it to false unconditionally when receiving a request, or use some other mechanism to distinguish internal from external services.

I'm wondering if a whitelist (regexp?) to identify internal services would be cleaner than the x-internal-request bit here. For WMF-internal services, this could be something like `/^http:\/\/[^/]+\.wmnet(:[0-9]+)?\//`.

lib/restbase.js
```diff
@@ -271,8 +278,16 @@ RESTBase.prototype._request = function(req) {
         var childRESTBase = this.makeChild(req);
         childReq.params = match.params;
 
+        var accessRestricted = match.permissions
```
gwicke: This should probably go into `makeChild`.

lib/restbase.js
```diff
@@ -39,6 +40,11 @@ function RESTBase(options, req) {
         this._recursionDepth = par._recursionDepth + 1;
         this._priv = par._priv;
         this.rb_config = this._priv.options.conf;
+        this._rootReq = par._rootReq || req;
```
gwicke: For performance, it is better to make sure that properties are assigned in the same order in both branches, so that the object shape ends up being the same.

lib/rbUtil.js
```diff
@@ -207,6 +207,22 @@ rbUtil.parseETag = function(etag) {
...
+rbUtil.copyForwardedCookies = function(restbase, req) {
+    if (restbase._rootReq
```
gwicke: Do you think it would be worth extracting the cookie once, when assigning a new rootreq?

lib/rbUtil.js
gwicke: I proposed yet another option in https://github.com/wikimedia/restbase/pull/272/files#r37483785. Feel free to disregard if this is getting a bit too much ;)

config.example.yaml
```diff
@@ -190,6 +190,17 @@ templates:
           /{api:sys}:
             x-subspec: *wp/sys/1.0.0
 
+# Should be used instead of wp-default-1.0.0 for private wiki domains
+  wp-secure-1.0.0: &wp/secure/1.0.0
+    x-subspecs:
+      - paths:
+          /{api:v1}:
+            x-subspec: *wp/content/1.0.0
+            security:
```
gwicke: According to the swagger spec, the `security` stanza should contain one or more [securityRequirementObject](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#securityRequirementObject)s.

So, perhaps something like this?

```yaml
security:
  - mediawiki_auth:
    - read
```

Pchelolo:
Updated according to all comments. Also while working on this I found a bug in router https://phabricator.wikimedia.org/T109702 that was related, so fixed it together with this one, as it's dependent.

lib/restbase.js
```diff
@@ -67,10 +76,25 @@ function RESTBase(options, req) {
...
     }
-
 }
 
+RESTBase.prototype.isInternalRequest = function(req) {
+    if (this.rb_config.internal_request_whitelist) {
+        var uri = req.uri.toString();
+        return this.rb_config.internal_request_whitelist.some(function(regex) {
+            if (/^\/.+\/$/.test(regex)) {
+                return new RegExp(regex.substring(1, regex.length - 1)).test(uri);
```
gwicke: It would be cheaper to build a single regexp once at startup, and then just call that.

Example: 

```javascript
var regexps = ['/foo/', '/bar/'].map(function(s) {
  new RegExp(return s.substring(1, regex.length - 1)).toString();
});
var whitelist = new RegExp(regexps.join('|'));
```

config.example.yaml
```diff
@@ -213,6 +225,9 @@ services:
       salt: secret
       default_page_size: 250
       user_agent: RESTBase
+      internal_request_whitelist:
```
gwicke: I am actually wondering if we should move this to the security object, as it is closely tied to the mediawiki_auth cookie scheme. @Pchelolo  @d00rman @eevans, any thoughts?

In any case, we can do that in a follow-up as well.

Pchelolo: @gwicke We don't declare a special spec for a [security requirement scheme](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#securitySchemeObject) now, because swagger doesn't allow custom object types and what they have doesn't fit our needs: they have only 'basic', 'apiKey' and 'oAuth2' types, and what we have is non of these. Also, although internal requests are now used only in security part, it's not just about it, I could easily imagine us using it for some other stuff like counters/logging/whatever else.

gwicke: Those are different notions of internal vs. external, and I think it's not wise to mix them. Which services to forward MediaWiki authentication details to depends on the auth scheme. For example, a signed oauth token with limited permissions might be safe to forward to some services, while forwarding a general MW cookie would not be.

We already have

Pchelolo: @gwicke I don't agree with your understanding of swagger doc wording.

[Here](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#securitySchemeObject) validity column is related to the type of security definition object where the particular field can be added (e.g. `in` field is only valid in an object of `apiKey` type). To support my understanding [here](https://github.com/swagger-api/swagger-spec/blob/master/schemas/v2.0/schema.json#L1226) is the complete JSON schema of the security definition object. Please look at each object type definition: the `type` field there is restricted to a single-value enum and only possible values are `basic`, `apiKey` and `oath2`, so if we say that `type` is `mediawiki` we would get an invalid spec.

config.example.wikimedia.yaml
```diff
@@ -36,15 +36,21 @@ templates:
...
-      - mediaWikiAuth:
-        - user:read
     x-subspecs:
       - mediawiki/v1/content
       - mediawiki_v1_graphoid
       - mediawiki/v1/mobileapps
     # - mediawiki/v1/revision-scoring
+    securityDefinitions:
+      mediawiki_auth:
+        description: Checks permissions using MW api
```
gwicke: This description should ideally be useful to third-party developers. Perhaps something like "MediaWiki cookie authentication", so that they get an idea what kind of cookie they are supposed to pass in here?

config.example.wikimedia.yaml
```diff
@@ -36,15 +36,21 @@ templates:
...
+        type: apiKey
+        in: header
+        name: cookie
+        x-internal-request-whitelist:
+          - /http:\/\/[a-zA-Z0-9\.]+\/w\/api\.php/
+          - http://parsoid-lb.eqiad.wikimedia.org
```
gwicke: This looks nice overall!

PR #273, opened 2015-07-03
Added rawcontinue parameter to MediaWiki query requests,

Pchelolo:
MediaWiki API have deployed a breaking change in paging support, which breaks all restless listings.

(see https://lists.wikimedia.org/pipermail/wikitech-l/2015-June/081931.html)

### DIFF
--- a/mods/action.js
+++ b/mods/action.js
@@ -155,7 +155,7 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
     body.action = defBody.action;
     body.format = body.format || defBody.format || 'json';
     body.formatversion = body.formatversion || defBody.formatversion || 1;
-    if (defBody.rawcontinue) {
+    if (defBody.rawcontinue && !body.continue) {
         body.rawcontinue = defBody.rawcontinue;
     }
     req.method = 'post';
--- a/mods/action.js
+++ b/mods/action.js
@@ -155,7 +155,7 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
     body.action = defBody.action;
     body.format = body.format || defBody.format || 'json';
     body.formatversion = body.formatversion || defBody.formatversion || 1;
-    if (defBody.rawcontinue !== undefined) {
+    if (defBody.rawcontinue) {
         body.rawcontinue = defBody.rawcontinue;
     }
     req.method = 'post';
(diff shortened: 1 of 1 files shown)


mods/action.js
```diff
@@ -162,7 +165,8 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
 ActionService.prototype.query = function(restbase, req) {
     return this._doRequest(restbase, req, {
         action: 'query',
-        format: 'json'
+        format: 'json',
+        rawcontinue: ''
```
d00rman: As per https://lists.wikimedia.org/pipermail/wikitech-l/2015-June/081931.html it should be

```js
rawcontinue: 1
```

mods/action.js
```diff
@@ -155,6 +155,9 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
     body.action = defBody.action;
     body.format = body.format || defBody.format || 'json';
     body.formatversion = body.formatversion || defBody.formatversion || 1;
+    if (defBody.rawcontinue !== undefined) {
```
d00rman: We also might want to check (temporarily) for `req.body.continue`: if the caller provided that one, then we do not need to append `rawcontinue`

mods/action.js
Pchelolo: Right, stupid me)

mods/action.js
Pchelolo: In case of req.body.continue is provided the response is returned in another format we cannot parse, so we need raw continue either way

mods/action.js
```diff
@@ -155,6 +155,9 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
     body.action = defBody.action;
     body.format = body.format || defBody.format || 'json';
     body.formatversion = body.formatversion || defBody.formatversion || 1;
+    if (defBody.rawcontinue && !body.continue) {
```
d00rman: the `continue` param must be an empty string on first invocation, so this will evaluate to true always. Use:

```js
if (!body.hasOwnProperty('continue') && defBody.rawcontinue)
```

Pchelolo: @d00rman I need to read a book about JS)

d00rman:
Looking good :+1:

PR #274, opened 2015-07-03
Allow overlapping spec paths if methods are different.

Pchelolo:
In case paths are overlapping in a spec we throw an error indicating
that, but in case methods are not overlapping it's incorrect.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -236,6 +236,11 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
                 specRoot.paths[prefixPath][methodName] = method;
             }
 
+            if (node.value.methods[methodName]) {
+                throw new Error('Trying to re-define existing method '
+                    + node.value.path + ':' + methodName);
+            }
+
             var backendRequest = method && method['x-backend-request'];
             if (backendRequest) {
                 // Set up a templated backend request handler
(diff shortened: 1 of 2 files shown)


lib/router.js
```diff
@@ -236,6 +236,11 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
                 specRoot.paths[prefixPath][methodName] = method;
             }
 
+            if (node.value.methods[methodName]) {
+                throw new Error('Trying to re-define existing method '
+                    + node.value.path + ':' + methodName);
+            }
+
```
Pchelolo: Before the fix we were failing way earlier, before loading modules and sub specs, but it does not matter how early or late to fail the startup.

Pchelolo:
The tests are failing because https://github.com/wikimedia/restbase/pull/273 is not released yet. After it's integrated this should be rebased and retested

Pchelolo:
Rebased against master with https://github.com/wikimedia/restbase/pull/273

d00rman:
LGTM, but please add a test or two in `features/router/buildTree.js`

Pchelolo:
@d00rman Added two tests: for overlapping methods and for additional methods. Also rebased on top of master.

PR #275, opened 2015-07-06
Bump content type version in RESTBase

Pchelolo:
Parsoid switched html content-type version to 1.1.0.
We have to request parsoid to rerender html in case an older version
is found in storage. The html content-type is moved to config, and tests are updated to use it instead of hard-coded strings. In case data stored in restbase doesn't match expected content-type version it's pulled from parsoid again. 

This PR only fixes tests and doesn't affect actual workflow because we cannot afford more stress on cassandra. Second PR will be submitted when this one is settled.

Bug: https://phabricator.wikimedia.org/T104462

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -157,10 +157,12 @@ services:
       spec: *spec
       salt: secret
       default_page_size: 1
-      contentTypes:
-        html: text/html;profile=mediawiki.org/specs/html/1.1.0;charset=utf-8
-        data-parsoid: application/json;profile=mediawiki.org/specs/data-parsoid/0.0.1;charset=utf-8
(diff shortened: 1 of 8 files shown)


config.test.yaml
```diff
@@ -157,6 +157,10 @@ services:
       spec: *spec
       salt: secret
       default_page_size: 1
+      contentTypes:
```
d00rman: I'd prefer for this to be inside a `test` hash so that it is clear is pertains only to tests, something like:

```yaml
test:
  content_types:
    ...
```

(also note how we do not use camel case in the config)

d00rman:
LGTM.

/cc @eevans @gwicke

d00rman:
Merged since tests fail everywhere because of the new content type being given by Parsoid.

@Pchelolo please rebase the other PRs on top of this one

PR #276, opened 2015-07-06
Expose RESTBase monitoring examples in Swagger spec

Pchelolo:
Added x-amples to several endpoints to allow enabling monitoring.
Covered endpoints:
- /page/title/{title}
- /page/html/{title}
- /page/data-parsoid/{title}
- /page/revision/{revision}
- /transform/wikitext/to/html{/title}{/revision}
- /page/graph/png/{title}/{revision}/{graph_id}

This gives fine coverage while not creating too much load.
Also a monitoring.js test was added which runs all tests based on x-amples. The test is mostly copy-pasted from https://github.com/wikimedia/service-template-node/pull/39 and in future this should be probably replaced by an actual library that would be used for monitoring. 

The actual body is checked only partially to make it independent from any changes that might happen to the page. More specific matching would be added when there's an actual monitoring lib in place.

Bug: https://phabricator.wikimedia.org/T104850

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -154,6 +154,7 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
             specRoot = Object.assign({}, subSpec);
             specRoot.paths = {};
             specRoot.definitions = {};
+            specRoot['x-default-params'] = {};
             specRoot.basePath = specRootBasePath + prefixPath;
             prefixPath = '';
             // XXX: The basePath is incorrect when shared between domains. Set
@@ -265,12 +266,16 @@ Router.prototype._handleSwaggerSpec = function (rootNode, spec, operations, spec
         specRoot = spec;
         if (!spec.paths) { spec.paths = {}; }
         if (!spec.definitions) { spec.definitions = {}; }
+        if (!spec['x-default-params']) {spec['x-default-params'] = {}; }
         if (!spec.basePath) { spec.basePath  = prefixPath || ''; }
         prefixPath = '';
     }
     if (spec.definitions) {
         // Merge definitions
         Object.assign(specRoot.definitions, spec.definitions);
+    }
+    if (spec['x-default-params']) {
+        Object.assign(specRoot['x-default-params'], spec['x-default-params']);
     }
     var self = this;
     function handlePaths (paths) {
--- a/lib/router.js
+++ b/lib/router.js
@@ -266,7 +266,7 @@ Router.prototype._handleSwaggerSpec = function (rootNode, spec, operations, spec
         specRoot = spec;
         if (!spec.paths) { spec.paths = {}; }
         if (!spec.definitions) { spec.definitions = {}; }
-        if (!spec['x-default-params']) {spec['x-default-params'] = {}; }
+        if (!spec['x-default-params']) { spec['x-default-params'] = {}; }
         if (!spec.basePath) { spec.basePath  = prefixPath || ''; }
         prefixPath = '';
     }
--- a/mods/simple_service.js
+++ b/mods/simple_service.js
@@ -55,9 +55,6 @@ SimpleService.prototype.processSpec = function(spec) {
                 }
 
                 function backendRequest() {
-                    if (req.params.domain === 'en.wikipedia.test.local') {
-                        req.params.domain = 'en.wikipedia.org';
-                    }
                     var beReq = {
                         uri: backendUriTemplate.toString({
                             params: req.params
--- a/mods/simple_service.js
+++ b/mods/simple_service.js
@@ -55,6 +55,9 @@ SimpleService.prototype.processSpec = function(spec) {
                 }
 
                 function backendRequest() {
+                    if (req.params.domain === 'en.wikipedia.test.local') {
+                        req.params.domain = 'en.wikipedia.org';
+                    }
                     var beReq = {
                         uri: backendUriTemplate.toString({
                             params: req.params
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -96,12 +96,27 @@ paths:
       x-backend-request:
         uri: /{domain}/sys/page_revisions/page/{title}
       x-amples:
-        - title: Should get a latest revision of a page by title
+        - title: Get rev of by title from MW
           request:
             params:
               title: Foobar
             headers:
               cache-control: no-cache
+          response:
+            status: 200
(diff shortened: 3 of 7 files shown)


lib/router.js
```diff
@@ -265,6 +266,7 @@ Router.prototype._handleSwaggerSpec = function (rootNode, spec, operations, spec
         specRoot = spec;
         if (!spec.paths) { spec.paths = {}; }
         if (!spec.definitions) { spec.definitions = {}; }
+        if (!spec['x-default-params']) {spec['x-default-params'] = {}; }
```
gwicke: Minor: insert a space after { for symmetry.

specs/mediawiki/v1/content.yaml
```diff
@@ -94,6 +95,24 @@ paths:
             $ref: '#/definitions/problem'
       x-backend-request:
         uri: /{domain}/sys/page_revisions/page/{title}
+      x-amples:
```
d00rman: It would be good to put another example that fetches the latest rev directly from storage after this one.

specs/mediawiki/v1/content.yaml
```diff
@@ -202,6 +221,19 @@ paths:
...
+          request:
+            params:
+              title: Foobar
+            headers:
+              cache-control: no-cache
+          response:
+            status: 200
+            headers:
+              etag: /.+/
+              content-type: text/html
```
d00rman: Should probably be `/text\/html/`

specs/mediawiki/v1/content.yaml
```diff
@@ -202,6 +221,19 @@ paths:
         # work in current Chrome and Firefox.
         # See also http://tools.ietf.org/html/rfc7231#section-7.1.2
         uri: /{domain}/sys/parsoid/html/{title}
+      x-amples:
```
d00rman: Also, add another example for fetching directly from storage.

specs/mediawiki/v1/content.yaml
```diff
@@ -422,6 +454,22 @@ paths:
             $ref: '#/definitions/problem'
       x-backend-request:
         uri: /{domain}/sys/parsoid/data-parsoid/{title}
+      x-amples:
+        - title: Should get a latest data-parsoid of a page by title
+          request:
+            params:
+              title: Foobar
+            headers:
+              cache-control: no-cache
```
d00rman: No need for this header, the previous test for `html/{title}` already fetches the newest page bundle from Parsoid

specs/mediawiki/v1/content.yaml
```diff
@@ -740,6 +803,21 @@ paths:
...
+            body:
+              wikitext: == Heading ==
+              bodyOnly: true
+          response:
+            status: 200
+            headers:
+              content-type: text/html
```
d00rman: Idem, `/text\/html/`

specs/mediawiki_v1_graphoid.yaml
```diff
@@ -50,3 +50,15 @@ paths:
 
       x-backend-request:
         uri: /{domain}/sys/graphoid/v1/png/{title}/{revision}/{graph_id}
+
+      x-amples:
+        - title: Should get a graph from graphoid
```
d00rman: Nice one!

specs/mediawiki/v1/content.yaml
Pchelolo: @d00rman Could be both ways, in the test we check both a regexp and if the actual header starts with what's declared. Could expand this to full content-type with regexed version

d00rman:
LGTM modulo small in-lined comments.

Also, it would probably be a good idea for the tests to have shorter titles as they will be used in alerting.

specs/mediawiki/v1/content.yaml
d00rman: Let's be precise, since we use regex notation elsewhere. The monitoring utility could (and probably should) interpret this as *I want the field to be exactly this* as-is.

>  in the test we check both a regexp and if the actual header starts with what's declared

Yup, but that's more of a utility on our end. The actual production monitor should be clear on what we mean where.

Pchelolo:
The fix is updated according to @d00rman comments. The tests are failing now because of parsoid content-type mismatch, but they would pass after this PR is rebase on top of https://github.com/wikimedia/restbase/pull/275

mods/simple_service.js
```diff
@@ -55,6 +55,9 @@ SimpleService.prototype.processSpec = function(spec) {
                 }
 
                 function backendRequest() {
+                    if (req.params.domain === 'en.wikipedia.test.local') {
```
d00rman: Not sure this is a good idea. We could circumvent this easily by fixing the domain part in the graphoid module request uri.

specs/mediawiki/v1/content.yaml
d00rman: Not sure how this comment ended up here...

specs/mediawiki/v1/content.yaml
```diff
@@ -202,6 +236,29 @@ paths:
         # work in current Chrome and Firefox.
         # See also http://tools.ietf.org/html/rfc7231#section-7.1.2
         uri: /{domain}/sys/parsoid/html/{title}
+      x-amples:
+        - title: Get html by title from Parsoid
```
d00rman: Nope. this one is going to the storage, there are no `no-cache` headers. Invert the two tests - fetching from Parsoid should come first

specs/mediawiki/v1/content.yaml
```diff
@@ -740,6 +826,21 @@ paths:
             $ref: '#/definitions/problem'
       x-backend-request:
         uri: /{domain}/sys/parsoid/transform/wikitext/to/html{/title}{/revision}
+      x-amples:
+        - title: Transform wikitext to html
+          request:
+            params:
+              title: Foobar
+            headers:
+              cache-control: no-cache
```
d00rman: Not sure this header is needed.

d00rman:
Also, for all public API endpoints that are not covered with `x-amples` tests, each method needs to contain `x-monitor: false` so that the production monitor knows it shouldn't try to check the request/response pair.

specs/mediawiki/v1/content.yaml
Pchelolo: @d00rman again, comments seem to be moving around

Pchelolo:
@d00rman Updated according to your comments, rebased on master.

PR #277, opened 2015-07-08
Improve RESTBase CSP headers: use 'self' instead of *, allow inline s…

Pchelolo:
'self' doesn't work for localhost tests, so actual domains were added.
This should've been as easy as adding a domain name, but looks like Safari
doesn't support subdomains and lack of protocol, so for en.wikipedia.org
we actually need `http://*.wikipedia.org https://*.wikipedia.org`

Also 'unsafe-inline' attribute added for html that we've got from Parsoid and
therefor know it's sanitazed and safe.

Bug: https://phabricator.wikimedia.org/T104913

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -57,19 +57,21 @@ rbUtil.wildcardSubdomain = function(domain) {
 /**
  * Constructs Content-Security-Policy header to send in response
  *
- * @param allowInline if true 'unsafe-inline' is added to style-src
+ * @param domain the domain to allow. If undefined, '*' is allowed
+ * @param options options containing the following fields:
+ *                - allowInline - if true 'unsafe-inline' is added to style-src
  * @returns {string} CSP header value
  */
-rbUtil.constructCSP = function(req, allowInline) {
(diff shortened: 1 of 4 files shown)


lib/rbUtil.js
```diff
@@ -40,6 +40,38 @@ rbUtil.parseURL = function parseURL (uri) {
...
+    }
+};
+
+/**
+ * Constructs Content-Security-Policy header to send in response
+ *
+ * @param allowInline if true 'unsafe-inline' is added to style-src
+ * @returns {string} CSP header value
+ */
+rbUtil.constructCSP = function(req, allowInline) {
```
gwicke: Minor nit: It might be slightly cleaner to only pass in the domain, instead of the entire req. This will keep this utility function independent of the req structure, and make it clearer which part of the request is actually used.

Additionally, it might be clearer to accept an options object with an `allowInline` field. The advantage is that call sites are somewhat self-documenting that way, which is better than a boolean arg.

Pchelolo: @gwicke fixed

PR #278, opened 2015-07-13
Switch to the new MW API continuation mode

Pchelolo:
Switch to the new MW API continue. Action module doesn't set rawcontinue any more,
but sets a continue parameter instead. All paces using MW API continuation were updated.
Places where we make a request to storage are not updated.

Docs says we need to provide everything returned in the previous continue field to the new request, so `Object.assign()` is used.

Also parsoid provided paging info to rev_key_values, but it was not honoured. 

Bug: https://phabricator.wikimedia.org/T104703

### DIFF
--- a/mods/action.js
+++ b/mods/action.js
@@ -130,7 +130,7 @@ function buildQueryResponse(res) {
     // XXX: Clean this up!
     res.body = {
         items: newBody,
-        next: res.body["query-continue"]
+        next: res.body["continue"]
     };
     return res;
 }
@@ -155,8 +155,8 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
     body.action = defBody.action;
     body.format = body.format || defBody.format || 'json';
     body.formatversion = body.formatversion || defBody.formatversion || 1;
-    if (defBody.rawcontinue && !body.hasOwnProperty('continue')) {
-        body.rawcontinue = defBody.rawcontinue;
+    if (!body.hasOwnProperty('continue')) {
+        body.continue = '';
     }
     req.method = 'post';
     return restbase[req.method](req).then(cont);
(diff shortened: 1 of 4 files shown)


d00rman:
I might be missing something, but I think the end result is not well-handled. Concretely, the [MW API docs](https://www.mediawiki.org/wiki/API:Query#Continuing_queries) specifies that the query should be continued until `continue` is no longer in the result set. However, we do not check that here and in that case return an empty `next:` field in the response. However, if the client supplies that, either (a) a new continuation loop could start; or (b) the client could get a result set it had gotten previously. Both outcomes are wrong. We need to explicitly check for this edge case.

mods/page_revisions.js
```diff
@@ -136,13 +136,12 @@ PRS.prototype.listTitles = function(restbase, req, options) {
...
             prop: 'revisions',
-            format: 'json',
-            gapcontinue: ''
+            format: 'json'
         }
     };
 
     if (req.query.page) {
-        listReq.body.gapcontinue = restbase.decodeToken(req.query.page);
+        Object.assign(listReq.body, restbase.decodeToken(req.query.page));
```
gwicke: Nice trick ;)

mods/action.js
```diff
@@ -130,7 +130,7 @@ function buildQueryResponse(res) {
     // XXX: Clean this up!
     res.body = {
         items: newBody,
-        next: res.body["query-continue"]
+        next: res.body["continue"]
```
gwicke: Minor: This can be written more cleanly as `res.body.continue`.

gwicke:
> However, we do not check that here and in that case return an empty next: field in the response.

To me it looks like body.next is checked in a few places. Do you see one where it isn't?

d00rman:
> To me it looks like body.next is checked where it needs to be. Do you see a case where it should be, but isn't?

Duh, completely neglected to see the `if(body.next)` blocks wrapping these... Don't know how that happened :P

PR #279, opened 2015-07-20
Bump restbase-mod-table-cassandra dependency to v0.7.6

gwicke:
Change-Id: Ie72f1f6b8fbce16e59f2513ff1a88f763ea59610


gwicke:
Self-merged, as it's trivial (would pull latest version anyway) & tests are passing.

PR #280, opened 2015-07-22
Fix transparent revision retrieval without explicit revision id

gwicke:
The page deletion fix broke the transparent retrieval of revision metadata
if no explicit revision id was supplied. This patch restores that
functionality by catching a 404 from the internal storage request and calling fetchAndStoreMWRevision.

Change-Id: I3ad0cdf193497141caf9200db685d98ba8674cb0

### DIFF
--- a/mods/page_revisions.js
+++ b/mods/page_revisions.js
@@ -352,6 +352,12 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
                         }
                     });
                 }
+            })
+            .catch(function(e) {
+                if (e.status !== 404) {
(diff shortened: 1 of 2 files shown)


gwicke:
Self-merged this hot-fix for a speedy deploy, as this is the entry point actually hit by VE. New pages won't load properly in VE right now.

gwicke:
Deployed as a hotfix.

d00rman:
LGTM

Pchelolo:
@gwicke LGTM too

PR #281, opened 2015-07-22
MobileApps service endpoints

d00rman:
This PR adds the MobileApps service public endpoints and its back-end service. Only two routes are added, currently reflecting the existing ones in the service:

- `/{domain}/v1/page/mobile-html-full/{title}` for the *standard* version
- `/{domain}/v1/page/mobile-json-lite/{title}` for the *lite* version

Things to note:

- There is no `{/revision}` component at the moment, due to the usage of MW *mobileview* API which [does not permit selecting a particular revision](https://phabricator.wikimedia.org/T106143#1460715).
- The public endpoints are based on [T103811](https://phabricator.wikimedia.org/T103811) and as such can be subject to change.
- Before deploying this, the MobileApps service must be deployed and we need to have the config part in ops/puppet in place.
- Because the service will not have a public DNS record, we are using the labs instance in the repo config(s).

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -125,6 +125,10 @@ templates:
                   get:
                     backend_request:
                       uri: http://appservice.wmflabs.org/{domain}/v1/mobile/app/page/html/{title}
+                /v1/lite/{title}:
+                  get:
+                    backend_request:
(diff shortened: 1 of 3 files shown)


d00rman:
cc @gwicke @berndsi @dbrant

PR #283, opened 2015-07-23
Templates support

Pchelolo:
Made more general request reemploying support. On startup template for each request is compiled - separate URI templating lib, and a tassembly for each '{...}' template. When request comes, we use an array of compiled tassembly functions to substitute all templates in a new request.

~~Important note, this PR depends on https://github.com/gwicke/tassembly/pull/1~~ 
The PR is rebased on current master

Bug: https://phabricator.wikimedia.org/T86737

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -112,6 +112,10 @@ templates:
                   get:
                     backend_request:
                       uri: http://graphoid.wikimedia.org/en.wikipedia.org/v1/png/{title}/{revision}/{graph_id}
+                      headers:
+                        accept: '{accept}'
+                        accept-language: '{accept-language}'
+                      body: '{$.request.body}'
 
       /{module:mobileapps}:
         x-modules:
@@ -147,6 +151,10 @@ templates:
                   get:
                     backend_request:
                       uri: http://en.wikipedia.org/wiki/{+key}
+                      headers:
+                        accept: '{accept}'
+                        accept-language: '{accept-language}'
+                      body: '{$.request.body}'
                     storage: 
                       no-cache_refresh: true
                       bucket_request: 
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -112,9 +112,6 @@ templates:
                   get:
                     backend_request:
                       uri: http://graphoid.wikimedia.org/en.wikipedia.org/v1/png/{title}/{revision}/{graph_id}
-                      headers:
-                        accept: '{accept}'
-                        accept-language: '{accept-language}'
                       body: '{$.request.body}'
 
       /{module:mobileapps}:
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -4,45 +4,55 @@ var URI = require('swagger-router').URI;
 var rbUtil = require('./rbUtil');
 var TAssembly = require('tassembly');
 
-function _setAtPath(obj, value, path) {
-    var i;
-    path = path.split('.');
-    for (i = 0; i < path.length - 1; i++) {
-        obj = obj[path[i]];
-    }
-    if (value !== undefined) {
-        obj[path[i]] = value;
-    } else {
-        delete obj[path[i]];
-    }
+function _setAtPathFun(path) {
+    var ref = path.split('.').map(function(elem) {
+        return '["' + elem + '"]';
+    }).join('');
+    var code =  'if (value !== undefined) {\n' +
+                '   obj' + ref + ' = value;\n' +
+                '} else {\n' +
+                '   delete obj' + ref +';\n' +
+                '}';
+    /* jslint evil: true */
+    return new Function('obj', 'value', code);
 }
 
 function _createTemplateResolvers(subspec, reqPart, path, resolvers) {
+    var resolveTemplate = function(templateSpec, templatePath) {
+        if (templateSpec instanceof Object) {
+            _createTemplateResolvers(templateSpec, reqPart, templatePath, resolvers);
+        } else if (/^\{[^}]+}$/.test(templateSpec)) {
+            var template = /^\{([^}]+)}$/.exec(templateSpec)[1];
+            if (template.indexOf('$.req') === 0) {
+                template = template.replace('$.req', 'm');
+            } else {
+                template = 'm.' + reqPart + '.' + template;
+            }
+            var getValue = TAssembly.compile([['raw', template]], {
+                errorHandler: function() { return undefined; },
+                identityCallback: true
+            });
+            var setValue = _setAtPathFun(templatePath);
+            resolvers.push(function(newReq, req) {
+                var value = getValue(req);
+                setValue(newReq, value);
+            });
+        }
+    };
+
     if (!resolvers) {
         resolvers = [];
     }
     if (!path) {
         path = reqPart;
     }
-    Object.keys(subspec).forEach(function(key) {
-        if (subspec[key] instanceof Object) {
-            _createTemplateResolvers(subspec[key], reqPart, path + '.' + key, resolvers);
-        } else if (/^\{[^}]+}$/.test(subspec[key])) {
-            var template = /^\{([^}]+)}$/.exec(subspec[key])[1];
-            if (template.indexOf('$.req') === 0) {
-                template.replace('$.req', 'm');
-            } else {
-                template = 'm.' + reqPart + '.' + template;
-            }
-            var compiled = TAssembly.compile([['raw', template]], {
-                cb: function(val) { return val; }
-            });
-            resolvers.push(function(newReq, req) {
-                var value = compiled(req);
-                _setAtPath(newReq, value, path + '.' + key);
-            });
-        }
-    });
+    if (subspec instanceof Object) {
+        Object.keys(subspec).forEach(function(key) {
+            resolveTemplate(subspec[key], path + '.' + key);
+        });
+    } else {
+        resolveTemplate(subspec, path);
+    }
     return resolvers;
 }
 
@@ -68,13 +78,13 @@ function Template(spec) {
 }
 
 Template.prototype.eval = function(req) {
-    var self = this;
-    var newReq = self.spec;
-    newReq.uri = self._evalUri(req);
-    if (self._evalMethod) {
-        newReq.method = self._evalMethod(req);
+    var newReq = this.spec;
+    newReq.uri = this._evalUri(req);
+    newReq.params = req.params;
+    if (this._evalMethod) {
+        newReq.method = this._evalMethod(req);
     }
-    self.resolvers.forEach(function(resolver) {
+    this.resolvers.forEach(function(resolver) {
         resolver(newReq, req);
     });
     return newReq;
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -1,0 +1,83 @@
+"use strict";
+
+var URI = require('swagger-router').URI;
+var rbUtil = require('./rbUtil');
+var TAssembly = require('tassembly');
+
+function _setAtPath(obj, value, path) {
+    var i;
+    path = path.split('.');
+    for (i = 0; i < path.length - 1; i++) {
+        obj = obj[path[i]];
+    }
+    if (value !== undefined) {
+        obj[path[i]] = value;
+    } else {
+        delete obj[path[i]];
+    }
+}
+
+function _createTemplateResolvers(subspec, reqPart, path, resolvers) {
+    if (!resolvers) {
+        resolvers = [];
+    }
+    if (!path) {
+        path = reqPart;
+    }
+    Object.keys(subspec).forEach(function(key) {
+        if (subspec[key] instanceof Object) {
+            _createTemplateResolvers(subspec[key], reqPart, path + '.' + key, resolvers);
+        } else if (/^\{[^}]+}$/.test(subspec[key])) {
+            var template = /^\{([^}]+)}$/.exec(subspec[key])[1];
+            if (template.indexOf('$.req') === 0) {
+                template.replace('$.req', 'm');
+            } else {
+                template = 'm.' + reqPart + '.' + template;
+            }
+            var compiled = TAssembly.compile([['raw', template]], {
+                cb: function(val) { return val; }
+            });
+            resolvers.push(function(newReq, req) {
+                var value = compiled(req);
+                _setAtPath(newReq, value, path + '.' + key);
+            });
+        }
+    });
+    return resolvers;
+}
+
+function Template(spec) {
+    var uriTemplate = new URI(spec.uri, {}, true);
+    var self = this;
+
+    self.spec = spec;
+    self._evalUri = function(req) {
+        return uriTemplate.expand(req.params);
+    };
+    if (!spec.method) {
+        self._evalMethod = function(req) {
+            return req.method;
+        };
+    }
+    self.resolvers = [];
+    ['headers', 'body', 'query'].forEach(function(reqPart) {
+        if (spec[reqPart]) {
+            self.resolvers = self.resolvers.concat(_createTemplateResolvers(spec[reqPart], reqPart));
+        }
+    });
+}
+
+Template.prototype.eval = function(req) {
+    var self = this;
+    var newReq = self.spec;
+    newReq.uri = self._evalUri(req);
+    if (self._evalMethod) {
+        newReq.method = self._evalMethod(req);
+    }
+    self.resolvers.forEach(function(resolver) {
+        resolver(newReq, req);
+    });
+    return newReq;
+};
+
+module.exports = Template;
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -4,6 +4,10 @@ var URI = require('swagger-router').URI;
 var rbUtil = require('./rbUtil');
 var TAssembly = require('tassembly');
 
+/**
+ * Creates a function that sets a value at the path
+ * or deletes if undefined values is provided.
+ */
 function _setAtPathFun(path) {
     var ref = path.split('.').map(function(elem) {
         return '["' + elem + '"]';
@@ -56,6 +60,11 @@ function _createTemplateResolvers(subspec, reqPart, path, resolvers) {
     return resolvers;
 }
 
+/**
+ * Creates and compiles a new Template object using the provided JSON spec
+ *
+ * @param spec request spec
+ */
 function Template(spec) {
     var uriTemplate = new URI(spec.uri, {}, true);
     var self = this;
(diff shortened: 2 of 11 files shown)


Pchelolo:
СС @gwicke @d00rman @eevans

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+                '   obj' + ref + ' = value;\n' +
+                '} else {\n' +
+                '   delete obj' + ref +';\n' +
+                '}';
+    /* jslint evil: true */
+    return new Function('obj', 'value', code);
+}
+
+function _createTemplateResolvers(subspec, reqPart, path, resolvers) {
+    var resolveTemplate = function(templateSpec, templatePath) {
```
d00rman: It seems to me this doesn't have to be a `var`, `function resolverTemplate(...)` is highly recommended

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+
+var URI = require('swagger-router').URI;
+var rbUtil = require('./rbUtil');
+var TAssembly = require('tassembly');
+
+/**
+ * Creates a function that sets a value at the path
+ * or deletes if undefined values is provided.
+ */
+function _setAtPathFun(path) {
```
d00rman: Minor nit: lose the `Fun` part in the name. It's clear this is a function, and `Fun` kind of makes me think of fun, not functions :)

Oh, and also, please document the `path` param in the description above.

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+    } else {
+        resolveTemplate(subspec, path);
+    }
+    return resolvers;
+}
+
+/**
+ * Creates and compiles a new Template object using the provided JSON spec
+ *
+ * @param spec request spec
```
d00rman: Sorry, but this is not really useful. Is that an object? What kind? What does *request spec* refer to? Is that the template provided in the Swagger spec? Only a part of it? etc...

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+    var code =  'if (value !== undefined) {\n' +
+                '   obj' + ref + ' = value;\n' +
+                '} else {\n' +
+                '   delete obj' + ref +';\n' +
+                '}';
+    /* jslint evil: true */
+    return new Function('obj', 'value', code);
+}
+
+function _createTemplateResolvers(subspec, reqPart, path, resolvers) {
```
d00rman: Comments, params, return values. Please :)

specs/mediawiki/v1/content.yaml
```diff
@@ -220,6 +229,10 @@ paths:
           description: The page title.
           type: string
           required: true
+        - name: sections
```
d00rman: Nice catch!

specs/mediawiki/v1/content.yaml
```diff
@@ -399,6 +426,10 @@ paths:
           in: path
           description: The revision's time ID
           type: string
+        - name: sections
+          in: query
+          description: Comma-separated list of dection IDs
```
d00rman: *section IDs*

test/features/router/misc.js
```diff
@@ -84,4 +86,92 @@ describe('router - misc', function() {
...
+                'global': '{$.request.headers.name-with-dashes}'
+            },
+            body: {
+                'object': '{object}',
+                'global': '{$.request.params.domain}',
+                'added': 'addedValue',
+                'nested': {
+                    'a': {
+                        'b': {
+                            c: '{a.b.c}'
```
d00rman: Could perhaps use a different depth and names, e.g. `{one.two}`, so that we are sure the mapping is not done by depth equality.

d00rman:
Some nits and comments here and there, but all in all it seems like good stuff! I'll have to take a second, more thorough look though.

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+var URI = require('swagger-router').URI;
+var rbUtil = require('./rbUtil');
+var TAssembly = require('tassembly');
+
+/**
+ * Creates a function that sets a value at the path
+ * or deletes if undefined values is provided.
+ */
+function _setAtPathFun(path) {
+    var ref = path.split('.').map(function(elem) {
```
gwicke: It would be nice to also validate the path to be of the kind of simple form we currently support. Something like `/^(\$\.)?([a-zA-Z][a-zA-Z0-9-]*)(\.[a-zA-Z][a-zA-Z0-9-]*)*$/` ?

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function _createTemplateResolvers(subspec, reqPart, path, resolvers) {
+    var resolveTemplate = function(templateSpec, templatePath) {
+        if (templateSpec instanceof Object) {
+            _createTemplateResolvers(templateSpec, reqPart, templatePath, resolvers);
+        } else if (/^\{[^}]+}$/.test(templateSpec)) {
+            var template = /^\{([^}]+)}$/.exec(templateSpec)[1];
+            if (template.indexOf('$.request') === 0) {
+                template = template.replace('$.request', 'm');
```
gwicke: I think it probably makes more sense to replace '$.' with 'rm.'

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+    return new Function('obj', 'value', code);
+}
+
+function _createTemplateResolvers(subspec, reqPart, path, resolvers) {
+    var resolveTemplate = function(templateSpec, templatePath) {
+        if (templateSpec instanceof Object) {
+            _createTemplateResolvers(templateSpec, reqPart, templatePath, resolvers);
+        } else if (/^\{[^}]+}$/.test(templateSpec)) {
+            var template = /^\{([^}]+)}$/.exec(templateSpec)[1];
+            if (template.indexOf('$.request') === 0) {
```
gwicke: You can just match for '$.' here, to make it general for all globals.

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+                template = template.replace('$.request', 'm');
+            } else {
+                template = 'm.' + reqPart + '.' + template;
```
gwicke: Just 'm' should be enough here. We can then set the model we pass in to the reqPart of the original request, which is slightly cleaner IMHO.

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+            }
+            var getValue = TAssembly.compile([['raw', template]], {
+                errorHandler: function() { return undefined; },
+                identityCallback: true
```
gwicke: See my comment re the callback in https://github.com/gwicke/tassembly/pull/1.

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+    };
+
+    if (!resolvers) {
+        resolvers = [];
+    }
+    if (!path) {
+        path = reqPart;
+    }
+    if (subspec instanceof Object) {
+        Object.keys(subspec).forEach(function(key) {
```
gwicke: It might be easier to reason about this if it was written in a map style, with each resolveTemplate call *returning* its resolver.

Pchelolo: @gwicke not quite sure what did you mean by `map style`?

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+    self.spec = spec;
+    self._evalUri = function(req) {
+        return uriTemplate.expand(req.params);
+    };
+    if (!spec.method) {
+        self._evalMethod = function(req) {
+            return req.method;
+        };
+    }
+    self.resolvers = [];
```
gwicke: Maybe change this upside down by iterating over Object.keys of the template, and special-casing uri & method?

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Evaluates the compiled template using the provided request
+ *
+ * @param req a request object where to take data from
+ * @returns {*} a new request object with all templates either substituted or dropped
+ */
+Template.prototype.eval = function(req) {
+    // Deep copy the spec, because it's faster than traverse it trying to find added values
+    var newReq = JSON.parse(JSON.stringify(this.spec));
```
gwicke: This would be quite expensive for each request, especially considering that most templated content is going to be replaced anyway. Maybe start with an empty object instead, and iterate over the template without modifying it?

lib/reqTemplate.js
```diff
@@ -0,0 +1,109 @@
...
+    newReq.uri = this._evalUri(req);
+    newReq.params = req.params;
+    if (this._evalMethod) {
+        newReq.method = this._evalMethod(req);
+    }
+    this.resolvers.forEach(function(resolver) {
+        resolver(newReq, req);
+    });
```
gwicke: Future optimization idea: Compile the resolver calls into one function, so that they can potentially be inlined by the JIT.

specs/mediawiki/v1/content.yaml
```diff
@@ -220,6 +229,10 @@ paths:
           description: The page title.
           type: string
           required: true
+        - name: sections
+          in: query
+          description: Comma-separated list of dection IDs
```
gwicke: Minor: s/dection/section/

lib/reqTemplate.js
gwicke: @Pchelolo, something like 

```javascript
var resolvers = Object.keys(subspec).map(makeResolver);
```
with makeResolver being a wrapper around resolveTemplate taking the `key` parameter.

gwicke:
@Pchelolo: Sorry for the flood of comments! Happy to chat about it tomorrow.

lib/reqTemplate.js
Pchelolo: @gwicke I wanted to do like that in the beginning, but tassebly gives us a ready function, not the code, so each resolver consists of 2 compiled functions: getValue from tassembly and setValue. Or I didn't understand what do you mean here?

test/features/router/misc.js
```diff
@@ -84,4 +86,89 @@ describe('router - misc', function() {
...
+                'global': '{$.request.headers.name-with-dashes}'
+            },
+            body: {
+                'object': '{object}',
+                'global': '{$.request.params.domain}',
+                'added': 'addedValue',
+                'nested': {
+                    'one': {
+                        'two': {
+                            'tree': '{a.b.c}'
```
d00rman: I meant have also different depth levels, not just names.

Pchelolo: @d00rman It has, here it's level 4 (note nested), and in body it's level 3 (no nested)

lib/reqTemplate.js
```diff
@@ -0,0 +1,132 @@
...
+     */
+    function makeResolvers(templateSpec, templatePath) {
+        var setValue = _setAtPath(templatePath);
+        if (templateSpec instanceof Object) {
+            return _createTemplateResolvers(templateSpec, reqPart, templatePath);
+        } else if (/^\{[^}]+}$/.test(templateSpec)) {
+            var template = /^\{([^}]+)}$/.exec(templateSpec)[1];
+            var buildModel;
+
+            if (!/^(\$\.)?([a-zA-Z][a-zA-Z0-9-]*)(\.[a-zA-Z][a-zA-Z0-9-]*)*$/.test(template)) {
```
d00rman: Underscores should be supported as well

lib/reqTemplate.js
```diff
@@ -0,0 +1,132 @@
...
+var TAssembly = require('tassembly');
+
+/**
+ * Creates a function that sets a value at the path and initializes all objects along the path.
+ *
+ * @param path the path within an object separated with dots {e.g. test.body.a.b.c}
+ * @returns {Function} the function to set a value at path and initialize object along the path if needed
+ * @private
+ */
+function _setAtPath(path) {
```
d00rman: The function assumes `Object`s. What about arrays? Should we complicate the code to include them as well? Does it make sense to cover something like:

```yaml
body:
  - {a.b}
  - {a.c}
```

lib/reqTemplate.js
```diff
@@ -0,0 +1,132 @@
...
+            return _createTemplateResolvers(templateSpec, reqPart, templatePath);
+        } else if (/^\{[^}]+}$/.test(templateSpec)) {
+            var template = /^\{([^}]+)}$/.exec(templateSpec)[1];
+            var buildModel;
+
+            if (!/^(\$\.)?([a-zA-Z][a-zA-Z0-9-]*)(\.[a-zA-Z][a-zA-Z0-9-]*)*$/.test(template)) {
+                throw new Error('Invalid template ' + template);
+            }
+
+            if (template.indexOf('$.') === 0) {
```
d00rman: `template[0] === '$'` should be faster (as we know from the regex above no other combo with `$` is allowed)

lib/reqTemplate.js
```diff
@@ -0,0 +1,132 @@
...
+                template = template.replace('$', 'rm');
```
d00rman: Probably faster to `'$' + template.slice(1)`

d00rman: Ups, I meant `'rm' + template.slice(1)`

test/features/router/misc.js
d00rman: Hehe, missed the `nested` :P

lib/reqTemplate.js
```diff
@@ -0,0 +1,143 @@
...
+     */
+    function makeResolvers(templateSpec, templatePath) {
+        var setValue = _setAtPath(templatePath, origSpec);
+        if (templateSpec instanceof Object) {
+            return _createTemplateResolvers(origSpec, templateSpec, reqPart, templatePath);
+        } else if (/^\{[^}]+}$/.test(templateSpec)) {
+            var template = /^\{([^}]+)}$/.exec(templateSpec)[1];
+            var buildModel;
+
+            if (!/^(\$\.)?([a-zA-Z0-9-_]*)(\.[a-zA-Z0-9-_]*)*$/.test(template)) {
```
gwicke: I think we should require each component to start with a letter, and not a number.

gwicke:
Just double-checked the performance impact of this, and it does indeed seem to be really low. Definitely less than 5%.

specs/mediawiki/v1/content.yaml
```diff
@@ -788,6 +849,8 @@ paths:
             $ref: '#/definitions/problem'
       x-backend-request:
         uri: /{domain}/sys/parsoid/transform/html/to/wikitext{/title}{/revision}
+        body:
+          html: '{html}'
```
d00rman: We need to whitelist `oldid` and `scrubWikitext` as well, since these are sent by VE, cf. https://github.com/wikimedia/mediawiki-extensions-VisualEditor/blob/master/ApiVisualEditor.php#L158-L160

Pchelolo: @d00rman Interesting. I'll also check if I can add tests for those too.

specs/mediawiki/v1/content.yaml
```diff
@@ -840,6 +903,9 @@ paths:
             $ref: '#/definitions/problem'
       x-backend-request:
         uri: /{domain}/sys/parsoid/transform/wikitext/to/html{/title}{/revision}
+        body:
+          wikitext: '{wikitext}'
+          bodyOnly: '{bodyOnly}'
```
d00rman: Unfortunately, VE uses `body` for compatibility with systems that run Parsoid, but not RESTBase, cf. https://github.com/wikimedia/mediawiki-extensions-VisualEditor/blob/master/ApiVisualEditor.php#L181

gwicke: But, the RESTBase interface translates that, so that this should work as-is: https://github.com/wikimedia/mediawiki/blob/b79196778ed7c4f49ae09bee9ba16f861a6e3760/includes/libs/virtualrest/RestbaseVirtualRESTService.php#L153

config.test.yaml
```diff
@@ -111,6 +111,10 @@ templates:
                   get:
                     backend_request:
                       uri: http://graphoid.wikimedia.org/en.wikipedia.org/v1/png/{title}/{revision}/{graph_id}
+                      headers:
```
gwicke: Is the backend response differing based on accept / accept-language headers?

d00rman: Nope, it doesn't

specs/mediawiki/v1/content.yaml
d00rman: Oh right :)

config.test.yaml
Pchelolo: @gwicke @d00rman These headers were white-listed in the simple_service implementation, so I thought these might be needed. I'll remove them as you say these are unnecessary.

specs/mediawiki/v1/content.yaml
d00rman: Actually, `oldid` is being translated by the VRS, so we need to add just the `scrubWikitext` flag

gwicke:
I think this is ready to be merged.

For deployment, I think we'll have to update the production config in sync with the roll-out. We can do so manually by disabling puppet across the cluster, and then manually re-enabling & running puppet in sync with the code deploy, one node at a time. Any other ideas?

d00rman:
> I think this is ready to be merged.

I'm going to give it a second look, but I think so too.

> For deployment, I think we'll have to update the production config in sync with the roll-out. We can do so manually by disabling puppet across the cluster, and then manually re-enabling & running puppet in sync with the code deploy, one node at a time. Any other ideas?

AFAICT the only (relevant) change in the config is adding `body: {$.request.body}` to the Graphoid back-end module. In this instance I'd say we're fine with applying the config change with Puppet and then do a normal rolling deploy of the code. The downside I see is that some (occasional) workers might fail to start if they are respawned between the config change application and rolling deploy.

> 
> Remark on deployments: We could also move the RESTBase config templating to Ansible to avoid the coordination issue. However, we'd then have to pull in some of the hiera data. In practice it would probably be sufficient to copy over https://github.com/wikimedia/operations-puppet/blob/production/hieradata/role/common/restbase.yaml to ansible group vars (until this is moved to etcd), so maybe it's actually fairly easy. The biggest issue might be push-back from ops.

That might be a good idea. Whether there or create another config-only repo (but I'm not sure this is actually a good approach - we've seen MW config deployment problems, but that's another issue entirely). However, we might need to think about Beta and (eventually) Staging as well. I guess the simply approach there could be adding the list of hosts and their respective configs in there too. The thing to be aware is that names and IPs tend to be less stable than in prod.

As for ops, I think they might be on board with this since they support moving configs out of `ops/puppet`.

gwicke:
> I guess the simply approach there could be adding the list of hosts and their respective configs in there too. The thing to be aware is that names and IPs tend to be less stable than in prod.

We are already maintaining these anyway, and it's not that hard really.

gwicke:
> AFAICT the only (relevant) change in the config is adding body: {$.request.body} to the Graphoid back-end module.

A lot of the other changes are mandatory too (header forwarding, for example), but it should be possible to deploy the config changes before the deploy, without breaking the old code.

d00rman:
> but it should be possible to deploy the config changes before the code, without breaking the old code.

Exactly my point. The 2-3 minutes lag between config and code deploy should keep the number of 5xx responses to a minimum, if any at all. I don't envision that in that interval most of the 6*32 RB processes die and don't come back up.

d00rman:
LGTM as-is overall.

gwicke:
Okay, went ahead and merged it so that we can thoroughly test it before the next deploy.

PR #284, opened 2015-07-23
Count public & private requests

gwicke:
Keep a simplistic count of requests from public & private networks.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -192,6 +192,16 @@ function handleRequest (opts, req, resp) {
     };
     reqOpts.log = reqOpts.logger.log.bind(reqOpts.logger);
 
+    // Simplistic count of requests from public & private networks
+    var xff = req.headers['x-forwarded-for'];
+    var remoteAddr = xff && xff.split(',')[0].trim()
+        || req.socket.remoteAddress;
+    if (/^(?:10|127)\./.test(remoteAddr)) {
+        reqOpts.metrics.increment('requests.private');
+    } else {
+        reqOpts.metrics.increment('requests.public');
+    }
+
     // Create a new, clean request object
     var urlData = rbUtil.parseURL(req.url);
 


lib/server.js
```diff
@@ -192,6 +192,14 @@ function handleRequest (opts, req, resp) {
     };
     reqOpts.log = reqOpts.logger.log.bind(reqOpts.logger);
 
+    // Simplistic count of requests from public & private networks
+    var remoteAddr = req.socket.remoteAddress;
+    if (/^(?:10|127)\./.test(remoteAddr)) {
```
Pchelolo: What about 172.16.x.x - 172.31.x.x and 192.168.x.x?

d00rman: `10.x.x.x` as well

d00rman: Ups, it's there already. :P

lib/server.js
```diff
@@ -192,6 +192,14 @@ function handleRequest (opts, req, resp) {
     };
     reqOpts.log = reqOpts.logger.log.bind(reqOpts.logger);
 
+    // Simplistic count of requests from public & private networks
+    var remoteAddr = req.socket.remoteAddress;
```
d00rman: `remoteAddress` will report the connecting *client* which will be the LVS (or Varnish). In that respect, you are right to count `127.x.x.x`, but `10.x.x.x` will cover all of the remaining requests, which means we'll have 100% of private requests.

Varnish sets the `X-Forwarded-For` header, so we should use that one, and, if it's missing, use the socket remote address.

lib/server.js
gwicke: We don't use the others ;) But sure, lets add them too.

lib/server.js
gwicke: Hmm, yeah. Good point ;)

gwicke:
Added support for XFF as well.

PR #285, opened 2015-07-28
Pass through scrubWikitext to Parsoid

gwicke:
Parsoid added an API flag to fix up HTML in ways that results in cleaner wikitext.
While it's not clear that this is going to be the longer-term API, we agreed
to pass this through as-is without documenting it to allow VE and other
clients to take advantage of the feature quickly.

We can then discuss the longer-term API design in
https://phabricator.wikimedia.org/T106909.

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -482,6 +482,13 @@ PSP.transformRevision = function (restbase, req, from, to) {
             body2[from] = req.body[from];
         }
 
+        // For now, simply pass this through.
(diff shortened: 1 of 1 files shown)


arlolra:
+1

d00rman:
LGTM, but if VE uses the `VirtualRESTService` interface, we should add this there as well.

d00rman:
> but if VE uses the VirtualRESTService interface, we should add this there as well.

No, actually no need for that, the VRS transmits the body as-is.

PR #286, opened 2015-07-29
Created SQLite backend testing mode

Pchelolo:
Now we are able to test restbase with sqlite backend and cassandra backend. 

Simple npm test would check if cassandra is present and run against it, but if it's not, it would run tests against sqlite. In node > 0.10, we could specify what's being tests bug doing `npm test -- sqlite|cassandra|all`. I node 0.10 the same can be achieved by running `run_tests.sh test sqlite|cassandra|all`. Travis is changed to tests both backends.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,6 @@
     "swagger-test": "0.2.0",
     "url-template": "^2.0.6",
     "nock": "^2.6.0",
-    "restbase-mod-table-sqlite": "git+https://github.com/wikimedia/restbase-mod-table-sqlite#master"
+    "restbase-mod-table-sqlite": "^0.1.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "service-runner",
     "test": "sh test/utils/run_tests.sh test",
-    "coverage": "sh test/utils/run_test.sh coverage",
+    "coverage": "sh test/utils/run_tests.sh coverage",
     "coveralls": "cat ./coverage/lcov.info | coveralls"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "main": "server.js",
   "scripts": {
     "start": "service-runner",
-    "test": "sh test/utils/cleandb.sh && mocha",
-    "coverage": "sh test/utils/cleandb.sh && istanbul cover _mocha -- -R spec",
+    "test": "sh test/utils/run_tests.sh test",
+    "coverage": "sh test/utils/run_test.sh coverage",
     "coveralls": "cat ./coverage/lcov.info | coveralls"
   },
   "repository": {
@@ -54,6 +54,7 @@
     "mocha-lcov-reporter": "^0.0.2",
     "swagger-test": "0.2.0",
     "url-template": "^2.0.6",
-    "nock": "^2.6.0"
+    "nock": "^2.6.0",
+    "restbase-mod-table-sqlite": "git+https://github.com/wikimedia/restbase-mod-table-sqlite#master"
   }
 }
--- a/test/utils/run_tests.sh
+++ b/test/utils/run_tests.sh
@@ -50,6 +50,5 @@ fi
 
 if [ "$2" = "all" ]
 then
-    runTest "sqlite" $1
-    runTest "cassandra" $1
+    runTest "sqlite" $1 && runTest "cassandra" $1
 fi
--- a/test/utils/run_tests.sh
+++ b/test/utils/run_tests.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-PATH=$PATH:node_modules/.bin
+mod_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../.. && pwd )/node_modules
+mocha="$mod_dir"/mocha/bin/mocha
+istambul="$mod_dir"/istambul/bin/istambul
 
 runTest ( ) {
     if [ "$1" = "sqlite" ]
@@ -14,18 +16,16 @@ runTest ( ) {
         sh ./test/utils/cleandb.sh
     fi
 
-    if [ "$2" = "test" ]
+    if [[ "$2" = "test" ]]
     then
-        mocha
-    fi
-
-    if [ "$2" = "coverage" ]
+        ${mocha}
+    elif [[ "$2" = "coverage" ]]
     then
-        istanbul cover node_modules/.bin/_mocha -- -R spec
+        ${istanbul} cover node_modules/.bin/_mocha -- -R spec
     fi
 }
 
-if [ -z ${2+x} ]
+if [[ -z ${2+x} ]]
 then
     # no concrete backend is provided, check for cassandra
     `echo exit;` | cqlsh
@@ -38,17 +38,16 @@ then
     fi
 fi
 
-if [ "$2" = "sqlite" ]
+if [[ "$2" = "sqlite" ]]
 then
     runTest "sqlite" $1
-fi
-
-if [ "$2" = "cassandra" ]
+elif [[ "$2" = "cassandra" ]]
 then
     runTest "cassandra" $1
-fi
-
-if [ "$2" = "all" ]
+elif [[ "$2" = "all" ]]
 then
     runTest "sqlite" $1 && runTest "cassandra" $1
+else
+    echo "Invalid  testing mode"
+    exit 1
 fi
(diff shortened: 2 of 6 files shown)


test/utils/server.js
```diff
@@ -14,13 +14,28 @@ var hostPort  = 'http://localhost:7231';
...
 
+function loadConfig(path) {
+    var confString = fs.readFileSync(path).toString();
+    var backendImpl = process.env.RB_TEST_BACKEND;
+    if (backendImpl) {
+        if (backendImpl !== 'cassandra' && backendImpl !== 'sqlite') {
+            throw new Error('Invalid RB_TEST_BACKEND env variable value. Allowed values: "cassandra", "sqlite"');
+        }
+        if (backendImpl === 'sqlite') {
+            confString = confString.replace('restbase-mod-table-cassandra', 'restbase-mod-table-sqlite')
```
gwicke: ewwww ;) Can we replace this in the structure, rather than doing a string replace?

Pchelolo: @gwicke In the structure we ether need to 'know' the exact place where to make a replace (which's bad as we might change it and forget to change it here), or walk over all the structure (which looked slower to me). I'll test which way is faster

gwicke: Hmm, good point. I guess the string is unique enough and it's a test config only. So, let me retract my opposition here.

package.json
```diff
@@ -54,6 +54,7 @@
     "mocha-lcov-reporter": "^0.0.2",
     "swagger-test": "0.2.0",
     "url-template": "^2.0.6",
-    "nock": "^2.6.0"
+    "nock": "^2.6.0",
+    "restbase-mod-table-sqlite": "git+https://github.com/wikimedia/restbase-mod-table-sqlite#master"
```
d00rman: How about releasing `restbase-mod-table-sqlite` and using the version here?

Pchelolo: @d00rman There's one PR we are still pending there: https://github.com/wikimedia/restbase-mod-table-sqlite/pull/6 After that's resolved, we could release it

test/utils/run_tests.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+PATH=$PATH:node_modules/.bin
```
d00rman: This will be useless on non-symlink systems, e.g. MediaWiki-Vagrant where dependencies are installed without script symlinks. IMHO, a better variant would be to set the concrete path to the scripts we need, e.g.

```sh
mod_dir=$(cd ${0} && pwd)/node_modules
mocha=${mod_dir}/mocha/bin/mocha
```

*etc*, and then use that instead.

test/utils/run_tests.sh
```diff
@@ -0,0 +1,54 @@
...
+        echo "Running with SQLite backend"
+        export RB_TEST_BACKEND=sqlite
+        rm -f restbase
+    else
+        echo "Running with Cassandra backend"
+        export RB_TEST_BACKEND=cassandra
+        sh ./test/utils/cleandb.sh
+    fi
+
+    if [ "$2" = "test" ]
```
d00rman: Better to compact these to `if` blocks into `if`-`elsif` - `else` so that we can be sure a good command has been given (it's also more readable)

test/utils/run_tests.sh
```diff
@@ -0,0 +1,54 @@
...
+    if [ "$?" -eq 0 ]
+    then
+        runTest "cassandra" $1
+    else
+        echo "Cassandra not available. Using SQLite backed for tests"
+        runTest "sqlite" $1
+    fi
+fi
+
+if [ "$2" = "sqlite" ]
```
d00rman: Same as above - use `if` - `elsif` - `elsif` - `else`

test/utils/run_tests.sh
d00rman: Ups, that should have been:

```sh
mod_dir=$(cd ${0}/../.. && pwd)/node_modules
```

package.json
d00rman: kk

test/utils/run_tests.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+PATH=$PATH:node_modules/.bin
+
+runTest ( ) {
+    if [ "$1" = "sqlite" ]
```
d00rman: Since the shebang is set on `bash`, let's use *bash-esque* constructs and turn this into:

```sh
if [[ "$1" == "sqlite" ]]; then
  ...
```

Elsewhere in the file too :)

d00rman:
When I run this locally with Cassandra running, I'm getting:

```
$ npm test

> sh test/utils/run_tests.sh test

Running with Cassandra backend
...
  98 passing (1m)

Invalid  testing mode
npm ERR! Test failed.  See above for more details.
npm ERR! not ok code 0
```

test/utils/run_tests.sh
```diff
@@ -0,0 +1,53 @@
...
+    elif [ "$2" = "coverage" ]
+    then
+        ${istanbul} cover node_modules/.bin/_mocha -- -R spec
+    fi
+}
+
+if [ -z ${2+x} ]
+then
+    # no concrete backend is provided, check for cassandra
+    `echo exit;` | cqlsh
```
d00rman: Append `2>/dev/null` to get rid of those connection errors.

test/utils/run_tests.sh
d00rman: Actually, this also needs to be compacted into the same `if` above

d00rman: That's why I'm getting these `Invalid testing mode` errors.

test/utils/run_tests.sh
```diff
@@ -0,0 +1,53 @@
...
+then
+    runTest "sqlite" $1
+elif [ "$2" = "cassandra" ]
+then
+    runTest "cassandra" $1
+elif [ "$2" = "all" ]
+then
+    runTest "sqlite" $1 && runTest "cassandra" $1
+else
+    echo "Invalid  testing mode"
```
d00rman: Extra space

test/utils/run_tests.sh
```diff
@@ -0,0 +1,53 @@
...
+    if [ "$2" = "test" ]
+    then
+        ${mocha}
+    elif [ "$2" = "coverage" ]
+    then
+        ${istanbul} cover node_modules/.bin/_mocha -- -R spec
+    fi
+}
+
+if [ -z ${2+x} ]
```
d00rman: The standard *sh* fashion of testing this is: `if [ "x$2" == "x" ]`

test/utils/run_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+        rm -f restbase
+    else
+        echo "Running with Cassandra backend"
+        export RB_TEST_BACKEND=cassandra
+        sh ./test/utils/cleandb.sh
+    fi
+
+    if [ "$2" = "test" ]
+    then
+        ${mocha}
```
d00rman: Should be quoted in case `${mod_dir}` has got spaces or other special chars

test/utils/run_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+    elif [ "$2" = "coverage" ]
+    then
+        ${istanbul} cover node_modules/.bin/_mocha -- -R spec
```
d00rman: Idem - `"${istanbul}"`

test/utils/run_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+    fi
+elif [ "$2" = "sqlite" ]
+then
+    runTest "sqlite" $1
+elif [ "$2" = "cassandra" ]
+then
+    runTest "cassandra" $1
+elif [ "$2" = "all" ]
+then
+    runTest "sqlite" $1 && runTest "cassandra" $1
```
d00rman: I'd prefer to have them run sequentially without one being dependent on the other. If sqlite tests fail, we still want to know whether cassandra tests pass:

```sh
runTest sqlite $1
runTest cassandra $1
```

d00rman:
LGTM now, still some minor nits. In general, I'd really prefer for the script to use the `if [ ... ]; then` syntax - wasting a line on `then` feels somewhat unnatural to me. Thoughts?

test/utils/run_tests.sh
```diff
@@ -0,0 +1,44 @@
...
+    fi
+
+    if [ "$2" = "test" ]; then
+        "${mocha}"
+    elif [ "$2" = "coverage" ]; then
+        "${istanbul}" cover node_modules/.bin/_mocha -- -R spec
+    fi
+}
+
+if [ "x$2" == "x" ]; then
```
d00rman: Ups, that should have been

```
if [ "x$2" = "x" ]; then
```

I got carried away by the *bashism* in my earlier comment.

Pchelolo:
@d00rman Updated, I think now all comments have been fixed

PR #288, opened 2015-07-30
Insecure POST request to MW api

Pchelolo:
Please note, this is not a finished job, I'm making this PR as a reference for discussion!

Although this one looked simple, it uncovered some issues in URI tempting. In the ticket @gwicke suggests to use http://api.svc.eqiad.wmnet/w/api.php as an API uri, which is good for prod, but useless for tests and manual testing, bc it's an internal uri. So, in this PR I'm enabled request tempting for apiUrl in the action module. 

The problem is that in `config.yaml` meant to be used for manual testing/non-prod env, the uri template still has to be `http://{domain}/w/api.php'. But URI object in swagger-router doesn't support that, more that that, templates like this completely break swagger-router.URI, so domain be substituted before creating a template. Previously it was solved by these lines: https://github.com/wikimedia/restbase/blob/master/mods/action.js#L111-L114 , but now this means we would have to recompile the template on each request, which is really expensive. I've tried to use reggae-based solution and recompile the URI template on each template evaluation, but it's really slow. In this PR I make templates support '{...}' syntax and add a $.default_uri magic variable, which is really ugly, but it's the most obvious and fastest solution I've found. Would be really great to hear some opinions on this.

PS the test should still be created for host header, but I've postponed it until we decide on how to proceed with this.

Bug: https://phabricator.wikimedia.org/T107030


Pchelolo:
@gwicke @d00rman @eevans Would be really great to hear your opinions on how to solve this little issue.

d00rman:
Looks to me like that is a bug in the request templating, and not swagger-router. We should add a `params` field to the magic `$.request` templating object which would hold the key => value pairs of the request's URI parameters and then could do:

```yaml
uri: 'http://{$.request.params.domain}/w/api.php'
headers:
  Host: '${$.request.params.domain}'
```

Pchelolo:
@d00rman Nope. The problem is that swagger-router.URI.expand method only handles path elements, and not the `protocol` or `host` url parts. And even bigger problem is that it feeds those parts into native node's `url` object, which breaks completely if we have {....} as a host, so `http://{domain}/a/b/c` becomes `http:///%7Bdomain%7D/a/b/c` and we can't efficiently recover it

d00rman:
You are right @Pchelolo . I'd go for the KISS method. We know we will not need it in prod. So, we play the system by keeping the [ActionService.proto.apiURI](https://github.com/wikimedia/restbase/blob/master/mods/action.js#L111-L114) which we be used only locally until such time as we fix SwaggerRouter.URI to allow for domains to be params as well.

Pchelolo:
@d00rman So, adding support for both apiURI and apiRequest. Will update the PR.

gwicke:
It would be nice to expand swagger-router to support templating in other parts of the URI as well. We could even consider integrating the URI templating part with the general request templating functionality, possibly as a separate ([request-template](https://www.npmjs.com/package/request-template)?) library.

config.example.yaml
```diff
@@ -100,7 +100,12 @@ templates:
           - name: action
             type: file
             options:
-              apiURI: http://{domain}/w/api.php
+              apiRequest:
+                method: post
+                uri: '{$.default_uri}'
```
gwicke: For third-party users this is probably puzzling. What is this default_uri and where does it come from? I think it's fine for this patch, but it might be worth revisiting this later.

gwicke:
While re-reading the summary message the "reggae-based solution" caught my eye.. @Pchelolo, what does that involve? ;)

Pchelolo:
@gwicke OMG, you've merged that... I was not a finished job, I didn't make a test yet(

reggae = regex, autocorrect..

d00rman:
> you've merged that... I was not a finished job, I didn't make a test yet(

You can always create a follow-up PR and correct anything you need. Namely, relying on `apiTemplate` instead of compiling it every time could be good.

> reggae = regex, autocorrect..

lool [damn you autocorrect!](http://www.damnyouautocorrect.com/)

d00rman:
> you've merged that... I was not a finished job, I didn't make a test yet(

You can always create a follow-up PR and correct anything you need. Namely, relying on `apiTemplate` instead of compiling it every time could be good.

> reggae = regex, autocorrect..

lool [damn you autocorrect!](http://www.damnyouautocorrect.com/)

PR #289, opened 2015-07-30
Ensure the User-Agent header is set when doing external requests

d00rman:


d00rman:
cc @gwicke @eevans @Pchelolo

lib/restbase.js
```diff
@@ -164,6 +165,9 @@ RESTBase.prototype.defaultListingHandler = function(value, restbase, req) {
 
 // Special handling for external web requests
 RESTBase.prototype.defaultWebRequestHandler = function(req) {
+    // enforce the usage of UA
+    req.headers = req.headers || {};
+    req.headers['User-Agent'] = req.headers['User-Agent'] || this.rb_config.user_agent;
```
gwicke: lowercase `'user-agent'`

d00rman: All incoming header names are lowercased, however outgoing ones should be [RFC 7231-compliant](https://tools.ietf.org/html/rfc7231#section-5.5.3) (heh, notice the number of the RFC as being identical to the RB port :P). Hm, looking around now, it's true, we send out lowercased headers.

I agree consistency matters, but the larger question is: should we Correctly-Capitalise all of the outgoing headers? It's a standard to do so after all.

eevans: > I agree consistency matters, but the larger question is: should we Correctly-Capitalise all of the outgoing headers? It's a standard to do so after all.

Unless there is a compelling reason to lower-case them, I think we should honor the standard (even if it means opening a separate issue to fix existing lower-cased headers).

d00rman: > Unless there is a compelling reason to lower-case them, I think we should honor the standard (even if it means opening a separate issue to fix existing lower-cased headers).

I agree. We have a precedent too - [PR 214](https://github.com/wikimedia/restbase/pull/214) which introduced the `x-request-id` header (note the lowercasing).

The options seem to be:
1. Merge this as-is and take care of the rest in a follow up
2. Lower-case `User-Agent` (to `user-agent`) here and then fix everything in a follow-up
3. Just lower-case the UA header and keep the others lowercased as well

My vote goes to option (1)

eevans: On Thu, Jul 30, 2015 at 12:16 PM, Marko Obrovac <notifications@github.com>
wrote:

> In lib/restbase.js
> <https://github.com/wikimedia/restbase/pull/289#discussion_r35895162>:
>
> > @@ -164,6 +165,9 @@ RESTBase.prototype.defaultListingHandler = function(value, restbase, req) {
> >
> >  // Special handling for external web requests
> >  RESTBase.prototype.defaultWebRequestHandler = function(req) {
> > +    // enforce the usage of UA
> > +    req.headers = req.headers || {};
> > +    req.headers['User-Agent'] = req.headers['User-Agent'] || this.rb_config.user_agent;
>
> Unless there is a compelling reason to lower-case them, I think we should
> honor the standard (even if it means opening a separate issue to fix
> existing lower-cased headers).
>
> I agree. We have a precedent too - PR 214
> <https://github.com/wikimedia/restbase/pull/214> which introduced the
> x-request-id header (note the lowercasing).
>
> The options seem to be:
> 1. Merge this as-is and take care of the rest in a follow up
> 2. Lower-case User-Agent (to user-agent) here and then fix everything in
> a follow-up
> 3. Just lower-case the UA header and keep the others lowercased as well
>
> My vote goes to option (1)
>
+1


-- 
Eric Evans
eevans@wikimedia.org

gwicke: > notice the number of the RFC as being identical to the RB port

No coincidence ;)

+1 for fixing in a follow-up. RFC 7230 states: "Each header field consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace."

gwicke: Created a task to follow up in https://phabricator.wikimedia.org/T107509.

PR #290, opened 2015-07-31
TEST: Fix rb-mod-table-cass version to 0.7.6

d00rman:
Travis seems to be having problems today with the back-end module
version 0.7.7, so let's see if reverting to 0.7.6 helps.


d00rman:
This was just a test PR to see if 0.7.6 would fail as well. It's not so now we know that transitioning from 0.7.6 to 0.7.7 causes travis to fail.

PR #291, opened 2015-07-31
Add Cassandra version output to travis YAML

d00rman:


d00rman:
CASSANDRA_CONF and CLASSPATH are not set, giving up on this idea.

PR #292, opened 2015-07-31
Disable timeouts in tests

gwicke:
I think @cscott was right about timeouts being mostly annoying and very rarely
useful. They aren't deterministic, and easily trigger when Travis or some
external service is slow. They are essentially worthless for proper
performance regression testing.

So, this patch disables timeouts for our tests in general.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "server.js",
   "scripts": {
     "start": "service-runner",
-    "test": "sh test/utils/cleandb.sh && mocha",
+    "test": "sh test/utils/cleandb.sh && mocha --no-timeouts",
     "coverage": "sh test/utils/cleandb.sh && istanbul cover _mocha -- -R spec",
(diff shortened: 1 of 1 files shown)


d00rman:
Still witnessing the exact same failure in Travis. All RB PRs today are failing with Cassandra time outs, except #290 , so that makes me believe it's a (weird) regression from 0.7.6 to 0.7.7

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server.js",
   "scripts": {
     "start": "service-runner",
-    "test": "sh test/utils/cleandb.sh && mocha",
+    "test": "sh test/utils/cleandb.sh && mocha --no-timeouts",
```
d00rman: We have `test5/mocha.opts` so let's put it there?

d00rman:
It turns out test time-outs have been disabled for a while now, due to #263

PR #293, opened 2015-08-01
More efficient revision query

gwicke:


gwicke:
Still timing out the background update query, see https://phabricator.wikimedia.org/T107640

d00rman:
LGTM except the `--no-timeouts` issue I mentioned in #292

Pchelolo:
LGTM too, only thing is `--no-timeouts` wouldn't work any more, because the testing infrastructure was changed when SQLite testing mode was created.

gwicke:
Okay, dropped the timeout patch from this PR.

PR #294, opened 2015-08-03
Action module: Be backwards-compatible with apiURI-style configs

d00rman:
We have switched to request templating for `mods/action.js` configuration. However, some configs might still have the old-style `apiURI` config option. Translate that to the new style, since reading the old `config.yaml` file with the current RESTBase master fails to start the service entirely.

Note: this is a work-around, not a proper solution, as it assumes that either:
- there is no host templating in the URI, in which case the original `apiURI` string is used; or
- the URI host is templated and the rest of the URI is `/w/api.php`

### DIFF
--- a/mods/action.js
+++ b/mods/action.js
@@ -107,6 +107,8 @@ function apiError(apiErr) {
 function ActionService (options) {
     // be backwards-compatible with apiURI-style configs
     if(!options.apiRequest && options.apiURI) {
+        // log a deprecation warning
+        options.log('warn/actionService', 'The config options for this module have changed. Please use the apiRequest template stanza');
         options.apiRequest = {
             method: 'post',
             // TODO: assume the URI is in the form https?://{domain}/w/api.php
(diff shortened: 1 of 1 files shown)


mods/action.js
```diff
@@ -105,6 +105,24 @@ function apiError(apiErr) {
  * Action module code
  */
 function ActionService (options) {
+    // be backwards-compatible with apiURI-style configs
+    if(!options.apiRequest && options.apiURI) {
```
Pchelolo: Might be nice to print out a warning that 'You are using an outdated config format, please update your config ...'  so that the user could know that he needs to take an action. Otherwise LGTM

d00rman: @Pchelolo done. Also added a bail-out if neither are found in the config

Pchelolo:
Please note that in https://github.com/wikimedia/restbase/pull/295 I'm changing this logic too, so these 2 PRs should be handled together.

d00rman:
cc @gwicke @eevans

gwicke:
@Pchelolo, do you see this patch changing with #295? This patch would simplify the migration to the new configs, so it would be great to get it in soon.

Pchelolo:
@gwicke with https://github.com/wikimedia/restbase/pull/295 we would get rid of default_url and use http://{domain}/w/api.php instead

PR #295, opened 2015-08-03
New method for URI templating

Pchelolo:
Added support for substring templating e.g. 'string {this_is_replaced_by_value}'. This is used to template URIs with plain tassembly without using the URI.expand method. This is not only more efficient (~20% perf improvement on my laptop) but also we now support host templating. With this new functionality we can get rid of a hack with default_uri template in the action service.

### DIFF
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -37,6 +37,15 @@ function _setAtPath(path, spec) {
     return new Function('obj', 'value', code);
 }
 
+/**
+ * Validates and prepares the template string - resolves local references
+ * and wraps property access to ["..."]
+ *
+ * @param {string} template an original template string
+ * @param {string} defaultLookupPath a default path withing the context which
+ *                 to look if template is local
+ * @returns {string} validated and
+ */
 function prepareTemplateString(template, defaultLookupPath) {
     if (!/^(\$\.)?([a-zA-Z][a-zA-Z0-9-_]*)(\.[a-zA-Z][a-zA-Z0-9-_]*)*$/.test(template)) {
         throw new Error('Invalid template ' + template);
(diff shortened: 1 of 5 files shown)


lib/reqTemplate.js
```diff
@@ -157,6 +220,19 @@ function Template(spec) {
 }
 
 /**
+ * Adds common functions to the context.func object,
+ * so that they could be used in templates as rm.func.funcName
+ *
+ * @param {object} context a request context to put common functions to
+ */
+function appendCommonOperators(context) {
```
gwicke: Normally the model is reserved for data. It might be cleaner to use `rc.g` here to define a global encodeURIComponent function at compile time. That should also make it more likely that this function is inlined by the JIT.

lib/reqTemplate.js
```diff
@@ -55,14 +54,45 @@ function compileTemplate(templateSpec, setValue, reqPart) {
...
+function compileTemplate(templateSpec, setValue, reqPart) {
+    var template;
+    var templatedString;
+    var prevIndex = 0;
+    var completeTemplate = /^\{([^}]+)}$/.exec(templateSpec);
+    if (completeTemplate && completeTemplate.length > 0) {
+        template = [['raw', prepareTemplateString(completeTemplate[1], reqPart)]];
+    } else {
+        template = [];
+        templateSpec.replace(/\{([^}]+)}/g, function(full, spec, index) {
```
gwicke: You aren't really replacing here. Instead, you should be able to use `.exec` like this (untested):

```javascript
var re = /\{([^}]+)}/g;
var template = [];
var i = 0;
var match;
do {
  match = re.exec(templateSpec);
  if (match) {
    if (match.index !== i) {
      template.push(templateSpec.substring(i, match.index);
    }
    template.push(['raw', prepareTemplateString(match[1], reqPart)]);
  } else if (i !== templateSpec.length) {
    template.push(templateSpec.substring(i, templateSpec.length);
  }
} while (match)
```

lib/reqTemplate.js
```diff
@@ -55,14 +54,45 @@ function compileTemplate(templateSpec, setValue, reqPart) {
...
+            if (prevIndex !== index) {
+                template.push(templateSpec.substring(prevIndex, index));
+            }
+            templatedString = prepareTemplateString(spec, reqPart);
+            template.push(['text', templatedString]);
```
gwicke: I think you want 'raw' here, not 'text'. With 'text' you'll get HTML escaping.

lib/reqTemplate.js
```diff
@@ -121,6 +151,40 @@ function _createTemplateResolvers(origSpec, subspec, reqPart, path) {
...
+        };
+        return compileTemplate(spec.uri, setValue, 'params');
+    } else if (/^(?:https?:\/\/)?\{[^\/]+}\//.test(spec.uri)) {
+        // The host is templated - replace it with TAssembly and use URI.expand for path templates
+        var hostTemplate = /^((?:https?:\/\/)?\{[^\/]+}\/)/.exec(spec.uri)[1];
+        return compileTemplate(hostTemplate, setter, 'params').map(function(resolver) {
+            return function(newReq, context) {
+                var uriSpec;
+                resolver(newReq, context);
+                uriSpec = newReq.uri + spec.uri.substr(hostTemplate.length);
```
gwicke: This looks rather expensive, with a new URI template being created for each expansion. Maybe you could reuse the path template somewhat like this, by updating `.urlObj`:

```function
var pathTemplate = new URI(path, {}, true);
return function(newReq, context) {
  var newReq = resolver(newReq, context);
  var newUri = pathTemplate.expand(context.request.params);
  newUri.urlObj = url.parse(newPrefix.uri);
  setter(..)
}
```

Pchelolo:
@gwicke Updated according to your comments. An idea with pathTemplate is really good, almost 40% performance improvement.

PR #296, opened 2015-08-05
Support ipv6 embedded addresses in private / public IP classifier

gwicke:
Production hosts running iojs return ipv6 addresses for socket.remoteAddress.
Those are typically derived from ipv4 addresses, so can be matched with a
simple extension to the ipv4 regexp.


gwicke:
See also: https://en.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses

PR #297, opened 2015-08-05
Minor config example cleanup from production template

gwicke:


d00rman:
Could be good to also add the `apiURI => apiRequest` change for the action module. Apart from that, LGTM

gwicke:
> Could be good to also add the apiURI => apiRequest change for the action module.

@Pchelolo already added that in his PR.

PR #298, opened 2015-08-05
Enforce code style

Pchelolo:
Most of the stuff fixed automatically, then I've looked over the changes and all seems to be fine.
The jscs config is the same as in parsoid, but we don't verify that comments are starting with an upper-case letter, because we have several places with commented-out code that could be useful for debugging.

Bug: https://phabricator.wikimedia.org/T107950

### DIFF
--- a/mods/action.js
+++ b/mods/action.js
@@ -10,75 +10,75 @@ var Template = require('../lib/reqTemplate');
  * Error translation
  */
 var errDefs = {
-    '400': { status: 400, type: 'invalid_request' },
-    '401': { status: 401, type: 'unauthorized' },
-    '403': { status: 403, type: 'access_denied#edit' },
-    '409': { status: 409, type: 'conflict' },
-    '413': { status: 413, type: 'too_large' },
-    '429': { status: 429, type: 'rate_exceeded' },
-    '500': { status: 500, type: 'server_error' },
-    '501': { status: 501, type: 'not_supported' }
+    400: { status: 400, type: 'invalid_request', },
+    401: { status: 401, type: 'unauthorized', },
+    403: { status: 403, type: 'access_denied#edit', },
+    409: { status: 409, type: 'conflict', },
+    413: { status: 413, type: 'too_large', },
+    429: { status: 429, type: 'rate_exceeded', },
+    500: { status: 500, type: 'server_error', },
+    501: { status: 501, type: 'not_supported', },
 };
 
 var errCodes = {
     /* 400 - bad request */
-    'articleexists': errDefs['400'],
-    'badformat': errDefs['400'],
-    'badmd5': errDefs['400'],
-    'badtoken': errDefs['400'],
-    'invalidparammix': errDefs['400'],
-    'invalidsection': errDefs['400'],
-    'invalidtitle': errDefs['400'],
-    'invaliduser': errDefs['400'],
-    'missingparam': errDefs['400'],
-    'missingtitle': errDefs['400'],
-    'nosuchpageid': errDefs['400'],
-    'nosuchrcid': errDefs['400'],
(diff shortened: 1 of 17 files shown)


d00rman:
I really really have a problem with trailing commas. I find them too confusing and I think they hurt readability.

Pchelolo:
We could then use "disallowTrailingComma": true and remove all of them.

mods/action.js
```diff
@@ -10,75 +10,75 @@ var Template = require('../lib/reqTemplate');
...
 var errDefs = {
-    '400': { status: 400, type: 'invalid_request' },
-    '401': { status: 401, type: 'unauthorized' },
-    '403': { status: 403, type: 'access_denied#edit' },
-    '409': { status: 409, type: 'conflict' },
-    '413': { status: 413, type: 'too_large' },
-    '429': { status: 429, type: 'rate_exceeded' },
-    '500': { status: 500, type: 'server_error' },
-    '501': { status: 501, type: 'not_supported' }
+    400: { status: 400, type: 'invalid_request'},
```
d00rman: Why? If there is a space after the opening brace, I'd expect there to be one at the end as well (symmetry, OCD and all) :)

d00rman:
> We could then use "disallowTrailingComma": true and remove all of them.

Thanks for that! Looks much nicer now. But, still, some bike-shedding comments... Code face-lifting is never as easy as it seems.

Also, this is a good opportunity to get rid of all of those ugly `console.log()` calls that are commented out. Thoughts?

d00rman:
@Pchelolo mind rebasing on top of master so we have the io.js tests passing?

Pchelolo:
@d00rman Updated, rebased, removed commented-out code

d00rman:
Looks nice now, thnx @Pchelolo !

Bike-shed: what's the exact benefit of Capitalised comments? Makes sense if you have `/* */` blocks that explain stuff, but capitalising couple-of-word comments seems a bit excessive.

Pchelolo:
@d00rman not sure what's the reason for comment capitalisation. We could switch off that check.

gwicke:
I like how trailing commas simplify editing and diffing. When adding a new array entry or property, only a single line needs to be touched, and only a single line shows up dirty in the diff.

Given that @d00rman feels strongly about not using them, could we perhaps instate an 'anything goes' policy wrt to trailing commas?

Pchelolo:
@gwicke ok, if @d00rman wouldn't be too upset about it)

gwicke:
I'm also :+1: on ignoring comment capitalization.

eevans:
@d00rman 
> I really really have a problem with trailing commas. I find them too confusing and I think they hurt readability.

I'm a little surprised that this doesn't appeal to your need for symmetry (because it does mine :)).  I'd also echo @gwicke's sentiment that from a practical standpoint, it creates a smaller more readable diff later if the structure is extended.

That said, I don't know that I'd reject a patch during code-review because it did *not* have a trailing comma, so I'm in favor of having the code-checker ignore one.

eevans:
> not sure what's the reason for comment capitalisation. We could switch off that check.

Personally, I think you *should* use capitalization (and punctuation) when formatting comments (for the same reasons that you would in documentation, correspondence, etc).  And I say this with the full knowledge that if we audited my changesets, it'd likely paint me a hypocrite. :)

That said, I can't see myself pushing back on a change during code-review because of this, so I agree that the code checker should ignore it.

But I'd leave in place the capitalization changes that are already in this PR, because they make the code better.

Pchelolo:
Updated the PR - we now ignore comment issues and trailing commas

eevans:
:+1: from me; Thanks @Pchelolo!

PR #299, opened 2015-08-05
Use io.js v2.5.0 explicitly for Travis

d00rman:


Pchelolo:
LGTM, but let's wait if it also looks good to travis

PR #300, opened 2015-08-05
Fixed action service backwards compatibility support

Pchelolo:
We don't have default_uri any more and we can now template hosts in a uri, so we can use apiURI directly when supporting old configs


gwicke:
It looks like this needs a rebase.

Pchelolo:
@gwicke rebased

PR #301, opened 2015-08-05
Generalize content-addressable POST request storage

Pchelolo:
Adding the implementation that would allow to create request chains with config. 

Notes:
1. We should provide a Template to resources now.
2. The catch condition function is created as `new Function` because it's faster than reconstructing it on each request, but I'm not sure this optimisation is needed.

PS: this PR still lacks automatic testing coverage, but I think it's better to discuss the code and then add testing for different corner cases, as these corner cases might change during the review/discussion.

Bug: https://phabricator.wikimedia.org/T105975

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -8,6 +8,7 @@ var P = require('bluebird');
 var URI = require('swagger-router').URI;
 var uuid   = require('cassandra-uuid').TimeUuid;
 var rbUtil = require('../lib/rbUtil');
+var Template = require('../lib/reqTemplate');
 
 // TODO: move tests & spec to separate npm module!
 var yaml = require('js-yaml');
@@ -635,7 +636,7 @@ module.exports = function(options) {
         operations: ps.operations,
         // Dynamic resource dependencies, specific to implementation
         resources: [
-            {
+            new Template({
                 uri: '/{domain}/sys/key_rev_value/parsoid.html',
                 body: {
                     revisionRetentionPolicy: {
@@ -646,14 +647,14 @@ module.exports = function(options) {
                     valueType: 'blob',
                     version: 1
                 }
-            },
-            {
+            }),
+            new Template({
                 uri: '/{domain}/sys/key_rev_value/parsoid.wikitext',
                 body: {
                     valueType: 'blob'
                 }
-            },
-            {
+            }),
+            new Template({
                 uri: '/{domain}/sys/key_rev_value/parsoid.data-parsoid',
                 body: {
                     revisionRetentionPolicy: {
@@ -664,8 +665,8 @@ module.exports = function(options) {
                     valueType: 'json',
                     version: 1
                 }
-            },
-            {
+            }),
+            new Template({
                 uri: '/{domain}/sys/key_rev_value/parsoid.section.offsets',
                 body: {
                     revisionRetentionPolicy: {
@@ -676,13 +677,13 @@ module.exports = function(options) {
                     valueType: 'json',
                     version: 1
                 }
-            },
-            {
+            }),
+            new Template({
                 uri: '/{domain}/sys/key_rev_value/parsoid.data-mw',
                 body: {
                     valueType: 'json'
                 }
-            }
+            })
         ]
     };
 };
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -8,7 +8,6 @@ var P = require('bluebird');
 var URI = require('swagger-router').URI;
 var uuid   = require('cassandra-uuid').TimeUuid;
 var rbUtil = require('../lib/rbUtil');
-var Template = require('../lib/reqTemplate');
 
 // TODO: move tests & spec to separate npm module!
 var yaml = require('js-yaml');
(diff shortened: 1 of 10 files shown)


Pchelolo:
I've benchmarked this solution yesterday with testservice, and compared to old testservice. 
- Cache-hits have ~12% performance improvement
- Cache-miss benchmarked with local http server - for 100k requests we have degradation from 500 seconds to 520 seconds, but I suppose it's acceptable.

mods/parsoid.js
```diff
@@ -8,6 +8,7 @@ var P = require('bluebird');
 var URI = require('swagger-router').URI;
 var uuid   = require('cassandra-uuid').TimeUuid;
 var rbUtil = require('../lib/rbUtil');
+var Template = require('../lib/reqTemplate');
```
gwicke: Could we keep this out of the modules, and perform the Template wrapping in the module loading code in the router instead? I think that would keep the interface cleaner.

mods/simple_service.js
```diff
@@ -19,74 +18,308 @@ function SimpleService(options) {
...
+ *
+ * @param catchDefinition the condition in the format of 'catch' and 'return_if' stanza
+ * @returns {Function} a function that verifies the condition
+ */
+function compileCatchFunction(catchDefinition) {
+    function createCondition(catchCond, option) {
+        if (catchCond === 'status') {
+            if (/^[0-9]+$/.test(option.toString())) {
+                return '(res["' + catchCond + '"] === ' + option + ')';
+            } else {
```
gwicke: Perhaps worth checking that the condition is actually of the form `/^[0-9]xx$/i`?

mods/simple_service.js
```diff
@@ -19,74 +18,308 @@ function SimpleService(options) {
...
+        throw new Error('Invalid spec. ' +
+            'return_if should have a matching request. Step: ' + JSON.stringify(stepConf));
+    }
+
+    // Only supply 'catch' when 'request' is specified
+    if (Object.keys(stepConf).some(function(requestName) {
+        return stepConf[requestName].catch && !stepConf[requestName].request;
+    })) {
+        throw new Error('Invalid spec. ' +
+            'catch should have a matching request. Step: ' + JSON.stringify(stepConf));
```
gwicke: Hmm.. should we allow a bare catch too, so that we can set up a catch-all error handler?

Pchelolo: @gwicke there's a problem with bare catch: 
 - we cant's really match stuff other then a current request result, because it's implemented like this
 - if we have a catch, the request chain return undefined, which is hard to handle.

So, it's quite hard to handle, and the `catch` impl would be way slower, so I'd better not include it until we need it.

What do you think?

gwicke: I'm actually wondering how we'll template error responses. It seems we dropped that capability in the syntax reshuffle.

I guess the question is whether we eventually want to allow blocks like
```yaml
- return_if_error:
    return_if:
      $.earlier_request.status: 333
    return:
      body: ....
```
or, for catch,
```yaml
- catch_any_error:
    catch:
- ...
```

mods/simple_service.js
```diff
@@ -19,74 +18,308 @@ function SimpleService(options) {
...
+        });
+    });
+}
+
+/**
+ * Prepares a request chain on startup: compiles request/response templates,
+ * creates catch conditions verifiers.
+ * @param conf request chain configuration
+ *
+ * @returns {Array} an array of prepared steps, each containing an
```
d00rman: *containing an* ... ?

mods/simple_service.js
```diff
@@ -19,74 +18,308 @@ function SimpleService(options) {
...
+ */
+function prepareRequestChain(conf) {
+
+    /**
+     * Creates a function gene
```
d00rman: *gene* ?

mods/simple_service.js
```diff
@@ -19,74 +18,308 @@ function SimpleService(options) {
...
+                requestSpec.request = function(restbase, context) {
+                    return template.eval(context);
+                };
+            }
+        }
+    }
+
+    function prepareCatch(requestSpec, requestConf) {
+        if (requestConf.catch) {
+            var isCatched = compileCatchFunction(requestConf.catch);
```
d00rman: *isCaught*

mods/simple_service.js
gwicke: In any case, we can add this later.

gwicke:
As a next step, I think it would be nice to move this functionality from the simple_service module into RESTBase proper. That way, we can use the same templating functionality instead of `x-backend-request` in top-level specs.

Pchelolo:
Updated the fix according to comments. I think it's better to postpone moving whole RB on this, it should be tested better first.

gwicke:
LGTM. @d00rman @eevans, any objections to merging this?

d00rman:
How about adding a few tests related to this feature before merging? That could come in handy later on should any regressions arise.

Pchelolo:
@d00rman ok, I'll add some tests. Need to cover parallel solution and validations.

Pchelolo:
@d00rman Added tests for spec validation and parallel requests

Pchelolo:
@d00rman Great idea with x-monitor. Forgot about this feature, fixed now.

gwicke:
Congratulations, @Pchelolo!

PR #302, opened 2015-08-07
Modify MobileApps routes

d00rman:
We have settled on the following routes for the MobileApps service:
- `/page/mobile-html`
- `/page/mobile-html-sections`
- `/page/mobile-html-sections-lead`
- `/page/mobile-html-sections-remaining`
- `/page/mobile-text`

The content specification and configuration files have been changed accordingly.

Bug: [T102270](https://phabricator.wikimedia.org/T102270)
Bug: [T104719](https://phabricator.wikimedia.org/T104719)

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -130,23 +130,23 @@ templates:
                 /v1/html/{title}:
                   get:
                     backend_request:
-                      uri: http://appservice.wmflabs.org/en.wikipedia.org/v1/page/mobile-html/{title}
+                      uri: http://appservice.wmflabs.org/{domain}/v1/page/mobile-html/{title}
                 /v1/sections/{title}:
                   get:
                     backend_request:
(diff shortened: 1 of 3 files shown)


d00rman:
cc @gwicke @berndsi

config.example.yaml
```diff
@@ -128,14 +127,26 @@ templates:
             type: file
             options:
               paths:
-                /v1/std/html/{title}:
+                /v1/html/{title}:
                   get:
                     backend_request:
-                      uri: http://appservice.wmflabs.org/{domain}/v1/mobile/app/page/html/{title}
-                /v1/lite/{title}:
+                      uri: http://appservice.wmflabs.org/en.wikipedia.org/v1/page/mobile-html/{title}
```
Pchelolo: Why are you setting en.wikipodia.org in the example config? It makes sense in tests, but here we can provide a real domain.

d00rman: Duh, copy/paste fail. I've never been good at that. Thnx @Pchelolo for noticing!

PR #303, opened 2015-08-07
Hook up Mathoid in RESTBase

d00rman:
This PR adds the public API end-points and the back-end module definition for proxying calls to Mathoid - the math rendering service.

Two public routes are added:
- `/media/math/{format}/{formula}`
- `/media/math/{format}/{formula}/{type}`

The first renders a formula given in TeX format into the specified format, while the latter allows a client to specify the input type as well.

Note: depends on #302 since they modify the same files.

Bug: [T102030](https://phabricator.wikimedia.org/T102030)

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -44,6 +44,7 @@ templates:
       - mediawiki/v1/content
       - mediawiki_v1_graphoid
       - mediawiki/v1/mobileapps
+      - mediawiki/v1/mathoid
     # - mediawiki/v1/revision-scoring
 
   wmf-sys-1.0.0: &wp/sys/1.0.0
@@ -119,6 +120,22 @@ templates:
                   get:
                     backend_request:
                       uri: http://graphoid.wikimedia.org/{domain}/v1/png/{title}/{revision}/{graph_id}
+
+      /{module:mathoid}:
+        x-modules:
+          - name: simple_service
+            version: 1.0.0
+            type: file
+            options:
+              paths:
+                /v1/{format}/{formula}/{type}:
+                  get:
+                    backend_request:
+                      method: post
+                      uri: http://mathoid-tester.wmflabs.org/{format}
+                      body:
+                        q: '{$.request.params.formula}'
+                        type: '{$.request.params.type}'
 
       /{module:mobileapps}:
         x-modules:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -43,6 +43,7 @@ templates:
       - mediawiki/v1/content
       - mediawiki_v1_graphoid
       - mediawiki/v1/mobileapps
+      - mediawiki/v1/mathoid
       - test
 
   wmf-sys-1.0.0: &wp/sys/1.0.0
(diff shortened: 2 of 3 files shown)


d00rman:
cc @eevans @gwicke @Pchelolo @physikerwelt

config.example.yaml
```diff
@@ -119,7 +120,22 @@ templates:
...
+            type: file
+            options:
+              paths:
+                /v1/{format}/{formula}/{type}:
+                  get:
+                    backend_request:
+                      method: post
+                      uri: http://mathoid-tester.wmflabs.org/{format}
+                      body:
+                        q: '{$.request.params.formula}'
```
Pchelolo: @d00rman I assume the formula should be URL-encoded somehow and mathoid understands the URL-encoded format, right? There are lost of non-url-friendly characters in latex format

config.example.yaml
```diff
@@ -119,7 +120,22 @@ templates:
...
+          - name: simple_service
+            version: 1.0.0
+            type: file
+            options:
+              paths:
+                /v1/{format}/{formula}/{type}:
+                  get:
+                    backend_request:
+                      method: post
+                      uri: http://mathoid-tester.wmflabs.org/{format}
```
physikerwelt: is mathoid-tester already up and running?

config.example.yaml
physikerwelt: What happens with very long formulae?

config.example.yaml
d00rman: Yup. That's just a proxy for `mathosphere.math.eqiad.wmflabs:10042` (for some reason I couldn't get `math.beta.wmflabs.org:10042` to work)

config.example.yaml
d00rman: @Pchelolo yup, it has to be URI-encoded, so we're safe.

@physikerwelt the lowest maximum URL length is 2KB (old IE & IIS), modern browsers and servers accept arbitrary lengths, so we should be good here. Do you know of any real-world case where the formula would be longer than that?

gwicke: Yeah, this won't work for large formulas larger than ~1k with escaping. I thought the plan was to use [POST storage](https://phabricator.wikimedia.org/T105975)?

physikerwelt: @gwicke @d00rman Yes I think we need post storage. Even though for most cases 2kb are enough.

d00rman: In practice, the 2K limit is not a real restriction for this particular case. I agree that POST storage would be a better fit for this. This was more meant as a *lets-get-started-on-this*-kind-of PR than a complete solution.

gwicke: The first version of POST storage has now landed. Should we tweak this PR to use it?

d00rman: Yup. I'll follow up on this.

config.example.yaml
physikerwelt: Ok that's good to know. I normally regulary delete labs-instances and create new ones.
Maybe we can use http://mathoid.testme.wmflabs.org which is the URL currently used by the math extension
See also
https://phabricator.wikimedia.org/T87390

physikerwelt:
I would add to the commit message that the input type can be for example tex or inline-tex which defines the size of the LaTeX operators.

config.example.yaml
d00rman: We can default to whichever you think it's best as long as the instance is regularly kept updated.

physikerwelt: I'd vote for http://mathoid.testme.wmflabs.org/
Note the port here is 80 and not 10042

d00rman:
Outdated, so closing. Will open a new one.

PR #304, opened 2015-08-12
Added support for http HEAD method

Pchelolo:
If the HEAD request comes, we check for a handler, and if it's not found, check for a GET handler and run a request. Upon response - remove the body and content-length headers.

### DIFF
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -39,6 +39,7 @@ function RESTBase(options, req) {
         this._recursionDepth = par._recursionDepth + 1;
         this._priv = par._priv;
         this.rb_config = this._priv.options.conf;
+        this._rootReq = par._rootReq || req;
     } else {
         // Brand new instance
         this.log = options.log; // Logging method
@@ -65,6 +66,7 @@ function RESTBase(options, req) {
 
         this.rb_config = options.conf;
         this.rb_config.user_agent = this.rb_config.user_agent || 'RESTBase';
+        this._rootReq = null;
     }
 
 }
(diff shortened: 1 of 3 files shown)


lib/restbase.js
```diff
@@ -243,6 +243,10 @@ RESTBase.prototype._request = function(req) {
     var match = priv.router.route(childReq.uri);
     var methods = match && match.value && match.value.methods;
     var handler = methods && (methods[childReq.method] || methods.all);
+    if (childReq.method === 'head' && !handler) {
```
d00rman: I'm wondering if we should store the original request's method (in the Restbase object) and propagate it. Then we could decide on the right handler every time, not just until we fail for the first time. Consider a situation where the public endpoint, *PE*, calls an internal one, *S1*, which calls another one, *S2*. Now, if *S2* had both `GET` and `HEAD` handlers, but only `GET`s were registered for *S1*, *S2*'s `HEAD` handler would not be called. Instead, the `GET` one would be used, potentially entailing a wrongful execution. Hence, *S2* should be able to use the original request's method as well when deciding which handler to use.

Pchelolo: @d00rman Hm. Nice idea, but this would imply a little perf overhead with checking `originalMethod` on every request. Do you really see any viable use-cases where separate `HEAD` handler would be needed?

d00rman: Putting a Boolean will work just as well. If we change `childReq`'s method, we might as well set a simple Boolean and check it on each request. If the method hasn't changed, no Boolean property is set, which will evaluate to `false` (well, `not true` to be precise).

Pchelolo: But this would make the req obj not monomorphic, but, actually we can fight this problem too, so I do agree, lets do it, I'll prepare a new PR tomorrow

d00rman: No, it wouldn't if the boolean is placed in `childRestbase`. Then, the child instance could decide by looking at the value of `this._priv.magic_boolean`

Pchelolo: I mean we change the shape of an object depending on a condition === polymorphic object, but we can easily avoid that by machining `amagic_bool = null` in coonstructor

Pchelolo:
@d00rman Updated, now we also check if original request method was head. the _rootRequest part is needed in https://github.com/wikimedia/restbase/pull/272 , so it's fine to store the whole request

d00rman:
Nice one @Pchelolo !

PR #305, opened 2015-08-12
Added tests agains MW API agains test & test2 wikis

Pchelolo:
We need to perform tests related to MW API against test.wikipedia.org & test2.wikipedia.org too, as they contain newer versions of the API. 

I've chosen revisions tests and page_save tests and the work with MW API the most. Ideas which other tests to include are very welcome. 

2 tests in page_save are excluded, because they fail against test.wikipedia.org which is returning 'This action has been automatically identified as harmful.' Need to resolve it, but better do it separately.

Bug: https://phabricator.wikimedia.org/T103104

### DIFF
--- a/test/features/pagecontent/revisions.js
+++ b/test/features/pagecontent/revisions.js
@@ -11,7 +11,7 @@ var pagingToken = '';
 function generateTests(options) {
 
     var apiRequestTemplate = server.config.conf.templates['wmf-sys-1.0.0']
-                .paths['/{module:action}']['x-modules'][0]['options']['apiRequest'];
+                ['paths']['/{module:action}']['x-modules'][0]['options']['apiRequest'];
     var prevUri = apiRequestTemplate.uri;
     var prevHost = apiRequestTemplate.headers.host;
 
--- a/test/features/pagecontent/revisions.js
+++ b/test/features/pagecontent/revisions.js
@@ -8,36 +8,37 @@ var preq   = require('preq');
 var server = require('../../utils/server.js');
 var pagingToken = '';
 
-describe('revision requests', function() {
+function generateTests(options) {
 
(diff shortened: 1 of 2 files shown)


Pchelolo:
I had to revert changes in page_save tests, because we constantly get 429 error on test wikis and timeouts do not help.

test/features/pagecontent/revisions.js
```diff
@@ -8,36 +8,37 @@ var preq   = require('preq');
...
-    var pageLastRev = 653530930;
-
-    this.timeout(20000);
-
-    before(function () { return server.start(); });
+function generateTests(options) {
+
+    var apiRequestTemplate = server.config.conf.templates['wmf-sys-1.0.0']
+                .paths['/{module:action}']['x-modules'][0]
+                .options.apiRequest;
```
d00rman: Since this string is long already, I think it would be slightly more readable if, starting with `wmf-sys-1.0.0`, all of the properties were in *hash notation* (*`['prop']`*). That could provide a visual aid in identifying the complete path sought.

test/features/pagecontent/revisions.js
```diff
@@ -8,36 +8,37 @@ var preq   = require('preq');
...
+    var prevUri = apiRequestTemplate.uri;
+    var prevHost = apiRequestTemplate.headers.host;
+
+    before(function () {
+        apiRequestTemplate.uri = 'http://' + options.apiDomain + '/w/api.php';
```
d00rman: Better to use `HTTPS` right away to avoid an unnecessary redirect.

d00rman:
Nice work! A couple of nits, otherwise LGTM.

Pchelolo:
@d00rman updated, thank you

PR #307, opened 2015-08-16
Don't perform expensive no-cache instance tests

gwicke:
While doing some latency testing I noticed that the title Foobar was really
slow to load (500ms). I eventually traced this down to many thousands of
revision entries (possibly hundreds of thousands), which are caused by x-ample
tests forcing revision refreshes with no-cache.

Looking at the config, I found another x-ample that continuously re-renders
Foobar's HTML. This only avoided filling up revision storage as the content of
Foobar does not change much between renders, and we have implemented an
optimization that avoids storing unchanged revisions.

In both cases, there is really not much of a point in testing re-renders
separately for each instance. We have plenty of data on backend connections
that we can set alerts on. Having RESTBase alerts go off when the Action API
is in trouble is confusing, which is the last thing we need when things aren't
working so well.

### DIFF
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -104,23 +104,23 @@ paths:
           if-unmodified-since: '{if-unmodified-since}'
       x-monitor: true
       x-amples:
-        - title: Get rev of by title from MW
-          request:
-            params:
-              title: Foobar
-            headers:
-              cache-control: no-cache
-          response:
-            status: 200
-            headers:
-              etag: /.+/
-              content-type: application/json
-            body:
-              items:
-                - title: 'Foobar'
-                  rev: /\d+/
-                  tid: /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/
-                  comment: /.*/
+        #- title: Get rev of by title from MW
+        #  request:
+        #    params:
+        #      title: Foobar
+        #    headers:
+        #      cache-control: no-cache
+        #  response:
+        #    status: 200
+        #    headers:
+        #      etag: /.+/
+        #      content-type: application/json
+        #    body:
+        #      items:
+        #        - title: 'Foobar'
+        #          rev: /\d+/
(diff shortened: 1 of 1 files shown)


lavagetto:
@gwicke makes sense, +1

Health checks should just verify that the software is running correctly, and be idempotent (meaning they don't cause any modification with their calls).

Pchelolo:
@gwicke LGTM

d00rman:
This is sort of a nit-pick, but ...

> Health checks should just verify that the software is running correctly, and be idempotent (meaning they don't cause any modification with their calls).

... I also think they should be stateless, in that they do not depend on the state of the underlying storage. The first `/page/title/{title}` `no-cache` `x-ample` (i.e. the one commented out in this PR) guarantees that as it'll get the revision info. The tests are passing now (and will do so in production as well) because other tests, executed before the monitoring ones, obtain revision info for Foobar. Long story short, nothing will break if we merge this, but the monitoring will not potentially work on a completely fresh RESTBase install (as in, installing a new RB cluster with empty storage). Therefore, I'll merge this one as we can revisit this later.

> Having RESTBase alerts go off when the Action API is in trouble is confusing, which is the last thing we need when things aren't working so well.

RESTBase functions correctly *iff* its interaction with back-end services (the number of which is growing, yay) is working, so I'd say this is precisely what we want. On that note, all of the monitoring `x-amples` test precisely that - that the storage, MW API, Parsoid, Graphoid and MobileApps respond as intended and that RESTBase handles the responses as it should.

gwicke:
> Long story short, nothing will break if we merge this, but the monitoring will not potentially work on a completely fresh RESTBase install

If the revision is not in storage, then it will be fetched transparently. The `no-cache` header is not needed to make this happen.

> RESTBase functions correctly iff its interaction with back-end services (the number of which is growing, yay) is working, so I'd say this is precisely what we want

I guess the question is whether the RESTBase tests should cover our entire infrastructure, or whether they should focus on RESTBase itself. From a 'alerts have gone off' perspective, I think the latter is less confusing. It's better to set up more focused alerts per backend (based on our normal metrics), and clearly say 'Action API error rate exceeded' or 'Parsoid latency threshold exceeded', rather than 'RESTBase broken'.

PR #308, opened 2015-08-17
Updated spec to be supported in newer swagger-ui

Pchelolo:
Mobileapps routes has been added recently, we've hit a swagger bug there, so it was decided to upgrade our swagger-ui fork from upstream. 

Now a complete mess of forks and branches begin:
1. A branch with an upgraded swagger-ui is [here](https://github.com/wikimedia/swagger-ui/tree/us) but it's not mergeable with master, because it was made by reset to forks `HEAD~4` (undo all stuff in fork, because it's not needed any more), pull upstream up to [the latest commit](https://github.com/swagger-api/swagger-ui/commit/8e72906c911a02853b87f5241d157e83e2bf7377) to the latest 2.1.1 release, and a new commit, that uses our [new fork for swagger-js](https://github.com/wikimedia/swagger-js), because the functionality we need to patch is now in that lib. This PR is not mergable, I could either make mergable by adding reverse patches, or just rewrite master with -F, alternatively we could point to the newer brunch.
2. Restbase docs spec was a little bit wrong and wasn't parable by newer version, so a couple of simple fixes added, also our dynamic overrides needed to be patched a little bit
3. The desired functionality, that was added to [the fork](https://github.com/wikimedia/swagger-ui/pull/1), now moved [here](https://github.com/wikimedia/swagger-js/pull/1)

Sorry, it's a complete mess of PRs, repos and forks, but swagger guys moved the functionality we need to patch deeper, so now we have 2 levels of forks, but all works really well now. 

Bug: https://phabricator.wikimedia.org/T109311

### DIFF
--- a/lib/swaggerUI.js
+++ b/lib/swaggerUI.js
@@ -21,9 +21,9 @@ function staticServe(restbase, req) {
                         '<title>RESTBase docs</title>')
                 // Replace the default url with ours, switch off validation &
                 // limit the size of documents to apply syntax highlighting to
-                .replace(/sorter *: "alpha"/, 'sorter: "alpha", ' +
-                    'url: "?spec", validatorUrl: null, ' +
-                    'highlightSizeThreshold: 10000, docExpansion: "list",');
+                .replace(/Sorter: "alpha"/, 'Sorter: "alpha", ' + 'validatorUrl: null, ' +
+                    'highlightSizeThreshold: 10000, docExpansion: "list"')
+                .replace(/ url: url,/, 'url: "?spec",');
         }
 
         var contentType = 'text/html';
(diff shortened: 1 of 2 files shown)


gwicke:
Just tested this locally by patching package.json to point to the new branch, and it looks good to me.

@Pchelolo @eevans @d00rman, if you don't have any objections I'll merge this later, and force-push the swagger-ui branch to match the 'us' branch.

Pchelolo:
@gwicke regarding the force-push - that was my original plan to proceed, so I'm OK with that. Also I'll subscribe to swagger-ui & swagger-js releases to keep our forks in sync (so that we never get 400 commits back again)

gwicke:
Did the force-push and verified that a new install works as expected. Nice work, @Pchelolo !

PR #309, opened 2015-08-20
Add example config for local wikis

gwicke:
Add an example config for typical local RESTBase installs:

- use `http://localhost/w/api.php` as the API URI
- use `localhost` as the domain
- use `http://localhost:8142/` as the parsoid url (matching the Debian
  package, plus a comment about using port 8000 if using a git checkout)
- configure the sqlite backend

Task: https://phabricator.wikimedia.org/T109669

### DIFF
--- a/config.example.localhost.yaml
+++ b/config.example.localhost.yaml
@@ -1,0 +1,116 @@
+# RESTBase config for small wiki installs
+#
+# - sqlite backend
+# - parsoid at http://localhost:8142
+# - wiki at http://localhost/w/api.php
+# 
+# Quick setup:
+# - npm install
+#   If you see errors about sqlite, you might have to `apt-get install
+#   libsqlite3-dev`.
+# - cp config.example.localhost.yaml config.yaml
+# - double-check and possibly modify lines marked with XXX, then start restbase with
+#
+#   node server
+#
+# - If all went well, http://localhost:7231/localhost/v1/page/html/Main_Page
+# should show your wiki's [[Main Page]].
+
+
+info:
+  name: restbase
+
+templates:
+  # The public API is defined here.
+  wmf-content-1.0.0: &wp/content/1.0.0
+    swagger: '2.0'
+    info:
+      version: 1.0.0-beta
+      title: Wikimedia REST API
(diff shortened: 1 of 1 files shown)


gwicke:
@Krenair proposed to make a simple config like this the new `config.example.yaml`. I think that is a good idea, is it should help third-party users to get started without being confused by mobile app services, cassandra etc. We could consider renaming the current .example to config.wmf.example.yaml, or something along those lines. @d00rman @Pchelolo @eevans, any thoughts?

config.example.localhost.yaml
```diff
@@ -0,0 +1,116 @@
+# RESTBase config for small wiki installs
+#
+# - sqlite backend
+# - parsoid at http://localhost:8142
+# - wiki at http://localhost/w/api.php
+# 
+# Quick setup:
+# - npm install
+# # needs libsqlite3-dev:
+# - npm install restbase-mod-table-sqlite
```
d00rman: `npm install` already does that

config.example.localhost.yaml
```diff
@@ -0,0 +1,116 @@
...
+# - cp config.example.localhost.yaml config.localhost.yaml
+# - double-check and possibly modify lines marked with XXX, then start restbase with
+#
+#   node server -c config.localhost.yaml
```
d00rman: Since this is for small installs, it'd be probably easier for them to:

```
ln -s config.localhost.yaml config.yaml
nodejs server.js
```

d00rman:
Very minor nits, LGTM otherwise. This is more or less the config we are actually supplying for MW-Vagrant.

config.example.localhost.yaml
gwicke: Strangely, it did not for me the first time round. It did work in the second attempt, though.

config.example.localhost.yaml
gwicke: Yeah.

gwicke:
@d00rman @Pchelolo, any objections against moving this to config.example.yaml?

Pchelolo:
@gwicke and moving config.examle.yaml -> config.<xxx mwf?>.example.yaml?

PR #311, opened 2015-08-23
Quote the profile in content-type headers

cscott:
According to RFC 7231 section 3.1.1, parameter values are either tokens or quoted strings.  The forward-slash character ("/") is not allowed in a token (defined in RFC 7230, section 3.2.6).  We need to make the parameter value a quoted string.

This parsing restriction is actually enforced by the npm `content-type` module (version 1.0.1) which is used by Express 4.x to manipulate content type headers.  The unmodified content-type causes an exception to be thrown.

See also: https://gerrit.wikimedia.org/r/233188

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -263,9 +263,9 @@ services:
 
 test:
   content_types:
-    html: text/html;profile=mediawiki.org/specs/html/1.1.0;charset=utf-8
-    data-parsoid: application/json;profile=mediawiki.org/specs/data-parsoid/0.0.1;charset=utf-8
-    wikitext: text/plain;profile=mediawiki.org/specs/wikitext/1.0.0;charset=utf-8
+    html: text/html;profile="mediawiki.org/specs/html/1.1.0";charset=utf-8
+    data-parsoid: application/json;profile="mediawiki.org/specs/data-parsoid/0.0.1"
+    wikitext: text/plain;profile="mediawiki.org/specs/wikitext/1.0.0";charset=utf-8
 
 logging:
   name: restbase
--- a/doc/HandlerConfiguration.md
+++ b/doc/HandlerConfiguration.md
@@ -18,7 +18,7 @@ Example for a bucket handler:
       default:
         description: Unexpected error
         schema: { $ref: Error }
-    produces: text/html;profile=mw.org/specs/html/1.0
+    produces: text/html;profile="mw.org/specs/html/1.0"
 
     request_handler:
     - if:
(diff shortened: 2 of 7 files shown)


d00rman:
Makes sense and LGTM, but we'll have to wait for the production Parsoid to be updated

gwicke:
:+1: 

Those profile strings could use some standardization. Should we switch to "mediawiki.org/...." ?

cscott:
+1 from me on `mediawiki.org` without the protocol.

Unless we actually want to treat the profile as a URI and host a placeholder page there, in which case protocol would be necessary (and should be https in all cases).  Is there a standard for "profile" parameters?

gwicke:
I am not aware of any standard that absolutely spells it out for the Content-Type header, but examples like [this](http://amundsen.com/media-types/collection/format/) do use a full href / url. We could use "https://mediawiki.org/..." if we care about locking down the protocol. I am not aware of any automated use of those URLs, and the content at the other end does not need to be in any particular format. In practice, either "mediawiki.org/..." or "https://mediawiki.org/..." will probably work.

cscott:
The purpose of putting a placeholder page there is just to act as a de facto reservation.  If everyone demonstrates control over the URL of the profile (by putting up a page at the URL) before publishing material referencing it, then there's no chance that two independent mediawiki teams could inadvertently choose the same profile string for two different purposes.

Of course, this is more important for (say) ibm.com with tens of thousands of employees than it is for mediawiki.org.  But it's still good practice to (at least) demonstrate control of the referenced URL in some way, or (at best) describe the profile.  A page on mediawiki.org would be a really good URL string.

If you don't plan to host content related to the profile accessible via https, then you are really using a URN not a URL.  URNs don't have protocol portions.

cscott:
Production parsoid is updated.  That doesn't seem to be sufficient to make the test suite pass, though. Maybe some caching involved?

Pchelolo:
@cscott RB caches are dropped before testing. Looks like now all failing tests are expecting data-parsoid content-type as `application/json;profile=\"mediawiki.org/specs/data-parsoid/0.0.1\";charset=utf-8`, but I don't see if it's updated in [parsoid fix](https://gerrit.wikimedia.org/r/#/c/233188/). Should that be updated in parsoid too?

cscott:
@Pchelolo good catch!  Fixed in https://gerrit.wikimedia.org/r/233601 ; hopefully we'll get that deployed tomorrow.

gwicke:
@cscott, I was wondering if we should move the specs to actual wiki pages, for example https://mediawiki.org/wiki/Specs/html/1.0.0. We could simply store a snapshot of that spec version there. Releasing a new spec version would involve creating a new subpage, and perhaps updating a changelog.

In any case, we should perhaps take that discussion to phabricator. It's pretty orthogonal to this patch.

cscott:
+1 on T110126.  Just for sanity's sake, though, I think we should merge this first (since the parsoid side of the changes have already been merged and will be deployed tomorrow), and then treat T110126 as a separate follow-on task.

gwicke:
> I think we should merge this first

Yes, absolutely. Just waiting for the Parsoid deploy, so that we can make sure that this passes.

cscott:
Parsoid has been deployed (with a cherry-pick fix).  I updated the patch slightly to remove a `charset=utf-8` which gwicke didn't like.  Let's see if travis likes this now!

gwicke:
This has been rebased & merged in #313.

Thank you, @cscott !

PR #312, opened 2015-08-24
Use the Parsoid v3 API

cscott:
Needs love & attention wrt to the sections API.  Parsoid v3 returns section offsets without the <body> wrapper; some of this code might need to be updated accordingly?

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -265,7 +265,7 @@ test:
   content_types:
     html: text/html;profile="mediawiki.org/specs/html/1.1.0";charset=utf-8
     data-parsoid: application/json;profile="mediawiki.org/specs/data-parsoid/0.0.1"
-    wikitext: text/plain;profile="mediawiki.org/specs/wikitext/1.0.0";charset=utf-8
+    wikitext: text/plain; charset=utf-8; profile="mediawiki.org/specs/wikitext/1.0.0"
 
 logging:
   name: restbase
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -141,7 +141,7 @@ PSP.pagebundle = function(restbase, req) {
     // TODO: Pass in current or predecessor version data if available
     var newReq = Object.assign({}, req);
     if (!newReq.method) { newReq.method = 'get'; }
-    newReq.uri = this.parsoidHost + '/v2/' + domain + '/pagebundle/'
+    newReq.uri = this.parsoidHost + '/' + domain + '/v3/page/pagebundle/'
         + encodeURIComponent(rbUtil.normalizeTitle(rp.title)) + '/' + rp.revision;
     return restbase.request(newReq);
 };
@@ -511,11 +511,8 @@ PSP.transformRevision = function(restbase, req, from, to) {
 
 PSP.callParsoidTransform = function callParsoidTransform(restbase, req, from, to) {
     var rp = req.params;
-    // Parsoid currently spells 'wikitext' as 'wt'
     var parsoidTo = to;
-    if (to === 'wikitext') {
-        parsoidTo = 'wt';
-    } else if (to === 'html') {
+    if (to === 'html') {
         // Retrieve pagebundle whenever we want HTML
         parsoidTo = 'pagebundle';
     }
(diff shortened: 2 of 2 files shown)


cscott:
Parsoid v3 has been deployed now.

mods/parsoid.js
```diff
@@ -618,7 +614,7 @@ PSP.makeTransform = function(from, to) {
             var innerRes = res.body[to];
             innerRes.status = 200;
             // Handle bodyOnly flag
-            if (to === 'html' && req.body.bodyOnly) {
+            if (from === 'sections' && to === 'html' && req.body.bodyOnly) {
```
Pchelolo: Why is that added? In the [docs](https://rest.wikimedia.org/en.wikipedia.org/v1/?doc#!/Transforms/post_transform_html_to_html_title_revision) we state that 'bodyOnly' is valid for other `from` formats as well.

mods/parsoid.js
```diff
@@ -618,7 +614,7 @@ PSP.makeTransform = function(from, to) {
             var innerRes = res.body[to];
```
Pchelolo: The bug is around here: for `to === 'wikitext'` Parsoid returns an unwrapped response.  From what I can tell from testing, this should look like

```
var innerRes = {};
if (to === 'wikitext') {
    innerRes.body = res.body;
    innerRes.headers = {
        'content-type': res.headers['content-type']
    }
} else {
    innerRes = res.body[to];
}
```
And also the content-type header is different from what we expect, the charset is the second param, not the third.

Pchelolo:
@cscott Unfortunately the tests still fail. I've left a couple of comments inline

mods/parsoid.js
cscott: Because you pass `bodyOnly` along in the request to Parsoid, and (in v3) it has already stripped the body tag from the output when you get it back.

But the `sections` stuff generates its own body tag around the output, which then has to be manually stripped here.  This is a bug, but like I said, I don't fully understand what the `sections` stuff is doing and could use some help.  I'm worried I could screw up the offsets if I just blindly remove `<body>` from the code which generates the sections output.

mods/parsoid.js
cscott: Yes, express reorders the charset.  Can't be helped.

cscott:
Updated the patch and the tests pass.  Could still use some review of the `sections` stuff, since I suspect that the need for the `cheapBodyInnerHTML` function could be removed entirely with some refactoring.  But perhaps that can be a follow-up patch by someone who understands `sections` better than I.

mods/parsoid.js
```diff
@@ -615,10 +611,18 @@ PSP.makeTransform = function(from, to) {
         return transform
         .then(function(res) {
             // Unwrap to the flat response format
-            var innerRes = res.body[to];
+            var innerRes = {};
+            if (to === 'wikitext') {
```
gwicke: Why is there a special case for wikitext?

mods/parsoid.js
```diff
@@ -615,10 +611,18 @@ PSP.makeTransform = function(from, to) {
...
+                innerRes.headers = {
+                    'content-type': res.headers['content-type']
+                };
+            } else {
+                innerRes = res.body[to];
+            }
             innerRes.status = 200;
             // Handle bodyOnly flag
-            if (to === 'html' && req.body.bodyOnly) {
+            if (from === 'sections' && to === 'html' && req.body.bodyOnly) {
```
gwicke: Why is this limited to `from === 'sections'`?

cscott: See above: because the section extraction code adds a <body> around its result, which then has to be stripped off.  Parsoid no longer does this.  Help is requested to fix up the section extraction code, since I don't understand it well enough (and my daughter is going to be born within a matter of hours now).

mods/parsoid.js
cscott: It should probably be `if (to !== 'pagebundle')` instead.  The `html` and `wikitext` formats return the results in the main body of the response (as does RESTBase).  It's only the `pagebundle` format that returns a JSON object wrapping the content instead.  Again, help refactoring would be appreciated if you want to turn the default case around and specially-handle only the pagebundle case.

config.test.yaml
```diff
@@ -265,7 +265,7 @@ test:
   content_types:
     html: text/html;profile="mediawiki.org/specs/html/1.1.0";charset=utf-8
     data-parsoid: application/json;profile="mediawiki.org/specs/data-parsoid/0.0.1"
-    wikitext: text/plain;profile="mediawiki.org/specs/wikitext/1.0.0";charset=utf-8
+    wikitext: text/plain; charset=utf-8; profile="mediawiki.org/specs/wikitext/1.0.0"
```
gwicke: This seems like a random change. Could you keep the content-type as emitted by Parsoid the same?

cscott: No, it's not me reordering it, it's express.  It would require an upstream patch to express.

gwicke: I see. In that case, I guess we should change our content types across the board, and create similar normalization logic for compatibility checks.

cscott: You can use the same `content-type` npm module to do the normalization (and the comparison, I think).

gwicke: This is now tracked in https://phabricator.wikimedia.org/T116333.

Pchelolo:
This PR is outdated, closing it in favour of https://github.com/wikimedia/restbase/pull/395

PR #313, opened 2015-08-25
Fix profile quoting to match latest Parsoid

gwicke:
This rebases and updates #313 to match latest Parsoid output.


eevans:
LGTM

PR #315, opened 2015-08-26
Support $.request.hash template

Pchelolo:
A simple fix to support calculation of `'{$.request.hash}'`  templates to calculate hashes of the request.

### DIFF
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -83,7 +83,17 @@ function compileTemplate(templateSpec, setValue, reqPart) {
     // Special case - lazy-evaluation of a request hash property
     if (templateSpec === '{$.request.hash}') {
         return function(newReq, context) {
+            // Need to remove x-request-id header if it's present
+            var reqId = context.request.headers && context.request.headers['x-request-id'];
+            if (reqId) {
+                delete context.request.headers['x-request-id'];
+            }
+
             setValue(newReq, sha1(stringify(context.request)));
+
+            if (reqId) {
+                context.request.headers['x-request-id'] = reqId;
+            }
         };
     }
 
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -4,6 +4,8 @@ var URI = require('swagger-router').URI;
 var rbUtil = require('./rbUtil');
 var TAssembly = require('tassembly');
 var url = require('url');
+var sha1 = require('sha1');
+var stringify = require('json-stable-stringify');
 
 /**
  * Creates a function that sets a value at the path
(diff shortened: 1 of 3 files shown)


lib/reqTemplate.js
```diff
@@ -159,6 +162,21 @@ function _createTemplateResolvers(origSpec, subspec, reqPart, path) {
...
 }
 
+function calculateHash(context) {
+    // Need to remove x-request-id header if it's present
+    var reqId = context.request.headers && context.request.headers['x-request-id'];
+    if (reqId) {
+        delete context.request.headers['x-request-id'];
+    }
+
+    var result = sha1(stringify(context.request));
```
d00rman: We should ensure consistent hashing. That means checking for standard request components (uri, headers, body, query) so as to avoid different hashes in situations where some are missing. E.g., these two requests will yield different hashes even though they should be treated equally:

```yaml
request1:
  uri: /a/simple/path
  headers: {}
  body: ''

request2:
  uri: /a/simple/path
  body: {}
```

test/features/router/misc.js
```diff
@@ -259,6 +259,73 @@ describe('router - misc', function() {
...
+        var result1 = requestTemplate.eval({
+            request: {
+                method: 'post',
+                headers: {
+                    'x-request-id': '10'
+                },
+                body: 'a'
+            }
+        });
+        assert.deepEqual(result1.body, '19a5337cc49833b0923ac4b6d72744bf8e915de9');
```
d00rman: Instead of asserting equivalence with a specific SHA1, it'd be better to compare the hashes of `result1` and `result2`, as we are really interested in the fact they are equal.

lib/reqTemplate.js
```diff
@@ -159,6 +162,28 @@ function _createTemplateResolvers(origSpec, subspec, reqPart, path) {
...
+function calculateHash(context) {
+    var req = context.request;
+    // Ensure consistent hashing
+    req.headers = req.headers || {};
+    req.body = req.body || {};
+    req.query = req.query || {};
+    req.params = req.params || {};
+
+    // Need to remove x-request-id header if it's present
+    var reqId = req.headers && req.headers['x-request-id'];
```
d00rman: Not necessary to check for `req.headers` now that we ensure the property exists.

gwicke:
It might be cleaner to implement this as a thin wrapper module on top of a key_value bucket, perhaps `post_value`? The hash would be calculated on POST, and there would be one entry point to retrieve by hash. The bucket name would be passed through to the underlying key_value bucket.

Pchelolo:
@gwicke Yes, that's true, we've got out of the original plan. I will update the PR to implement a post_data module.

Pchelolo:
Actually, the new approach is completely different and not related to this PR, so I'd better close this one and create a new one with `post_data` module.

PR #316, opened 2015-08-31
Made swagger spec more valid

Pchelolo:
There is an [online tool](https://github.com/swagger-api/validator-badge) for swagger spec validation. I shows errors on our spec, so in this PR it's made more valid. One of the visible consequences: now the [/page/revision/{revision}](https://rest.wikimedia.org/en.wikipedia.org/v1/?doc#!/Page_content/get_page_revision_revision) endpoint docs correctly show the result items spec. 

Additionally, fixed a small issue in router - it didn't verify, that the 'prefixPath' exist, so all the contents of the spec `info` field ended up in the paths object under `""` key. 

Unfortunately, the spec validator doesn't support optional path elements (e.g. `{/tid}`), so our spec is still invalid even after this fix, so we can't put a nice valid badge on the spec page. We could fork the validator and add support for that, but it's too low-priority task to do it.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -223,7 +223,7 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
     // Process HTTP method stanzas ('get', 'put' etc)
     .then(function() {
         // Register the path in the specRoot
-        if (specRoot && !specRoot.paths[prefixPath]) {
+        if (specRoot && !specRoot.paths[prefixPath] && prefixPath) {
             specRoot.paths[prefixPath] = {};
         }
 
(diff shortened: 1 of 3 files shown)


Pchelolo:
@gwicke Regarding yesterday's question about prefixPath. It's empty, iff the spec is root, or loaded as an `x-subspec`. In both cases the `pathspec` variable contains a complete swagger spec, so we can't have any method definitions on the top level there. So everything is fine with the current solution.

gwicke:
@Pchelolo, so in that case, the spec we are trying to load is invalid? Should we warn or error out?

Pchelolo:
@gwicke no, it's valid. It can have properties (like `info`), but it can't have method definitions on top level. This means we won't loose anything with this check.

PR #317, opened 2015-08-31
Implemented prototype of varnish purge for testing and benchmarking.

Pchelolo:
NOTE: This is an experimental PR, added only for testing and running benchmarks of mobile-apps endpoints. This code should not get to production and this PR should be reverted when testing/benchmarks are done, because without a proper dependency tracking system we can't create a real solution for this issue.

Here we enable varnish caches for mobile apps endpoints and set up background HTCP purges when the page is updated. Ideally, nothing should be hard-coded, and dependencies between endpoints should be stored in the dependency tracking system. 

This PR is only for testing, so I think making the code better is worthless as it will be removed soon. Correct me if I'm wrong, and I'll make this completely spec-driven without hard-coded endpoints and varnish IPs

Bug: https://phabricator.wikimedia.org/T109742

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -51,13 +51,13 @@ function ParsoidService(options) {
 
     // THIS IS EXPERIMENTAL AND ADDED FOR TESTING PURPOSE!
     // SHOULD BE REWRITTEN WHEN DEPENDENCY TRACKING SYSTEM IS IMPLEMENTED!
-    self.purgeOnUpdate = options.purge_on_update;
+    self.purgeOnUpdate = !!options.purge;
     if (self.purgeOnUpdate) {
         self.purger = new Purger({
             routes: [
                 {
-                    host: '239.128.0.112',
-                    port: 4827
+                    host: options.purge.host || '239.128.0.112',
+                    port: options.purge.port || 4827
                 }
             ]
         });
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -21,36 +21,21 @@ var Template = require('../lib/reqTemplate');
 var cacheURIs = [
     // /page/mobile-html/{title}
     new Template({
-        uri: 'https://rest.wikimedia.org/{domain}/v1/page/mobile-html/{title}'
-    }),
-    new Template({
         uri: 'https://{domain}/api/rest_v1/page/mobile-html/{title}'
     }),
     // /page/mobile-html-sections/{title}
-    new Template({
-        uri: 'https://rest.wikimedia.org/{domain}/v1/page/mobile-html-sections/{title}'
-    }),
     new Template({
         uri: 'https://{domain}/api/rest_v1/page/mobile-html-sections/{title}'
     }),
     // /page/mobile-html-sections-lead/{title}
     new Template({
-        uri: 'https://rest.wikimedia.org/{domain}/v1/page/mobile-html-sections-lead/{title}'
-    }),
-    new Template({
         uri: 'https://{domain}/api/rest_v1/page/mobile-html-sections-lead/{title}'
     }),
     // /page/mobile-html-sections-remaining/{title}
     new Template({
-        uri: 'https://rest.wikimedia.org/{domain}/v1/page/mobile-html-sections-remaining/{title}'
-    }),
-    new Template({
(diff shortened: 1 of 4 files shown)


mods/parsoid.js
```diff
@@ -14,6 +14,28 @@ var yaml = require('js-yaml');
 var fs = require('fs');
 var spec = yaml.safeLoad(fs.readFileSync(__dirname + '/parsoid.yaml'));
 
+// THIS IS EXPERIMENTAL AND ADDED FOR TESTING PURPOSE!
+// SHOULD BE REWRITTEN WHEN DEPENDENCY TRACKING SYSTEM IS IMPLEMENTED!
+var Purger = require('htcp-purge');
+var Template = require('../lib/reqTemplate');
+var cacheURIs = [
+    // /page/mobile-html/{title}
+    new Template({uri: 'https://rest.wikimedia.org/{domain}/v1/page/mobile-html/{title}'}),
```
gwicke: I don't think we should care about the rest.wikimedia.org entry points. Those are deprecated & will be switched off.

gwicke:
It would be nice if we could switch purging on/off with a flag, defaulting to false. That way, we could start testing this in staging without a need to revert it before the next deploy.

mods/parsoid.js
Pchelolo: @gwicke we've chatted with bblack about it and he suggested to purge them too, as he's not sure which URIs are used by varnish - original or rewritten.

mods/simple_service.js
```diff
@@ -325,7 +326,14 @@ SimpleService.prototype.processSpec = function(spec) {
                 var context = {
                     request: req
                 };
-                return generatePromiseChain(restbase, context, requestChain);
+                return generatePromiseChain(restbase, context, requestChain)
+                .then(function(res) {
+                    if (self.cache && res && /^2..$/.test('' + res.status)) {
```
gwicke: This will need some rework before we can enable it in prod. Could you add a todo comment to that effect, so that we don't forget to re-do this?

mods/parsoid.js
```diff
@@ -14,13 +14,70 @@ var yaml = require('js-yaml');
...
 var spec = yaml.safeLoad(fs.readFileSync(__dirname + '/parsoid.yaml'));
 
+// THIS IS EXPERIMENTAL AND ADDED FOR TESTING PURPOSE!
+// SHOULD BE REWRITTEN WHEN DEPENDENCY TRACKING SYSTEM IS IMPLEMENTED!
+var Purger = require('htcp-purge');
+var Template = require('../lib/reqTemplate');
+var cacheURIs = [
+    // /page/mobile-html/{title}
+    new Template({
+        uri: 'https://rest.wikimedia.org/{domain}/v1/page/mobile-html/{title}'
```
gwicke: Could you drop the rest.wikimedia.org entries here?

gwicke:
Before we enable this, it would be useful to test the HTCP purge logic manually (with a small test script around the module) in labs. Afaik there is a HTCP purge daemon running there (possibly unicast, though), along with Varnishes we can log requests from (by running `vanishncsa`). By testing in labs first, we can make sure that the HTCP daemon doesn't explode when it sees our lovingly crafted packets. Having that happen in production would make for a bit more hectic than necessary.

mods/parsoid.js
```diff
@@ -14,13 +14,55 @@ var yaml = require('js-yaml');
...
         || 'http://parsoid-lb.eqiad.wikimedia.org';
+
+    // THIS IS EXPERIMENTAL AND ADDED FOR TESTING PURPOSE!
+    // SHOULD BE REWRITTEN WHEN DEPENDENCY TRACKING SYSTEM IS IMPLEMENTED!
+    self.purgeOnUpdate = options.purge_on_update;
+    if (self.purgeOnUpdate) {
+        self.purger = new Purger({
+            routes: [
+                {
+                    host: '239.128.0.112',
```
d00rman: It would be nice to have the host and port configurable, so that we can test this in labs before going out with it. Have them as optional params to the parsoid module conf would be ideal

d00rman:
We have tested the `htcp-purge` package in labs and it works flawlessly. I'm merging this in so that I can conduct further tests in labs.

PR #318, opened 2015-09-01
Implemented post_data module as a simple wrapper over key_value

Pchelolo:
Implementation of a simple post_data module wrapping key_value bucket storage module.

Bug: https://phabricator.wikimedia.org/T105975

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -143,7 +143,7 @@ templates:
                             method: put
                             uri: /{domain}/sys/post_data/post.test/
                             headers: '{$.request.headers}'
-                            body: '{$.request}'
+                            body: '{$.request.body}'
                 /storage/{hash}:
                   get:
                     on_setup:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -141,7 +141,7 @@ templates:
                       - put_to_storage:
                           request:
                             method: put
-                            uri: /{domain}/sys/post_data/post.test/data
+                            uri: /{domain}/sys/post_data/post.test/
                             headers: '{$.request.headers}'
                             body: '{$.request}'
                 /storage/{hash}:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -128,7 +128,7 @@ templates:
                           request:
                             uri: http://graphoid.wikimedia.org/en.wikipedia.org/v1/png/{title}/{revision}/{graph_id}
 
-      /{module:post_data_sevice}:
+      /{module:post_data_service}:
         x-modules:
           - name: simple_service
             version: 1.0.0
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -141,9 +141,9 @@ templates:
                       - put_to_storage:
                           request:
                             method: put
-                            uri: /{domain}/sys/post_data/post.test/store
+                            uri: /{domain}/sys/post_data/post.test/data
                             headers: '{$.request.headers}'
-                            body: '{$.request.body}'
+                            body: '{$.request}'
                 /storage/{hash}:
                   get:
                     on_setup:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -91,6 +91,13 @@ templates:
             version: 1.0.0
             type: file
 
+      /{module:post_data}:
+        x-modules:
+          - name: post_data
+            version: 1.0.0
+            type: file
+
+
       /{module:key_rev_value}:
         x-modules:
           - name: key_rev_value
@@ -120,6 +127,33 @@ templates:
                       - get_from_graphoid:
                           request:
                             uri: http://graphoid.wikimedia.org/en.wikipedia.org/v1/png/{title}/{revision}/{graph_id}
+
+      /{module:post_data_sevice}:
+        x-modules:
+          - name: simple_service
+            version: 1.0.0
+            type: file
+            options:
+              paths:
+                /storage:
+                  put:
+                    on_request:
+                      - put_to_storage:
+                          request:
+                            method: put
+                            uri: /{domain}/sys/post_data/post.test/store
+                            headers: '{$.request.headers}'
+                            body: '{$.request.body}'
+                /storage/{hash}:
+                  get:
+                    on_setup:
+                      - init_storage:
+                          uri: /{domain}/sys/post_data/post.test
+                    on_request:
+                      - get_from_storage:
+                          request:
+                            uri: /{domain}/sys/post_data/post.test/{hash}
+
 
       /{module:mobileapps}:
         x-modules:
--- a/mods/post_data.js
+++ b/mods/post_data.js
@@ -90,7 +90,7 @@ PostDataBucket.prototype.putRevision = function(restbase, req) {
             'content-type': 'application/json',
             'cache-control': req.headers && req.headers['cache-control']
         },
-        body: req
+        body: req.body
     })
     .then(function(res) {
         return {
--- a/mods/post_data.js
+++ b/mods/post_data.js
@@ -8,7 +8,7 @@
 var yaml = require('js-yaml');
 var fs = require('fs');
 var spec = yaml.safeLoad(fs.readFileSync(__dirname + '/post_data.yaml'));
-var sha1 = require('sha1');
+var crypto = require('crypto');
 var stringify = require('json-stable-stringify');
 var URI = require('swagger-router').URI;
 
@@ -48,7 +48,10 @@ PostDataBucket.prototype.getRevision = function(restbase, req) {
         path.push(rp.tid);
     }
     return restbase.get({
-        uri: new URI(path)
+        uri: new URI(path),
+        headers: {
+            'cache-control': req.headers && req.headers['cache-control']
+        }
     });
 };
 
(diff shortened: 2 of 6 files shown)


mods/post_data.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+  /{bucket}/{key}/:
+    get:
+      operationId: listRevisions
+
+  /{bucket}/{key}{/tid}:
+    get:
+      operationId: getRevision
+
+  /{bucket}/data{/tid}:
```
Pchelolo: The `data` part is extremely ugly, but without a required `hash` parameter we can't distinguish this endpoint with bucket creation endpoint.

Pchelolo:
PS: shall library which I've used is the fastest one available according to my simple benchmarks.
PPS: I can't verify the value of returned sha1 hash, because in node 0.10 default `connection` header value is `close`, and in 0.12+ `keep-alive`.

mods/post_data.js
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Key-value bucket handler
+ */
+
+// TODO: move to separate spec package
+var yaml = require('js-yaml');
+var fs = require('fs');
+var spec = yaml.safeLoad(fs.readFileSync(__dirname + '/post_data.yaml'));
+var sha1 = require('sha1');
```
gwicke: Is this faster than the built-in https://nodejs.org/api/crypto.html ?

mods/post_data.yaml
gwicke: How about using POST to `/{bucket}/` ?

mods/post_data.js
```diff
@@ -0,0 +1,120 @@
...
+    if (reqId) {
+        req.headers['x-request-id'] = reqId;
+    }
+    return result;
+}
+
+PostDataBucket.prototype.putRevision = function(restbase, req) {
+    var rp = req.params;
+    var key = calculateHash(req);
+    return restbase.put({
```
gwicke: This should only happen if the content isn't stored yet.

Pchelolo:
@gwicke updated. Regarding the `crypto` module, it's funny how I always try to find something on npmjs without checking the built in solutions.. crypto is indeed way faster.

mods/post_data.js
```diff
@@ -0,0 +1,142 @@
...
+
+PostDataBucket.prototype.listRevisions = function(restbase, req) {
+    var rp = req.params;
+    return restbase.get({
+        uri: new URI([rp.domain, 'sys', 'key_value', rp.bucket, ''])
+    });
+};
+
+function calculateHash(req) {
+    // Ensure consistent hashing
```
gwicke: Since we are interested in POST data, how about hashing the body only? Do you see use cases that would benefit from storing / hashing the other request parts?

mods/post_data.yaml
```diff
@@ -0,0 +1,27 @@
...
+    put:
+      operationId: createBucket
+
+  /{bucket}/:
+    get:
+      operationId: listBucket
+
+  /{bucket}/{key}/:
+    get:
+      operationId: listRevisions
```
gwicke: Is there actually a use case for having multiple revisions of the same content?

mods/post_data.js
```diff
@@ -0,0 +1,142 @@
...
+
+    if (reqId) {
+        req.headers['x-request-id'] = reqId;
+    }
+    return result;
+}
+
+PostDataBucket.prototype.putRevision = function(restbase, req) {
+    var rp = req.params;
+    var key = calculateHash(req);
```
gwicke: It might be slightly more efficient to JSON.stringify only once, and pass that string to calculateHash instead.

gwicke:
> PPS: I can't verify the value of returned sha1 hash, because in node 0.10 default connection header value is close, and in 0.12+ keep-alive, so the resulting hashes are different.

Only hashing & storing the body would avoid this.

d00rman:
> Only hashing & storing the body would avoid this.

+1. We have discussed this a couple of weeks ago and me too, I'm inclined to hash the body only. It might be worth somehow integrating only the URI in there too, though, to avoid clashes like:

```yaml
r1:
  uri: /my/uri
  body:
    common_param1: val1
    common_param2: val2
r2:
  uri: /my/uri/optional_path
  body:
    common_param1: val1
    common_param2: val2
```

mods/post_data.js
Pchelolo: @gwicke And where do we stringify the second time?

mods/post_data.js
```diff
@@ -0,0 +1,124 @@
...
+        uri: new URI(path),
+        headers: {
+            'cache-control': req.headers && req.headers['cache-control']
+        }
+    });
+};
+
+function prepareStoredReq(req) {
+    var storedReq = {};
+    storedReq.uri = req.body && req.body.uri && req.body.uri.toString();
```
d00rman: Sightly more efficient and readable:

```js
var storedReq = { uri: '', body: {} };
if (!req.body) {
  return storedReq;
}
storedReq.uri = req.body.uri && req.body.uri.toString() || '';
storedReq.body = req.body.body || {};
```

This also makes me think - should we really allow `req.body.uri` to be left unspecified?

Pchelolo: @d00rman in a normal usage pattern req.body.uri should never be undefined, so I'd say let's throw an error in case `!req.body || !req.body.uri`

d00rman: Yeah, I'd add `req.body.body` to the list as well, the rationale being that if there is no body data, what is it that the user wants to store?

Pchelolo: @d00rman I could imagine a useful empty `req.body.body`. It could represent something like an empty graph, or an empty table (if we had some table-rendering service). So, although it's not used now, it's possible that empty body would be a legal request for some service.

d00rman: I do agree that `{}` might be a valid value, but I don't the same holds true for `undefined`. To me, supplying an empty value is really different from not supplying a value at all (note that `!req.body.body` will return `false` for `req.body.body = {}`)

gwicke:
Why would sharing post storage between two URLs be a bad thing?

Pchelolo:
@gwicke because we a sharing the content by a hash too, so if the resulting content is different depending on some optional path element, we would end up effectively ignoring it for content requests.

gwicke:
@Pchelolo, content is stored in different buckets, though. Even the post data might be one bucket per service, so that we can effectively keep track of usage & delete buckets that aren't needed any more.

Pchelolo:
@gwicke why in different buckets? Suppose we have `graphoid.png` bucket - we store all content there. Suppose also we would have some optional path parameter that affects the graphed result - this path parameter should be reflected in the hash. It's not the case now, but it could be. On the other hand, the same concern was about request headers - if at some point we have a header, that affects the content, generated by a service, it should go into the hash calculation. 

How about storing/hashing the whole (and unchanged) `req.body` in the post_data, and what gets there would be controlled by upper-level services? By default the would be `'{$.request.body}'`, but we will have flexibility in this case.

gwicke:
This task is just about storing the POST data, not about how that data is used. As discussed [in the task](https://phabricator.wikimedia.org/T105975), the POST data can later be dereferenced in order to convert a GET request (with headers, parameters, whatever) into a POST request. That content request can do whatever it wants, including parametrizing the content generation based on query strings, headers, or the URL.

Pchelolo:
@gwicke Updated. Now we store only the 'body' of a request, that came to POST data and store it completely. An up-level service can decide what to put there - the whole request, only body, or may be even some specific part of the body.

mods/post_data.js
```diff
@@ -0,0 +1,117 @@
...
+function calculateHash(storedData) {
+    return crypto.createHash('sha1')
+                 .update(stringify(storedData))
+                 .digest('hex');
+}
+
+PostDataBucket.prototype.putRevision = function(restbase, req) {
+    var rp = req.params;
+    var storedData = req.body || {};
+    var key = calculateHash(storedData);
```
gwicke: Minor: Perhaps stringify *before* passing the buffer to the hash function? That way you avoid having to stringify twice.

gwicke: Nevermind, since the valueType is defined as JSON we can't really avoid the second stringify.

config.test.yaml
```diff
@@ -121,6 +128,33 @@ templates:
                           request:
                             uri: http://graphoid.wikimedia.org/en.wikipedia.org/v1/png/{title}/{revision}/{graph_id}
 
+      /{module:post_data_sevice}:
```
gwicke: service

specs/test.yaml
```diff
@@ -34,6 +34,20 @@ paths:
                 headers:
                   content-type: /^text\/html.+/
 
+  /{module:post_data}/storage:
+    post:
+      x-backend-request:
+        method: put
+        uri: /{domain}/sys/post_data_sevice/storage
```
gwicke: service

specs/test.yaml
```diff
@@ -34,6 +34,20 @@ paths:
...
+        headers: '{$.request.headers}'
+        body: '{$.request.body}'
+      x-monitor: false
+
+  /{module:post_data}/storage/{hash}:
+    get:
+      x-backend-request:
+        uri: /{domain}/sys/post_data_sevice/storage/{hash}
```
gwicke: service

PR #319, opened 2015-09-02
Added page save api

Pchelolo:
Added public endpoints for Wikitext and HTML saving

Bug: https://phabricator.wikimedia.org/T101501


Pchelolo:
/cc @gwicke @d00rman @eevans 

PS. One problem with this path, is that with added tests we constantly get 429 from MW API. May be there's a way to increate a rate for us?

gwicke:
@Pchelolo, we ran into that issue before when we introduced sqlite & thus doubled the number of save tests. My proposal would be to switch the save tests to http://en.wikipedia.beta.wmflabs.org/wiki/Main_Page, as I *think* that wiki is not rate limited. It is also meant for this kind of testing, so we don't have to use our user pages. One downside is that this wiki will likely be less reliable than enwiki, but overall it will probably still be an okay trade-off.

Pchelolo:
This PR depends on new SQLite version: https://github.com/wikimedia/restbase-mod-table-sqlite/pull/16

specs/mediawiki/v1/content.yaml
```diff
@@ -289,69 +289,75 @@ paths:
...
-#            $ref: '#/definitions/problem'
-#      x-backend-request:
-#        uri: /{domain}/sys/page_save/html/{title}
+    post:
+      tags:
+        - Page content
+      description: >
+        Save a new revision of a page given in HTML format
+
+        Note that the title must exist. See [the MediaWiki DOM
```
gwicke: We should also support new page creation.

d00rman:
LGTM :+1:

mods/page_save.js
```diff
@@ -124,16 +165,20 @@ PageSave.prototype.saveHtml = function(restbase, req) {
...
     this._checkParams(req.body);
     // First transform the HTML to wikitext via the parsoid module
+    var path = [rp.domain, 'sys', 'parsoid', 'transform', 'html', 'to', 'wikitext', title];
+    var baseRevision = this._getBaseRevision(req);
+    if (baseRevision) {
+        path.push(baseRevision);
+    }
     return restbase.post({
-        uri: new URI([rp.domain, 'sys', 'parsoid', 'transform', 'html', 'to', 'wikitext', title]),
+        uri: new URI(path),
```
gwicke: This should also set `headers: { 'if-match': req.body.base_etag || headers['if-match'] }`, so that the html2wt end point can use that instead of the meta tag hack.

specs/mediawiki/v1/content.yaml
```diff
@@ -289,69 +289,75 @@ paths:
...
+        to detect edit conflicts.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      parameters:
+        - name: title
+          in: path
+          description: The page title.
+          type: string
+          required: true
+        - name: baseETag
```
gwicke: Should we use snake_case or CamelCase for post parameters? To me, snake_case seems potentially slightly less error prone, but we have started to use CamelCase in some instances (bodyOnly is the main one that comes to mind).

d00rman: I prefer *snake_case*, but `bodyOnly` makes me wonder. If we can identify all clients using this flag, then I'd go with *snake_case* and rename `bodyOnly` to `body_only`. Optionally, we could support both for a period (effectively deprecating `bodyOnly`).

Pchelolo:
@gwicke @d00rman Updated.

Pchelolo:
Updated. Renamed parameters according to what have been decided in the ticket, rebased on master.

lib/rbUtil.js
```diff
@@ -162,6 +162,22 @@ rbUtil.generateRequestId = function() {
     return uuid.now().toString();
 };
 
+// Create a uniform but shallow request object copy with sane defaults. This
```
d00rman: Minor nit: use `/** */` when commenting functions / methods.

mods/action.js
```diff
@@ -181,6 +181,13 @@ function buildEditResponse(res) {
 }
 
 ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
+    if (/\.test\.local$/.test(req.params.domain)) {
```
d00rman: We have to stop introducing these test hacks... Can't we have two action modules in the test config - one for en.wp and the other to be a *normal* one, and then use one or the other during the tests?

mods/page_save.js
```diff
@@ -36,11 +36,49 @@ function PageSave(options) {
     };
 }
 
-PageSave.prototype._getRevInfo = function(restbase, req) {
+PageSave.prototype._getStartTimestamp = function(req) {
+    if (req.headers && req.headers['if-match']) {
+        var etag = rbUtil.parseETag(req.headers['if-match']);
+        if (!etag || !TimeUuid.test(etag.tid) || !/^(?:[0-9]+)$/.test(etag.rev)) {
+            throw new rbUtil.HTTPError({
```
d00rman: I'm ambivalent here. On the one hand, it's good to inform the client their header is not correct. On the other, we default to using the current *ts*.

d00rman: On second thought, let's keep it as-is for now, it enforces correctness.

mods/page_save.js
```diff
@@ -36,11 +36,49 @@ function PageSave(options) {
...
+                }
+            });
+        } else {
+            return TimeUuid.fromString(etag.tid).getDate().toISOString();
+        }
+    }
+    return undefined;
+};
+
+PageSave.prototype._getBaseRevision = function(req) {
```
d00rman: Euh, isn't the base revision ID contained in the ETag?

Pchelolo: @d00rman The if-match header contains a latest revision, base revision is in base_etag.

mods/page_save.js
```diff
@@ -76,28 +114,30 @@ PageSave.prototype._checkParams = function(params) {
             body: {
                 type: 'invalid_request',
                 title: 'Missing parameters',
-                description: 'The html/wikitext and token parameters are required'
+                description: 'The html/wikitext and CSRF token parameters are required'
```
d00rman: `csrf_token` - better to supply the exact name sought.

specs/mediawiki/v1/content.yaml
```diff
@@ -289,69 +289,75 @@ paths:
...
+      consumes:
+        - multipart/form-data
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: The existing revision of the page matches the sent text
+        '201':
+          description: A new revision of the page has been created
+        default:
```
d00rman: Also add `400`

mods/page_save.js
d00rman: Nevermind me. All good.

specs/mediawiki/v1/content.yaml
```diff
@@ -743,69 +749,78 @@ paths:
...
-#        uri: /{domain}/sys/page_save/wikitext/{title}
+
+  /{module:page}/wikitext/{title}:
+    post:
+      tags:
+        - Page content
+      description: >
+        Save a new revision of a page.
+
+        For new pages, or when editting the latest revision of a page,
```
d00rman: This description is much clearer than the one for `POST /page/html/`. Perhaps integrate this there as well?

specs/mediawiki/v1/content.yaml
```diff
@@ -743,69 +749,78 @@ paths:
...
+      consumes:
+        - multipart/form-data
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: The existing revision of the page matches the sent text
+        '201':
+          description: A new revision of the page has been created
+        default:
```
d00rman: Idem, add `400`

mods/action.js
Pchelolo: @d00rman I don't like these too, but how do you imaging it? We would then need 2 different sys specs for different domains, which would be extremely verbose.

gwicke: I'm pretty sure you looked into this before, so excuse me asking again: what stops us from using http://en.wikipedia.beta.wmflabs.org/wiki/Main_Page for tests in general? That could potentially avoid the need for hacks like this, and additionally avoid running into the edit limits on enwiki when running tests (yes, sometimes Travis is fast....).

Pchelolo: @gwicke Switching all test to labs is a amount of work, because we need to create same pages as we use for testing, alter all revision numbers, and sometimes pages need to have a pretty specific history. Additionally, to avoid hacks like this there is another PR: https://github.com/wikimedia/restbase/pull/324

d00rman: I've just merged #324 so we can get rid of these hacks. We can discuss alternatives to it later. *beta* does seem like a good idea, especially now with shorter MW deployment trains.

mods/page_save.js
```diff
@@ -124,16 +165,23 @@ PageSave.prototype.saveHtml = function(restbase, req) {
...
+    }
     return restbase.post({
-        uri: new URI([rp.domain, 'sys', 'parsoid', 'transform', 'html', 'to', 'wikitext', title]),
+        uri: new URI(path),
         body: {
-            revision: req.body.revision,
             html: req.body.html
+        },
+        headers: {
+            'if-match': req.body.base_etag || req.headers['if-match']
```
d00rman: I think this should be added *iff* `req.body.base_etag || req.headers['if-match']` evaluates to a truth-y value.

d00rman: Ah, no, it should be ok, since now the transform endpoints require a tid to be supplied. Disregard.

d00rman:
LGTM, but I find the tests somewhat confusing - some use en.wp, others use beta, but this is part of another discussion.

Let's merge this and make some people happy?

Pchelolo:
@d00rman Practice shows, that en.wiki can survive 2 edit tests, if we add more we get 429. But en.wiki is more stable, than labs wiki, so we use it as much as we can -  2 tests on en.wiki, and 2 on labs. It's a compromise.

d00rman:
> Practice shows, that en.wiki can survive 2 edit tests, if we add more we get 429. But en.wiki is more stable, than labs wiki, so we use it as much as we can - 2 tests on en.wiki, and 2 on labs. It's a compromise.

Yeah, I know the rationale behind this choice, just stating it can be confusing and that we need to find an alternative, more stable solution. But, as I said, this PR is not the place to deal with that ;)

PR #320, opened 2015-09-02
Fill the latest_rev field

Pchelolo:
In page_revisions spec and table we have a `latest_rev` field, which might be useful. However now it's never set. This patch sets it when we get a new revision for a title, so for historic revisions it will still be `null`, but at least it will be set for newer ones. 

This is a (little) part of the work under https://phabricator.wikimedia.org/T106455


Pchelolo:
Please disregard this PR for some time, apparently adding a static column is supported in cassandra, so we can also fix and fill the `latest_tid` field.

gwicke:
I'll close this PR for now, as it doesn't seem to be needed. Please re-open if you disagree.

PR #323, opened 2015-09-09
Deprecate bodyOnly and replace it with body_only

Pchelolo:
After some discussion it's been decided that we need to use snake_case in POST request parameters. The only one that wasn't following this rule is `bodyOnly`. This PR deprecated the `bodyOnly` parameter and replaces it with `body_only`.

### DIFF
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -887,6 +887,11 @@ paths:
           required: true
         - name: bodyOnly
           in: formData
+          description: Deprecated. Use `body_only`
+          type: boolean
+          required: false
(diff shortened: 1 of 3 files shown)


specs/mediawiki/v1/content.yaml
```diff
@@ -910,6 +915,7 @@ paths:
         body:
           wikitext: '{wikitext}'
           bodyOnly: '{bodyOnly}'
+          body_only: '{body_only}'
```
d00rman: This is an example where it'd be nice to have expressions in the templating mechanism.

```yaml
body_only: '{body_only || bodyOnly}'
```

PR #326, opened 2015-09-10
Made request block definitions generally available

Pchelolo:
After this PR we can use simple_service backend request definitions generally instead of `x-backend-request`. For testing modules and endpoints this allowed to move all declarations to the test.yam spec directly.


config.example.wikimedia.yaml
```diff
@@ -119,8 +119,8 @@ templates:
               paths:
                 /v1/png/{title}/{revision}/{graph_id}:
                   get:
-                    on_request:
-                      - get_from_graphoid:
+                    x-backend-request:
```
gwicke: Is this intended?

gwicke:
A couple of comments:

## Internal vs. external APIs, and the future of the simple-service module

Originally, I assumed that we would get rid of the simple_service module once we have all that general request templating functionality in RESTBase proper. This PR shies away from that step. Let me address some possible issues related to this:

- The size of the main spec might become a bit too large if everything is specified inline.
  - While this wouldn't necessarily be an issue for external spec users (we can filter out those stanzas the same way we do with modules), it does make specs quite unwieldy for editing. We can however use sub-modules to keep individual module size manageable.
- Separation of internal form external APIs. By having an indirection via an internal API, we can change the external API without changing the internal one, giving us a more stable internal API to reference internally.
  - This concern motivated the /sys/ namespace originally, and I think it has worked very well for general infrastructure like table storage and buckets. However, many of the services we are now hooking up are more specialized and / or are changing quickly in any case, thus rendering this issue less pressing. We haven't really used simple_service instances much internally. Given the limited benefit, I think that the extra cost incurred by the need to set up two separate interfaces might not be worth it for simple_service use cases.

Does this address the concerns you had about removing simple_service?

## Bikeshedding

The current naming choices are:

- `x-backend-request` for the request handler
- `x-backend-setup` for the setup handler

I actually liked the `on_setup` / `on_request` naming, as those cast these sections more accurately as event handlers. Those names also didn't imply that there would be only a single backend request, which `x-backend-request` (now incorrectly) does. However, we now need the x- prefix if we want to keep the raw spec valid. With filtering, I would actually argue that this doesn't really matter, as only the public / merged / filtered spec really needs to validate. So, some proposals:

"on_"
```yaml
get:
  on_setup:
    - ....
  on_request:
    - ....
```

"x-restbase"
```yaml
get:
  x-restbase:
    on_setup:
      - ....
    on_request:
```

"cute"
```yaml
get:
  x-setup:
    - ....
  x-ecute:
    - ....
```

"handlers"
```yaml
get:
  x-setup-handler:
    - ...
  x-request-handler:
    - ....
```

config.example.wikimedia.yaml
Pchelolo: @gwicke Oh, I thought I've triple-checked that, obviously it's a flow..

Pchelolo:
@gwicke The name of the stanza must have an `x-` prefix, because it's not standard in the swagger config and it will appear on the `top` level, not inside other x-prefixed object. From all the options I like `x-request-handler` and `x-setup-handler` the most. 

I'm in favour of removing the simple_service, and filtering-out all `x-...` parts from the resulting spec we expose on `?spec` endpoint. In that case we move all handler specifications to "front-end" specs (e.g. content.yaml etc) - this makes config.yaml smaller, and gives a clear view of want's happening right from the "front-end" spec. 

Also, we shouldn't forget about the speed-up we get from eliminating one additional hop (`api -> sys -> backend` would become `api -> backend`)

@d00rman @eevans What are your thoughts on this?

doc/HandlerConfiguration.md
```diff
@@ -1,92 +1,170 @@
-# Declarative proxy handler definition
-- Provide a clear spec of the request flow for a given entry point
-- Abstract common operations
-- Make it easy to port to another environment later
+# Declarative request handler definition
```
d00rman: Wow! Great docs!!!

lib/router.js
```diff
@@ -242,13 +236,16 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
                     + node.value.path + ':' + methodName);
             }
 
-            var backendRequest = method && method['x-backend-request'];
+            var backendSetup = method && method['x-setup-handler'];
+            if (backendSetup) {
+                handlerTemplate.parseSetupConfig(backendSetup).forEach(function(resource) {
```
d00rman: Might be simpler:

```js
Array.prototype.push.apply(node.value.resources, handlerTemplate.parseSetupConfig(backendSetup));
```

doc/HandlerConfiguration.md
Pchelolo: @d00rman haha, this is because you wrote it (https://phabricator.wikimedia.org/T105975#1499210)

d00rman:
## Simple Service

Here's my *pro* / *contra* list with regards to removing `simple_service`.

### Pro

- Less internal hops. As @Pchelolo notes, We'd lose one redirect that way, which at large scale becomes a significant improvement.
- Direct back-end referencing. This PR allows us now to directly write the request for the back-end service right in the route handler, which (to some extent) minimises the chance of getting it wrong.
- Less config changes. Removing `simple_service` and moving most of the request-handling configuration to the API spec means that most of these things will become an integral part of code deploys.

### Contra

- Front-end / back-end separation. To steal @gwicke's words, *by having an indirection via an internal API, we can change the external API without changing the internal one, giving us a more stable API to reference internally.* Additionally, having this separation allows us to reason about public and internal routes individually. @gwicke mentions that `simple_service` could be removed on the basis that back-end service endpoints / definitions change often (at least in the initial phase). To me that seems like a strong reason to keep this separation, because we can ensure / enforce public API stability while accommodating back-end service changes. At the same time, we also control the internal API endpoints, which can be referenced by other (internal) modules.
- Spec length. Our specs are really getting longer and longer - both for individual endpoints and for the spec as a whole. I agree that we could (and should) separate different parts of the overall spec into multiple files. However, that does not *cure* the problem of the length of individual endpoints. What I mean here is that in `vim` I can see around 50 lines per screen, and there are endpoints which do not fit into that. IMHO, that's a clear signal there's something wrong here :)

### Discussion

I must admit I personally like `simple_service` and the power it gives us to hook up back-end services (and/or have the ability to define simple modules) without resorting to coding. Generalising this approach and allowing users to do so directly in the endpoint definitions is a *good thing*(TM). However, I think we should keep `simple_service` around. My rationale is that keeping it would allow us to decide where and when to use back-end definitions and when to set it up directly in the public endpoint definition.

## Bike-shedding

I'll be short here :) While I do like `x-request-handler` and `x-setup-handler`, since we already have `x-amples` I'm voting for the *cute* variant (`x-etup` and `x-ecute`).

doc/HandlerConfiguration.md
d00rman: LOL. I'm an egoist, what can I say hahaha

gwicke:
> I'm in favour of removing the simple_service, and filtering-out all x-... parts from the resulting spec we expose on ?spec endpoint.

This would also remove x-monitor and x-amples, which would break our monitoring setup.

gwicke:
@d00rman: I don't see those two points as a strong reason for keeping simple_service as a module:

- If we wish, we can continue to keep the current front-end / back-end separation where it makes sense.
- The spec size will be slightly shorter without a [ module wrapper](https://github.com/wikimedia/restbase/blob/master/config.test.yaml#L129-L134).

Regarding naming: If we want to stick with x- for internal options, then I think I'd prefer the "handlers" option. It's definitely not as cute, but clearer and more descriptive.

gwicke:
Summary of the discussion:

- @d00rman is okay with removing the module, as the indirection is still available.
- @Pchelolo, @eevans and myself are favoring the "handlers" option. I guess we aren't into cute very much.

d00rman:
`x-request-handler` and `x-setup-handler` are fine by me as well, I don't feel too strongly about it either way (not in the bike-shedding mood, I guess :P). They're clear enough, albeit a bit long to type, but let's not get into that discussion :)

Pchelolo:
@gwicke @d00rman @eevans I've updated the PR - removed a simple_service and set up mobileApps and graphed handlers directly.

PR #327, opened 2015-09-11
Added delay between running Cassandra and SQLite tests

Pchelolo:
429 errors in travis became extremely annoying, so we need to add a 1 minute delay between running SQLite and Cassandra tests.


Pchelolo:
I'm closing this PR as delays doesn't seem to help in travis.

PR #328, opened 2015-09-11
Use nock for some of the save_api tests

Pchelolo:
429 errors on save tests are very annoying, and even setting a 90 second delay between running tests on different backends doesn't help, so we switch some of the tests to nock.

### DIFF
--- a/mods/action.js
+++ b/mods/action.js
@@ -187,7 +187,7 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
     apiRequest.body.action = defBody.action;
     apiRequest.body.format = apiRequest.body.format || defBody.format || 'json';
     apiRequest.body.formatversion = apiRequest.body.formatversion || defBody.formatversion || 1;
-    if (!apiRequest.body.hasOwnProperty('continue')) {
+    if (!apiRequest.body.hasOwnProperty('continue') && apiRequest.action === 'query') {
         apiRequest.body.continue = '';
     }
     return restbase.request(apiRequest).then(cont);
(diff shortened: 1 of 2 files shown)


test/features/pagecontent/save_api.js
```diff
@@ -110,6 +111,23 @@ describe('page save api', function() {
...
 
     it('fail for bad token', function() {
+        var apiURI = server.config
+            .conf.templates['wmf-sys-1.0.0']
+            .paths['/{module:action}']['x-modules'][0].options.apiRequest.uri;
+        apiURI = apiURI.replace('{domain}', 'en.wikipedia.beta.wmflabs.org');
+
+        nock.enableNetConnect();
+        var api = nock(apiURI)
+            // The first request should return a page so that we store it.
```
d00rman: Indentation. Also, the comment doesn't seem to be in line with what it returns.

The same comment applies to all nock additions in this PR.

mods/action.js
```diff
@@ -187,7 +187,7 @@ ActionService.prototype._doRequest = function(restbase, req, defBody, cont) {
     apiRequest.body.action = defBody.action;
     apiRequest.body.format = apiRequest.body.format || defBody.format || 'json';
     apiRequest.body.formatversion = apiRequest.body.formatversion || defBody.formatversion || 1;
-    if (!apiRequest.body.hasOwnProperty('continue')) {
+    if (!apiRequest.body.hasOwnProperty('continue') && apiRequest.action === 'query') {
```
Pchelolo: Please note this change, it's not really related to the task, but we get a warning from the API when we add `continue` for non-query requests

PR #329, opened 2015-09-11
Enabled tests with node 4.0 and removed io.js

Pchelolo:
node.js 4.0 is now available, making io.js fork stale, so we need to use that for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
     - "0.10"
     - "0.12"
-    - "iojs-v2.5.0"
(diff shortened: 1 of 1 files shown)


gwicke:
I'm +1 on adding node 4, but I think there is value in continuing testing io.js 2.5, as that's what we are testing in production. Also, node 4 is a newer version of the io.js codebase, and not a different fork.

Pchelolo:
Unfortunately, travis tooling is not ready for node 4.0 yet, it can't install heap dump. Probably have to wait some more time, closing this PR.

PR #332, opened 2015-09-14
Pass x-forwarded-for header to backend services

Pchelolo:
We need to pass `x-forwarded-for` header to backend services as RB works like a proxy for them.

Bugs:
https://phabricator.wikimedia.org/T112536

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -275,13 +275,19 @@ rbUtil.parseETag = function(etag) {
  * Copies forwarded headers from restbase to request.
  * If the same header was already set it takes precedence over
  * the forwarded header.
+ *
+ * @param restbase a restbase instance to take the forwarded headers from
+ * @param req a request where to copy the headers
+ * @param headers array of header names to copy
  */
-rbUtil.copyForwardedCookies = function(restbase, req) {
-    if (restbase._rootReq
-            && restbase._forwardedCookies
-            && !req.headers.cookie) {
+rbUtil.copyForwardedHeaders = function(restbase, req, headers) {
+    if (restbase._rootReq && restbase._forwardedHeaders) {
         req.headers = req.headers || {};
-        req.headers.cookie = restbase._forwardedCookies;
+        headers.filter(function(header) {
+            return !req.headers[header] && restbase._forwardedHeaders[header];
+        }).forEach(function(header) {
+            req.headers[header] = restbase._forwardedHeaders[header];
+        });
     }
     return req;
 };
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -283,10 +283,10 @@ rbUtil.parseETag = function(etag) {
 rbUtil.copyForwardedHeaders = function(restbase, req, headers) {
     if (restbase._rootReq && restbase._forwardedHeaders) {
         req.headers = req.headers || {};
-        headers.filter(function(header) {
-            return !req.headers[header] && restbase._forwardedHeaders[header];
-        }).forEach(function(header) {
-            req.headers[header] = restbase._forwardedHeaders[header];
+        headers.forEach(function(header) {
+            if (!req.headers[header] && restbase._forwardedHeaders[header]) {
+                req.headers[header] = restbase._forwardedHeaders[header];
+            }
         });
     }
     return req;
--- a/test/features/pagecontent/save_api.js
+++ b/test/features/pagecontent/save_api.js
@@ -348,7 +348,11 @@ describe('page save api', function() {
 
     it('save HTML', function() {
         nock.enableNetConnect();
-        var api = nock(prodApiURI)
+        var api = nock(prodApiURI, {
+            reqheaders: {
+                'x-forwarded-for': '123.123.123.123'
+            }
+        })
         // Mock MW API success response
         .post('')
         .reply(200, {
@@ -369,6 +373,9 @@ describe('page save api', function() {
             assert.deepEqual(res.status, 200, 'Could not retrieve test page!');
             return preq.post({
                 uri: htmlUri,
+                headers: {
+                    'x-forwarded-for': '123.123.123.123'
+                },
                 body: {
                     html: res.body.replace(/\<\/body\>/,
                         '<p>Generated via direct HTML save! Random ' + Math.floor(Math.random() * 32768) + ' </p></body>'),
@@ -407,13 +414,13 @@ describe('page save api', function() {
             assert.deepEqual(res.status, 200, 'Could not retrieve test page!');
             return preq.post({
                 uri: htmlUri,
+                headers: {
+                    'if-match': lastHTMLETag
+                },
                 body: {
                     html: res.body.replace(/\<\/body\>/, '<p>Old revision edit that should detect conflict!</p></body>'),
                     csrf_token: htmlToken,
                     base_etag: oldHTMLEtag
-                },
-                headers: {
-                    'if-match': lastHTMLETag
                 }
             });
         }).then(function(res) {
--- a/test/features/pagecontent/save_api.js
+++ b/test/features/pagecontent/save_api.js
@@ -350,7 +350,8 @@ describe('page save api', function() {
         nock.enableNetConnect();
         var api = nock(prodApiURI, {
             reqheaders: {
-                'x-forwarded-for': '123.123.123.123'
+                'x-forwarded-for': '123.123.123.123',
+                cookie: 'test'
             }
         })
         // Mock MW API success response
@@ -374,7 +375,8 @@ describe('page save api', function() {
             return preq.post({
                 uri: htmlUri,
                 headers: {
-                    'x-forwarded-for': '123.123.123.123'
+                    'x-forwarded-for': '123.123.123.123',
+                    cookie: 'test'
                 },
                 body: {
                     html: res.body.replace(/\<\/body\>/,
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -45,7 +45,7 @@ var securityDefinitionCreators = {
         return {
             prepareRequest: function(restbase, req) {
                 if (isInternalRequest(req)) {
-                    rbUtil.copyForwardedHeaders(restbase, req, ['cookie']);
+                    rbUtil.copyForwardedHeaders(restbase, req, ['cookie', 'x-forwarded-for']);
                 }
             },
             checkPermissions: function(restbase, req, permDefinitions) {
@@ -59,7 +59,7 @@ var securityDefinitionCreators = {
 
                 // If nothing is applicable to the current method - skip the check
                 if (permissions.length === 0) {
-                    return;
+                    return P.resolve();
                 }
 
                 var checkReq = {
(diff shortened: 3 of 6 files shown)


config.example.wikimedia.yaml
```diff
@@ -108,6 +108,7 @@ templates:
                 uri: 'http://{domain}/w/api.php'
                 headers:
                   host: '{$.request.params.domain}'
+                  x-forwarded-for: '{$.request.headers.x-forwarded-for}'
```
gwicke: Would just `{x-forwarded-for}` work here as well?

mods/page_save.js
```diff
@@ -120,12 +120,30 @@ PageSave.prototype._checkParams = function(params) {
...
+    var xff = req.headers['x-forwarded-for'];
+    var remoteAddr = xff && xff.split(',')[0].trim()
+        || req.socket && req.socket.remoteAddress;
+    if (!/^(?:::ffff:)?(?:10|127)\./.test(remoteAddr)) {
+        throw new rbUtil.HTTPError({
+            status: 403,
+            body: {
+                type: 'access_denied#revision',
+                title: 'Access to resource denied',
+                description: 'Only internal clients are allowed to use save api'
```
d00rman: IMHO, *This client is not allowed to use the save API* is better, as it still tells the client they can't do that, but doesn't expose the security reason behind it (which might incite them to try to get around it) :)

mods/page_save.js
```diff
@@ -154,7 +172,8 @@ PageSave.prototype.saveWikitext = function(restbase, req) {
         return restbase.post({
             uri: new URI([rp.domain, 'sys', 'action', 'edit']),
             headers: {
-                cookie: req.headers.cookie
+                cookie: req.headers.cookie,
+                'x-forwarded-for': req.headers['x-forwarded-for']
```
d00rman: ... `|| req.socket.remoteAddress`

mods/page_save.js
```diff
@@ -178,7 +199,8 @@ PageSave.prototype.saveHtml = function(restbase, req) {
             html: req.body.html
         },
         headers: {
-            'if-match': req.body.base_etag || req.headers['if-match']
+            'if-match': req.body.base_etag || req.headers['if-match'],
+            'x-forwarded-for': req.headers['x-forwarded-for']
```
d00rman: Idem.

config.example.wikimedia.yaml
d00rman: It would, but personally I find this much clearer.

config.test.yaml
```diff
@@ -159,6 +159,7 @@ templates:
                 uri: http://{domain}/w/api.php
                 headers:
                   host: '{$.request.params.domain}'
+                  x-forwarded-for: '{$.request.headers.x-forwarded-for}'
```
d00rman: Why there's no `cookie` header here as well?

test/features/pagecontent/save_api.js
```diff
@@ -251,10 +272,28 @@ describe('page save api', function() {
         });
     });
 
+    it('should forbid external usage', function() {
```
d00rman: The fact that this test passes means that there's something wrong with the code :) Hint: it's actually related to my comment about adding `|| req.socket.remoteAddress`

d00rman:
While this PR looks ok to me (modulo some minor in-lined comments), I think we can do better than this. We are working around a concrete *problem*, but IMHO we should be supplying back-end services with `X-Forwarded-For` header as a rule, since technically RESTBase acts a kind of proxy.

mods/page_save.js
gwicke: I'm not sure that exposing internal IPs is actually a good idea. It might be a better idea to provide a fixed fall-back IP like 127.0.0.1 instead.

test/features/pagecontent/save_api.js
```diff
@@ -91,6 +91,9 @@ describe('page save api', function() {
     it('fail for missing content', function() {
         return preq.post({
             uri: uri,
+            headers: {
+                'x-forwarded-for': '127.0.0.1'
```
gwicke: Do we need this? What happens when we omit the header?

test/features/pagecontent/save_api.js
gwicke: The test looks okay to me, but I might be missing something.

test/features/pagecontent/save_api.js
```diff
@@ -251,10 +272,28 @@ describe('page save api', function() {
...
+            assert.deepEqual(e.status, 403);
+        });
+    });
+
     it('save page', function() {
         // Leave 1 test unmocked for sanity check.
         return preq.post({
             uri: uri,
+            headers: {
+                'x-forwarded-for': '127.0.0.1'
```
gwicke: Do we need this? What happens if we omit it?

gwicke:
@d00rman, we should not forward XFF headers to external services that we wouldn't feel comfortable sending other user data to. We already have a whitelist for services we trust with cookies, which I think is a good starting point for XFF header forwarding as well. A service that doesn't need authentication should also have no business looking at XFF headers.

test/features/pagecontent/save_api.js
d00rman: I don't see how this test should pass. So, the idea is that RESTBase should allow page saves for internal clients and fail for others. But ... in test mode, there are no external clients, only a local one. Hence, if this test is passing, something's wrong in the code, as `req.socket.remoteAddress` in this case must be equal to `127.0.0.1`

test/features/pagecontent/save_api.js
d00rman: We shouldn't need them, and in fact, removing them makes these tests fail.

mods/page_save.js
d00rman: Hm. If the API doesn't do anything sensible when XFF is missing then we should ask ourselves what's the point of this PR, shouldn't we?

You are right, though, that we should supply an agreed-upon value in case XFF is missing, so that it can be white-listed on the API servers and that we don't hit any limits or restrictions imposed on individual IPs.

test/features/pagecontent/save_api.js
gwicke: > So, the idea is that RESTBase should allow page saves for internal clients and fail for others.

No, the POST end point restriction is not related to XFF forwarding. They are two independent issues that just ended up in the same PR.

gwicke:
I think my original title in https://phabricator.wikimedia.org/T112539 was a bit confusing, and lead to this mix-up between XFF forwarding and POST storage access restrictions.

There was no intention to restrict page saving to internal IPs. Only POST storage end points need to be restricted. If possible without too much pain, it might be helpful to tease apart the two issues into separate PRs.

mods/page_save.js
gwicke: My preference would really be to let the action API handle it, so that such handling is uniform across clients.

Pchelolo:
@gwicke The cookies are really security-sensitive, but I'm not sure I understand what sensitive info we expose with an `x-forwarded-for` header?

gwicke:
@Pchelolo, we would be exposing client IPs along with specific requests to external services we don't trust with log-in data. Client IPs can be used to identify users, and are thus considered sensitive. User-identifying information like IP addresses is covered by https://wikimediafoundation.org/wiki/Privacy_policy.

Pchelolo:
@gwicke ok, but we can't really reuse the same whitelist from the mediawiki_auth service, because they are kinda hidden in the securityDefinition, so it wouldn't be very obvious why are they used in an unrelated part of the code. How about making a global whitelist `trusted endpoints` and reuse in both places?

gwicke:
@Pchelolo, couldn't we piggy-back on the cookie forwarding on the way out? I think for now it's reasonable to equate "can be trusted with authentication" with "can be trusted with authentication and other sensitive client data like the IP".

Pchelolo:
@gwicke we can't reuse the cookie-forwarding as it's called on `iff` the auth it enabled, so this forwarding should be done separately like in this PR. And AuthService is not even created (config parsed etc) if it's not enabled, so we can't pull a list of trusted service from it's config. That's why I proposed moving this list to the top level

lib/rbUtil.js
```diff
@@ -231,13 +231,19 @@ rbUtil.parseETag = function(etag) {
...
  */
-rbUtil.copyForwardedCookies = function(restbase, req) {
-    if (restbase._rootReq
-            && restbase._forwardedCookies
-            && !req.headers.cookie) {
+rbUtil.copyForwardedHeaders = function(restbase, req, headers) {
+    if (restbase._rootReq && restbase._forwardedHeaders) {
         req.headers = req.headers || {};
-        req.headers.cookie = restbase._forwardedCookies;
+        headers.filter(function(header) {
```
gwicke: Sligtly simpler alternative: 

```javascrip
headers.forEach(function(header) {
  if (!req.headers[header] && restbase._forwardedHeaders[header]) {
    req.headers[header] = restbase._forwardedHeaders[header];
  }
});
```

gwicke:
> we can't reuse the cookie-forwarding as it's called on iff the auth it enabled, so this forwarding should be done separately like in this PR

We'll want to forward authentication cookies through the save API (without requiring specific rights), which happens to be the main use case we need XFF for right now. Do you have other use cases in mind where XFF will be needed, but auth should not be supplied?

Pchelolo:
@gwicke @d00rman had an idea that we need to forward XFF generally as RB acts like a proxy for all services, so technically we should forward (or even create our own) XFF

gwicke:
@Pchelolo @d00rman, sensitive info like user IPs should only be forwarded on an as-needed basis. Most services don't need to know client IPs, so we shouldn't be sending this information to them.

gwicke:
It looks like we should have discussed this more thoroughly on https://phabricator.wikimedia.org/T112536 before diving into the implementation.

Pchelolo:
@gwicke I've just got a better plan to support this: as we need to provide the cookie to these endpoints for unrestricted domains too, we have to add a `security` stanza to the `post` endpoints. So, the `mediawiki_auth` security implementation can pass `x-forwarded-for` together with a cookie. Or it could be configurable. 

But support for adding security definitions to individual methods is added in https://github.com/wikimedia/restbase/pull/336, so the PR now depends on that one.

lib/rbUtil.js
gwicke: Also, it should do less allocation ;)

Pchelolo:
New approach. We reuse auth service here so forward `cookie` and `x-forwarded-for` header to MW API on page save, so that the edit author could be seen correctly

lib/rbUtil.js
```diff
@@ -275,13 +275,19 @@ rbUtil.parseETag = function(etag) {
...
  */
-rbUtil.copyForwardedCookies = function(restbase, req) {
-    if (restbase._rootReq
-            && restbase._forwardedCookies
-            && !req.headers.cookie) {
+rbUtil.copyForwardedHeaders = function(restbase, req, headers) {
+    if (restbase._rootReq && restbase._forwardedHeaders) {
         req.headers = req.headers || {};
-        req.headers.cookie = restbase._forwardedCookies;
+        headers.forEach(function(header) {
```
d00rman: Does it make sense here to allow `headers` not to be passed in, in which case all of the `restbase._forwardedHeaders` would be used?

Pchelolo: @d00rman hm, we don't have such a use-case yet, so what's the point in complicating the code? If we need that at some point, it would be easy to add.

d00rman:
LGTM

gwicke:
LGTM as well. Will merge in a bit if there aren't any objections.

PR #333, opened 2015-09-14
Tighten down CSP headers

gwicke:
- Create a single method to encapsulate all CSP header logic in one place.
- Set really strict CSP headers on anything but text/html and image/svg.
- Consistently lowercase header names to avoid confusion and subtle bugs.

Task: https://phabricator.wikimedia.org/T112586

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -63,7 +63,7 @@ var cspHeaderVariants = [
     'content-security-policy',
     // For IE 10 & 11
     'x-content-security-policy',
-    // For Chrome <= v25 (seems to be <1% traffic; should we still care?)
+    // For Chrome <= v25 (<<1% traffic; todo: revisit support)
     'x-webkit-csp',
 ];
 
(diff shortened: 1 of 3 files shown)


lib/rbUtil.js
```diff
@@ -54,24 +54,65 @@ rbUtil.wildcardSubdomain = function(domain) {
...
-        styleSource = '*';
+        // Other content: Disallow everything, especially framing to avoid
+        // re-dressing attacks.
+        csp = "default-src 'none'; frame-ancestors 'none'";
+    }
+    rh['x-xss-protection'] = '1; mode=block';
+
+    // Let backend services override CSP headers
+    // XXX: Re-consider this policy
+    if (!rh['content-security-policy']) {
```
Pchelolo: Shouldn't we set variants with a custom header value too? (e.g. copy existing CSP header to x-CSP and x-webkit-CSP)

lib/rbUtil.js
```diff
@@ -54,24 +54,65 @@ rbUtil.wildcardSubdomain = function(domain) {
...
+function isHtmlSvgContent(res) {
+    return res.headers
+        && /^(?:text\/html|image\/svg)/i.test(res.headers['content-type']);
+}
+
+var cspHeaderVariants = [
+    'content-security-policy',
+    // For IE 10 & 11
+    'x-content-security-policy',
+    // For Chrome <= v25 (seems to be <1% traffic; should we still care?)
```
d00rman: Yes, I think we should ;)

lib/rbUtil.js
d00rman: Not sure I understand @Pchelolo . That's precisely what's being done here.

Pchelolo: @d00rman I mean if we have a 'content-security-policy' header set, we should also ensure that other variants are equal to it, e.g.
```
 else {
   scpHeaderVariants.forEach(function(name) {
      rh[name] = rh['content-security-policy'];
   });
}

Comment (d00rman) [inline lib/rbUtil.js]: Indeed @Pchelolo, good catch! That was in effect previously, but has been removed in this PR. A slightly more optimised version:

```js
if (rh['content-security-policy']) {
  csp = rh['content-security-policy'];
}
cspHeaderVariants.forEach(function(name) {
  rh[name] = csp;
});
```

PR #334, opened 2015-09-14
Document Cassandra datacenter configuration

eevans:
This links to the documentation added in https://github.com/wikimedia/restbase-mod-table-cassandra/pull/147, and so ought to be merged after it.

References to the `keyspace` directive were also removed here.  There isn't likely to be any legit reason that a user would need to configure this themselves (and it defaults to `system` if unset).

Bug: https://phabricator.wikimedia.org/T76494

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -65,10 +65,10 @@ templates:
           - name: restbase-mod-table-cassandra
             version: 1.0.0
             type: npm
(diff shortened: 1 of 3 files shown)


d00rman:
+1e6 ! We should document all of the params like this (and do so in the respective module as well actually). Minor nit: should this be copy/pasted into `config.example.yaml` as well for reference?

gwicke:
Nice docs!

I wonder if it would be more discoverable if we moved this to a README section in restbase-mod-table-cassandra, so that we can link to it from the config examples & puppet?

PR #336, opened 2015-09-15
Check client IP on requests to POST storage

Pchelolo:
This PR is very tightly coupled with https://github.com/wikimedia/restbase/pull/332. 

It's more a prototype to start the discussion, than a final implementation. A couple of notes:
1. The `client_ip` securityDefinition is set on the `/sys` subspec, so it doesn't get to the publicly exposed spec, leaving it valid (this securityDefinition is invalid from swagger perspective) while we can still use it on the /sys level. When we need to add such a feature on /api level we can move it (and rename to x-securityDefinitions).
2. If we state, that RB works as a proxy, the correct way to work with an `x-forwarded-for` header would be to add request's `remoteAddress` to the incoming header's proxy list, or set it as a client ip if no header was passed. 
3. RB can make requests upon startup and these shouldn't be checked.
4. Parallel access checks are only valid for GET and HEAD, other methods should use sequential checks.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -85,11 +85,13 @@ var securityDefinitionCreators = {
             }
         };
     },
-    client_ip: function(definition) {
+    header_match: function(definition) {
+        var errorMessage = definition['x-error-message']
+            || 'This client is not allowed to use the endpoint';
         var whitelistMap = {};
-        Object.keys(definition['x-client-whitelists']).forEach(function(whitelistName) {
+        Object.keys(definition['x-whitelists']).forEach(function(whitelistName) {
             whitelistMap[whitelistName]
-            = constructInternalRequestRegex(definition['x-client-whitelists'][whitelistName]);
+            = constructInternalRequestRegex(definition['x-whitelists'][whitelistName]);
         });
 
         return {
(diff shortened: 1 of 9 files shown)


Pchelolo:
@gwicke I've updated the PR to make a generic header checker and user `x-rb-client-ip` header.

d00rman:
Hmm, if this is merged, #339 won't work as expected any more /me thinks :(

Pchelolo:
@d00rman Indeed.. The argumentation for this is in [this ticket](https://phabricator.wikimedia.org/T112539) and it's also valid, as it's really easy to fill up the storage with junk in case the post_storage is opened to the world. 

Conservative rate-limiting could help, but we don't have such a feature, or we could set some TTL on the stored POST data, but as I understand, the latter isn't an option, as we want post data forever?

d00rman:
> @d00rman Indeed.. The argumentation for this is in this ticket and it's also valid, as it's really easy to fill up the storage with junk in case the post_storage is opened to the world.

That was more me thinking out loud than trying to say not to merge this. I completely agree with securing the POST endpoints. But that means I have to rethink completely the approach for #339 as that means that external clients will never be able to make a legitimate request to Mathoid (in the current form of the PR and with this PR merged).

> Conservative rate-limiting could help, but we don't have such a feature, or we could set some TTL on the stored POST data, but as I understand, the latter isn't an option, as we want post data forever?

Neither option would really help us if we are DDoS'ed, e.g.

lib/restbase.js
```diff
@@ -325,13 +325,20 @@ RESTBase.prototype._request = function(req) {
...
 
         if (accessRestricted) {
-            return P.all([
-                reqHandlerPromise,
-                childRESTBase._authService.checkPermissions(childRESTBase, childReq)
-            ])
-            .then(function(res) {
-                return res[0];
-            });
+            if (req.method === 'get' || req.method === 'head') {
```
d00rman: Why limit `GET`s as well? `POST`s are a real problem here. Say that an internal service `POST`s something and gets back the hash for it. Then, external clients should be able to retrieve it, shouldn't they?

d00rman: I realise this is a general *check permissions* routine, but for `POST`s I think we should separate that.

Pchelolo: @d00rman right, I'll move the security check to POST endpoint, but it depends on a little bug-fix in router: https://github.com/wikimedia/swagger-router/pull/21

Pchelolo: @d00rman now we check access to post storage only on PUT, get's are allowed unconditionally.

lib/auth.js
```diff
@@ -84,6 +97,57 @@ var securityDefinitionCreators = {
...
+                            && requirementDefinition.method !== req.method) {
+                        return;
+                    }
+
+                    var headerMatchRequirement = requirementDefinition.value;
+                    var headerName = headerMatchRequirement.header;
+                    var headerValue = req.headers && req.headers[headerName]
+                            || restbase._rootReq.headers && restbase._rootReq.headers[headerName];
+
+                    headerMatchRequirement.patterns.forEach(function(permission) {
```
gwicke: The use of 'permission' seems to be a bit confusing here. Do you think 'pattern' would be clearer?

gwicke:
LGTM overall, only one minor naming nit inline.

PR #337, opened 2015-09-15
Minor templating improvements

d00rman:
This PR makes some small improvements to the handler-templating engine:

- **Handle `xxx` conditions properly**. Previously, only one `x` was replaced by a `.` and the construct `/^.$/.test(a_string)` does not seem to pass the syntax check when used inside `new Function(code)`. Both of these bugs have been corrected.
- **Allow URIs to have absolute templates**. Absolute URI templating is needed for cases where one wants to use the POST request's hash string as the primary key for a *key_value* bucket storing responses from back-end services. The PR adds the capability without hurting performance in cases where this is not needed.

### DIFF
--- a/test/features/router/handlerTemplate.js
+++ b/test/features/router/handlerTemplate.js
@@ -124,7 +124,7 @@ describe('handler template', function () {
                 return_if: {
                     status: '5xx'
                 },
-                return: '$.request'
+                return: '{$.request}'
             }
         }]);
     });
--- a/test/features/router/handlerTemplate.js
+++ b/test/features/router/handlerTemplate.js
@@ -115,6 +115,20 @@ describe('handler template', function () {
         }, /^Invalid spec\. Either request or return must be specified\..*/);
     });
 
+    it('validates config: compiles a valid condition function', function() {
+        handlerTemplate.createHandler([{
+            get_one: {
+                request: {
+                    uri: '/my/path'
+                },
+                return_if: {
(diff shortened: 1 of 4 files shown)


test/features/router/handlerTemplate.js
```diff
@@ -115,6 +115,20 @@ describe('handler template', function () {
...
+    it('validates config: compiles a valid condition function', function() {
+        handlerTemplate.createHandler([{
+            get_one: {
+                request: {
+                    uri: '/my/path'
+                },
+                return_if: {
+                    status: '5xx'
+                },
+                return: '$.request'
```
Pchelolo: It doesn't matter for this test, but correct is `'{$.request}'`

Pchelolo:
@d00rman It's very strange that `/^.$/.test(a_string)` doesn't pass a syntax check and a variant with a `match` does. Kinda looks like a bug in JS :)

LGTM, except one super-minor comment in the test.

test/features/router/handlerTemplate.js
d00rman: yup!

d00rman:
> It's very strange that /^.$/.test(a_string) doesn't pass a syntax check and a variant with a match does.

Indeed it is rather strange. With `/^$/.test(blah)` I kept getting `SyntaxError: Unexpected token }`. No idea where is that `}` pulled from, though.

PR #338, opened 2015-09-15
Added support for default values in templates

Pchelolo:
Added support for `{$.request.body.xxx || "defaultVal"}` in request templates.

But please be aware, that although it's a handy feature, it's quite slow. Template evaluation benchmarks shows ~5-7 times performance impact comparing to plain template without a default. But we can live with it, as impact on a endpoint latency is close to 0.

### DIFF
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -92,12 +92,15 @@ function compileTemplate(templateSpec, setValue, reqPart) {
             res += '' + bit;
         }
     };
+    var errorHandler = function() { return undefined; };
+
     var resolveTemplate = TAssembly.compile(template, {
         nestedTemplate: true,
-        cb: callback
+        cb: callback,
+        errorHandler: errorHandler
     });
     var options = {
-        errorHandler: function() { return undefined; },
+        errorHandler: errorHandler,
         cb: callback,
         globals: {
             default: function(val, defVal) {
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -52,10 +52,6 @@ function _setAtPath(path, spec) {
  * @returns {string} validated and
  */
 function prepareTemplateString(template, defaultLookupPath) {
-    if (!/^(\$\.)?([a-zA-Z][a-zA-Z0-9-_]*)(\.[a-zA-Z][a-zA-Z0-9-_]*)*$/.test(template)) {
-        throw new Error('Invalid template ' + template);
-    }
-
     if (template[0] === '$') {
         template = 'rm' + template.slice(1);
     } else {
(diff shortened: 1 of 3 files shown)


gwicke:
Tassembly supports using global functions, and calling `default(value, defaultValue)` similar to [Jinja2](http://jinja.pocoo.org/docs/dev/templates/#default) will likely be faster than if / ifnot. If we hook up that default function generally, then we can likely also avoid the need for the custom || syntax.

lib/reqTemplate.js
```diff
@@ -75,12 +71,32 @@ function prepareTemplateString(template, defaultLookupPath) {
  * @returns {Function} a template resolver for the provided template
  */
 function compileTemplate(templateSpec, setValue, reqPart) {
+    function createTAssemblyTemplate(template) {
+        if (/^(?:\$\.)?(?:[a-zA-Z][\w-]*)(?:\.[a-zA-Z][\w-]*)*$/.test(template)) {
```
d00rman: This RegExp could use some comments. At least what is it checking for. Ideally, though, each of its parts would be commented properly.

lib/reqTemplate.js
```diff
@@ -75,12 +71,32 @@ function prepareTemplateString(template, defaultLookupPath) {
...
+            // It's an ordinary template without default value
+            return [['raw', prepareTemplateString(template, reqPart)]];
+        }
+        var condTemplate = new RegExp('^((?:\\$\.)?(?:[a-zA-Z][\\w-]*)(?:\\.[a-zA-Z][\\w-]*)*)' +
```
d00rman: Idem. This one is even longer and with a lot of back-slashes. Not easy to read at all. Let's help fellow coders :)

Pchelolo:
@gwicke ok, I'll try to check that solution too.

d00rman:
> Tassembly supports using global functions, and calling default(value, defaultValue) similar to Jinja2 will likely be faster than if / ifnot. If we hook up that default function generally, then we can likely also avoid the need for the custom || syntax, but using {default($.request.foo, 'bar')} directly.

I was actually wondering today whether it'd be better to use `{$.blah || "yey'}` or the function version `{default($.request, "yey")}`. I guess one issue with using the function equivalent is (maybe) the fact that it'd questionable whether the current syntax, which allows stuff like `$.request.headers.x-forwarded-for`, would still remain valid (wrt. e.g. parsing).

gwicke:
>  I guess one issue with using the function equivalent is (maybe) the fact that it'd questionable whether the current syntax, which allows stuff like $.request.headers.x-forwarded-for, would still remain valid (wrt. e.g. parsing).

That's fully supported in tassembly, and there is also [an expression parser](https://github.com/gwicke/knockoff/blob/master/KnockoutExpressionParser.pegjs) to validate complex constructs.

d00rman:
> That's fully supported in tassembly, and there is also an expression parser to validate complex constructs.

Yup, but I meant more in the direction that if we start allowing function signatures, it'll be a question of time before we possibly introduce some other expressions, which would then force us to reconsider the current policy. But, perhaps I'm rushing into things here :)

gwicke:
See https://github.com/gwicke/template-expressions for a stand-alone expression compiler.

d00rman:
Actually, hm, perhaps going for functions straight away is the way to go. I have just come across another use case that will be useful - headers merging:

```yaml
req1:
  request:
    uri: /path/no1
req2:
  request:
    uri: /path/no2
  return:
    status: 200
    headers: '{merge($.req1.headers, $.req2.headers)}'
```

d00rman:
Nice :+1: AFAICT, this can be merged.

Pchelolo:
@gwicke Any objections agains merging it?

lib/reqTemplate.js
```diff
@@ -75,42 +83,58 @@ function prepareTemplateString(template, defaultLookupPath) {
...
-            if (res === undefined) {
-                res = bit;
-            } else {
-                res += '' + bit;
+        cb: callback,
+        globals: {
+            default: function(val, defVal) {
+                return val || defVal;
+            },
+            merge: function(destination, source) {
```
d00rman: Question: would something like this work:

```yaml
reqX:
  request:
    uri: /my/path/
  return:
    headers: '{$$.merge($.reqX.headers, {'x-my-header', $.request.headers.x-my-header})}'
```

where `{'x-my-header', $.request.headers.x-my-header}` represents an object containing only one header we'd need to add?

(sorry for being late with this, but it has just occurred to me I'll need exactly that)

Pchelolo: @d00rman It should, but I've just checked and it doesn't.. Let me handle it tomorrow morning

lib/reqTemplate.js
```diff
@@ -42,28 +42,36 @@ function _setAtPath(path, spec) {
...
- *
- * @param {string} template an original template string
- * @param {string} defaultLookupPath a default path withing the context which
- *                 to look if template is local
- * @returns {string} validated and
- */
-function prepareTemplateString(template, defaultLookupPath) {
-    if (!/^(\$\.)?([a-zA-Z][a-zA-Z0-9-_]*)(\.[a-zA-Z][a-zA-Z0-9-_]*)*$/.test(template)) {
-        throw new Error('Invalid template ' + template);
+function splitAndPrepareTAsseblyTemplate(templateSpec) {
```
gwicke: This won't handle string literals that contain '{' or '}'.

d00rman: Touché. On the upside, it will figure out the brackets are unbalanced. Albeit, adding a matching bracket inside the string will pass unnoticed:

```js
foo { '{}' }
```

gwicke: Any kind of looking inside string literals is a bug to me.

d00rman: Yup, me too.

PR #340, opened 2015-09-17
Grouped mobileapps sys endpoints together

Pchelolo:
Without the simple_service mobiles /sys endpoints look really ugly. Here I've grouped them together using the `x-subspec` stanza.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -117,36 +117,41 @@ templates:
                 request:
                   uri: http://graphoid.wikimedia.org/{domain}/v1/png/{title}/{revision}/{graph_id}
 
-      /{module:mobileapps}/v1/html/{title}:
-        get:
-          x-request-handler:
-            - get_from_backend:
-                request:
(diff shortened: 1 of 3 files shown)


config.example.wikimedia.yaml
```diff
@@ -117,36 +117,41 @@ templates:
...
-            - get_from_backend:
-                request:
-                  uri: http://appservice.wmflabs.org/{domain}/v1/page/mobile-html-sections-remaining/{title}
-      /{module:mobileapps}/v1/text/{title}:
-        get:
-          x-request-handler:
-            - get_from_backend:
-                request:
-                  uri: http://appservice.wmflabs.org/{domain}/v1/page/mobile-text/{title}
+      /{module:mobileapps}/v1:
```
d00rman: Euh, why do we have `v1` in `/sys`? What's the point of that? I guess I should be able to answer that question, but can't :P Would you mind removing it and adjusting the paths in mediawiki/v1/mobileapps.yaml ?

Pchelolo: @d00rman Sure. I thought there was some purpose for that :)

Pchelolo:
@d00rman done

PR #341, opened 2015-09-18
Removed the delay from resource handler and check resources length

Pchelolo:
Recently we've had to make resources array initialised by default, so `delay(0)` is called on every node now, making startup very slow. While debugging this, me and @d00rman tested startup without the `delay(0)` and it seems to work fine both locally and in staging, so this PR also removes that workaround.


d00rman:
LGTM

PR #343, opened 2015-09-18
Add analytics pageview module, spec, and tests

madhuvishy:
Bug on phabricator: T107053

These would be the frontend to backend mappings we need on all wikis (basically just making the {domain} map implicitly to the {project} parameter on our backend:

      /{module:stats}/views/v1/per-article/{access}/{agent}/{article}/{granularity}/{start}/{end}
        get:
          x-request-handler:
            - get_view_stats_per_article
                request:
                  uri: http://wikimedia.org/stats/views/v1/per-article/{domain}/{access}/{agent}/{article}/{granularity}/{start}/{end}

      /{module:stats}/views/v1/per-project/{access}/{agent}/{granularity}/{start}/{end}
        get:
          x-request-handler:
            - get_view_stats_per_project
                request:
                  uri: http://wikimedia.org/stats/views/v1/per-project/{domain}/{access}/{agent}/{granularity}/{start}/{end}

      /{module:stats}/views/v1/top/{access}/{year}/{month}/{day}
        get:
          x-request-handler:
            - get_view_top_stats
                request:
                  uri: http://wikimedia.org/stats/views/v1/top/{domain}/{access}/{year}/{month}/{day}

And for the global wikimedia.org we could just pass whatever comes after stats/* straight through, there would be no rewriting.  This is where people could pass values such as "all-projects" for the {project} parameter.

### DIFF
--- a/config.test.analytics.yaml
+++ b/config.test.analytics.yaml
@@ -1,0 +1,124 @@
+# RESTBase config for Analytics APIs
+
+info:
+  name: restbase
+
+
+templates:
+
+  wmf-content-1.0.0: &wp/content/1.0.0
+    swagger: '2.0'
+    # swagger options, overriding the shared ones from the merged specs (?)
+    info:
+      version: 1.0.0-beta
+      title: Wikimedia Analytics Pageviews REST API
+      description: >
+          This API aims to provide coherent and low-latency access to
+          Wikimedia content and services. It is currently in beta testing, so
+          things aren't completely locked down yet. Each entry point has
+          explicit stability markers to inform you about development status
+          and change policy, according to [our API version
+          policy](https://www.mediawiki.org/wiki/API_versioning).
+
+          ### High-volume access
+            - Don't perform more than 500 requests/s to this API.
+            - Set a unique `User-Agent` header that allows us to contact you
+              quickly.  Email addresses or URLs of contact pages work well.
+            - Consider using our [HTML
+              dumps](https://phabricator.wikimedia.org/T17017) once they
+              become available.
+
+      termsOfService: https://wikimediafoundation.org/wiki/Terms_of_Use
+      contact:
+        name: the Wikimedia Services team
+        url: http://mediawiki.org/wiki/RESTBase
+      license:
+        name: Apache2
+        url: http://www.apache.org/licenses/LICENSE-2.0
+    security:
+      # ACLs for public *.wikipedia.org wikis
+      - mediaWikiAuth:
+        - user:read
+    x-subspecs:
+      - analytics_v1_pageviews
+      - test_analytics
+
+  wmf-sys-1.0.0: &wp/sys/1.0.0
+    info: 
+      title: Default MediaWiki sys API module
+      version: 1.0.0
+    paths:
+      /{module:table}: &wp/sys/table # Can use this anchor to share the table
+        x-modules:
+          # There can be multiple modules too per stanza, as long as the
+          # exported symbols don't conflict. The operationIds from the spec
+          # will be resolved against all of the modules.
+          - name: restbase-mod-table-cassandra
+            version: 1.0.0
+            type: npm
+            options: # Passed to the module constructor
+              conf:
+                hosts: [localhost]
+                keyspace: system
+                username: cassandra
+                password: cassandra
+                defaultConsistency: one # or 'one' for single-node testing
+                storage_groups:
+                  - name: test.group.local
+                    domains:
+                      - /test\..*\.org$/
+                      - /test\.local$/
+                  - name: default.group.local
+                    domains: /./
+
+      /{module:pageviews}:
+        x-modules:
+            - name: pageviews
+              version: 1.0.0
+              type: file
+
+  wp-default-1.0.0: &wp/default/1.0.0
+    x-subspecs:
+      - paths:
+          /{api:v1}:
+            x-subspec: *wp/content/1.0.0
+      - paths:
+          /{api:sys}:
+            x-subspec: *wp/sys/1.0.0
+
+
+spec: &spec
+  title: "The RESTBase root"
+  # Some more general RESTBase info
+  paths:
+    # test domain
+    /{domain:en.wikipedia.test.local}: *wp/default/1.0.0
+
+
+services:
+  - name: restbase
+    module: ./lib/server
+    conf: 
+      port: 7231
+      spec: *spec
+      salt: secret
+      default_page_size: 1
+
+test:
+  content_types:
+    html: text/html;profile=mediawiki.org/specs/html/1.1.0;charset=utf-8
+    wikitext: text/plain;profile=mediawiki.org/specs/wikitext/1.0.0;charset=utf-8
+
+logging:
+  name: restbase
+  level: info
+  #streams:
+  ## XXX: Use gelf-stream -> logstash
+  #- type: gelf
+  #  host: <%= @logstash_host %>
+  #  port: <%= @logstash_port %>
+
+metrics:
+  #type: txstatsd
+  #host: localhost
+  #port: 8125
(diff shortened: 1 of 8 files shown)


config.test.analytics.yaml
```diff
@@ -0,0 +1,124 @@
+# RESTBase config for Analytics APIs
```
Pchelolo: There's no need to define a separate `config.test.analytics.yaml`, as it's not picked up anywhere, therefor tests are failing in travis. Just adding 

```yaml
    /{module:pageviews}:
        x-modules:
            - name: pageviews
              version: 1.0.0
              type: file
```
to the `config.test.yaml` would be enough.

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,291 @@
...
+  termsofservice: https://github.com/wikimedia/restbase#restbase
+  contact:
+    name: Analytics
+    email: analytics@lists.wikimedia.org
+    url: https://www.mediawiki.org/wiki/Analytics
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+x-default-params:
+  domain: analytics.wikimedia.org
```
Pchelolo: This is not needed here. It's used in monitoring to pre-set the default domain in a monitor. Since you are not using monitoring (`x-monitor: false`) this is not needed.

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,291 @@
...
+      responses:
+        '200':
+          description: The list of values
+          schema:
+            $ref: '#/definitions/article'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-backend-request:
```
Pchelolo: This code requires a rebase with `wikimedia/restbase` master, because `x-backend-request` stanza was removed and replaced with `x-request-handler`.
See [docs](https://github.com/wikimedia/restbase/blob/master/doc/HandlerConfiguration.md) for examples of current handler configuration options.

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,291 @@
...
+        uri: /{domain}/sys/pageviews/per-article/{project}/{access}/{agent}/{article}/{granularity}/{start}/{end}
+      x-monitor: false
+
+  /{module:pageviews}/per-project/{project}/{access}/{agent}/{granularity}/{start}/{end}:
+    get:
+      tags:
+        - Pageviews data
+        - per-project
```
Pchelolo: I'd avoid too many tags, because in [swagger-driven docs](https://rest.wikimedia.org/en.wikipedia.org/v1/?doc) each tag becomes a separate section, so it's better to keep all page views endpoints under a single tag.

d00rman:
I'm quite confused. This PR seems to aim at integrating the *pageviews* logic completely into the upstream RESTBase. Alas, you will have your own RESTBase instance set up in the analytics cluster.

If you want to make your module publicly available, I think it would better to package and publish it as a separate npm module and document it there (cf [the Cassandra module](https://www.npmjs.com/package/restbase-mod-table-cassandra) and [the SQLite module](https://www.npmjs.com/package/restbase-mod-table-sqlite) for examples of stand-alone RESTBase modules).

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,291 @@
...
+      parameters:
+        - name: project
+          in: path
+          description: Name of the project
+          type: string
+          required: true
+        - name: access
+          in: path
+          description: Type of access method [all-access|desktop|mobile-app|mobile-web]
+          type: string
```
Pchelolo: In case a string is actually a enum, you could add a [swagger enum](http://stackoverflow.com/questions/27603871/how-to-define-enum-in-swagger-io) to get a nice choice box in resulting docs.

mods/pageviews.yaml
```diff
@@ -0,0 +1,27 @@
...
+    /per-project/{project}/{access}/{agent}/{granularity}/{start}/{end}:
+        get:
+          summary: query pageviews per project
+          operationId: pageviewsForProjects
+    /top/{project}/{access}/{year}/{month}/{day}:
+        get:
+          summary: query top pageviews
+          operationId: pageviewsForTops
+
+    # These three endpoints are just for fake data:
```
Pchelolo: All of these endpoints could be removed from the module and be completely spec-driven in the `analytics_test` config ([See docs](https://github.com/wikimedia/restbase/blob/master/doc/HandlerConfiguration.md))

mods/pageviews.js
```diff
@@ -0,0 +1,297 @@
...
+
+    return {
+        spec: spec,
+        operations: {
+            pageviewsForArticle: pjvs.pageviewsForArticle.bind(pjvs),
+            pageviewsForProjects: pjvs.pageviewsForProjects.bind(pjvs),
+            pageviewsForTops: pjvs.pageviewsForTops.bind(pjvs),
+
+            // These operations just insert fake data, uncomment them above to play
+            insertPageviewsForArticleTestData: pjvs.insertPageviewsForArticleTestData.bind(pjvs),
```
Pchelolo: These could be removed and mocked only by a spec.

Pchelolo:
One more confusion is that I didn't find any references to the backend analytics service, so it's not clear to me, how the data would be pulled into cassandra storage?

d00rman:
> it's not clear to me, how the data would be pulled into cassandra storage?

To answer your question @Pchelolo and clarify further my previous comment, there will soon be an additional RESTBase instance in WMF production to fulfil Analitics team's needs (*pageviews* API); cf. [Gerrit patch 231574](https://gerrit.wikimedia.org/r/#/c/231574/) which adds its production configuration. The idea is that the principal RESTBase instance only exposes the public API and simply forwards all of the requests to the internal *pageviews* RESTBase instance. As such, the only thing that should be in this PR is the specification that exposes the public API endpoints and their connection to the *pageviews* RESTBase instance.

d00rman:
> If you want to make your module publicly available, I think it would better to package and publish it as a separate npm module and document it there

Follow-up: AFAIC, I'm not against adding `mods/pageviews.js` (and related tests) per se. I do mind, however, putting additional example configs in the repo, as I really think this will become messy quite soon.

Regarding the public routes themselves, the `v1` should be evicted, as the final URI will contain it twice. Also, as per [T103811](https://phabricator.wikimedia.org/T103811), the three routes have different scopes:

- `/{module:stats}/views/v1/per-article/{access}/{agent}/{article}/{granularity}/{start}/{end}` should really be `/{module:page}/title/{title}/stats/{access}/{agent}/{granularity}/{start}/{end}` so that we end up with a URI  like `https://en.wikipedia.org/api/rest_v1/page/title/Foobar/stats/{access}/{agent}/{granularity}/{start}/{end}`
- `/{module:stats}/views/v1/per-project/{access}/{agent}/{granularity}/{start}/{end}` should become `/{module:data}/stats/views/{access}/{agent}/{granularity}/{start}/{end}`, which would effectively give clients an URI like `https://en.wikipedia.org/api/rest_v1/data/stats/views/{access}/{agent}/{granularity}/{start}/{end}`
- `/{module:stats}/views/v1/top/{access}/{year}/{month}/{day}` should, similarly to the previous one become `/{module:data}/stats/views/top/{access}/{year}/{month}/{day}` and should be added only to the global, domain-independent domain spec (cf PR #339 which proposes `wikimedia.org` as the global domain), so if a client wants to obtain top views for all projects, they'd compose a URI like `https://wikimedia.org/data/stats/views/top/{access}/{year}/{month}/{day}` (also, because this route will be available only in the global domain, I wonder where's the added value in having `top` in the route at all)

d00rman:
also let's cc @wikimedia/analytics and @wikimedia/services here :)

d00rman:
Closing in favour of #350

PR #344, opened 2015-09-19
Small promise chain hot path

gwicke:
Simple single-request handlers are by far the most common, so it's worth
optimizing for them. In local testing, this seems to improve throughput (as
measured by `ab -c 20 -n 10000
localhost:7231/en.wikipedia.org/v1/page/html/Foobar` from around 840 req/s to
about 870 req/s.

### DIFF
--- a/lib/handlerTemplate.js
+++ b/lib/handlerTemplate.js
@@ -264,11 +264,24 @@ function prepareRequestChain(conf) {
 function createHandler(spec) {
     validateSpec(spec);
     var requestChain = prepareRequestChain(spec);
-    return function(restbase, req) {
-        return generatePromiseChain(restbase, {
-            request: req
-        }, requestChain);
-    };
+    if (requestChain.length === 1
+            && requestChain[0].length === 1
+            && requestChain[0][0].request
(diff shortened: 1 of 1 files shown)


lib/handlerTemplate.js
```diff
@@ -264,11 +264,24 @@ function prepareRequestChain(conf) {
...
-        return generatePromiseChain(restbase, {
-            request: req
-        }, requestChain);
-    };
+    if (requestChain.length === 1
+            && requestChain[0].length === 1
+            && requestChain[0][0].request
+            && !requestChain[0][0].return
+            && !requestChain[0][0].catch) {
+        var handler = requestChain[0][0].request;
```
Pchelolo: Could be nice to wrap this ugly condition in a method with a descriptive name.

Pchelolo:
The same code have been added into https://github.com/wikimedia/restbase/pull/347, so we either need to close this one, or remove the same change from that PR

gwicke:
@Pchelolo, let me remove this from #347.

Pchelolo:
LGTM

PR #345, opened 2015-09-20
Allow HTTP pipelining

gwicke:
I don't actually recall why I added this header a long time ago. I do not
think that it still makes sense.

Clients like Varnish can get significantly better performance if pipelining is
enabled. Modern benchmark clients like wrk also get significantly better
throughput, and don't run out of connections.


Pchelolo:
LGTM

PR #347, opened 2015-09-21
Optimized templating

gwicke:
This PR optimizes request templating further. The templating micro-benchmark
is now about 2-3 times faster, with a further 3x speedup to be had if we
hoisted the try/catch one level.

However, profiles don't really show this code (before or after) as that much
of a bottleneck.  There are some gains to be made here for RESTBase as a
whole, but they will primarily benefit otherwise cheap entry points, where
request rewriting is a bigger part of the overall request cost. For most HTML
requests, much more time is spent in the Cassandra driver.

To summarize, I wouldn't call this change necessarily super-critical for
overall RESTBase performance, at this point.

- Don't silently ignore when entire sub-trees are undefined. This is still
  supported where needed (via default(), for example), but making this the
  default strikes me as dangerous.
- More aggressively expand all simple templates inline by rewriting local
  references to the path prefix. This isn't perfect yet, but does work rather
  well.
- Rename Template.eval to Template.expand; `eval` is not popular with jshint.

- More robust scope rewriting (see inline notes)
- Possibly, reduce the try/catch tax by moving that up to the request wrapper,
  similar to the technique described by the bluebird authors:
  https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax)
- Reconsider if we really want to default to `method: spec.method ||
  request.method || 'get'`, or just `method: spec.method || 'get'`. This
  strikes me as a little bit too much magic, and it's good to call out
  potentially state-changing requests explicitly.
- Remove commented-out inline microbenchmarking code.

### DIFF
--- a/lib/handlerTemplate.js
+++ b/lib/handlerTemplate.js
@@ -192,7 +192,11 @@ function prepareRequestChain(conf) {
         if (requestConf.request) {
             template = new Template(requestConf.request);
             requestSpec.request = function(restbase, context) {
-                var req = template.eval(context);
+                var req = template.expand(context);
+                if (!requestConf.request.method) {
+                    // Do we really want this?
+                    req.method = context.request.method || 'get';
+                }
                 return restbase.request(req)
                 .then(function(res) {
                     context[requestName] = res;
@@ -210,12 +214,12 @@ function prepareRequestChain(conf) {
                 requestSpec.request = function(restbase, context) {
                     return originalRequestHandler(restbase, context)
                     .then(function() {
-                        return template.eval(context);
+                        return template.expand(context);
                     });
                 };
             } else {
                 requestSpec.request = function(restbase, context) {
-                    return template.eval(context);
+                    return template.expand(context);
                 };
             }
         }
--- a/lib/reqTemplate.js
+++ b/lib/reqTemplate.js
@@ -4,45 +4,31 @@ var URI = require('swagger-router').URI;
 var TAssembly = require('tassembly');
 var url = require('url');
 var expressionCompiler = require('template-expression-compiler');
+require('core-js/shim');
 
-/**
- * Creates a function that sets a value at the path
- * and initializes all objects/arrays along the path.
- *
- * @param {string} path the path within an object/array separated
- *        with dots {e.g. test.body.a.b.c}
- * @param {object} spec spec element that contains provided path
- * @returns {Function} the function to set a value at path
- *          and initialize object along the path if needed
- * @private
- */
-function _setAtPath(path, spec) {
-    var pathArr = path.split('.');
-    var refArr = [];
-    var subspec = spec;
-    var code;
+var globalMethods = {
+    default: function(val, defVal) {
+        return val || defVal;
+    },
+    merge: function(destination, source) {
+        destination = destination || {};
+        source = source || {};
 
-    for (var idx = 0; idx < pathArr.length; idx++) {
-        var curRef = {};
-        curRef.ref = Array.isArray(subspec) ? '[' + pathArr[idx] + ']' : '["' + pathArr[idx] + '"]';
-        subspec = subspec[pathArr[idx]];
-        curRef.initializer = Array.isArray(subspec) ? '[]' : '{}';
-        refArr.push(curRef);
+        if (typeof destination !== 'object' || typeof source !== 'object') {
+            throw new Error('Illegal spec. Merge source and destination must be objects');
+        }
+
+        var result = Object.assign({}, destination);
+        Object.keys(source).forEach(function(keyName) {
+            if (result[keyName] === undefined) {
+                result[keyName] = source[keyName];
+            }
+        });
+        return result;
     }
+};
 
-    code =  'if (value !== undefined) {\n';
-    for (var i = 1; i < refArr.length; i++) {
-        var ref = refArr.slice(0, i).map(function(elem) { return elem.ref; }).join('');
-        code += 'obj' + ref + ' = obj' + ref + ' || ' + refArr[i - 1].initializer + ';\n';
-    }
-    code += 'obj' + refArr.map(function(elem) { return elem.ref; }).join('') + ' = value;\n';
-    code += '}';
-
-    /* jslint evil: true */
-    return new Function('obj', 'value', code);
-}
-
-function splitAndPrepareTAsseblyTemplate(templateSpec) {
+function splitAndPrepareTAsseblyTemplate(templateSpec, part) {
     var result = [];
     var templateNest = 0;
     var startIndex = 0;
@@ -59,7 +45,11 @@ function splitAndPrepareTAsseblyTemplate(templateSpec) {
         } else if (templateSpec[index] === '}') {
             if (templateNest === 1) { // The current template is finished
                 currentTemplate = templateSpec.substring(startIndex, index);
-                result.push(['raw', expressionCompiler.parse(currentTemplate)]);
+                var compiledExpression = expressionCompiler.parse(currentTemplate);
+                // FIXME: Rewrite path prefixes in expressionCompiler!
+                compiledExpression = compiledExpression.replace(/([,(\[:])m\./g,
+                            '$1rm.' + part + '.');
+                result.push(['raw', compiledExpression]);
                 startIndex = index + 1;
             } // Or and object literal finished
             templateNest--;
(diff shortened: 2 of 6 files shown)


Pchelolo:
On the last week I was trying to convince everyone that this code is 3 times slower, than the current implementation, however, over the weekends I've realised, that it was one of the most stupid things I've ever done :) The problem was that I've incorrectly benchmarked the baseline version (current), and so all my benchmarks for the new version couldn't achieve performance even close to the incorrect baseline. 

Sorry for that, shame on me.

lib/reqTemplate.js
```diff
@@ -230,44 +212,63 @@ function createURIResolver(spec) {
...
-            this.resolvers.forEach(function(resolver) {
-                resolver(newReq, context);
-            });
-            return newReq;
-        };
+    var completeTAssemblyTemplate;
+    try {
+        completeTAssemblyTemplate = expressionCompiler.parse(spec);
+    } catch (e) {
+        console.log('COMPILE ERROR');
```
Pchelolo: Do we really want this to go to console?

lib/reqTemplate.js
```diff
@@ -230,44 +212,63 @@ function createURIResolver(spec) {
...
+
+    var res;
+    var callback = function(bit) {
+        if (res === undefined) {
+            res = bit;
+        } else {
+            res += '' + bit;
+        }
+    };
+    var resolver = TAssembly.compile([['raw', completeTAssemblyTemplate]], {
```
Pchelolo: As I understand, we don't specify a different model and rootModel, so we can avoid compiling it as a nestedTemplate, this would simplify the code a bit.

lib/reqTemplate.js
```diff
@@ -230,44 +212,63 @@ function createURIResolver(spec) {
  *              fields that couldn't be resolved from original request would be ignored.
  */
 function Template(spec) {
+    spec = Object.assign({}, spec);
```
Pchelolo: Actually this should've been a deep clone. We are manipulating the spec here, replacing templates to TAssembly equivalent, and we don't want to expose that on a spec.

lib/reqTemplate.js
```diff
@@ -75,16 +65,53 @@ function splitAndPrepareTAsseblyTemplate(templateSpec) {
...
+            var uriTemplate = new URI(uri, {}, true);
+            return function(context) {
+                return uriTemplate.expand(context.rm.request.params);
+            };
+        })(spec.uri);
+    }
+}
+
+function errorHandler(e) {
+    console.error('ERR', e);
```
Pchelolo: Do we want `console.error` in production?

lib/reqTemplate.js
```diff
@@ -4,45 +4,31 @@ var URI = require('swagger-router').URI;
 var TAssembly = require('tassembly');
 var url = require('url');
 var expressionCompiler = require('template-expression-compiler');
+require('core-js/shim');
```
Pchelolo: This as needed only for a benchmark to work, in production it's not needed as already required way before this code

lib/handlerTemplate.js
```diff
@@ -264,11 +268,24 @@ function prepareRequestChain(conf) {
...
-    return function(restbase, req) {
-        return generatePromiseChain(restbase, {
-            request: req
-        }, requestChain);
-    };
+    if (requestChain.length === 1
+            && requestChain[0].length === 1
+            && requestChain[0][0].request
+            && !requestChain[0][0].return
+            && !requestChain[0][0].catch) {
```
Pchelolo: Could be nice to wrap this ugly condition in a function with a descriptive name.

lib/handlerTemplate.js
```diff
@@ -192,7 +192,11 @@ function prepareRequestChain(conf) {
         if (requestConf.request) {
             template = new Template(requestConf.request);
             requestSpec.request = function(restbase, context) {
-                var req = template.eval(context);
+                var req = template.expand(context);
+                if (!requestConf.request.method) {
+                    // Do we really want this?
+                    req.method = context.request.method || 'get';
```
Pchelolo: I'd say no, because now the template itself defaults to `get` whatever happens

lib/reqTemplate.js
```diff
@@ -75,16 +65,53 @@ function splitAndPrepareTAsseblyTemplate(templateSpec) {
...
+            newUri.urlObj = url.parse(newHost + path);
+            return newUri;
+        };
+    } else {
+        return (function(uri) {
+            var uriTemplate = new URI(uri, {}, true);
+            return function(context) {
+                return uriTemplate.expand(context.rm.request.params);
+            };
+        })(spec.uri);
```
Pchelolo: This self-invoking closure was added to avoid seeing original spec's elements rewrites, so as now it's cloned, I'm quite sure it's not needed, but not 100% sure.

gwicke:
> The problem was that I've incorrectly benchmarked the baseline version

The benchmark looked correct to me, I got similar results with the first version. I did find a performance bottleneck in how tassembly handled the try/catch in expression expansions, which I have since fixed. That combined with some more aggressive static prefixing got us most of the speed-up.

lib/reqTemplate.js
gwicke: You are right, let me remove this.

lib/reqTemplate.js
gwicke: Yeah, that's definitely worth looking into. I mostly went for the `nesteTemplate` mode for a little bit extra performance, but you are right that it might not be worth it.

lib/handlerTemplate.js
gwicke: Yeah, but it seems that quite a few of our specs currently rely on this behavior. Many POST end points (transforms, page save api) don't seem to specify a method in the x-request-handler.

I'm definitely +1 on keeping things simple & explicit, and only offering 'get' fall-back, but we should get agreement in the team before adjusting the specs.

@d00rman, @eevans: What's your take on this?

lib/handlerTemplate.js
gwicke: Yeah, did so in the other PR & removed the code change from this one.

lib/reqTemplate.js
gwicke: {{done}}

lib/reqTemplate.js
gwicke: Removed it.

lib/reqTemplate.js
gwicke: Yeah, it's not needed any more.

gwicke:
@Pchelolo, I saw that you earlier also had a question about spec cloning. `replaceComplexTemplates` now recursively clones the object structure while processing it, so a separate clone pass shouldn't be needed any more.

Pchelolo:
@gwicke yep, I've removed that comment when I noticed that `replaceComlexTemplates` is cloning the input.

lib/handlerTemplate.js
d00rman: I agree too that:
- we should keep things simple
- provide the fall-back in only one place
- the fall-back to be `GET`

Pchelolo: Actually, since in templating code we are already making a fallback (`spec.method = "'" + (spec.method || 'get') + "'";`), this branch would never execute. Or am I missing something?

gwicke: To be clear, the proposal is to

1) make all non-GET methods explicit in the specs (POST primarily);
2) remove the `|| req.method` fall-back.

d00rman: I like the current behaviour actually. If there, use the provided method. Otherwise, use the original request's method (effectively reusing the route's method from the spec). Finally, use `GET`. It makes the most sense to me.

I can't see an added benefit in making the method explicit in the spec. Every time.

gwicke: > I can't see an added benefit in making the method explicit in the spec. 

I do think that there is value in explicitly documenting potentially state changing backend requests. Most end points are GET only, so in most cases we don't need to mention the method & can rely on the `get` default.

d00rman: Right, so we do agree that if the method is `GET` there should be no need for the user to specify the `method:` stanza in the config. Why then:

> 1) make all non-GET methods explicit in the specs (POST primarily);

?

gwicke: > if the default is to use the request's method?

That's your proposal, I believe.

d00rman: > That's your proposal, I believe.

If I'm reading the code right, [it's already so](https://github.com/wikimedia/restbase/blob/56a45db06ab04820b1e3612183faf6109422e2a1/lib/reqTemplate.js#L264)

d00rman: On second thought, let's make `GET` the default if the method is not specified, regardless of the scope's method. The less magic, the better; there's already too much logic and too many things to keep track of in the config. It's risking to bite us (you know where) if we keep introducing mystic stuff  :P

gwicke: @d00rman, okay, great ;) Let me tackle that in a separate PR.

d00rman:
LGTM

PR #348, opened 2015-09-21
Re-revert the removal of v1 is mobileapps sys endpoints

Pchelolo:
In https://github.com/wikimedia/restbase/pull/342 @gwicke have reverted a change that removed /v1 from sys mobile endpoints, because it added yet another config change after staging testing and before the deploy. Now we can re-revert that change.


gwicke:
It seems to make most sense to merge this once we are ready to test & roll out these config changes. We should also minimize labs & vagrant breakage by updating those in sync.

d00rman:
There's already a MW-Vagrant patch waiting and as soon as we have a corresponding change in ops/puppet this is going out.

PR #349, opened 2015-09-21
Added support for cache-control: no-cache in /html/{title} endpoint

Pchelolo:
There's a bug on phabricator about adding support for `cache-control: no-cache` header to the `/html/{title}` endpoint. It's quite simple to implement now, because `/title/{title}` also supports the no-cache header.

This impl is not perfect, because it requires re-render even if the content hasn't been changed. It could be fixed by using the `page_touched` property from MW API, but we currently don't expose it from the `page_revisions` module. Exposing it would be one more nice use for page-related metadata when it's finished. Another option is to start exposing it by adding a static (?) column in the page_revisions table, but as I understand this property doesn't have much value for not-last revisions.

Bug: https://phabricator.wikimedia.org/T97387

### DIFF
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -220,10 +220,13 @@ paths:
       tags:
         - Page content
       description: >
-        Retrieve the latest html for a page title.
+        Retrieve the latest html for a page title available in storage.
 
         If you know the revision as well, then please use the
         `{title}/{revision}` end point instead for better performance.
+
+        Supply the `Cache-Control: no-cache` header to force a fetch
+        of the most recent version.
 
         See [the MediaWiki DOM
         spec](https://www.mediawiki.org/wiki/Parsoid/MediaWiki_DOM_spec) for a
--- a/mods/page_revisions.js
+++ b/mods/page_revisions.js
@@ -178,6 +178,20 @@ PRS.prototype.listTitles = function(restbase, req, options) {
     });
 };
 
+/**
+ * Checks if two revisions are the same, ignoring different tid values.
+ * @private
+ */
+PRS.prototype._checkSameRev = function(firstRev, secondRev) {
+    return !Object.keys(firstRev).some(function(attrName) {
+        // We don't really care if an empty value is null, or undefined, or other falsy
+        if (attrName === 'tid' || !firstRev[attrName] || !secondRev[attrName]) {
+            return false;
+        }
+        return firstRev[attrName] !== secondRev[attrName];
+    });
+};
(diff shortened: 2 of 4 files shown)


specs/mediawiki/v1/content.yaml
```diff
@@ -220,11 +220,13 @@ paths:
       tags:
         - Page content
       description: >
-        Retrieve the latest html for a page title.
+        Retrieve the latest html for a page title among known to RESTBase.
```
d00rman: `s/among//`

d00rman: Also, `s/known to RESTBase/available in storage/` (clients shouldn't care it's RESTBase responding to the requests)

d00rman:
LGTM modulo minor comment.

> This impl is not perfect, because it requires re-render even if the content hasn't been changed. It could be fixed by using the page_touched property from MW API, but we currently don't expose it from the page_revisions module.

We couldn't detect needed re-renders if we actually do it that way. If an included template has changed, `page_touched` isn't updated, but the rendering of the page has.

gwicke:
As implemented, it looks like this would swamp the revision table with lots of duplicate revision entries, one for each request. Given that we have about 200 re-render requests per second, that would quickly lead to a very bloated revision table.

I think we would at least need to avoid saving duplicate information. 

Out-of-date page metadata doesn't seem to be such a big issue in practice either. I expected a few isolated complaints about individual pages by now, but haven't heard of any.

d00rman:
> As implemented, it looks like this would swamp the revision table with lots of duplicate revision entries, one for each request. Given that we have about 200 re-render requests per second, that would quickly lead to a very bloated revision table.

Aren't we upserting in the revision table? If not, we should. Re-renders of a revision's info doesn't make much sense to me (i.e. let's drop the TID from the hash there)

gwicke:
No, that has several deeper consequences that we probably don't want.

Pchelolo:
Actually, since we provide support for `cache-control: no-cache` header on `/page/title/{title}` endpoint, checking revisions for equality should have been done long time ago. I'll update the PR.

Pchelolo:
Updated the PR and added an equality check.

d00rman:
It seems this PR is based on outdated version of #350 but should be based on master.

Pchelolo:
@d00rman I've rebased the PR on current master, not sure how did PR 350 get in here, I might've messed something up with branches locally.

d00rman:
LGTM but needs a rebase

Pchelolo:
@d00rman rebased

PR #350, opened 2015-09-21
Add analytics pageview module, spec, and tests

milimetric:
This commit just adds the backend necessary to serve analytics
data.  We will add any code needed to configure a front-end in a later
pull request.

Bug: T107053


Pchelolo:
@milimetric Hm, the PRs seem to to get duplicated (https://github.com/wikimedia/restbase/pull/343) Can I close the first one? 

I think this would be easier to fix if I just create a PR for your branch :) I'll do it today

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,295 @@
...
+          description: The list of values
+          schema:
+            $ref: '#/definitions/article'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_backend:
+          request:
```
Pchelolo: The source of your problem is indentation here. Should be:
```yaml
x-request-handler:
  - get_from_backend:
      # Note how `request` is one-tab more indented than it your code.
      request:
        uri: ...
```

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,295 @@
...
+          description: The list of values
+          schema:
+            $ref: '#/definitions/project'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_backend:
+          request:
```
Pchelolo: Idem

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,295 @@
...
+          description: The list of top articles in the project
+          schema:
+            $ref: '#/definitions/tops'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_backend:
+          request:
```
Pchelolo: Idem

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,295 @@
...
+      parameters:
+        - name: project
+          in: path
+          description: Name of the project [all-projects|project-name]
+          type: string
+          required: true
+        - name: access
+          in: path
+          description: Type of access method [all-access|desktop|mobile-app|mobile-web]
+          type: string
```
Pchelolo: For parameters where there's a fixed set of possible values, use swagger's `enum` stanza to specify them all.

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,295 @@
...
+        items:
+          type: string
+
+  article:
+    required:
+      - items
+    properties:
+      items:
+        type: array
+        properties:
```
Pchelolo: This is a invalid swagger. Should be 
```yaml
properties:
  items:
    type: array
    items:
      properties:
        ...
```
Alternatively you could put the description of an individual object to a separate definition and use `$ref` stanza like [here](https://github.com/wikimedia/restbase/blob/master/specs/mediawiki/v1/content.yaml#L1177)

This comment applies to many places in this file, to check that you've fixed all of them - start the service and load [docs](http://localhost:7231/en.wikipedia.org/v1/?doc#/) - if it loads - all is fine.

specs/test.yaml
```diff
@@ -92,3 +92,60 @@ paths:
             request:
               uri: /{domain}/sys/post_data/post.test/{hash}
       x-monitor: false
+
+  /{module:post_data}/insert-per-article/{project}/{access}/{agent}/{article}/{granularity}/{timestamp}/{views}:
```
Pchelolo: It's not `{module:post_data}`, it's `{module:pageviews}`

specs/test.yaml
```diff
@@ -92,3 +92,60 @@ paths:
...
+                  project: '{$.request.params.project}'
+                  access: '{$.request.params.access}'
+                  agent: '{$.request.params.agent}'
+                  article: '{$.request.params.article}'
+                  granularity: '{$.request.params.granularity}'
+                  timestamp: '{$.request.params.timestamp}'
+                  views: '{$.request.params.views}'
+      x-monitor: false
+
+  /{module:post_data}/insert-per-project/{project}/{access}/{agent}/{granularity}/{timestamp}/{views}:
```
Pchelolo: Idem

specs/test.yaml
```diff
@@ -92,3 +92,60 @@ paths:
...
+                attributes:
+                  project: '{$.request.params.project}'
+                  access: '{$.request.params.access}'
+                  year: '{$.request.params.year}'
+                  month: '{$.request.params.month}'
+                  day: '{$.request.params.day}'
+                  rank: '{$.request.params.article}'
+                  article: '{$.request.params.article}'
+                  views: '{$.request.params.views}'
+      x-monitor: false
```
Pchelolo: There's no `rank` or `views` columns in the tops table, so you are inserting just `articles`. These should be manually `JSON.stringify`ed in the test code.

config.test.yaml
```diff
@@ -39,6 +39,7 @@ templates:
       - mediawiki/v1/content
       - mediawiki_v1_graphoid
       - mediawiki/v1/mobileapps
+      - analytics_v1_pageviews
```
Pchelolo: We should think carefully on the naming scheme, because we will have 2 analytics configs, and we really don't want to mess them up.

I propose the following:
1. `/analytics/v1/pageviews.yaml` - the whole analytics folder would belong to backend analytics RESTBase.
2. `/data/v1/analytics.yaml` - here we would put all publicly-exposed endpoints in the front-end RESTBase.

@gwicke @d00rman @eevans Any thoughts on the naming scheme?

.gitignore
```diff
@@ -5,3 +5,6 @@ node_modules
 npm-debug.log
 # WebStorm IDE files
 .idea/*
+*.iml
+
+*swp
```
Pchelolo: What is this file about?

config.test.yaml
```diff
@@ -171,6 +172,12 @@ templates:
           - name: page_save
             type: file
 
+      /{module:pageviews}:
+        x-modules:
+            - name: pageviews
```
Pchelolo: This is indented one tab more than it should be.

Pchelolo:
I think after fixing errors I've pointed the tests should pass. I'll take one more closer look after that, but in general this all looks good. The only big think left is bike shedding on spec's naming.

config.test.yaml
d00rman: I agree. For the second, though, I'd name it `/data/v1/stats/views.yaml` as their endpoints will likely end up as `/{domain}/v1/data/stats/views/...`

d00rman: Filed as [T113349](https://phabricator.wikimedia.org/T113349)

gwicke:
This kind of brings up the question of whether RESTBase should eventually contain and test a union of all possible RESTBase configs. Right now we bundle the Wikimedia sub-specs & test them along with RESTBase-the-framework, but at some point we might want to separate the config from the code.

.gitignore
milimetric: adding *swp here means ignoring the temporary files that vim creates (I use vim, so I added this here.  In gerrit, it probably would've been a separate change :))

config.test.yaml
milimetric: Once you guys decide, I'm happy to switch it either way.

specs/analytics_v1_pageviews.yaml
milimetric: oops.  I noticed the error was about a syntax problem, but I didn't know how to go from there to where the problem was.  Do you all use a linter or just know what problems to look for?

specs/test.yaml
milimetric: doh, your example explained that very well, I got confused and thought the insert code was still executing even though I deleted it.  Man, I gotta say I suck at this kind of coding.  Do you have a way of testing small changes to see if they worked or are you guys just familiar with the whole system and don't run into small problems?  I'm a front-end developer so I'm used to the "change / run small test" cycle :)

specs/test.yaml
milimetric: Done

specs/test.yaml
milimetric: Done

specs/analytics_v1_pageviews.yaml
milimetric: Done.  docs loads ok now.

specs/analytics_v1_pageviews.yaml
milimetric: Done.

specs/analytics_v1_pageviews.yaml
milimetric: Done.

config.test.yaml
milimetric: Done.

specs/analytics_v1_pageviews.yaml
milimetric: Done.

specs/analytics_v1_pageviews.yaml
Pchelolo: @milimetric The error message was indeed not too descriptive, I would've get confused if I didn't know what's going on. We should make an error message more self-explanatory

specs/test.yaml
Pchelolo: @milimetric I simply run `npm test` all the time. BTW, if you don't have cassandra installed, you could run `npm test -- sqlite`. It also speeds up local testing a bit

.gitignore
Pchelolo: ok, just I'm not a vim user (still no idea how to exit it :) ), so I was wondering what it's about

specs/test.yaml
```diff
@@ -92,3 +92,58 @@ paths:
...
+                attributes:
+                  project: '{$.request.params.project}'
+                  access: '{$.request.params.access}'
+                  agent: '{$.request.params.agent}'
+                  granularity: '{$.request.params.granularity}'
+                  timestamp: '{$.request.params.timestamp}'
+                  views: '{$.request.params.views}'
+      x-monitor: false
+
+  /{module:pageviews}/insert-top/{project}/{access}/{year}/{month}/{day}/{rank}/{article}/{views}:
```
Pchelolo: this should become `/{module:pageviews}/insert-top/{project}/{access}/{year}/{month}/{day}/{data}` and everything will go well

.gitignore
milimetric: :q  to exit.  I know how you feel.  I once accidentally opened "ed" and had to shut off the computer :P

mods/pageviews.js
```diff
@@ -0,0 +1,217 @@
...
+
+// Pageviews Service
+function PJVS(options) {
+    this.options = options;
+    this.log = options.log || function() {};
+}
+
+
+var tables = {
+        article: 'pageviews.per.article',
```
d00rman: Nit: indentation

mods/pageviews.js
```diff
@@ -0,0 +1,217 @@
...
+        project: 'pageviews.per.project',
+        tops: 'top.pageviews',
+    };
```
d00rman: Idem

mods/pageviews.js
```diff
@@ -0,0 +1,217 @@
...
+var tableURI = function(domain, tableName) {
+        return new URI([domain, 'sys', 'table', tableName, '']);
+    };
```
d00rman: Idem both the above lines.

mods/pageviews.js
```diff
@@ -0,0 +1,217 @@
...
+var tableSchemas = {
+        article: {
```
d00rman: Yup, you've guessed it - indentation :)

mods/pageviews.js
```diff
@@ -0,0 +1,217 @@
...
+    };
+
+/* general handler functions */
+var queryCatcher = function(e) {
+        if (e.status !== 404) {
+            throw e;
+        }
+    };
+var queryResponser = function(res) {
+        // always return at least an empty array so that queries for non-existing data don't error
```
d00rman: Hm, I wonder if this is the correct way to go about this. I agree that having no data for a specific period for an article/project can result in an *empty* 200 OK response, but receiving the same response when asking for data for an article/project that doesn't even exist is wrong and not RESTful, but rather MW API-ish.

milimetric: Yeah, I was torn about this as well.  I propose we leave it like this as an experiment and talk to people as they use the service.  Sadly, there's no way to say "this data should exist but doesn't".  I leaned towards the empty result because I find that easier to deal with on the front end (in dashboards, etc.) in cases where I'm expecting data to be there but it's just not ready yet or something.

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,297 @@
...
+          required: true
+        - name: agent
+          in: path
+          description: Type of user agent
+          type: string
+          enum: ['all-agents', 'user', 'spider', 'bot']
+          required: true
+        - name: granularity
+          in: path
+          description: Time granularity of the response data [hourly|daily]
```
Pchelolo: Could also become a enum

specs/analytics_v1_pageviews.yaml
```diff
@@ -0,0 +1,297 @@
...
+          enum: ['all-agents', 'user', 'spider', 'bot']
+          required: true
+        - name: article
+          in: path
+          description: Name of the article for which to get data
+          type: string
+          required: true
+        - name: granularity
+          in: path
+          description: Time granularity of the response data [hourly|daily]
```
Pchelolo: Could also become a enum

Pchelolo:
Except for two tiny nits this all looks good to me. (given that we didn't finish choosing the spec directory structure vs plain files, but that could be added/fixed separately.

.gitignore
d00rman: > adding *swp here means ignoring the temporary files that vim creates (I use vim, so I added this here. In gerrit, it probably would've been a separate change :))

Tip:

```sh
mkdir -p ~/tmp/vim
echo "set backupdir=~/tmp/vim" >> ~/.vimrc
echo "set directory=~/tmp/vim//" >> ~/.vimrc
```

mods/pageviews.js
d00rman: I understand the motivation, but, frankly, swallowing errors because of a dashboard's niceties is wrong. All the more so because we are talking about a back-end service/module. We are very far away from the presentation layer here.

IMHO, the separation is quite good if you think about it:

- wrong date or param => receive a 400 (bad request)
- all good, but no data for that period => receive a 200 with an empty set
- no such project/article (resource in general) => receive a 404

> Sadly, there's no way to say "this data should exist but doesn't"

That is partly covered by the second bullet point above, but in general that's the paradigm of a stateless service - a resource is either there or not - there is no previous state form which it can deduce the appropriate *action*. In the REST philosophy it is the client's job to determine the appropriate action to undertake, given that it and only it has the context of the request.

mods/pageviews.js
```diff
@@ -0,0 +1,217 @@
...
+            { attribute: 'access', type: 'hash' },
+            { attribute: 'year', type: 'hash' },
+            { attribute: 'month', type: 'hash' },
+            { attribute: 'day', type: 'hash' },
+        ]
+    }
+};
+
+/* general handler functions */
+var queryCatcher = function(e) {
```
d00rman: Indentation for the whole function.

mods/pageviews.js
```diff
@@ -0,0 +1,217 @@
...
+        if (e.status !== 404) {
+            throw e;
+        }
+    };
+var queryResponser = function(res) {
```
d00rman: Idem, indentation for the whole function.

config.test.yaml
Pchelolo: @milimetric We've settled on a directory-based structure of configs, so this should go to `analytics/v1/pageviews.yaml`

mods/pageviews.js
```diff
@@ -0,0 +1,277 @@
...
+            }
+        });
+    }
+};
+
+/**
+ * Cleans the project parameter so it can be passed in as either en.wikipedia.org or en.wikipedia
+ */
+var stripOrgFromProject = function(rp) {
+    rp.project = rp.project.replace(/^(.*).org$/, '$1');
```
Pchelolo: The dot in `.org` should be escaped. Also, I think a cleaner way of doing the would be `rp.project.replace(/\.org$/, '')`, but it's up to you.

d00rman: +1 to both.

mods/pageviews.js
```diff
@@ -0,0 +1,277 @@
...
+};
+
+var validateStartAndEnd = function(rp) {
+    var errors = [];
+
+    stripOrgFromProject(rp);
+
+    if (!/^[0-9]{10}$/.test(rp.start)) {
```
d00rman: 10 digits? What kind of time stamp is that?

mods/pageviews.js
```diff
@@ -0,0 +1,277 @@
...
+
+    throwIfNeeded(errors);
+};
+
+var validateYearMonthDay = function(rp) {
+    var errors = [];
+
+    stripOrgFromProject(rp);
+
+    if (rp.year === 'all-years') {
```
d00rman: Hm, not sure it'd be wise to do that, as it means the client's request params would actually get overwritten.

mods/pageviews.js
```diff
@@ -0,0 +1,277 @@
...
+        rp.month = 'all-months';
+    }
+    if (rp.month === 'all-months') {
+        rp.day = 'all-days';
+    }
+
+    if (rp.year !== 'all-years' && !/^[0-9]{4}$/.test(rp.year)) {
```
d00rman: A sanity check for values would be good for all three blocks here (as in, I can put in only digits, but that's not enough for the date to be valid).

mods/pageviews.js
Pchelolo: @d00rman YYYYMMDDHH - 10 digits. It's reflected in the endpoint docs.

Actually, we could make a better regex for that, because we know the possible range for some numbers. Something like that could work: `/^[1-2][0-9]{3}[0-1][0-9][0-3][0-9][0-2][0-9]$/`

d00rman: > YYYYMMDDHH - 10 digits. It's reflected in the endpoint docs.

Ah I see. Thnx @Pchelolo 

> Actually, we could make a better regex for that, because we know the possible range for some numbers. Something like that could work: /^[1-2][0-9]{3}[0-1][0-9][0-3][0-9][0-2][0-9]$/

We might, but the TS still needs to be checked. Even your RegExp allows incorrect dates like `2987193929`.

mods/pageviews.js
milimetric: I was going to add a note, maybe I should have.  The only way to fully validate the dates would be to include another library, because basic checking still wouldn't consider leap years, leap seconds, etc.  Given that, I just do basic validation and assume a 404 is good enough for when people send invalid dates.  If you'd like to see another library here, that's cool with me, I just didn't want to add to the dependency tree.

mods/pageviews.js
milimetric: I'll document this better in the parameters, to explain the behavior.  But we didn't want to go down the weird path of returning the month of July for "all years".  So is that ok if the behavior is explained but stays the same?

mods/pageviews.js
milimetric: Same comment as below, basically I think dates are too hard to validate.  Between my basic check and Petr's, I think I'd favor his.

mods/pageviews.js
d00rman: I agree that having *july for all years* makes no sense at all, but I'd be inclined to say that such requests should be rejected precisely because they validity is questionable. I'm also thinking about caching: if requests to `/all-years/07` and `/all-years/all-months` are equivalent, there should really be only one entry in the cache, not two (or 365 of them in the extreme case).

mods/pageviews.js
d00rman: Now that we have `validateTimestamp` we should make it work for this case too, i.e. supply it `rp.year`, `rp.month` and `rp.day` and see if the date is valid.

mods/pageviews.js
```diff
@@ -0,0 +1,302 @@
...
+    if (!/^[0-9]{10}$/.test(timestamp)) {
+        return false;
+    }
+
+    var year = timestamp.substring(0, 4);
+    var month = timestamp.substring(4, 6);
+    var day = timestamp.substring(6, 8);
+    var hour = timestamp.substring(8, 10);
+
+    var dt = new Date([year, month, day].join('-') + ' ' + hour + ':00:00 UTC');
```
d00rman: `new Date(year, month, day, hour)` should work just as well.

mods/pageviews.js
```diff
@@ -0,0 +1,302 @@
+'use strict';
+
+/**
+ * Pageviews API module
+ *
+ * Main tasks:
+ * - TBD
```
d00rman: Mind expanding this slightly?

specs/analytics/v1/pageviews.yaml
```diff
@@ -0,0 +1,299 @@
...
+              uri: /{domain}/sys/pageviews/per-project/{project}/{access}/{agent}/{granularity}/{start}/{end}
+      x-monitor: false
+
+  /{module:pageviews}/top/{project}/{access}/{year}/{month}/{day}:
+    get:
+      tags:
+        - Pageviews data
+        - Top articles
+      description: >
+        List pageviews for a given project, over a given time period.
```
d00rman: Nit: the desc seems to be c/p-ed from the previous endpoint, but *top* could use some clarification

d00rman:
Modulo some minor comments and nits, LGTM. @wikimedia/services thoughts?

mods/pageviews.js
milimetric: no, that doesn't work, specifying UTC is important here

milimetric:
Thanks Marko, I think I patched everything except the date construction
suggestion.  That needs to use the weird string concat so we can specify
'UTC', otherwise this validation won't work when it runs in different time
zones.  Even though all our servers use UTC (thank deity), testing would be
affected.

mods/pageviews.js
```diff
@@ -0,0 +1,309 @@
...
+
+    if (rp.month !== 'all-months' && !validDate) {
+        errors.push('month must be "all-months" or a valid 2-digit month');
+    }
+
+    if (rp.day !== 'all-days' && !validDate) {
+        errors.push('day must be "all-days" or a valid 2-digit day');
+    }
+
+    throwIfNeeded(errors);
```
Pchelolo: I'm not sure we need to provide such level of detail here. For 2014-01-49 all 3 errors would happen, while year and month are valid. I think just returning smth like 'Invalid timestamp' would be enough

Pchelolo:
LGTM :+1:

mods/pageviews.js
```diff
@@ -0,0 +1,313 @@
...
+            { attribute: 'year', type: 'hash' },
+            { attribute: 'month', type: 'hash' },
+            { attribute: 'day', type: 'hash' },
+        ]
+    }
+};
+
+/**
+ * general handler functions */
+var queryResponser = function(res) {
```
d00rman: Nit: should probably be `queryResponder` (sorry, my OCD doesn't allow me to ignore this) :P

mods/pageviews.js
d00rman: Touché

mods/pageviews.js
milimetric: :) Sorry for the pun.  It's not quite a responder because it's not responding to queries, it's just fixing the response.  I'll name it accordingly.

d00rman:
LGTM too :+1:

d00rman:
@milimetric just rebase on top of master and we're GTG!

PR #351, opened 2015-09-22
Parsoid mod: Add warn log lines for transfroms to wikitext

d00rman:
Somewhere in the VE <> VRS <> RESTBase <> Parsoid pipeline something breaks and 400 errors are returned to VE. This is a temporary patch for RESTBase so that it emits warn-level log lines when such a situation occurs.

Note: this patch only lets us get around the production logger warn-level setting because of which the emitted errors do not make it through to logstash. That means that RESTBase will in reality emit two log lines for each of the errors targeted here. Hence, as soon as we figure out what needs to be fixed this commit should be reverted.

Bug: [T112339](https://phabricator.wikimedia.org/T112339)

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -17,6 +17,7 @@ var spec = yaml.safeLoad(fs.readFileSync(__dirname + '/parsoid.yaml'));
 
 function ParsoidService(options) {
     options = options || {};
+    this.log = options.log || function() {};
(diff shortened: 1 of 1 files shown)


d00rman:
cc @gwicke @eevans @Pchelolo @subbuss

Pchelolo:
LGTM

eevans:
LGTM

PR #352, opened 2015-09-23
low-cased the swagger's CSP policy header

Pchelolo:
I've just noticed that PR https://github.com/wikimedia/restbase/pull/333 have completely broken the swagger ui documentation. The problem was that swaggerUI.js was setting a camel-cased CSP header, and so additional low-cased CSP header was added. Apparently the last one took precedence in Safari (may be in other browsers too), so the browser couldn't load any resources.


d00rman:
Nice!

PR #353, opened 2015-09-23
Swagger Docs: Update RESTBase URL

d00rman:
`restbase.org` has expired, so let's link the logo in the documentation to the mediawiki.org page about RESTBase.


d00rman:
cc @wikimedia/services

Pchelolo:
LGTM

PR #354, opened 2015-09-24
Restrict only user/comment information if needed

Pchelolo:
There are several types of revision deletion flags: `texthidden`, `sha1hidden`, `commenthidden`, `userhidden` and `suppressed`. Only for the first 2 types 403 should be returned by restbase. For `commenthidden` and `userhidden` we just should delete user or comment info. The `suppressed` flag states that other restrictions are applied to admins too, but as we can't distinguish admins, we can simply ignore it. 

Related:
Docs for revision deleting: https://www.mediawiki.org/wiki/Manual:RevisionDelete
MW API source, that sets the flags: https://doc.wikimedia.org/mediawiki-core/master/php/ApiQueryRevisionsBase_8php_source.html 

Bug: https://phabricator.wikimedia.org/T113117

### DIFF
--- a/test/features/pagecontent/revisions.js
+++ b/test/features/pagecontent/revisions.js
@@ -167,6 +167,19 @@ describe('revision requests with en.wikipedia.org', function() {
             assert.deepEqual(res.status, 403);
         });
     });
+
+    it('should restrict user and comment', function() {
+        return preq.get({
+            uri: server.config.bucketURL + '/title/User:Pchelolo%2fRestricted_Rev'
+        })
+        .then(function(res) {
+            assert.deepEqual(res.status, 200);
+            var item = res.body.items[0];
+            assert.deepEqual(item.user_id, null);
+            assert.deepEqual(item.user_text, null);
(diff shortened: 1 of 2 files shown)


test/features/pagecontent/revisions.js
```diff
@@ -167,6 +167,19 @@ describe('revision requests with en.wikipedia.org', function() {
...
+    it('should restrict user and comment', function() {
+        return preq.get({
+            uri: server.config.bucketURL + '/title/User:Pchelolo%2fRestricted_Rev'
+        })
+        .then(function(res) {
+            assert.deepEqual(res.status, 200);
+            var item = res.body.items[0];
+            assert.deepEqual(item.user_id, null);
+            assert.deepEqual(item.user_text, null);
+            assert.deepEqual(item.comment, null);
```
d00rman: For all of the three checks above: better to have:

```js
assert.deepEqual(!!item.comment, false);
```

so that the check passes regardless of the false-y value the field is set to (`null`, `undefined`, `''`, etc)

PR #355, opened 2015-09-24
Return 409 if a tranform request is done for a restricted revision

Pchelolo:
Right now if the page was deleted or a revision was restricted while it was edited in VE, transform endpoints return 400 errors, however this situation is more about an edit conflict.

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -707,14 +707,14 @@ PSP.makeTransform = function(from, to) {
         return transform
         .catch(function(e) {
             // In case a page was deleted/revision restricted while edit was happening,
-            // return 429 Conflict error instead of a general 400
+            // return 410 Gone or 409 Conflict error instead of a general 400
             var pageDeleted = e.status === 404 && e.body
                     && /Page was deleted/.test(e.body.description);
             var revisionRestricted = e.status === 403 && e.body
                     && /Access is restricted/.test(e.body.description);
             if (pageDeleted || revisionRestricted) {
                 throw new rbUtil.HTTPError({
-                    status: 429,
+                    status: pageDeleted ? 410 : 409,
                     body: {
                         type: 'revision#conflict',
                         title: 'Conflict detected',
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -716,6 +716,22 @@ PSP.makeTransform = function(from, to) {
                     status: e.status
                 });
             }
+            // In case the a page was deleted/revision restricted wkile edit was happening,
+            // return 429 Conflict error instead of a general 400
+            var pageDeleted = e.status === 404 && e.body
+                    && /Page was deleted/.test(e.body.description);
+            var revisionRestricted = e.status === 403 && e.body
+                    && /Access is restricted/.test(e.body.description);
+            if (pageDeleted || revisionRestricted) {
+                throw new rbUtil.HTTPError({
+                    status: 429,
+                    body: {
+                        type: 'revision#conflict',
+                        title: 'Conflict detected',
+                        description: e.body.description
+                    }
+                });
+            }
             throw e;
         })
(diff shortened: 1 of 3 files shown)


d00rman:
LGTM. @subbuss do you think it is a good practice to return a 429 CONFLICT when a transform for a deleted page or revdeleted revision is attempted?

mods/parsoid.js
```diff
@@ -715,6 +715,22 @@ PSP.makeTransform = function(from, to) {
                     status: e.status
                 });
             }
+            // In case the a page was deleted/revision restricted wkile edit was happening,
```
d00rman: Hmm, while it is true that the code above will disappear soon, currently it will log something other than 429 (either a 404 or a 403). Should we try to integrate these two so that we can easily match user reports with logs (which was the initial incentive to introduce this temp code) ?

Also, some small nits on this line:
`s/the//`
`s/wkile/while/`

d00rman:
@subbuss rightfully pointed out our mistake. 429 stands for *too many requests*, which has nothing to do with what we are trying to achieve here.

We have at our disposal 409 (*CONFLICT*), 410 (*GONE*) and 423 (*LOCKED*). Ideally, I'd vote for a 410 for deleted pages and 423 for revdeleteled revisions, but since 423 is a WebDAV code, I propose to go with 410 for deleted pages and 409 for revdeletions.

gwicke:
Similar to the HTML save API, we already ask clients to set an `If-Match` header specifying the original etag (revision/timeuuid) in transform end points. Right now, we still provide the meta tag hack fall-back (mostly for VE's benefit), but I think we should really phase that out in favor of working with `If-Match` only. On conflict (rev / page deletion), this will make a [412 Precondition Failed](http://tools.ietf.org/html/draft-ietf-httpbis-p4-conditional-22#section-4.2) response the most likely choice.

Pchelolo:
@d00rman hah, we've been fighting so much with 429 codes, that it just settled in my head..

@gwicke The `if-match` header is parsed (required) only in html/to/* endpoints, so it might be not passed on others, like wikitext to html transformation, and this is needed on all transformations.

d00rman:
Let's keep it like that for now, but I agree with @gwicke that in the long term 412 would be the way to go.

gwicke:
> @gwicke The if-match header is parsed (required) only in html/to/* endpoints, so it might be not passed on others, like wikitext to html transformation, and this is needed on all transformations.

The meta tag hack certainly does not work for wikitext, and [the end point does not support passing in the timeuuid at all right now](https://en.wikipedia.org/api/rest_v1/?doc#!/Transforms/post_transform_wikitext_to_html_title_revision). It is also not absolutely needed, as any html/data-parsoid copy will do for incremental re-parsing, as long as it's consistent.

It would however be more consistent to move to using the ETag for wikitext as well. We can also support passing it in the POST data with a mechanism like https://phabricator.wikimedia.org/T111748.

Pchelolo:
@gwicke Actually, you've made me think that we could be handling this in a wrong way. 

On html->* transforms, the tid of the original content is required now, but in case original html is not in storage, and we are generating it, we font's include that tid in request for revision info, and get the latest info version of revision info. In this particular case, the page's been deleted, so we get these errors. Shouldn't we instead get the revision info for the original tid? (the latest one before the original tid)

gwicke:
@Pchelolo, I guess you were thinking about a use case where somebody editing a now-deleted revision will still expect a successful save based on that revision? In that case, we don't really have a tid for the revision info -- all we have is the timestamp contained in that tid. It might be possible to retrieve the revision info based on that, but it's icky.

The risk I see with enabling such an edit is that the revision deleted content could be perpetuated in the edit. From a user perspective, triggering an edit conflict workflow would probably make most sense, which fits with the `If-Match` handling.

PR #356, opened 2015-09-24
Purging: Minor improvement in the promise chain

d00rman:
Purging is a conditional action which defaults to false. So, instead of creating an additional `.then()` block in which we check whether a purge should happen, do it vice-versa and append a `.then()` block only in case the option is actually active.


Pchelolo:
@d00rman Aha, now it's not that late so I get it :)  LGTM

PR #357, opened 2015-09-25
Added support for reverted order of properties in a timeuuid meta tag

Pchelolo:
Although we want to move to `if-match` header instead of the TimeUuid meta tag hack, right now we have to fix the meta regex, since we have it. In IE11 the order of properties in a meta tag is reversed, so it wasn't matching correctly. Also a trailing whitespace is possible, support for it was also added into the RegExp.

Bug: https://phabricator.wikimedia.org/T113163

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -509,8 +509,8 @@ PSP.transformRevision = function(restbase, req, from, to) {
             tid = rbUtil.parseETag(req.headers['if-match']).tid;
         } else if (req.body && req.body.html) {
             // Fall back to an inline meta tag in the HTML
-            var tidMatch = new RegExp('<meta\\s(?:content="([^"]+)"\\s)?' +
-                    'property="mw:TimeUuid"(?:\\scontent="([^"]+)")?\\s*\\/?>')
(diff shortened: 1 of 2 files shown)


mods/parsoid.js
```diff
@@ -509,9 +509,10 @@ PSP.transformRevision = function(restbase, req, from, to) {
             tid = rbUtil.parseETag(req.headers['if-match']).tid;
         } else if (req.body && req.body.html) {
             // Fall back to an inline meta tag in the HTML
-            var tidMatch = /<meta property="mw:TimeUuid" content="([^"]+)"\/?>/
-                                .exec(req.body.html);
-            tid = tidMatch && tidMatch[1];
+            var tidMatch = new RegExp('<meta\\s(?:content="([^"]+)"\\s)?' +
```
d00rman: Let's allow multiple spaces around the fields, so `\\s` -> `\\s+`

mods/parsoid.js
```diff
@@ -509,9 +509,10 @@ PSP.transformRevision = function(restbase, req, from, to) {
...
+                    'property="mw:TimeUuid"(?:\\scontent="([^"]+)")?\\s*\\/?>')
```
d00rman: Idem here for the first `\\s`

PR #359, opened 2015-09-29
Enable testing with node 4.1, and drop heapdump from devDependencies

gwicke:
- Start testing with node 4.1
- Drop the heapdump dependency. While heapdump@0.3.7 works fine locally (on Debian unstable), it fails to compile in travis. I didn't investigate deeply, but noticed that we install heapdump explicitly in our Dockerfile anyway, so don't necessarily need it as a devDependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
     - "0.10"
     - "0.12"
+    - "4.1"
     - "iojs-v2.5.0"
 
 sudo: false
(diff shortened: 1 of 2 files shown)


Pchelolo:
LGTM. Now it's totally possible that heap dump wouldn't be installed and SIGUSR2 will crash a worker, so I've created a PR to protect against that https://github.com/wikimedia/service-runner/pull/53

PR #361, opened 2015-10-02
update service-runner to 0.2.8

eevans:
Needed to make https://github.com/wikimedia/service-runner/pull/54
available to aqs.


d00rman:
Merging even though this is not strictly needed. Just forcing a deploy repo rebuild would work just as well once the module dep's new version is published.

eevans:
Yeah, I came to that conclusion, but only after letting fly with a pull
request. :)

PR #362, opened 2015-10-03
Log remaining bodyOnly uses / users

gwicke:
We are planning to switch off bodyOnly by the end of November (see
https://lists.wikimedia.org/pipermail/mediawiki-api/2015-October/003656.html),
and it would be nice to make sure that there are no remaining users of the old
form before removing it.

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -746,6 +746,10 @@ PSP.makeTransform = function(from, to) {
             // Handle body_only flag.
             // bodyOnly is deprecated and will be removed at some point
             if (to === 'html' && (req.body.body_only || req.body.bodyOnly)) {
+                // Log remaining bodyOnly uses / users
+                if (req.body.bodyOnly) {
+                    self.log('warn/parsoid/bodyonly', req.headers);
+                }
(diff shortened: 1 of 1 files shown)


mods/parsoid.js
```diff
@@ -746,6 +746,10 @@ PSP.makeTransform = function(from, to) {
             // Handle body_only flag.
             // bodyOnly is deprecated and will be removed at some point
             if (to === 'html' && (req.body.body_only || req.body.bodyOnly)) {
+                // Log remaining bodyOnly uses / users
+                if (req.body.bodyOnly) {
```
d00rman: Could you also please add a TODO / FIXME / XXX / TEMP comment here so that we remember to remove it ultimately?

d00rman:
LGTM, minor comment in-lined.

eevans:
I  concur w/ @d00rman re: a comment of some sort ("can be safely removed after..."), otherwise, LGTM

Pchelolo:
@d00rman is deprecating `scrubWikitext` too (https://github.com/wikimedia/restbase/pull/364) so we could log it's usages the same way.

PR #363, opened 2015-10-04
Parsoid mod: add ability to stash wt2html transforms

d00rman:
This PR extends the *wt2html* transformations API to allow clients to indicate to RESTBase they plan to reuse the transformation at a later time, in which case RESTBase saves them in special buckets.

Summary:
- add the `stash` body flag to `POST /transform/wikitext/to/html`
- create the `parsoid.stash.data-parsoid`, `parsoid.stash.wikitext` and `parsoid.stash.html` buckets on start-up
- save the client-sent *wikitext* together with the *data-parsoid* and *html* Parsoid responses for the transform when the client sets `stash: true` when `POST`ing and set the ETag to `<rev>/<tid>/stash`
- detect when a client wants to use the stashed transform in a request to `/transform/html/to/wikitext` and reuse the stashed *wikitext*, *data-parsoid* and *html* originals.

Note: we should merge wikimedia/restbase-mod-table-spec#21 before merging this one.

Bug: [T114548](https://phabricator.wikimedia.org/T114548)

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -600,8 +600,47 @@ PSP.transformRevision = function(restbase, req, from, to) {
 
 };
 
-PSP.callParsoidTransform = function callParsoidTransform(restbase, req, from, to) {
+PSP.stashTranform = function(restbase, req, transformPromise) {
+    // A stash has been requested. We need to store the wikitext sent by
+    // the client together with the page bundle returned by Parsoid, so it
+    // can be later reused when transforming back from HTML to wikitext
+    // cf https://phabricator.wikimedia.org/T114548
     var self = this;
+    var tid = uuid.now().toString();
+    var wtType = req.original && req.original.headers['content-type'] || 'text/plain';
+    return transformPromise.then(function(original) {
+        // Save the returned data-parsoid for the transform and
+        // the wikitext sent by the client
+        return P.all([
+            restbase.put({
+                uri: self.getBucketURI(rp, 'stash.data-parsoid', tid),
+                headers: original.body['data-parsoid'].headers,
+                body: original.body['data-parsoid'].body
+            }),
+            restbase.put({
+                uri: self.getBucketURI(rp, 'stash.wikitext', tid),
+                headers: { 'content-type': wtType },
+                body: req.body.wikitext
+            })
+        ])
+        // Save HTML last, so that any error in metadata storage suppresses
+        // HTML.
+        .then(function() {
+            return restbase.put({
+                uri: self.getBucketURI(rp, 'stash.html', tid),
+                headers: original.body.html.headers,
+                body: original.body.html.body
+            });
+        // Add the ETag to the original response so it can be propagated
+        // back to the client
+        }).then(function() {
+            original.body.html.headers.etag = rbUtil.makeETag(rp.revision, tid, 'stash');
+            return original;
+        });
+    });
+};
+
+PSP.callParsoidTransform = function callParsoidTransform(restbase, req, from, to) {
     var rp = req.params;
     // Parsoid currently spells 'wikitext' as 'wt'
     var parsoidTo = to;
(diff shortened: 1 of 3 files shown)


mods/parsoid.js
```diff
@@ -617,15 +621,54 @@ PSP.callParsoidTransform = function callParsoidTransform(restbase, req, from, to
...
-        uri: this.parsoidHost + '/v2/' + domain + '/'
+        uri: this.parsoidHost + '/v2/' + rp.domain + '/'
             + parsoidTo + parsoidExtraPath,
         headers: { 'content-type': 'application/json' },
         body: req.body
     };
-    return restbase.post(parsoidReq);
+
+    var transformPromise = restbase.post(parsoidReq);
+    if (from === 'wikitext' && to === 'html' && req.body.stash) {
```
gwicke: It might be worth moving this to a separate function, to avoid the stash handling obscuring the common flow too much.

specs/mediawiki/v1/content.yaml
```diff
@@ -975,6 +975,11 @@ paths:
           description: Return only `body.innerHTML`
           type: boolean
           required: false
+        - name: stash
+          in: formData
+          description: Whether to stash temporarily the result of the transformation
```
gwicke: "Whether to temporarily stash the result of the transformation."

gwicke:
Nice! LGTM, only small tiny things inline.

mods/parsoid.js
d00rman: I was thinking the same thing. Will do.

mods/parsoid.js
```diff
@@ -809,6 +852,37 @@ module.exports = function(options) {
...
+                        grace_ttl: 86400
+                    },
+                    valueType: 'blob',
+                    version: 1
+                }
+            },
+            {
+                uri: '/{domain}/sys/key_rev_value/parsoid.stash.wikitext',
+                body: {
+                    valueType: 'blob'
```
Pchelolo: Why don't we have a `revisionRetentionPolicy` for stashed wikitext too?

mods/parsoid.js
```diff
@@ -809,6 +852,37 @@ module.exports = function(options) {
...
+                }
+            },
+            {
+                uri: '/{domain}/sys/key_rev_value/parsoid.stash.data-parsoid',
+                body: {
+                    revisionRetentionPolicy: {
+                        type: 'latest',
+                        count: 1,
```
Pchelolo: As I understand, we don't need to store even 1 revision permanently, but set a `grace_ttl` immediately. May be we could try to use `count: 0`? I'm not sure that will work from the code perspective, we don't have any tests for this case.

d00rman: > May be we could try to use count: 0? I'm not sure that will work in backends, as we don't have any tests for this case..

Precisely because we don't know if it'd work ;)

mods/parsoid.js
d00rman: Ah right

mods/parsoid.js
Pchelolo: Let me check it out, I'll write a test.

Pchelolo: @d00rman works in `sqlite`, but doesn't work in `cassandra`. Looks like a bug, I'll have a look.

Pchelolo: @d00rman I've written a test, and then I didn't want through it away, one after another, so now it will work both in cassandra and SQLite (https://github.com/wikimedia/restbase-mod-table-cassandra/pull/156 & https://github.com/wikimedia/restbase-mod-table-spec/pull/20 )

mods/parsoid.js
```diff
@@ -592,6 +596,46 @@ PSP.transformRevision = function(restbase, req, from, to) {
 
 };
 
+PSP.stashTranform = function(restbase, req, transformPromise) {
+    // A stash has been requested. We need to store the wikitext sent by
+    // the client together with the page bundle returned by Parsoid, so it
+    // can be later reused when transforming back from HTML to wikitext
+    // cf https://phabricator.wikimedia.org/T114548
+    var self = this;
+    var tid = uuid.now().toString();
```
Pchelolo: `+ var rp = req.params;`

mods/parsoid.js
```diff
@@ -822,6 +901,43 @@ module.exports = function(options) {
...
+                    valueType: 'blob',
+                    version: 1
+                }
+            },
+            {
+                uri: '/{domain}/sys/key_rev_value/parsoid.stash.data-parsoid',
+                body: {
+                    revisionRetentionPolicy: {
+                        type: 'latest',
+                        count: 0,
```
Pchelolo: May be let's wait for this https://github.com/wikimedia/restbase-mod-table-spec/pull/21 before it's merged?

d00rman: Yes, but until then *latest 0* is the best choice ;)

d00rman:
Ready for review, @wikimedia/services 

Note 1: tests failing due to the page save API rate limit
Note 2: before merging this one we should probably resolve wikimedia/restbase-mod-table-spec#21

Pchelolo:
Could be really nice to add a couple of tests.

d00rman:
@wikimedia/services the PR is now using the `ttl` retention policy

mods/parsoid.js
```diff
@@ -498,16 +498,42 @@ PSP._getOriginalContent = function(restbase, req, revision, tid) {
 
 };
 
+PSP._getStashedContent = function(restbase, req, etag) {
```
gwicke: Eventually we should be able to merge this with _getOriginalContent, but for now this is the only one loading wikitext.

mods/parsoid.js
```diff
@@ -596,6 +631,47 @@ PSP.transformRevision = function(restbase, req, from, to) {
 
 };
 
+PSP.stashTranform = function(restbase, req, transformPromise) {
```
gwicke: s/Tranform/Transform/

gwicke:
LGTM overall, only one minor typo nit.

d00rman:
@Pchelolo I'll add some tests in a follow-up PR

PR #364, opened 2015-10-04
Transforms API: introduce scrub_wikitext and deprecate scrubWikitext

d00rman:
Bug: [T114620](https://phabricator.wikimedia.org/T114620)

### DIFF
--- a/specs/mediawiki/v1/content.yaml
+++ b/specs/mediawiki/v1/content.yaml
@@ -905,6 +905,16 @@ paths:
           description: The HTML to transform
           type: string
           required: true
+        - name: scrub_wikitext
+          in: formData
+          description: Normalise the DOM to yield cleaner wikitext?
+          type: boolean
+          required: false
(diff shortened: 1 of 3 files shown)


d00rman:
cc @wikimedia/services

specs/mediawiki/v1/content.yaml
```diff
@@ -905,6 +905,16 @@ paths:
           description: The HTML to transform
           type: string
           required: true
+        - name: scrub_wikitext
+          in: formData
+          description: Normalise the DOM to yield cleaner wikitext?
```
Pchelolo: Why is there a question mark in the end?

d00rman: Because the flag is a boolean, and its value answers that question :)

Pchelolo: Ah, now I understand.

Pchelolo:
LGTM :+1:

PR #365, opened 2015-10-05
Add a first-class representation of pages

Pchelolo:
This is ready for review.

This PR adds a `page_data` table, that stores `rename_to`, `rename_from`, `delete` and `undelete` events for a page title. Events are sorted by `tid`, so looking at the latest event in the page table, when pulling a revision, we can see what happened with a page. If event is `deleted` - page is deleted, return `404`, if it's `rename_to` - page was renamed and latest revisions under this title shouldn't be shown any more (by historic revisions should be). Storing the full history of page-related events gives us an opportunity to make nice features in future, like provide listings of page revisions that follow renames.

To detect deletes, we use the same logic as before. To detect renames, I'm adding a new header to the update hook `x-restbase-parenttitle`. It's sent on `move` events, and renames are stored. Undelete is detected, when we receive a revision from MW API (using `no-cache` header), but according to the page table, the revision is deleted. In than case we follow the page table back in time, find the previous delete event and restore the `good_after` timestamp. We can't simply drop it, because the page can be deleted, then another one created instead of it, deleted and then restored.

When revision is fetched, page data for it's title is fetched too (in parallel), and checked for renames and deletes. When the page table is empty, the overhead is really low, not more than 5%. Obviously, when there's a long renaming history, finding a real latest page title takes many turnarounds to the page table, so performance drops, but that's a corner case and wouldn't be observed frequently. On `html` endpoints, we check the revision with `getTitleRevision`, so content is hidden for them accordingly too.

Also, we've had a bug, that on renames a revision got duplicated in the old and new titles. Now it's fixed.

This was tested `a lot` with vagrant installation, I've created really long and complicated sequences of renames/deletes/undeletes/creating_pages_instead_of_deleted_ones and everything seem to work correctly.

Right now, this PR exposes a security vulnerability, which should be handled before it's release, and I wanted to chat about the options first. The `x-restbase-parentrevision` and `x-restbase-parenttitle` headers would be meaningful, and making them point to random places, an attacker could completely mess up the database. What I propose to do, is to restrict them to internal clients only (the hook). To do that, I'm going to remove these headers from the spec, but for internal clients we will include all provided headers. So the spec would white-list the safe headers, but internal clients would be able to by-pass the white-list. Thoughts?

I've tested this patch in vagrant without required updates to the hook, and there's no major issues in first merging and deploying this, and then replying the hook. Of course, it would be better to deploy hook changes sooner than later, but in that period nothing catastrophic would happen and no invalid data would be generated.

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -105,7 +105,6 @@ templates:
                 storage_groups:
                   - name: test.group.local
                     domains: /./
-                #unsupported_features: fallback
 
       /{module:page_revisions}: &wp/sys/page_revisions
         x-modules:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -105,6 +105,7 @@ templates:
                 storage_groups:
                   - name: test.group.local
                     domains: /./
+                #unsupported_features: fallback
 
       /{module:page_revisions}: &wp/sys/page_revisions
         x-modules:
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -309,6 +309,22 @@ rbUtil.normalizeContentType = function(res) {
             contentType.format(contentType.parse(res.headers['content-type']));
     }
 };
+// Create a uniform but shallow request object copy with sane defaults. This
+// keeps code dealing with this request monomorphic (good for perf), and
+// avoids subtle bugs when requests shared between recursive requests are
+// mutated in another control branch. At the very minimum, we are mutating the
+// .params property for each sub-request.
+rbUtil.cloneRequest = function(req) {
+    return {
+        uri: req.uri || req.url || null,
+        method: req.method || 'get',
+        headers: req.headers || {},
+        query: req.query || {},
+        body: req.body || null,
+        params: req.params || {}
+    };
+};
+
 
 /***
  * MediaWiki-specific functions
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -246,14 +246,6 @@ util.inherits(HTTPError, Error);
 
 rbUtil.HTTPError = HTTPError;
 
-function HTTPRedirect(response) {
-    Error.call(this);
-    this.name = this.constructor.name;
-    Object.assign(this, response);
-}
-util.inherits(HTTPRedirect, Error);
-rbUtil.HTTPRedirect = HTTPRedirect;
-
 // Create an etag value of the form
 // "<revision>/<tid>/<optional_suffix>"
 rbUtil.makeETag = function(rev, tid, suffix) {
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -309,22 +309,6 @@ rbUtil.normalizeContentType = function(res) {
             contentType.format(contentType.parse(res.headers['content-type']));
     }
 };
-// Create a uniform but shallow request object copy with sane defaults. This
-// keeps code dealing with this request monomorphic (good for perf), and
-// avoids subtle bugs when requests shared between recursive requests are
-// mutated in another control branch. At the very minimum, we are mutating the
-// .params property for each sub-request.
-rbUtil.cloneRequest = function(req) {
-    return {
-        uri: req.uri || req.url || null,
-        method: req.method || 'get',
-        headers: req.headers || {},
-        query: req.query || {},
-        body: req.body || null,
-        params: req.params || {}
-    };
-};
-
 
 /***
  * MediaWiki-specific functions
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -240,15 +240,19 @@ function HTTPError(response) {
     if (response.body && response.body.type) {
         this.message += ': ' + response.body.type;
     }
-
-    for (var key in response) {
-        this[key] = response[key];
-    }
+    Object.assign(this, response);
 }
 util.inherits(HTTPError, Error);
 
 rbUtil.HTTPError = HTTPError;
 
+function HTTPRedirect(response) {
+    Error.call(this);
+    this.name = this.constructor.name;
+    Object.assign(this, response);
+}
+util.inherits(HTTPRedirect, Error);
+rbUtil.HTTPRedirect = HTTPRedirect;
 
 // Create an etag value of the form
 // "<revision>/<tid>/<optional_suffix>"
--- a/lib/server.js
+++ b/lib/server.js
@@ -261,7 +261,7 @@ function handleRequest(opts, req, resp) {
         return handleResponse(reqOpts, newReq, resp, result);
     })
     .catch(function(e) {
-        if (!e || (e.name !== 'HTTPError' && e.name !== 'HTTPRedirect')) {
+        if (!e || e.name !== 'HTTPError') {
             var originalError = e;
             var stack = e && e.stack;
             e = new rbUtil.HTTPError({
--- a/lib/server.js
+++ b/lib/server.js
@@ -204,6 +204,11 @@ function handleRequest(opts, req, resp) {
         reqOpts.metrics.increment('requests.private');
     } else {
         reqOpts.metrics.increment('requests.public');
+        Object.keys(req.headers).forEach(function(headerName) {
+            if (/(?:^x-restbase)|(?:^x-rb)/.test(headerName)) {
+                delete newReq.headers[headerName];
+            }
+        });
     }
 
     // Create a new, clean request object
@@ -215,15 +220,6 @@ function handleRequest(opts, req, resp) {
         method: req.method.toLowerCase(),
         headers: req.headers
     };
-
-    // Remove x-restbase and x-rb headers if the request is not local
-    if (!/^(?:::ffff:)?(?:10|127)\./.test(remoteAddr) && newReq.headers) {
-        Object.keys(newReq.headers).forEach(function(headerName) {
-            if (/(?:^x-restbase)|(?:^x-rb)/.test(headerName)) {
-                delete newReq.headers[headerName];
-            }
-        });
-    }
 
     // Start off by parsing any POST data with BusBoy
     return rbUtil.parsePOST(req)
--- a/lib/server.js
+++ b/lib/server.js
@@ -216,6 +216,15 @@ function handleRequest(opts, req, resp) {
         headers: req.headers
     };
 
+    // Remove x-restbase and x-rb headers if the request is not local
+    if (!/^(?:::ffff:)?(?:10|127)\./.test(remoteAddr) && newReq.headers) {
+        Object.keys(newReq.headers).forEach(function(headerName) {
+            if (/(?:^x-restbase)|(?:^x-rb)/.test(headerName)) {
+                delete newReq.headers[headerName];
+            }
+        });
+    }
+
     // Start off by parsing any POST data with BusBoy
     return rbUtil.parsePOST(req)
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -265,7 +265,7 @@ function handleRequest(opts, req, resp) {
         return handleResponse(reqOpts, newReq, resp, result);
     })
     .catch(function(e) {
-        if (!e || e.name !== 'HTTPError') {
+        if (!e || (e.name !== 'HTTPError' && e.name !== 'HTTPRedirect')) {
             var originalError = e;
             var stack = e && e.stack;
             e = new rbUtil.HTTPError({
--- a/mods/page_revisions.js
+++ b/mods/page_revisions.js
@@ -63,12 +63,16 @@ PRS.prototype.getTableSchema = function() {
             user_text: 'string',
             timestamp: 'timestamp',
             comment: 'string',
-            redirect: 'boolean'
+            redirect: 'boolean',
+            // Static link to old title. Statics are per-partition,
+            // so effectively different for each title
+            renamed_from: 'string'
         },
         index: [
             { attribute: 'title', type: 'hash' },
             { attribute: 'rev', type: 'range', order: 'desc' },
             { attribute: 'latest_rev', type: 'static' },
+            { attribute: 'renamed_from', type: 'static' },
             { attribute: 'tid', type: 'range', order: 'desc' }
         ],
         secondaryIndexes: {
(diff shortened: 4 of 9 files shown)


lib/rbUtil.js
```diff
@@ -297,6 +297,22 @@ rbUtil.copyForwardedHeaders = function(restbase, req, headers) {
     return req;
 };
 
+// Create a uniform but shallow request object copy with sane defaults. This
+// keeps code dealing with this request monomorphic (good for perf), and
+// avoids subtle bugs when requests shared between recursive requests are
+// mutated in another control branch. At the very minimum, we are mutating the
+// .params property for each sub-request.
+rbUtil.cloneRequest = function(req) {
```
d00rman: This must be a c/p fail :) The function is already declared on L218.

mods/page_revisions.js
```diff
@@ -82,6 +83,29 @@ PRS.prototype.getTableSchema = function() {
     };
 };
 
+PRS.prototype.pageTableName = 'page_data';
+PRS.prototype.pageTableURI = function(domain) {
+    return new URI([domain, 'sys', 'table', this.tableName, '']);
```
d00rman: `s/tableName/pageTableName/`

mods/page_revisions.js
```diff
@@ -293,56 +425,72 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
...
+                    return self.fetchAndStoreMWRevision(restbase, req);
+                }),
+                pageData: pageDataRequest
+            })
+            .then(function(res) {
+                if (res.pageData
+                        && res.pageData.body.items
+                        && res.pageData.body.items.length) {
+                    var latestEvent = res.pageData.body.items[0];
+                    if (latestEvent.event_type === 'rename_to') {
```
d00rman: I'm wondering if this is the correct behaviour. For correctness' sake, if we say *page renamed to X* then we should ensure *X* is indeed the latest page name.

I'm also worried about the correctness of throwing a 404 at all here. I think the correct way to go about this is either *(a)* respond with a 301 and point to the latest title URI; or *(b)* supply the data about the revision and add an additional header indicating the client should use the new title. Either solution would allow us to preserve stable URIs and allow viewing historical content regardless of the point in time the URI was obtained.

gwicke:
> Right now, this PR exposes a security vulnerability, which should be handled before it's release, and I wanted to chat about the options first. The x-restbase-parentrevision and x-restbase-parenttitle headers would be meaningful, and making them point to random places, an attacker could completely mess up the database. What I propose to do, is to restrict them to internal clients only (the hook). To do that, I'm going to remove these headers from the spec, but for internal clients we will include all provided headers. So the spec would white-list the safe headers, but internal clients would be able to by-pass the white-list. Thoughts?

We discussed this yesterday, but let me summarize that discussion here. One option we considered is generally restricting x-rb-* and/or x-restbase-* headers in server.js, before any processing on the requests. As a first step, we could strip all headers matching that pattern if the request comes from an external client.

While not as fine-grained as per-endpoint checks, this should provide nice blanket protection at minimal cost.

mods/page_revisions.js
Pchelolo: @d00rman Regarding redirect vs 404, there's been a discussion [here](https://phabricator.wikimedia.org/T87393). The main concern was about clients which follow redirects automatically, and wouldn't notice it. That discussion is still unfinished, but I agree that a redirect would be nicer here.

Regarding pointing to the latest title, it's a hard task (both programmatically and computationally), but I think that indeed it would be more correct. We are asking about the latest data, so we need to provide the latest name..

Pchelolo:
I've just realised I have one more issue here - in case a new page is created instead of a renamed one, we wouldn't detect that and will still restrict access to it, needs an update.

Pchelolo:
Added tracking for a really latest title (e.g. rename A->B->C and on request to A it will report a rename to C), the last question left is how to handle renamed paged? Are we going to send a redirect, or 404, or something else. @wikimedia/services who votes for which solution?

d00rman:
IMHO, 404 is plain wrong and breaks historic views. I'm voting for redirects. Also, given that most pages are renamed only a few times during their lifetime, it'd be perfectly acceptable to redirect the client to the *newer* title, since in around 5 hops max they can reach the current page title name.

Pchelolo:
@d00rman I've already implemented `redirect to the most latest title` on the server side, which's way more efficient than `redirect to newer title` and client running back-and-forward, so I propose to leave that code.

But in general I agree, let's go with redirects. We might need to add `Location` header to https://github.com/wikimedia/restbase/pull/371

Pchelolo:
@d00rman Updated the patch, now we return redirects for renamed pages.

d00rman:
LGTM in general, but let's keep it open for a bit more, it's quite a big PR to swallow at once :)

Pchelolo:
@d00rman Sure, it's a complete mess..

Could you please also have a look on https://github.com/wikimedia/restbase/pull/349, these two are correlated, I wanted to rebase this on on 349 and check that current implementation works fine.

berndsi:
Just wanted to mention that I'm interested in trying this out. I'm esp. curious if this also handles manually added redirects, e.g. https://en.wikipedia.org/api/rest_v1/page/html/Obama? This is a blocker for our mobile content service moving to using Parsoid content instead of MW API. Thanks!

Pchelolo:
@berndsi Could you please describe how do you make a manual redirect, I'll test it out.

berndsi:
@Pchelolo See https://en.wikipedia.org/w/index.php?title=Obama&action=edit for an example. 
#REDIRECT [[Barack Obama]]

https://www.mediawiki.org/wiki/Parsoid/MediaWiki_DOM_spec#Redirects has some good info about it.

Pchelolo:
@berndsi Thank you, I'll try it out

berndsi:
There are more complicated ones, like https://en.wikipedia.org/w/index.php?title=Anion&redirect=no redirects to a section of the target article.

Even more info: https://en.wikipedia.org/wiki/Wikipedia:Redirect

Pchelolo:
@berndsi unfortunately, this will not support manually-created redirects (yet), only renames made with a move action. We depend on MW hooks to detect page renames, and a Move hook is not fired when a manual redirect is happening. This could be a useful feature to detect those cases too, but we'll tackle it later, this PR is already too big.

berndsi:
Bummer. Thanks for letting me know. Any estimate on when manual redirects
will be handled? Is there a Phab task?

Pchelolo:
@berndsi not yet, we need to finish up and stabilise at least this.

mods/page_revisions.js
```diff
@@ -255,8 +360,9 @@ PRS.prototype.fetchAndStoreMWRevision = function(restbase, req) {
             redirect: redirect
         };
 
+        var actions = [
         // Check if the same revision is already in storage
-        return restbase.get({
+        restbase.get({
```
d00rman: Indentation: this whole block is part of the `actions` array, so it should be indented properly.

d00rman:
One minor nit, but apart from that I think we're GTG here. @Pchelolo please rebase yet once again :/

Pchelolo:
To make this work, we need to make some changes in the hook: https://gerrit.wikimedia.org/r/#/c/249711/

lib/server.js
```diff
@@ -216,6 +216,15 @@ function handleRequest(opts, req, resp) {
         headers: req.headers
     };
 
+    // Remove x-restbase and x-rb headers if the request is not local
+    if (!/^(?:::ffff:)?(?:10|127)\./.test(remoteAddr) && newReq.headers) {
```
gwicke: Minor: It might be nice test for local / remote once. Perhaps move this to line 204?

mods/page_revisions.js
```diff
@@ -424,6 +558,105 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
...
+                    var rootPath = restbase._rootReq.uri.path;
+                    var newPath = [];
+                    rootPath.forEach(function(pathElement) {
+                        if (pathElement === rp.title) {
+                            newPath.push(encodeURIComponent(latestTitle));
+                        } else {
+                            newPath.push(pathElement);
+                        }
+                    });
+                    throw new rbUtil.HTTPRedirect({
```
gwicke: Will this work when the API is accessed at /api/rest_v1/ ?

lib/rbUtil.js
```diff
@@ -239,15 +239,19 @@ function HTTPError(response) {
...
-    for (var key in response) {
-        this[key] = response[key];
-    }
+    Object.assign(this, response);
 }
 util.inherits(HTTPError, Error);
 
 rbUtil.HTTPError = HTTPError;
 
+function HTTPRedirect(response) {
```
gwicke: I'm a bit torn on treating HTTPRedirect as an error. It might be convenient to short-cut the processing this way, but also assumes that code several layers deep knows about how a response is going to be used.

mods/page_revisions.js
```diff
@@ -82,6 +83,29 @@ PRS.prototype.getTableSchema = function() {
     };
 };
 
+PRS.prototype.pageTableName = 'page_data';
```
gwicke: Why not just `pages` ?

mods/page_revisions.js
```diff
@@ -351,66 +483,68 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
...
-                return self.fetchAndStoreMWRevision(restbase, req);
+            revisionRequest = P.props({
+                revisionInfo: getLatestTitleRevision()
+                .catch(function(e) {
+                    if (e.status !== 404) {
+                        throw e;
+                    }
+                    return self.fetchAndStoreMWRevision(restbase, req);
+                }),
+                pageData: pageDataRequest
```
gwicke: Could we potentially skip the revision check when the request is for the latest revision's HTML?

Pchelolo: @gwicke sorry, I didn't understand your idea

lib/rbUtil.js
gwicke: See also: https://phabricator.wikimedia.org/T118195

mods/page_revisions.js
Pchelolo: Actually, we did call `checkRevReturn` for the latest revision before the delete detection was added too: https://github.com/wikimedia/restbase/pull/269/files

Also, we still have to check for `page_deleted` restriction, because for older content the page table doesn't contain that information.

gwicke: We skipped the check for latest-revision requests in https://github.com/wikimedia/restbase/commit/da53d971, but started checking for it to catch page deletions in https://github.com/wikimedia/restbase/pull/269/files.

> Also, we still have to check for page_deleted restriction, because for older content the page table doesn't contain that information.

The point is about the *latest* revision, though. I agree that we'll need to check both tables for older revisions for now.

mods/page_revisions.js
```diff
@@ -335,14 +448,33 @@ PRS.prototype.fetchAndStoreMWRevision = function(restbase, req) {
...
     var rp = req.params;
+    var title = rbUtil.normalizeTitle(rp.title);
     var revisionRequest;
+    var pageDataRequest = restbase.get({
+        uri: self.pageTableURI(rp.domain),
+        body: {
+            table: self.pageTableName,
+            attributes: {
+                title: title
+            }
```
gwicke: This would select *all* entries ever. Can we limit this to 1?

mods/page_revisions.js
gwicke: I don't think it will, honestly. Maybe we could use a relative redirect instead?

mods/page_revisions.js
```diff
@@ -351,66 +483,68 @@ PRS.prototype.getTitleRevision = function(restbase, req) {
...
+                }
+                return self.fetchAndStoreMWRevision(restbase, req);
+            }),
+            pageData: pageDataRequest
         });
     } else if (!rp.revision) {
         if (req.headers && /no-cache/.test(req.headers['cache-control'])) {
-            revisionRequest = self.fetchAndStoreMWRevision(restbase, req)
+            revisionRequest = P.props({
+                revisionInfo: self.fetchAndStoreMWRevision(restbase, req),
```
gwicke: I think we can skip the revision request here.

mods/page_revisions.js
Pchelolo: @gwicke Ah, you mean in parsoid module. We still can't skip it. `wrapContentRequest` is checking for page deletions too, not only for revision restrictions.

mods/page_revisions.js
Pchelolo: For page undeletes we need to iterate back in history and find the previous delete, so for that part we need the full history. However indeed we could make it fetch full history only for `no-cache` requests when page undulation is checked.

lib/rbUtil.js
Pchelolo: > but also assumes that code several layers deep knows about how a response is going to be used, which it probably shouldn't need to know & care about.

I'm not sure what do you mean. The only change needed to support error-like redirects is adding 1 line in server to treat the redirect same way as HTTPError.

If we make a redirect normal response, we'll need to add A LOT of checks, as throwing a redirect instead of returning it, we can shortcut really deep promise chains. I've tried to add all those checks today,and the code doesn't seem to become more cleaner.

If we make a decision to switch to `Response`-like style in RB, we'd need to change the behaviour of `HTTPError` too, so `HTTPRedirect` would be a minor issue.

mods/page_revisions.js
gwicke: Right. The point is that we only need to check for page deletion when accessing the latest revision.

lib/rbUtil.js
gwicke: My concern is primarily about avoiding surprising, and making sure that we keep a sane layering of requests. Some of these requests might be just internal, and callers might not necessarily expect a redirect error to be thrown at them. It seems that this could lead to surprises when this happens while fetching the original HTML for a page, for example. Should the caller follow the redirect? Would the redirected URL make sense? How are redirects translated so that they make sense at the outer layer, taking into account possible rewrites?

I think we should think through these implications a bit more. I'm fine with doing a one-off change here if we can establish that surprises as described above won't happen.

gwicke:
We should check that redirect pages are still accessible, despite the new redirect functionality. VE currently expects that to make redirect pages editable, so we should be careful not to change this behavior without coordinating with them.

config.test.yaml
```diff
@@ -84,6 +84,7 @@ templates:
                 storage_groups:
                   - name: test.group.local
                     domains: /./
+                #unsupported_features: fallback
```
gwicke: Is this needed for anything?

gwicke:
See https://phabricator.wikimedia.org/T118548 for a discussion of the redirect question.

Pchelolo:
- Rebased on master
- Removed the redirection logic - now we just store the renames, but don't use that info until we decide how to work with it.
- Some general cleanup
- Tested with vagrant - looks OK

mods/page_revisions.js
Pchelolo: @gwicke Not really. Imagine the situation: you have a page with rev1, then you delete it, create a new page with the same name with rev2. The latest revision of a page is rev2, but the restricted (deleted) one is rev1 - not the latest.

mods/page_revisions.js
gwicke: Since undeletions should be rather rare compared to regular getTitleRevision requests, I think it's worth optimizing this. The difference should be significant for pages with a couple of entries.

Pchelolo:
This was implemented in another PR with restrictions table.

PR #366, opened 2015-10-06
Include more client info in log entries

gwicke:
Include all client headers (plus some RB-internal ones) in log entries.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -181,7 +181,16 @@ function handleRequest(opts, req, resp) {
             req: {
                 method: req.method.toLowerCase(),
                 uri: req.url,
-                headers: req.headers,
+                headers: {
+                    'cache-control': req.headers['cache-control'],
+                    'content-length': req.headers['content-length'],
+                    'content-type': req.headers['content-type'],
+                    'if-match': req.headers['if-match'],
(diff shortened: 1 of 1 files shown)


tstarling:
Looks good to me

lib/server.js
```diff
@@ -170,12 +170,18 @@ function handleRequest(opts, req, resp) {
...
+    req.headers['x-rb-client-ip'] = remoteAddr;
+
     var reqOpts = {
         conf: opts.conf,
         logger: opts.logger.child({
             req: {
                 method: req.method.toLowerCase(),
-                uri: req.url
+                uri: req.url,
+                headers: req.headers,
```
Pchelolo: Is it safe to log all headers? They might contain very sensitive information, like cookies, so if someone steal the logs they potentially get access to the accounts.

d00rman: Good point. It's not about stealing but about WM's policy of not keeping sensitive user info around, cf [T112072](https://phabricator.wikimedia.org/T112072)

gwicke:
@Pchelolo, I addressed your security concern by whitelisting the logged headers in the follow-up patch.

Pchelolo:
@gwicke LGTM :+1:

PR #368, opened 2015-10-06
Use lowercased User-Agent header

d00rman:
Internally, we are using only lowercased header names, so enforce this for the user agent string as well.

This is a follow-up on #367


d00rman:
cc @wikimedia/services

PR #369, opened 2015-10-08
Minor tweaks to the README

subbuss:
Also, the sentence '.. API powered by swagger specs ..' reads rather odd. Swagger is not powering the API. It is just generating an API front-end and docs. I wanted to tweak it, but I wasn't sure .. so I am leaving it as an exercise to the maintainers. :)


d00rman:
Effectively, it's really Swagger-driven in the sense that one defines the routes and the handlers for them in the Swagger spec only, no *server-side code* is needed to complete an action.

PR #370, opened 2015-10-09
Bug fix: Parsoid stashing: Ensure title/revision are present

d00rman:
There is a subtle bug in the Parsoid stashing logic whereby if the title and revision are not provided, a 500 is returned since the bucket URI has an undefined path component. This PR fixes that for both `/transform/wikitext/to/html` (storing the stash) and `/transform/html/to/wikitext` (retrieving the stashed content).


d00rman:
cc @wikimedia/services

Pchelolo:
LGTM, but could you please file a ticket to investigate if we could avoid requiring title&revision?

d00rman:
Asked in [T104663#1714853](https://phabricator.wikimedia.org/T104663#1714853)

PR #371, opened 2015-10-11
CORS: Add the Access-Control-Expose-Headers header

d00rman:
Cross-site requests need to be able to read the `ETag` header. In order to do so in browsers, RESTBase must send the `Access-Control-Expose-Headers` header and set its value to `etag`.

Bug: [T115165](https://phabricator.wikimedia.org/T115165)

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -34,6 +34,7 @@ function handleResponse(opts, req, resp, response) {
         rh['access-control-allow-origin'] = '*';
         rh['access-control-allow-methods'] = 'GET';
         rh['access-control-allow-headers'] = 'accept, content-type';
+        rh['access-control-expose-headers'] = 'etag';
 
         // Set up security headers
         // https://www.owasp.org/index.php/List_of_useful_HTTP_headers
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

lib/server.js
```diff
@@ -34,6 +34,7 @@ function handleResponse(opts, req, resp, response) {
         rh['access-control-allow-origin'] = '*';
         rh['access-control-allow-methods'] = 'GET';
```
Pchelolo: Semi-related thing: we support `HEAD` and `POST` on some endpoints too. May be we could allow them here too?

d00rman: Was thinking the same

d00rman: Actually, it's not needed. The [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests) says:

> Cross-site requests are preflighted like this since they may have implications to user data.  In particular, a request is preflighted if:
> - It uses methods other than GET, HEAD or POST.  Also, if POST is used to send request data with a Content-Type other than application/x-www-form-urlencoded, multipart/form-data, or text/plain, e.g. if the POST request sends an XML payload to the server using application/xml or text/xml, then the request is preflighted.
> - It sets custom headers in the request (e.g. the request uses a header such as X-PINGOTHER)

Pchelolo: Indeed.

Pchelolo:
LGTM :+1:

PR #372, opened 2015-10-11
MobileApps: Set up RESTBase storage

d00rman:
Thus far, requests for the Mobile Content Service were only proxied through RESTBase. This commit sets up RESTBase to store the response from the back-end service and serve it on subsequent requests.

First, the current revision ID for a page is retrieved from the storage. Then, the storage is checked for the content rendered by the Mobile Content Service for the returned revision ID and time UUID. If data for that combination exists, it is returned. Otherwise, a request is placed to the Mobile Content Service to render the page and the result is stored in RESTBase and returned to the client. Note that, if the `Cache-Control: no-cache` header is supplied, the storage-check step is skipped and the back-end service is contacted for a new render right away, thus enabling re-renders even when the revision info hasn't changed. This principle is applied to all of the routes exposed by RESTBase for the Mobile Content Service.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -152,34 +152,144 @@ templates:
           paths:
             /html/{title}:
               get:
+                x-setup-handler:
+                  - init:
+                      method: put
+                      uri: /{domain}/sys/key_rev_value/mobileapps.mobile-html
+                      body:
+                        revisionRetentionPolicy:
+                          type: 'latest'
+                          count: 1
+                          grace_ttl: 86400
+                        valueType: 'json'
+                        version: 1
                 x-request-handler:
-                  - get_from_backend:
+                  - handle_req:
                       request:
-                        uri: http://appservice.wmflabs.org/{domain}/v1/page/mobile-html/{title}
+                        method: get
+                        uri: /{domain}/sys/mobileapps/handling/mobile-html/{title}
+                        headers:
+                          cache-control: '{cache-control}'
             /sections/{title}:
               get:
+                x-setup-handler:
+                  - init:
+                      method: put
+                      uri: /{domain}/sys/key_rev_value/mobileapps.mobile-html-sections
+                      body:
+                        revisionRetentionPolicy:
+                          type: 'latest'
+                          count: 1
+                          grace_ttl: 86400
+                        valueType: 'json'
+                        version: 1
                 x-request-handler:
-                  - get_from_backend:
+                  - handle_req:
                       request:
-                        uri: http://appservice.wmflabs.org/{domain}/v1/page/mobile-html-sections/{title}
+                        method: get
+                        uri: /{domain}/sys/mobileapps/handling/mobile-html-sections/{title}
+                        headers:
+                          cache-control: '{cache-control}'
             /sections-lead/{title}:
               get:
+                x-setup-handler:
+                  - init:
+                      method: put
+                      uri: /{domain}/sys/key_rev_value/mobileapps.mobile-html-sections-lead
+                      body:
+                        revisionRetentionPolicy:
+                          type: 'latest'
+                          count: 1
+                          grace_ttl: 86400
+                        valueType: 'json'
+                        version: 1
                 x-request-handler:
-                  - get_from_backend:
+                  - handle_req:
                       request:
-                        uri: http://appservice.wmflabs.org/{domain}/v1/page/mobile-html-sections-lead/{title}
+                        method: get
+                        uri: /{domain}/sys/mobileapps/handling/mobile-html-sections-lead/{title}
+                        headers:
+                          cache-control: '{cache-control}'
             /sections-remaining/{title}:
               get:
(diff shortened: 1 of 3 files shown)


d00rman:

config.test.yaml
```diff
@@ -150,34 +150,144 @@ templates:
...
+                      request:
+                        method: get
+                        uri: /{domain}/sys/page_revisions/page/{title}
+                        headers:
+                          cache-control: '{cache-control}'
+                  - check_storage:
+                      request:
+                        method: get
+                        uri: /{domain}/sys/key_rev_value/mobileapps.{$.request.params.route}/{$.request.params.title}/{$.get_rev.body.items[0].rev}/{$.get_rev.body.items[0].tid}
+                      return_if:
```
Pchelolo: My be better to use short notation for `{route}` and `{title}` to avoid a super-long line?

config.test.yaml
```diff
@@ -150,34 +150,144 @@ templates:
...
+                        status: 200
+                        headers: '{$.check_storage.body.headers}'
+                        body: '{$.check_storage.body.body}'
+                      catch:
+                        status: 404
+                  - get_mobileapps:
+                      request:
+                        method: get
+                        uri: http://appservice.wmflabs.org/{domain}/v1/page/{$.request.params.route}/{$.request.params.title}
+                        headers:
```
Pchelolo: Idem

config.test.yaml
```diff
@@ -150,34 +150,144 @@ templates:
...
+                          x-restbase-etag: '{$.get_rev.body.items[0].rev}/{$.get_rev.body.items[0].tid}'
+                  - store:
```
Pchelolo: What's the `x-restbase-etag` header? Why is it called like this and not like a plain `etag`?

config.example.wikimedia.yaml
```diff
@@ -152,34 +152,144 @@ templates:
...
+                        version: 1
                 x-request-handler:
-                  - get_from_backend:
+                  - handle_req:
                       request:
-                        uri: http://appservice.wmflabs.org/{domain}/v1/page/mobile-html/{title}
+                        method: get
+                        uri: /{domain}/sys/mobileapps/handling/mobile-html/{title}
+                        headers:
+                          cache-control: '{cache-control}'
```
Pchelolo: For this to work correctly the following lines https://github.com/wikimedia/restbase/blob/master/mods/key_value.js#L186 should be added to `key_rev_value` too.

Pchelolo:
I'm not sure I understand, how is the storage going to be invalidated on new revisions, or, for example, on a template change, when parsoid html is rerendered. Who is going to send requests to all endpoints with `cache-control: no-cache`, an app?

Also, as more endpoints no use `cache-control: no-cache` on `/title/{title}` endpoints, we need to merge https://github.com/wikimedia/restbase/pull/349 to avoid polluting the storage with lots of identical revision entries. I'll rebase it in a moment.

config.example.wikimedia.yaml
d00rman: No, because we need the `cache-control` only for revision retrieval

Pchelolo: I that case, how is the storage going to be invalidated, if the render change is needed, but revision is the same (template change, or some other dependency change)

d00rman:
> I'm not sure I understand, how is the storage going to be invalidated on new revisions, or, for example, on a template change, when parsoid html is rerendered. Who is going to send requests to all endpoints with cache-control: no-cache, an app?

The idea here is that when retrieving the content for the mobile routes we need to know if we have it for the current revision. Hence the `cache-control` header forwarding.

I envision that the dependency-tracking system will request the new version, but it's unlikely `cache-control: no-cache` will actually be sent seeing that that will happen after the Parsoid routes have been triggered.

d00rman:
Updated the PR so that now storage is not checked at all when the `Cache-Control: no-cache` header is present, hence allowing for re-renders of the same revision to happen.

Pchelolo:
@d00rman nice, the idea with routing based on `cache-control` worked out.. LGTM in general, but, again, who will be responsible for sending `no-cache` requests to invalidate storage, RB?

d00rman:
> who will be responsible for sending no-cache requests to invalidate storage, RB?

The dependecy-tracking mechanism.

Pchelolo:
@d00rman and in the period between merging this and integrating dep tracking?

d00rman:
That shouldn't be a real problem since the service itself strips a lot of stuff away, such as infoboxes, so in general I'm inclined to say that re-renders of the same revision are only tangentially important for MobileApps.

Pchelolo:
@d00rman ok, if we see problems with it, we always can make an ad-hoc solution like the one with varnish purges. So, LGTM.

berndsi:
I don't have any issues with that in the interim. Just wanted to clarify @d00rman's comment: We don't remove all infoboxes. We only remove the navboxes.

PR #375, opened 2015-10-13
Move reqTemplate library to swagger-router

Pchelolo:
reqTemplate is a general purpose lib, so it should e moved to swagger-router.


lib/handlerTemplate.js
```diff
@@ -7,8 +7,8 @@
 
 var P = require('bluebird');
 var URI = require('swagger-router').URI;
+var Template = require('swagger-router').Template;
```
d00rman: Probably better to require `swagger-router` once and then set the short-cut variables.

lib/handlerTemplate.js
```diff
@@ -6,9 +6,10 @@
  */
 
 var P = require('bluebird');
-var URI = require('swagger-router').URI;
+var swaggerRouter = require('swagger-router');
+var URI = swaggerRouter.URI;
```
d00rman: Minor nit: let's place these two variables *after* all of the requires. I feel it increases readability.

PR #376, opened 2015-10-14
Nock page_save tests

Pchelolo:
Wiki maintainers are not happy about the page_save tests running against labs, because it's polluting the revision history. This PR switches all our save_api tests to nock. However, for sanity-testing, it might be needed to run them agains prod from time to time, so I've made a global `NOCK_TESTS` flag - setting it to false will make all tests run agains prod and labs.

Bug: https://phabricator.wikimedia.org/T115369


d00rman:
Neat trick with `NOCK_TESTS` @Pchelolo !

PR #377, opened 2015-10-14
Various fixes

Pchelolo:
Various fixes and improvements:
1. Improved test script to sum sqlite and cassandra exit codes
2. Removed temporary logging in parsoid transforms
3. Replace `x-forwarded-for` header with `x-client-ip`, which is the real trusted IP of the client. XFF header could be faked.

### DIFF
--- a/test/features/pagecontent/save_api.js
+++ b/test/features/pagecontent/save_api.js
@@ -418,6 +418,7 @@ describe('page save api', function() {
                     uri: htmlUri,
                     headers: {
                         'x-forwarded-for': '123.123.123.123',
+                        'x-client-ip': '123.123.123.123',
                         cookie: 'test'
                     },
                     body: {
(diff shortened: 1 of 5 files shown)


test/features/pagecontent/save_api.js
```diff
@@ -418,6 +418,7 @@ describe('page save api', function() {
                     uri: htmlUri,
                     headers: {
                         'x-forwarded-for': '123.123.123.123',
+                        'x-client-ip': '123.123.123.123',
```
d00rman: Why are both set here? Shouldn't we set only `x-client-ip` and then let RB copy it as `XFF` ?

Pchelolo: @d00rman This is not what we set, but what we expect, it's a `nock` configuration.

PR #379, opened 2015-10-16
Public AQS endpoints for global domain

Pchelolo:
1. Introduced `/metrics/pageviews` endpoints for global domain
2. Renamed `per-project` to `aggregate`
3. Moved pageviews module testing from `en.wikipedia.org` to global domain, so that it's fake data could be used in monitoring tests, and because in backend RB it's used on a global domain.
4. Introduced an `x-ample` for one of the page views endpoints
5. Made monitoring tests run for `wikimedia.org` domain as well as for `en.wikipedia.org`
6. Removed `per-article`, `per-project` and `tops` tags from page views endpoints. Swagger gives it's own section for each tag and it looks really ugly to have all these tags with a single endpoint. Could be added back when needed.

Note: in `config.example.wikimedia.yaml` the endpoints are proxied to frontend RESTBase, so testing pageviews endpoints manually with curl wouldn't be possible until this is deployed. It's kinda cyclic reference..

CC @wikimedia/services @milimetric

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -280,7 +280,6 @@ templates:
       /{module:pageviews}:
         x-modules:
           - name: pageviews
-            version: 1.0.0
             type: file
 
   wp-default-1.0.0: &wp/default/1.0.0
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -39,7 +39,6 @@ templates:
       - mediawiki/v1/graphoid
       - mediawiki/v1/mobileapps
       - media/v1/mathoid
-      - analytics/v1/pageviews
       - test
     securityDefinitions: &wp/content-security/1.0.0
       mediawiki_auth:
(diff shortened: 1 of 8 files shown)


milimetric:
This is awesome, thanks for fixing the tests and everything.  Much appreciated.

d00rman:
LGTM, but will need to rebased on top of current master.

milimetric:
Ok, I added a corresponding puppet config change: https://gerrit.wikimedia.org/r/#/c/245887/

I think there's no change required to the AQS config itself.  Also, this is the order of the deployment, right?

1. review and merge this PR
2. deploy AQS
3. deploy RESTBase
4. merge the puppet config

Pchelolo:
There's one test failing in travis. It's totally unrelated to this PR and will be fixed by https://github.com/wikimedia/swagger-router/pull/23

Pchelolo:
@d00rman rebased, fixed a failing test

config.test.yaml
```diff
@@ -283,7 +277,11 @@ templates:
                   status: 200
                   headers: '{$$.merge($.mathoid.body[$.request.params.format].headers,$$.merge($.mathoid.headers, {"x-resource-location": $.request.headers.x-resource-location}))}'
                   body: '{$.mathoid.body[$.request.params.format].body}'
-
+      /{module:pageviews}:
+        x-modules:
+          - name: pageviews
+            version: 1.0.0
```
gwicke: minor: The version isn't really doing much, so perhaps better to omit it.

PR #380, opened 2015-10-20
Set up a testing db name for sqlite testing mode

Pchelolo:
Super-minor: don't rely on default DB name for SQLite testing mode.


eevans:
LGTM

PR #381, opened 2015-10-21
Save space by flattening the per article table

milimetric:
We found that we're using up a lot more disk space than we thought, so
we're going to flatten the per article table.  This change needs to be
deployed so we can create the table on the AQS cluster and start loading
it.  Once it's loading up properly, we can finish the change by doing
all the TODOs in a separate Pull Request.

Please merge and update the gerrit RESTBase mirror and we'll do the rest
of the deploy.


.gitignore
```diff
@@ -10,3 +10,6 @@ npm-debug.log
 
 # vim temp files
 *swp
+
+# Dockerfile (used for deploy)
+Dockerfile
```
gwicke: This shouldn't be ignored, I think. It's part of the repo.

gwicke:
@milimetric, from looking at the schema I'm not sure how this will save a lot of space. The project seems to be part of both the old & new keys, so you are probably not creating a set of tables for each domain.

.gitignore
d00rman: Actually it should, as it is regenerated every time. Good catch @milimetric :)

d00rman:
> from looking at the schema I'm not sure how this will save a lot of space.

Me neither. From what I can tell, you seem to be coalescing what used to be multiple records into one. While that might save some space (index, SSTables, etc), IMHO it's unlikely to bring any drastic data-size reduction. @milimetric what's the reasoning behind this switch? Have you tried to compare the two approaches?

> The project seems to be part of both the old & new keys, so you are probably not creating a set of tables for each domain.

Yup, they have only one *domain* (in the RESTBase sense) and put everything there.

milimetric:
So, the reasoning is that we used to have 16 rows for all possible combinations of agent and access method, and we're collapsing that into 1 row.  So the space savings, unless I'm missing some cassandra magic, will be roughly:

16 rows * 7 columns / 1 row * 20 columns (a little more than a factor of 5)

All the columns have either short strings or integers in them, so they're pretty comparable.  Right now, we're at 2 Terrabytes for 10 days, so the current schema will make us run out of space with about 2 months of data.  If we get this 5-fold reduction, we should be ok for now.  But we'll still run out of space really soon, so we'll need to figure out a long term solution.

milimetric:
Ok so we talked some more and we think the shorter column names on this new commit along with reducing the number of rows will save space.  Compaction will probably give us a little bit but right now the rate of disk usage would mean we'd run out of space before we even loaded the data that we have, let alone new data.  So we need to try something.

PR #382, opened 2015-10-22
Added a simple request validator

Pchelolo:
A simple request validator based on json-schema validator. Currently, we only validate that all required parameters are present, but the validator is very extensible. Also fixed some tests a bit.

### DIFF
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -258,7 +258,6 @@ RESTBase.prototype._request = function(req) {
     }
 
     if (handler) {
-
         // Remove the /{domain}/ prefix, as it's not very useful in stats
         var statName = match.value.path.replace(/\/[^\/]+\//, '')
             + '.' + req.method.toUpperCase() + '.';
@@ -283,8 +282,18 @@ RESTBase.prototype._request = function(req) {
         }
 
         // Call the handler with P.try to catch synchronous exceptions.
-        var reqHandlerPromise = P.try(handler, [childRESTBase, childReq])
-        .then(function(res) {
+        var reqHandlerPromise;
+        if (handler.validator) {
+            reqHandlerPromise = P.try(function() {
+                return handler.validator.validate(childReq);
+            })
+            .then(function() {
+                return handler(childRESTBase, childReq);
+            });
+        } else {
+            reqHandlerPromise = P.try(handler, [childRESTBase, childReq]);
+        }
+        reqHandlerPromise = reqHandlerPromise.then(function(res) {
             // Record request metrics & log
             var statusClass = Math.floor(res.status / 100) + 'xx';
             self.metrics.endTiming([statName + statusClass, statName + 'ALL'], startTime);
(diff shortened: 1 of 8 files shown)


eevans:
Nice!  Do we know what impact this has on performance?

Pchelolo:
@eevans this code can make ~20 million validations per second on my machine, while I have approximately 1000 req/s, so, simple calculations shows, that the impact is about 0.005 :)

lib/validator.js
```diff
@@ -0,0 +1,64 @@
...
+        properties: {}
+    };
+
+    parameters.forEach(function(param) {
+        if (!schema.properties[inMapping[param.in]]) {
+            schema.properties[inMapping[param.in]] = {
+                type: 'object',
+                properties: {}
+            };
+            schema.required = schema.required || [];
```
gwicke: Would it be okay to initialize required to the empty array right away (around line 24)?

Pchelolo: @gwicke no, according to the son-schema spec, required array must have at least on element. If we preinitialise it, it wouldn't compile if there's no required fields.

gwicke: Actually, might be worth mentioning that in a comment, so that we remember later.

lib/validator.js
```diff
@@ -0,0 +1,64 @@
...
+            status: 400,
+            body: {
+                type: 'bad_request',
+                title: ajv.errorsText(this._validatorFunc.errors),
+                req: req
+            }
+        });
+    }
+};
+
```
gwicke: I love how clean this is! Great work!

eevans:
>> @eevans:
>> Do we know what impact this has on performance?
>@Pchelolo:
> this code can make ~20 million validations per second on my machine, while I have approximately 1000 req/s, so, simple calculations shows, that the impact is about 0.005% :)

So, for all intents and purposes... none.  :)  Awesome.

lib/restbase.js
```diff
@@ -283,8 +282,18 @@ RESTBase.prototype._request = function(req) {
...
         // Call the handler with P.try to catch synchronous exceptions.
-        var reqHandlerPromise = P.try(handler, [childRESTBase, childReq])
-        .then(function(res) {
+        var reqHandlerPromise;
+        if (handler.validator) {
+            reqHandlerPromise = P.try(function() {
+                return handler.validator.validate(childReq);
+            })
+            .then(function() {
+                return handler(childRESTBase, childReq);
```
d00rman: Shouldn't this be `return P.try(function() {handler(childRESTBase, chlidReq) })` to keep the same functionality as before (catching sync exceptions) ?

Pchelolo: @d00rman it's already in the promise, so it's equal to the previous functionality. The `P.try` was needed before to 'initialise' a promise chain.

d00rman: Ah right. From the [BB docs](https://github.com/petkaantonov/bluebird/blob/master/API.md#promisetryfunction-fn--arraydynamicdynamic-arguments--dynamic-ctx----promise):

> Start the chain of promises with `Promise.try`. Any synchronous exceptions will be turned into rejections on the returned promise.

PR #383, opened 2015-10-23
Added a x-host-basePath to the example wikimedia config

Pchelolo:
Swagger-UI doesn't work for global domain on https://wikimedia.org/api/rest_v1/?doc , because we've forgot to add `x-host-basePath` to the global domain content spec.


d00rman:
Very nice catch! I'll do a corresponding config change in prod

PR #387, opened 2015-10-26
More features added to the validator

Pchelolo:
Added more features:
- Support for type validators. Everything comes in as strings, so we can't use built-in json-schema type validation. Instead, validate with appropriate regex patterns
- Support additional features like `maxLength`, `minLength`, `pattern`, `enum` etc.
- Support for `body` parameters validated against a json-schema

The only problem is [weird api for sections transform](https://phabricator.wikimedia.org/T116336) that allows either form-data or application/json types to be passed in. Until we figure out what to do with that issue, we cannot validate `sections` input against a complete schema, so just validate that it has some field named `sections`. We have tests for both types of input, so backwards compatibility is preserved for now.

Benchmarks show, that, obviously, as more parameters are added and more complex requirements are added, the validator is getting slower, but even with a very complex and long list of params, the overhead is still almost unnoticeable.

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -7,7 +7,8 @@ var inMapping = {
     path: 'params',
     query: 'query',
     header: 'headers',
-    formData: 'body'
+    formData: 'body',
+    body: 'body'
 };
 
 /**
@@ -34,11 +35,28 @@ function convertToJsonSchema(parameters) {
             schema.required = schema.required || [];
             schema.required.push(inMapping[param.in]);
         }
-        var reqPartSchema = schema.properties[inMapping[param.in]];
-        reqPartSchema.properties[param.name] = {};
-        if (param.required) {
-            reqPartSchema.required = reqPartSchema.required || [];
-            reqPartSchema.required.push(param.name);
+        if (param.in !== 'body') {
+            var reqPartSchema = schema.properties[inMapping[param.in]];
+            var paramSchema = {};
+            // We can't type-check directly, because everything come in as a string
+            if (param.type === 'number' || param.type === 'integer') {
+                paramSchema.type = 'string';
+                paramSchema.pattern = '^\\d+$';
+            } else if (param.type === 'boolean') {
+                paramSchema.type = 'string';
+                paramSchema.pattern = '^(?:true)|(?:false)|(?:1)|(?:0)$';
+            } else if (param.type === 'string') {
+                paramSchema.type = 'string';
+                paramSchema.enum = param.enum;
+                paramSchema.maxLength = param.maxLength;
+                paramSchema.minLength = param.minLength;
+                paramSchema.pattern = param.pattern;
+            }
+            reqPartSchema.properties[param.name] = paramSchema;
+            if (param.required) {
+                reqPartSchema.required = reqPartSchema.required || [];
+                reqPartSchema.required.push(param.name);
+            }
         }
     });
 
@@ -52,6 +70,7 @@ var Validator = function(parameters) {
 
 Validator.prototype.validate = function(req) {
     if (!this._validatorFunc(req)) {
+        console.log(ajv.errorsText(this._validatorFunc.errors));
         throw new HTTPError({
             status: 400,
             body: {
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -28,7 +28,7 @@ Validator.prototype._convertToJsonSchema = function(parameters) {
         type: 'object',
         properties: {}
     };
-    
+
     parameters.forEach(function(param) {
         if (param.in !== 'body') {
             if (!schema.properties[inMapping[param.in]]) {
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -124,8 +124,8 @@ Validator.prototype._createTypeCoercionFunc = function(parameters) {
         var paramAccessor = 'req.' + inMapping[param.in] + '["' + param.name + '"]';
         var paramCoercionCode;
         var errorNotifier = 'throw new HTTPError({status:400,body:{type:"bad_request",'
-                + 'title:"data.' + inMapping[param.in] + '.' + param.name
-                + ' should be ' + param.type + '"}});\n';
+                + ' title:"Invalid parameters", detail: "data.'
+                + inMapping[param.in] + '.' + param.name + ' should be ' + param.type + '"}});\n';
 
         switch (param.type) {
             case 'integer':
@@ -137,10 +137,10 @@ Validator.prototype._createTypeCoercionFunc = function(parameters) {
                     + 'if (Number.isNaN(' + paramAccessor + ')) {\n' + errorNotifier + '}\n';
                 break;
             case 'boolean':
-                paramCoercionCode = 'if(!/^(?:true)|(?:false)|(?:1)|(?:0)$/.test('
+                paramCoercionCode = 'if(!/^true|false|1|0$/.test('
                     + paramAccessor + '.toString())) {\n'
                     + errorNotifier + '}\n'
-                    + paramAccessor + ' = /^(?:true)|(?:1)$/.test('
+                    + paramAccessor + ' = /^true|1$/.test('
                     + paramAccessor + '.toString());\n';
         }
 
@@ -157,7 +157,7 @@ Validator.prototype._createTypeCoercionFunc = function(parameters) {
         code += paramCoercionCode;
     });
     if (code && code.trim()) {
-        code += 'return req;\n';
+        code += '\nreturn req;\n';
         /*jshint evil:true */
         return new Function('req', 'HTTPError', code);
     } else {
@@ -178,7 +178,8 @@ Validator.prototype.validate = function(req) {
             status: 400,
             body: {
                 type: 'bad_request',
-                title: this._ajv.errorsText(this._validatorFunc.errors),
+                title: 'Invalid parameters',
+                detail: this._ajv.errorsText(this._validatorFunc.errors),
                 req: req
             }
         });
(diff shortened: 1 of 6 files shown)


eevans:
Should this have some docs?  Something under `doc/` maybe?

Pchelolo:
@eevans Not sure we need them, it's really close to [swagger specification](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md). I can just list features we support/don't support

lib/validator.js
```diff
@@ -17,37 +23,65 @@ var inMapping = {
...
+        if (param.in !== 'body') {
+            if (!schema.properties[inMapping[param.in]]) {
+                schema.properties[inMapping[param.in]] = {
+                    type: 'object',
+                    properties: {}
+                };
+                // 'required' array must have at least one element according to json-schema spec,
+                // se we can't preinitialize it.
+                schema.required = schema.required || [];
+                schema.required.push(inMapping[param.in]);
```
d00rman: Would be nice to ensure uniqueness for the contents of `schema.required`

specs/mediawiki/v1/content.yaml
```diff
@@ -1142,10 +1143,17 @@ paths:
           type: integer
           required: true
         - name: sections
-          in: formData
+          in: body
```
d00rman: It can be in both actually, depends on which way the client supplies it

d00rman:
I agree with @eevans, some docs would be nice for this.

specs/mediawiki/v1/content.yaml
Pchelolo: @d00rman there's [a ticket](https://phabricator.wikimedia.org/T116336) to figure out which of these to leave. 

The current situation (when we accept both an object and form-data with a single field) is impossible to express with swagger specification.

lib/validator.js
gwicke: @d00rman, are you worried about a parameter being defined multiple times in the swagger spec? Would that be valid?

lib/validator.js
```diff
@@ -17,37 +23,65 @@ var inMapping = {
...
+            }
+
+            var reqPartSchema = schema.properties[inMapping[param.in]];
+            var paramSchema = {};
+            // We can't type-check directly, because everything come in as a string
```
gwicke: An alternative to these string-based tests might be to attempt a coercion right away (using the rules set out below), followed by a validation against the actual schema types. The coercion could return errors of its own, so that we get the helpful error messages describing the correct syntax for string parameters.

Main advantage of this approach is a consistent coercion pass, which can further simplify the code consuming the post data. It would also systematically eliminate differences between multipart form submissions and JSON requests.

lib/validator.js
Pchelolo: @gwicke no, @d00rman is actually correct. If we have 2 different parameters in the request part (e.g. query), it will get to required multiple times. I'll fix that.

lib/validator.js
Pchelolo: @gwicke I've tried that approach, but, for `path` parameters it breaks routing, and I didn't really investigate further. We could still try that, either making `path` params an exception, or adding a `toString()` call in swagger-router somewhere around [here](https://github.com/wikimedia/swagger-router/blob/master/lib/router.js#L103). But, may be we'll have more implications - I'll check tomorrow.

Pchelolo:
@wikimedia/services I've made a new version with type coercion. This one supports almost everything in swagger specification, and automatically coerces bools/ints/numbers. The problem with `sections` api is still there, because, according to the swagger spec, we can't specify `object` type in `formData` field. But when we deprecate and removed form-data type for this endpoint, all should be fine. 

Performance is a bit worse then before, but it's expected, we are coercing the types now, but even for very complex validation, with ~10 integer types, we get around 3 million validations per second, so the impact on performance would still be unnoticeable.

When we deploy and stabilise this, as a next step we could remove some validations, made in the handler code.

lib/validator.js
```diff
@@ -15,48 +54,131 @@ var inMapping = {
...
+            var paramSchema = { type: param.type };
+            supportedValidators.forEach(function(validator) {
+                paramSchema[validator] = param[validator];
+            });
+            reqPartSchema.properties[param.name] = paramSchema;
+            if (param.required) {
+                reqPartSchema.required = reqPartSchema.required || [];
+                reqPartSchema.required.push(param.name);
+            }
+        } else {
```
d00rman: So if there are multiple parameters in the body, the last one would win here?

lib/validator.js
```diff
@@ -15,48 +54,131 @@ var inMapping = {
...
+    var code = '';
+    parameters.forEach(function(param) {
+        if (param.type === 'string') {
+            // Don't need to process strings
+            return;
+        }
+        var paramAccessor = 'req.' + inMapping[param.in] + '["' + param.name + '"]';
+        var paramCoercionCode;
+        var errorNotifier = 'throw new HTTPError({status:400,body:{type:"bad_request",'
+                + 'title:"data.' + inMapping[param.in] + '.' + param.name
```
d00rman: I'd put this as the `detail` field, while have something like *Invalid parameter specification* as the error `title`.

lib/validator.js
```diff
@@ -15,48 +54,131 @@ var inMapping = {
...
+            case 'integer':
+                paramCoercionCode = paramAccessor + ' = parseInt(' + paramAccessor + ');\n'
+                    + 'if (!Number.isInteger(' + paramAccessor + ')) {\n' + errorNotifier + '}\n';
+                break;
+            case 'number':
+                paramCoercionCode = paramAccessor + ' = parseFloat(' + paramAccessor + ');\n'
+                    + 'if (Number.isNaN(' + paramAccessor + ')) {\n' + errorNotifier + '}\n';
+                break;
+            case 'boolean':
+                paramCoercionCode = 'if(!/^(?:true)|(?:false)|(?:1)|(?:0)$/.test('
```
d00rman: `/^(?:true|false|0|1)$/` should work just as well.

lib/validator.js
```diff
@@ -15,48 +54,131 @@ var inMapping = {
...
+        if (!param.required) {
+            // If parameter is not required, don't try to coerce "undefined"
+            paramCoercionCode = 'if (' + paramAccessor + ' !== undefined) {\n'
+                + paramCoercionCode + '}\n';
+        }
+
+        code += paramCoercionCode;
+    });
+    if (code && code.trim()) {
+        code += 'return req;\n';
```
d00rman: Nit: add a space or a new line before `return` since you're trimming it in the line above.

lib/validator.js
d00rman: .. if a schema has not been defined, that is.

Pchelolo: According to the swagger specification, you can have only one `body` parameter - it represents the whole body, not some field in the body. And I assume our spec are valid from swagger perspective.

specs/mediawiki/v1/content.yaml
```diff
@@ -1142,10 +1143,17 @@ paths:
...
           description: Sections to transform
           type: string
           required: true
+          schema:
+            type: 'object'
+            properties:
+              sections:
```
d00rman: Should we also put `type: string` and perhaps a desc here?

Pchelolo: @d00rman right now we can't put a `string` there. Because we still support both `application/json` and `form-data` on this endpoint. In first case this will be an object, in the second case - a string. So, we first need to properly deprecate the `form-data` option, ensure nobody's using it, and only then we can set `type: object` here. 

As for description - yes, we could, but it wouldn't be seen anywhere in the swagger UI.

gwicke: Given how random those IDs are currently, I would be surprised if anybody was using the entry point. It's officially marked as unstable, which means that it might change with fairly quickly.

lib/validator.js
```diff
@@ -15,48 +54,132 @@ var inMapping = {
...
+    parameters.forEach(function(param) {
+        if (param.type === 'string') {
+            // Don't need to process strings
+            return;
+        }
+        var paramAccessor = 'req.' + inMapping[param.in] + '["' + param.name + '"]';
+        var paramCoercionCode;
+        var errorNotifier = 'throw new HTTPError({status:400,body:{type:"bad_request",'
+                + ' title:"Invalid parameters", detail: "data.'
+                + inMapping[param.in] + '.' + param.name + ' should be ' + param.type + '"}});\n';
```
gwicke: Given that most of these errors probably come from post or path parameters, could we provide users with some hints about which syntax would be valid? For 'integer' and 'float' this might be fairly obvious, but for the boolean it might be slightly less so.

lib/validator.js
```diff
@@ -15,48 +54,132 @@ var inMapping = {
...
+                paramCoercionCode = paramAccessor + ' = parseInt(' + paramAccessor + ');\n'
+                    + 'if (!Number.isInteger(' + paramAccessor + ')) {\n' + errorNotifier + '}\n';
+                break;
+            case 'number':
+                paramCoercionCode = paramAccessor + ' = parseFloat(' + paramAccessor + ');\n'
+                    + 'if (Number.isNaN(' + paramAccessor + ')) {\n' + errorNotifier + '}\n';
+                break;
+            case 'boolean':
+                paramCoercionCode = 'if(!/^true|false|1|0$/.test('
+                    + paramAccessor + '.toString())) {\n'
```
gwicke: Slightly safer might be to say paramAccessor + "''". That won't explode on methods that don't have a `.toString` method.

gwicke:
Some minor nits inline, but overall this looks good to me.

PR #388, opened 2015-10-26
Added graphoid endpoints

Pchelolo:
This is a work-in-progress PR and some features are not implemented yet.

Adding 2 public routes:
- `POST /media/graph/{format}?mode={store|preview}` - only for internal IPs
  - In store mode:
    - A request hash is calculated
    - Storage checked for already rendered content
    - Two requests are sent to graphoid for png and svg content. If error occurs - it's propagated back to user and nothing is stored.
    - Results for png and svg are stored
    - Request is stored
    - Result returned with `x-resource-location` header
  - In preview mode:
    - Requests are passed to graphoid png and svg endpoints
    - Results are stored. To avoid clashing hashes with `store` mode, special `stub` field is added to the data when a hash is calculated. The stub field is not stored.
    - Result is returned with `x-resource-location` header
- `GET /media/graph/{format}/{hash}`
  -  Content buckets for a given format is checked.
  - If not found - original request is loaded
  - Original request sent to graphoid to regenerate content
  - Result is stored and returned.

Both endpoints support `cache-control: no-cache` header to force rerendering.

What's left:
- Title and revision are not stored for debugging purposes yet - need to make some changes in post_data to not use them is hash calculation
- Monitoring should be added to endpoints
- Graphoid endpoints are not deployed yet, so all graphoid tests are failing.
- There's not way to set _TTL for storage yet, so right now everything is stored permanently, even `preview` content.


nyurik:
Would it make sense to have just one bucket for both store & preview?  I think we can simplify it to this code?

=== `POST /media/graph/{format}?mode=store|preview`:
* Calculate request hash:   hash = hash_function(post_body)
* Calculate storage TTL:   ttl = (mode == "store" || hash in store) ? infinite : 1 hour
* Request is passed to graphoid /all endpoint
** If graphoid returns error, it is propagated back to user and nothing is stored
* Store request
* Store png and svg results
* Result returned with x-resource-location header

=== `GET /media/graph/{format}/{hash}`
* Get image for {hash}.{format}
* If not found - original request is loaded
* Original request sent to graphoid to regenerate content
** If graphoid returns error, it is propagated back to user and nothing is stored. We are relying on Varnish to cache the error result for a while, and not to request new image right away.
* Result is stored and returned.

Questions:
* what is the exact format of the x-resource-location ?

Pchelolo:
> Would it make sense to have just one bucket for both store & preview? 
As far as I know, neither cassandra nor sqlite backend implements setting a TTL for individual rows. We can potentially implement it, but even if we had that, logic would be quite complicated - we don't ever want to decrease TTL on preview with hash matching some stored content, and with `cache-control:no-cache` support for rerendering it would be really hard to get it right.
> what is the exact format of the x-resource-location ?
Something like this: `x-resource-location: a4a7724d64be3bb1d17bd95732f44eb164612fee`

nyurik:
I might be reading it wrong, but from what I understood, the TTL is for each individual cell. So if you set data to have TTL = hundred years, it should work as permanent, right?    Also, for a POST request, you can always refresh the data, even if it exists in cache.

nyurik:
Update:  I spoke with our UX dev re "interactive mode" for the graphs, and I realized that the `GET /media/graph/{format}/{hash}` needs to be able to support `format=json` to allow users to click on the static graph to convert it into an interactive graph. On click, the client will download vega libraries and the graph spec from restbase.

Pchelolo:
The PR is ready for review!

nyurik:
The service has been deployed to production

config.example.wikimedia.yaml
```diff
@@ -449,6 +442,168 @@ templates:
             - get_from_backend:
                 request:
                   uri: https://wikimedia.org/api/rest_v1/metrics/pageviews/top/{+rest}
+      /{module:graphoid}/v1/store/{format}:
```
d00rman: An empty line between the pageviews and graphoid modules would increase readability.

config.example.wikimedia.yaml
```diff
@@ -449,6 +442,168 @@ templates:
...
+            - graphoid:
+                request:
+                  method: post
+                  uri: /{domain}/sys/graphoid/v1/service_request
+                  body: '{$.request.body}'
+            - store_png:
+                request:
+                  method: put
+                  uri: /{domain}/sys/key_value/graphoid.png/{$.request.headers.x-resource-location}
+                  headers: '{$$.merge($.graphoid.body.png.headers, {"x-resource-location": $.request.headers.x-resource-location})}'
```
d00rman: You need to strip the `content-encoding` header.

config.example.wikimedia.yaml
```diff
@@ -449,6 +442,168 @@ templates:
...
+                  body: '{$.graphoid.body.png.body}'
+              store_svg:
+                request:
+                  method: put
+                  uri: /{domain}/sys/key_value/graphoid.svg/{$.request.headers.x-resource-location}
+                  headers: '{$$.merge($.graphoid.body.svg.headers, {"x-resource-location": $.request.headers.x-resource-location})}'
```
d00rman: Idem.

config.example.wikimedia.yaml
```diff
@@ -449,6 +442,168 @@ templates:
...
+                  body: '{$.graphoid.body.svg.body}'
+            - return_result:
+                return:
+                  status: 200
+                  headers: '{$$.merge($.graphoid.body[$.request.params.format].headers, {"x-resource-location": $.request.headers.x-resource-location})}'
```
d00rman: Idem

specs/media/v1/graphoid.yaml
```diff
@@ -0,0 +1,128 @@
...
+          in: path
+          description: The format of the output image
+          type: string
+          enum:
+            - png
+            - svg
+          required: true
+        - name: mode
+          in: query
+          description: If `preview`, store content temporary. Default is `store`
```
d00rman: `s/temporary/temporarily/`

specs/media/v1/graphoid.yaml
```diff
@@ -0,0 +1,128 @@
...
+          enum:
+            - preview
+            - store
+        - name: content
+          in: body
+          description: Graph definition
+          required: true
+      responses:
+        '200':
+          description: The PNG render of the requested graph.
```
d00rman: "The render of the given graph in the requested format"

specs/media/v1/graphoid.yaml
```diff
@@ -0,0 +1,128 @@
...
+                - internal
+      x-request-handler:
+        - post_to_backend:
+            request:
+              method: post
+              uri: /wikimedia.org/sys/graphoid/v1/{$$.default($.request.query.mode, "store")}/{format}
+              headers:
+                cache-control: '{cache-control}'
+              body: '{$.request.body}'
+
```
d00rman: No need for this empty line here.

d00rman:
Does this PR depend on #387 ? If not, it shouldn't include its changes.

Pchelolo:
@d00rman It's not changes from https://github.com/wikimedia/restbase/pull/387, just 378 adds support for `body` type in validator (which we didn't use before, so didn't support), but to let this one proceed without 378, I'm just excluding `body` type from validation, because proper handling of this type needs more work. 

So, if we merge 378 before this one, I'll remove changes in validator, if this one goes before 378 - I'll rebase 378.

d00rman:
In that case I suggest we first deal with #378 and then continue here, as I find the current state of affairs rather confusing.

Pchelolo:
@d00rman kk, I'm working on 378 right now, so should be ready by the end of the day

config.example.wikimedia.yaml
```diff
@@ -450,6 +443,170 @@ templates:
...
+            - graphoid:
+                request:
+                  method: post
+                  uri: /{domain}/sys/graphoid/v1/service_request
+                  body: '{$.request.body}'
+            - store_png:
+                request:
+                  method: put
+                  uri: /{domain}/sys/key_value/graphoid.png/{$.request.headers.x-resource-location}
+                  headers: '{$$.merge($$.strip($.graphoid.body.png.headers, "content-encoding"), {"x-resource-location": $.request.headers.x-resource-location, "max-age": $.request.headers.max-age})}'
```
gwicke: This content-encoding stripping looks fairly repetitive. Could we strip those headers further down the stack?

It might also save some complexity to have a function to pick some keys from an object. Something like `$.pick($.request.headers, ["x-resource-location","max-age"])`.

config.example.wikimedia.yaml
```diff
@@ -450,6 +443,170 @@ templates:
                 request:
                   uri: https://wikimedia.org/api/rest_v1/metrics/pageviews/top/{+rest}
 
+      /{module:graphoid}/v1/store/{format}:
```
gwicke: We really need a way to pull in spec fragments like this from a separate file, passing in a config object.

config.example.wikimedia.yaml
```diff
@@ -450,6 +443,170 @@ templates:
...
+                request:
+                  method: put
+                  uri: /{domain}/sys/post_data/graphoid.request/hash
+                  body: '{$$.merge($.request.body, {"x-restbase-mode": "preview"})}'
+            - handle_req:
+                request:
+                  method: post
+                  uri: /{domain}/sys/graphoid/v1/handler/{$.request.params.format}
+                  headers:
+                    max-age: 3600
```
gwicke: It would be nice to use a shared stash for this, instead of relying on per-row TTLs.

config.example.wikimedia.yaml
```diff
@@ -450,6 +443,170 @@ templates:
...
+                  uri: /{domain}/sys/post_data/graphoid.request/{hash}
+            - handle_req:
+                request:
+                  method: post
+                  uri: /{domain}/sys/graphoid/v1/handler/{format}
+                  headers:
+                    cache-control: '{cache-control}'
+                    x-resource-location: '{$.request.params.hash}'
+                    content-type: application/json
+                  body: '{$.load_request.body}'
```
gwicke: This could also be where we make the graphoid request & store the response. We could then skip the entire block below, simplifying the overall stanza.

mods/key_value.js
```diff
@@ -250,11 +254,18 @@ KVBucket.prototype.putRevision = function(restbase, req) {
...
                 tid: tid,
                 value: req.body,
-                'content-type': req.headers['content-type']
-                // TODO: include other data!
+                'content-type': req.headers['content-type'],
+                headers: req.headers
             }
         }
     };
+    if (req.headers && req.headers['max-age']) {
```
gwicke: We can't sanely support per-row TTLs in the backend, so will probably have to replace this with a shared stash bucket with a retention policy.

mods/post_data.js
```diff
@@ -98,7 +98,16 @@ PostDataBucket.prototype.putRevision = function(restbase, req) {
             throw e;
         }
     });
+};
 
+PostDataBucket.prototype.calculateHash = function(restbase, req) {
```
gwicke: I almost wonder if we should hook this up as `$.hash()`, saving the extra endpoint.

specs/media/v1/graphoid.yaml
```diff
@@ -0,0 +1,141 @@
...
+            status: 200
+            headers:
+              content-type: /^image\/svg/
+            body: /.+/
+
+  /{module:media}/graph/{format}/{hash}:
+    get:
+      tags:
+        - Media
+        - Graph
```
gwicke: Perhaps only use "Media", to avoid making the the docs too verbose by listing things multiple times?

lib/validator.js
```diff
@@ -7,7 +7,8 @@ var inMapping = {
     path: 'params',
     query: 'query',
     header: 'headers',
-    formData: 'body'
+    formData: 'body',
+    body: 'body'
```
gwicke: This looks like something from #387.

gwicke: Ah, nevermind; you already discussed this before the latest amends.

d00rman:
Could use a rebase :)

PR #389, opened 2015-10-28
Do not DOS ourselves by re-rendering the same title in monitoring

gwicke:
Somehow we managed to repeat the earlier mistake of re-generating the same
title over & over again. At sufficiently high rates (hundreds of re-renders
per hour), this will overwhelm retention policies (once queries fail), and
then lead to extremely wide partitions, timeouts and OOM on query.

### DIFF
--- a/specs/mediawiki/v1/mobileapps.yaml
+++ b/specs/mediawiki/v1/mobileapps.yaml
@@ -42,18 +42,6 @@ paths:
                 cache-control: '{cache-control}'
       x-monitor: true
       x-amples:
-        - title: Get MobileApps Main Page
-          request:
-            params:
-              domain: en.wikipedia.org
-              title: Main_Page
-            headers:
-              cache-control: no-cache
-          response:
-            status: 200
-            headers:
-              content-type: /text\/html/
-            body: /body/
         - title: Get MobileApps Main Page from RB storage
           request:
             params:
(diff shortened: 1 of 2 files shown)


specs/mediawiki/v1/mobileapps.yaml
```diff
@@ -47,8 +47,6 @@ paths:
             params:
               domain: en.wikipedia.org
               title: Main_Page
-            headers:
-              cache-control: no-cache
           response:
```
Pchelolo: Now the next monitoring test is identical to this one, so the whole x-ample can be removed.

gwicke: Yeah, good point. Will amend.

Pchelolo:
And it looks like travis stopped support for io.js (https://travis-ci.org/wikimedia/restbase/jobs/87935430), so may be lets remove it as a part of this PR?

Pchelolo:
:+1:

gwicke:
There are some other issues here that we should look into:
- The retention policy is set up to keep renders for 24 hours. It might make sense to change this to only keep one render, or at least only the renders of a short period (like 10 minutes).
- We only want to retain renders of the latest revision. It might be worth considering to use key_value storage instead of key_rev_value.

eevans:
:+1:

PR #391, opened 2015-10-30
Patch to remove all-years parameter

Ironholds:
This should remove the all-years parameter entirely from the pageviews side of things (and also adds a unit test to protect against regressions).

There are test failures but given that they're unrelated to the actual patch I'm going to assume they're related to the absence of cassandra.

### DIFF
--- a/mods/pageviews.js
+++ b/mods/pageviews.js
@@ -228,11 +228,6 @@ var validateYearMonthDay = function(rp) {
 
     stripOrgFromProject(rp);
 
-    if (rp.year === 'all-years' && (rp.month !== 'all-months' || rp.day !== 'all-days')) {
-        errors.push(
-            'month must be "all-months" and day must be "all-days" when passing "all-years"'
-        );
-    }
     if (rp.month === 'all-months' && rp.day !== 'all-days') {
         errors.push('day must be "all-days" when passing "all-months"');
     }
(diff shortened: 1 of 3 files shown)


mods/pageviews.js
```diff
@@ -242,7 +237,6 @@ var validateYearMonthDay = function(rp) {
 
     // fake a timestamp in the YYYYMMDDHH format so we can reuse the validator
     var validDate = validateTimestamp(
-        ((rp.year === 'all-years') ? '2015' : rp.year) +
```
milimetric: So we still need to concatenate rp.year to the beginning of this string, otherwise validateTimestamp will always return false.

Ironholds: Done.

milimetric:
Thanks Oliver, this looks good now.  I am going to push more changes into it (or move it to my own pull request if I don't have rights to do that.  Because we don't want to deploy AQS and RESTBase too many times and I have one more spec change that needs to go out (removing the "hourly" granularity as an option, and fixing the 404s).

Ironholds:
Sounds good! (Wait, hourly is being removed? As a temporary or permanent change? And from which types of call?)

milimetric:
hourly is being removed from the per-article endpoint.  None of the
solutions we looked into (Cassandra, Druid, Elastic Search) will fit hourly
resolution data with the number of dimensions we currently have in
per-article.  So we're talking about adding a new endpoint called
per-article-hourly that might be able to hold just article title and view
count (that was the IRC chat yesterday).  Other options are to dump one
file per article title with its full history of hourly views.  And we have
a few other ways to skin the cat, so we're not abandoning it.  Just the
limited hardware we have right now won't hold hourly data for more than a
couple of months.  And if we index it in a fancier way like Druid and
Elastic, we don't have enough memory to query it anywhere near fast enough.

Ironholds:
Makes total sense (and makes my job easier).

d00rman:
@Ironholds the tests are failing because of another service being currently unavailable. @milimetric please open a new PR for that change.

PR #392, opened 2015-10-31
Revisions mod: Bug fix: Compare two revisions properly

d00rman:
Currently, when checking the equality of two revisions, attribute values are compared using the `!==` operator. However, the string representations of the `timestamp` attribute always differ because of a small discrepancy in the way storage and the MW API stringify them: one includes milliseconds while the other one doesn't, leading RESTBase to believe no two revisions are equal.

This PR corrects that by special-casing the `timestamp` attribute and casting both values to Date before comparison. Also, Arrays are differently checked for equivalence (number of elements, then each element separately).

### DIFF
--- a/mods/page_revisions.js
+++ b/mods/page_revisions.js
@@ -184,11 +184,29 @@ PRS.prototype.listTitles = function(restbase, req, options) {
  */
 PRS.prototype._checkSameRev = function(firstRev, secondRev) {
     return !Object.keys(firstRev).some(function(attrName) {
+        var firstVal = firstRev[attrName];
+        var secondVal = secondRev[attrName];
         // We don't really care if an empty value is null, or undefined, or other falsy
-        if (attrName === 'tid' || !firstRev[attrName] || !secondRev[attrName]) {
+        if (!firstVal || !secondVal || attrName === 'tid') {
+            return false;
+        } else if (attrName === 'timestamp') {
+            // 'timestamp' fields need to be parsed because Cassandra
+            // returns a ISO8601 ts which includes milliseconds, while
+            // the ts returned by MW API does not
+            return Date.parse(firstVal) !== Date.parse(secondVal);
+        } else if (Array.isArray(firstVal) || Array.isArray(secondVal)) {
+            // we need a special case for arrays (the 'restrictions' attribute)
+            if (Array.isArray(firstVal) && Array.isArray(secondVal)
+                    && firstVal.length === secondVal.length) {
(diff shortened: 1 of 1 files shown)


mods/page_revisions.js
```diff
@@ -184,11 +184,29 @@ PRS.prototype.listTitles = function(restbase, req, options) {
...
+        } else if (attrName === 'timestamp') {
+            // 'timestamp' fields need to be parsed because Cassandra
+            // returns a ISO8601 ts which includes milliseconds, while
+            // the ts returned by MW API does not
+            return Date.parse(firstVal) !== Date.parse(secondVal);
+        } else if (Array.isArray(firstVal) || Array.isArray(secondVal)) {
+            // we need a special case for arrays (the 'restrictions' attribute)
+            if (Array.isArray(firstVal) && Array.isArray(secondVal)
+                    && firstVal.length === secondVal.length) {
+                for (var idx = 0; idx < firstVal.length; idx++) {
```
gwicke: Minor: You could use `Array.every` here.

d00rman: True, but I'd still need the index to compare it to the other value, so I opted for the old-school style :)

gwicke:
Good catch!

Pchelolo:
@d00rman when arrays have different length, or in one object the property is an array, and in another - not an array - the check returns true, which is incorrect.

Pchelolo:
@d00rman another thing, is that I can't understand how does the page deleting test pass with the current array length bug. In theory, only the `restrictions` array length would differ when page deletion is spotted, so the test should fail. This means that either test is incorrect, or there's some other bug around this code. 

Could you leave fixing the array length bug to me? I wanna look at the deletion test too tomorrow, I've made that test, so I'm interested where did I make a mistake.

d00rman:
@Pchelolo ah indeed for the length difference. Well spotted! The incorrect outcome for valuesbeing something other than strings is not really important, as revisions have the `restrictions` array, but we can make that explicit (test for `attrName` instead of type)

gwicke:
This bug had quite serious performance implications on the cluster, and might have partly caused the OOMs we saw last week. Lets try to get a proper fix deployed asap.

In the meantime, after thorough testing in staging, I have gone ahead and truncated the biggest revision tables in production. This has quite significantly reduced iowait on restbase100[7-9] (see below), as well as visibly dropped p99 latency across all cassandra queries.

image

d00rman:
This fix was deployed as part of https://gerrit.wikimedia.org/r/#/c/250375/ .

PR #395, opened 2015-11-02
Parsoid: Switch to V3 api, adapt tests and normalise Content-Type

Pchelolo:
Switch RB to use [v3 Parsoid API](https://www.mediawiki.org/wiki/Parsoid/API).

Changes:
- Switch endpoints and adapt `nock` in a security test, which verifies that cookies are passed correctly.
- Don't deal with `body_only` in RESTBase at all, now parsoid supports it, let it handle the flag.
- `Content-Type` formatting. Enforce predictable order of fields in the `Content-Type` header from parsoid. It's arguable we need it right now, but it could be treated as preparation for [this](https://phabricator.wikimedia.org/T116333). Performance impact is fairly low - I get ~350.000 `normalizeContentType` calls per second, while an average request rate is lower that 1000 r/s on my machine. So, we get around 0.2% overhead. In future, apart from plain normalisation, we'd parse the content-type, and in case it's version is different from what's defined in a spec - rerender the stored content.

cc @wikimedia/services @cscott

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -211,7 +211,7 @@ PSP.pagebundle = function(restbase, req) {
     // TODO: Pass in current or predecessor version data if available
     var newReq = Object.assign({}, req);
     if (!newReq.method) { newReq.method = 'get'; }
-    newReq.uri = this.parsoidHost + '/v2/' + domain + '/pagebundle/'
+    newReq.uri = this.parsoidHost + '/' + domain + '/v3/page/pagebundle/'
         + encodeURIComponent(rbUtil.normalizeTitle(rp.title)) + '/' + rp.revision;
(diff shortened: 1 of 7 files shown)


d00rman:
LGTM :+1: @wikimedia/services let's merge this?

mods/parsoid.js
```diff
@@ -624,6 +625,11 @@ PSP.transformRevision = function(restbase, req, from, to) {
         if (req.body.scrubWikitext || req.body.scrub_wikitext) {
             body2.scrubWikitext = true;
         }
+
+        if (req.body.bodyOnly || req.body.body_only) {
+            body2.body_only = true;
```
gwicke: Could this cause a body-only version to be stashed?

PR #396, opened 2015-11-02
Clarify 404 errors and make room for bigger ints

milimetric:
* 404 errors get a more descriptive message
* int was too small for per-project monthly data, increasing to long
* 'hourly' granularity resulted in too much data, removing permanently
from per-article-flat, might add later in a different shape
* 'monthly' is being calculated for per-project, so add it to the query
options

NOTE: do not merge this until support for long columns is added

Bug: T117017

### DIFF
--- a/mods/pageviews.js
+++ b/mods/pageviews.js
@@ -61,7 +61,9 @@ var tableSchemas = {
             // the hourly timestamp will be stored as YYYYMMDDHH
             timestamp: 'string',
             views: 'int',
-            v: 'long'
+            // store this as a string because it's too big for an int
+            // and long/bigint are not supported in RESTBase at this time
+            v: 'string'
         },
         index: [
             { attribute: 'project', type: 'hash' },
@@ -315,8 +317,14 @@ PJVS.prototype.pageviewsForProjects = function(restbase, req) {
     return dataRequest.then(normalizeResponse).then(function(res) {
         if (res.body.items) {
             res.body.items.forEach(function(item) {
-                // prefer the long column if it's loaded
-                item.views = item.v || item.views;
+                // prefer the v column if it's loaded
+                if (item.v && item.v !== '') {
+                    try {
+                        item.views = parseInt(item.v);
+                    } catch (e) {
+                        item.views = null;
+                    }
+                }
                 delete item.v;
             });
         }
--- a/mods/pageviews.js
+++ b/mods/pageviews.js
@@ -63,7 +63,7 @@ var tableSchemas = {
             views: 'int',
             // store this as a string because it's too big for an int
             // and long/bigint are not supported in RESTBase at this time
-            v: 'string'
+            v: 'long'
         },
         index: [
             { attribute: 'project', type: 'hash' },
@@ -128,17 +128,12 @@ Object.keys(viewCountColumnsForArticleFlat).forEach(function(k) {
 
 var notFoundCatcher = function(e) {
     if (e.status === 404) {
-        throw new rbUtil.HTTPError({
-            status: 404,
-            body: {
-                type: 'error',
-                description: 'The date(s) you used are valid, but we either do ' +
+        e.body.description = 'The date(s) you used are valid, but we either do ' +
                              'not have data for those date(s), or the project ' +
                              'you asked for is not loaded yet.  Please check ' +
                              'https://wikimedia.org/api/rest_v1/?doc for more ' +
-                             'information.'
-            }
-        });
+                             'information.';
+        e.body.type = 'not_found';
     }
     throw e;
 };
@@ -318,9 +313,9 @@ PJVS.prototype.pageviewsForProjects = function(restbase, req) {
         if (res.body.items) {
             res.body.items.forEach(function(item) {
                 // prefer the v column if it's loaded
-                if (item.v && item.v !== '') {
+                if (item.hasOwnProperty('v') && item.v !== null) {
                     try {
-                        item.views = parseInt(item.v);
+                        item.views = parseInt(item.v, 10);
                     } catch (e) {
                         item.views = null;
                     }
--- a/mods/pageviews.js
+++ b/mods/pageviews.js
@@ -23,7 +23,6 @@ function PJVS(options) {
 
 
 var tables = {
-    article: 'pageviews.per.article',
     articleFlat: 'pageviews.per.article.flat',
     project: 'pageviews.per.project',
     tops: 'top.pageviews',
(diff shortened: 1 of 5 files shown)


Pchelolo:
Here's a PR that adds `long` support to storage https://github.com/wikimedia/restbase-mod-table-spec/pull/25

milimetric:
Thanks, Petr.  It seems to me the string type we're using now, with the
parseInt, is best for us.  That way our client doesn't have to parseInt for
one of our endpoints but not the others.  Do you agree or you think we
should change it to the new long -> string type your PR implements?

Pchelolo:
@milimetric Actually, I think using `long` is better, as you will get more efficient storage. 
From the client-side and code perspective, nothing would differ: the `long` type consumes and returns `string`, but internally, in cassandra table, it's mapped to a `bigint`, which should be more efficient. So, you don't need to make any changes in your current code, except making `v` a `long` after that PR is merged.

mods/pageviews.js
```diff
@@ -83,7 +60,10 @@ var tableSchemas = {
             granularity: 'string',
             // the hourly timestamp will be stored as YYYYMMDDHH
             timestamp: 'string',
-            views: 'int'
+            views: 'int',
+            // store this as a string because it's too big for an int
+            // and long/bigint are not supported in RESTBase at this time
+            v: 'string'
```
d00rman: Perhaps use a more descriptive column name?

mods/pageviews.js
```diff
@@ -146,6 +126,23 @@ Object.keys(viewCountColumnsForArticleFlat).forEach(function(k) {
     tableSchemas.articleFlat.attributes[viewCountColumnsForArticleFlat[k]] = 'int';
 });
 
+var notFoundCatcher = function(e) {
+    if (e.status === 404) {
+        throw new rbUtil.HTTPError({
```
d00rman: Should be enough to modify `e` directly, i.e. modify its `e.body.description` field.

mods/pageviews.js
```diff
@@ -146,6 +126,23 @@ Object.keys(viewCountColumnsForArticleFlat).forEach(function(k) {
...
+            status: 404,
+            body: {
+                type: 'error',
```
d00rman: `s/error/not_found`

test/features/pageviews/pageviews.js
```diff
@@ -88,6 +92,38 @@ describe('pageviews endpoints', function () {
         });
     });
 
+    it('should not die on bad data in the v column', function () {
+        return preq.post({
+            uri: server.config.globalURL +
+                 fix(insertProjectEndpoint, 'en.wikipedia', '2') +
+                 '/catch'
```
Pchelolo: This is not inserting not only to the `v` column, but to the `views` column to. `views` is integer, therefor you get tests failing inside the driver code attempting to convert this 'catch' to a number.

milimetric:
Oh, so you mean keep the parseInt(item.v) logic so the client gets the JS
Number, but just use 'long' in the table schema?  That makes sense, though
storage isn't an issue for this table, it's rolled up to the project level
so it'll always be tiny in comparison to the others.

Pchelolo:
> Oh, so you mean keep the parseInt(item.v) logic so the client gets the JS
Number, but just use 'long' in the table schema?

Exactly. This would probably win a little bit, but it's for free (the backends PR will probably settle soon)

Pchelolo:
@milimetric The `long` PRs have been merged. Please update cassandra backend version to 0.8.8 and sqlite to 0.1.11

d00rman:
> @milimetric The long PRs have been merged. Please update cassandra backend version to 0.8.8 and sqlite to 0.1.11

I.e. do `rm -rf node_modules && npm install`

mods/pageviews.js
milimetric: I'm using "v" for two reasons:

1. Cassandra stores column names with each row, so I've become very cautious about descriptive names
2. This is a helper column, and hopefully temporary.  It seemed from the discussions yesterday that even though RESTBase didn't support removing a column name, we could hack the table such that eventually we end up with views: 'long' instead of views: 'int'.  I get that having both views and v, with nulls in one or the other doesn't store any more data than just having a single column, but I also hate looking at code that has two columns which do the same thing.

mods/pageviews.js
milimetric: ah, I was trying to modify `e.description`, `e.body` is what I was missing.  Thanks, will update.

mods/pageviews.js
milimetric: done

test/features/pageviews/pageviews.js
milimetric: I added a second insert endpoint that lets me control if I'm inserting 'views' or 'v'.  I got this test to pass on SQLite, but it doesn't work in Cassandra (I guess because now v is also a bigint, so inserting 'catch' in either of them won't work).  So I'll just remove the test.

mods/pageviews.js
```diff
@@ -83,7 +60,10 @@ var tableSchemas = {
             granularity: 'string',
             // the hourly timestamp will be stored as YYYYMMDDHH
             timestamp: 'string',
-            views: 'int'
+            views: 'int',
+            // store this as a string because it's too big for an int
```
Pchelolo: The comment is outdated.

Pchelolo:
One tiny thing about an outdated comment, otherwise LGTM

milimetric:
thx, sorry 'bout that, fixed.

mods/pageviews.js
```diff
@@ -320,9 +305,25 @@ PJVS.prototype.pageviewsForProjects = function(restbase, req) {
...
 
-    });
+    }).catch(notFoundCatcher);
 
-    return dataRequest.then(normalizeResponse);
+    return dataRequest.then(normalizeResponse).then(function(res) {
+        if (res.body.items) {
+            res.body.items.forEach(function(item) {
+                // prefer the v column if it's loaded
+                if (item.hasOwnProperty('v') && item.v !== null) {
```
d00rman: `if (item.v)` should be enough given that by the time this code is deployed, the underlying table will be updated.

Also, it might be a good idea to run a script that goes through all of the records and moves the `views` values into `v`, so that such checks may be removed.

milimetric: Not true, if (item.v) fails for v === 0

d00rman: My point was you don't need the `hasOwnProperty` part

milimetric: I mean I could do `if ('v' in item)`, but eslint doesn't love that in browsers so I don't usually use it.  Since `if (item.v)` is not correct what else would you like to see here.  Also, what's wrong with `hasOwnProperty`, this is exactly what it was meant for.

As a sort of higher level point, I don't think it's too productive to discuss tiny details like this in code review.  If this introduced a problem with the code you could suggest a test that would fail because of it.  And I'd be happy to implement the test and the fix.  But if it works, we waste time by explaining to each other style nuances and correctness nuances that we might not have thought of.

d00rman: Heh, no, my point was exactly that you don't even need to test for `v`, as it will be there for sure (even though as you are writing the code it doesn't). So `if (item.v !== null)` will be enough.

milimetric:
Ok, I removed the check for v, can we please merge this now?

PR #397, opened 2015-11-04
Key_Value Module: Store all headers

Pchelolo:
As a part of graphoid PR (https://github.com/wikimedia/restbase/pull/388) we've made key_value module store all headers, not only `content-type`. But now this feature is needed for mobile-apps too, so I'm extracting this little part from the graphoid PR. It will be rebased when we merge this.

cc @wikimedia/services

### DIFF
--- a/mods/key_value.js
+++ b/mods/key_value.js
@@ -31,8 +31,9 @@ KVBucket.prototype.getBucketInfo = function(restbase, req, options) {
 };
 
 KVBucket.prototype.makeSchema = function(opts) {
-    opts.schemaVersion = 1;
+    opts.schemaVersion = 2;
     return {
+        version: opts.schemaVersion,
         options: {
             compression: [
                 {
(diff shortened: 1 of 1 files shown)


mods/key_value.js
```diff
@@ -126,6 +127,9 @@ function returnRevision(req) {
                 etag: rbUtil.makeETag(row.rev, row.tid),
                 'content-type': row['content-type']
             };
```
Pchelolo: This is still needed for backwards-compatibility with already stored data.

mods/key_value.js
```diff
@@ -31,7 +31,7 @@ KVBucket.prototype.getBucketInfo = function(restbase, req, options) {
 };
 
 KVBucket.prototype.makeSchema = function(opts) {
-    opts.schemaVersion = 1;
+    opts.schemaVersion = 2;
```
d00rman: You also need to add `version: opts.schemaVersion` to the returned object, otherwise the version is not propagated into the *table* module.

Pchelolo: @d00rman done

PR #398, opened 2015-11-04
Parsoid: Bug fix, sections request returns full content

Pchelolo:
In case `no-cache` header is provided, but new render doesn't differ from the old render, a request for sections returns full content. The bug was that we assumed we don't need to call `wrapContentRequest` in this case, because we don't need to check for revision access, but `wrapContentRequest` is also responsible for stripping-out the sections from full content.

### DIFF
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -118,10 +118,13 @@ PSP.purgeCaches = function(domain, title) {
  * the revision should be denied
  *
  * @param restbase RESTBase the Restbase router object
- * @param req Object the user request
- * @param promise Promise the promise object to wrap
+ * @param {Object} req the user request
+ * @param {Object} promise the promise object to wrap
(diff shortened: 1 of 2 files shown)


Pchelolo:
cc @wikimedia/services

mods/parsoid.js
```diff
@@ -264,7 +269,7 @@ function sameHtml(a, b) {
     return normalizeHtml(a) === normalizeHtml(b);
 }
 
-PSP.generateAndSave = function(restbase, req, format, currentContentRes) {
+PSP.generateAndSave = function(restbase, req, format, currentContentRes, currentContentTid) {
```
d00rman: Why is `currentContentTid` needed? We can get it from `req.params.tid`, can't we?

Pchelolo: ah, indeed.

PR #399, opened 2015-11-05
Validation: Validate response content-type

Pchelolo:
Now we can validate response content-type agains the `produces` list in the spec. In case on invalid return type, the request is repeated with `cache-control: no-cache` header.

Some points:
- Now only parsoid html version was updated, so we only have outdated content for html in storage, so validation was enabled only there
- Parsing a content-type is a pretty slow operation. Given that we don't have too many different types, I've made a `contentTypeCache` to speed things up a little bit.
- In case the request already had `no-cache` header, we just log a warning and return response as is.

Bug: https://phabricator.wikimedia.org/T116333


test/features/parsoid/ondemand/ondemand.js
```diff
@@ -255,4 +257,46 @@ describe('on-demand generation of html and data-parsoid', function() {
...
+            uri: pageUrl + '/html/User%3APchelolo%2FValidationTest'
+        })
+        .then(function(res) {
+            assert.deepEqual(res.status, 200);
+            var CT = contentType.parse(res.headers['content-type']);
+            assert.deepEqual(CT.type, 'text/html');
+            assert.deepEqual(CT.parameters.profile, 'mediawiki.org/specs/html/1.1.0');
+        })
+        .then(function() {api.done(); })
+        .finally(function() { nock.cleanAll();  });;
```
d00rman: Extra spaces, double semicolons.

lib/rbUtil.js
```diff
@@ -310,5 +311,41 @@ rbUtil.normalizeTitle = function(title) {
...
+ * @returns {boolean} whether the check succeed
+ */
+rbUtil.contentTypeMatch = function(res, allowedTypes) {
+    var resContentTypeString = res.headers && res.headers['content-type'];
+    if (resContentTypeString) {
+        var resContentType = rbUtil._parseContentType(resContentTypeString);
+        var expectedTypeVariants = allowedTypes.filter(function(typeString) {
+            return typeString.indexOf(resContentType.type) === 0;
+        })
+        .map(rbUtil._parseContentType);
```
d00rman: Could move it one line above after `})`.

Pchelolo:
BTW, we should forget, that we have quite a lot of html content with old content-type version, so deploying this would produce some additional load on the cluster.

Pchelolo:
Rebased on the current master

d00rman:
This should probably be moved over to hyperswitch now.

gwicke:
This looks a lot like a [filter](https://phabricator.wikimedia.org/T127132).

Pchelolo:
This was implemented by another PR

PR #400, opened 2015-11-06
Pageviews: Use DTCS for the article flat column family

d00rman:


d00rman:
cc @wikimedia/services @wikimedia/analytics

jobar:
LGTM !

Pchelolo:
LGTM, but we also need to update all the tables we've already switched to DTCS, because in @gwicke PR https://github.com/wikimedia/restbase-mod-table-cassandra/pull/162 the LCS is the default, so on the next deploy it will migrate all tables without hints to LCS

d00rman:
Yes, @Pchelolo , but we cannot do it here and now, there's too much data in prod, it would cause havoc. We need to manually switch them all and then put them on DTCS in the code

Pchelolo:
@d00rman I mean we cannot deploy RB until we do it, because as soon as we deploy, the ones currently switched will migrate back.

gwicke:
@Pchelolo, you are right that we need to avoid such a migration in prod. One way we can do that is to set the default options to match the new default in the schema validation / normalization pass. Since this is applied to stored schemas as well, it would effectively disable migrations on existing tables, until the options are explicitly changed to include a non-default 'updates' section.

d00rman:
@Pchelolo @gwicke I think we should just wait for the compaction to be over, then add DTCS to our tables and then deploy. Changing the default does not seem like the best idea at this point to me.

d00rman:
Upgrade works in Cassandra, and Travis seems not to work today. This is needed by AQS (currently down), so merging.

PR #401, opened 2015-11-09
Text Extracts: Implement a /page/summary/{title} endpoint

Pchelolo:
After a discussion in the ticket, it's been decided that the `summary` endpoint would provide the most value for the client. The summary would contain an article timestamp, first 5 sentences on the article text, and a 320px wide thumbnail.

The content is invalidated, when Parsoid detects that there's a new render with different HTML. It's probably possible to make more fine-granted invalidation, but we need dependency tracking for that. Please note, that current invalidation code is temporary, until we set up change propagation.

Sample response:
```json
{
  "title": "Foobar",
  "extract": "The terms foobar (/ˈfuːbɑr/), fubar, or foo, bar, baz...",
  "thumbnail": {
    "source": "http://upload.wikimedia.org/wikipedia/en/thumb/7/73/Smokeycover.jpg/229px-Smokeycover.jpg",
    "width": 229,
    "height": 320
  },
  "timestamp": "2015-10-23T19:06:48Z"
}
```

Bug: https://phabricator.wikimedia.org/T117082

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -166,8 +166,6 @@ templates:
                       request:
                         method: get
                         uri: /{domain}/sys/page_revisions/page/{title}
-                        headers:
-                          cache-control: '{cache-control}'
                   - process_request:
                       request:
                         method: get
@@ -182,9 +180,9 @@ templates:
                         uri: /{domain}/sys/action/query
                         body:
                           prop: 'extracts|pageimages'
-                          redirects: true
+                          redirects: ''
                           exsentences: 5
-                          explaintext: 'true'
+                          explaintext: ''
                           piprop: 'thumbnail'
                           pithumbsize: 320
                           titles: '{$.request.body.title}'
@@ -194,7 +192,7 @@ templates:
                         uri: /{domain}/sys/key_value/summary/{$.request.body.title}
                         headers:
                           content-type: application/json
-                          etag: '{$.request.body.rev}/{$.request.body.tid}'
+                          etag: '"{$.request.body.rev}/{$.request.body.tid}"'
                         body:
                           title: '{$.get_extracts.body.items[0].title}'
                           extract: '{$.get_extracts.body.items[0].extract}'
@@ -205,7 +203,7 @@ templates:
                         status: 200
                         headers:
                           content-type: application/json
-                          etag: '{$.request.body.rev}/{$.request.body.tid}'
+                          etag: '"{$.request.body.rev}/{$.request.body.tid}"'
                         body:
                           title: '{$.get_extracts.body.items[0].title}'
                           extract: '{$.get_extracts.body.items[0].extract}'
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -40,6 +40,7 @@ templates:
     x-host-basePath: /api/rest_v1
     x-subspecs:
       - mediawiki/v1/content
+      - mediawiki/v1/summary
       - mediawiki/v1/graphoid
       - mediawiki/v1/mobileapps
       - media/v1/mathoid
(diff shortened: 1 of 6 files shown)


Pchelolo:
Rebased on master, may be will help with travis

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,91 @@ templates:
...
+          paths:
+            /request/{title}:
+              get:
+                x-setup-handler:
+                  - init:
+                      method: 'put'
+                      uri: /{domain}/sys/key_value/summary
+                      body:
+                        valueType: 'json'
+                        version: 1
```
d00rman: Nit: not needed, `key_value` overrides it.

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,91 @@ templates:
...
+            /handler/no-cache:
+              get:
+                x-request-handler:
+                  - get_extracts:
+                      request:
+                        method: post
+                        uri: /{domain}/sys/action/query
+                        body:
+                          prop: 'extracts|pageimages'
+                          redirects: 'true'
```
d00rman: Why sending it as a string? Bool is no good?

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,91 @@ templates:
...
+                          exsentences: 5
+                          explaintext: 'true'
+                          piprop: 'thumbnail'
+                          pithumbsize: 320
+                          titles: '{$.request.body.title}'
```
d00rman: The `/handler/no-cache` route has no parameter named `title`. This should be taken from the body, also because the title supplied by the user might be a redirect.

d00rman: Duh, disregard, it's already used.

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,91 @@ templates:
...
+                          title: '{$.get_extracts.body.items[0].title}'
+                          extract: '{$.get_extracts.body.items[0].extract}'
+                          thumbnail: '{$$.default($.get_extracts.body.items[0].thumbnail, null)}'
+                          timestamp: '{$.request.body.timestamp}'
+                  - return_result:
+                      return:
+                        status: 200
+                        headers:
+                          content-type: application/json
+                          etag: '{$.request.body.rev}/{$.request.body.tid}'
```
d00rman: We usually surround ETags with double quotes. But since key_value returns an ETag itself, it'd be good to use that one.

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,91 @@ templates:
...
+                      request:
+                        method: get
+                        uri: /{domain}/sys/key_value/summary/{$.request.body.title}
+                      return_if:
+                        status: '2xx'
+                      return:
+                        status: 200
+                        headers:
+                          content-type: application/json
+                          etag: '{$.request.body.rev}/{$.request.body.tid}'
```
d00rman: Idem

mods/parsoid.js
```diff
@@ -338,6 +338,20 @@ PSP.generateAndSave = function(restbase, req, format, currentContentRes) {
             return self.wrapContentReq(restbase, req,
                 P.resolve(currentContentRes), format, rp.tid, true);
         } else {
+            // Temporary code: need to invalidate summary when a render changes.
```
d00rman: Maybe put a comment header that is more visible? Something like:

```js
/*** TEMP TEMP TEMP ***/
```

config.example.wikimedia.yaml
Pchelolo: Actually, `key_value` etag is not as good as this manually crafted etag. key_value doesn't have a notion of a revision, so it's etag is `undefined/<tid>`

d00rman: Yes, I know, I'm working on a patch to turn that into `0/<tid>` which isn't ideal either.

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,91 @@ templates:
...
+                          explaintext: 'true'
+                          piprop: 'thumbnail'
+                          pithumbsize: 320
+                          titles: '{$.request.body.title}'
+                  - store:
+                      request:
+                        method: put
+                        uri: /{domain}/sys/key_value/summary/{$.request.body.title}
+                        headers:
+                          content-type: application/json
```
d00rman: Since the ETag will be custom, could be good to store that as well now that key_value can store all of the headers

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,91 @@ templates:
...
+                x-request-handler:
+                  - check_storage:
+                      request:
+                        method: get
+                        uri: /{domain}/sys/key_value/summary/{$.request.body.title}
+                      return_if:
+                        status: '2xx'
+                      return:
+                        status: 200
+                        headers:
```
d00rman: If all of the necessary headers end up in key_value, we can simply return them

Pchelolo:
@d00rman I've updated the PR according to your comments, and rebased.

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,85 @@ templates:
...
+                      uri: /{domain}/sys/key_value/summary
+                      body:
+                        valueType: 'json'
+                x-request-handler:
+                  - get_revision:
+                      request:
+                        method: get
+                        uri: /{domain}/sys/page_revisions/page/{title}
+                        headers:
+                          cache-control: '{cache-control}'
```
d00rman: In 99% of the cases this call will come after the content has been refreshed, so we will have the newest rev info in storage already. Hence, I'm not sure it's worth supplying the `cache-control` header here and induce a superfluous call to the MW API.

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,85 @@ templates:
...
+                          piprop: 'thumbnail'
+                          pithumbsize: 320
+                          titles: '{$.request.body.title}'
+                  - store:
+                      request:
+                        method: put
+                        uri: /{domain}/sys/key_value/summary/{$.request.body.title}
+                        headers:
+                          content-type: application/json
+                          etag: '{$.request.body.rev}/{$.request.body.tid}'
```
d00rman: Double quotes around the etag, please, for consistency.

specs/mediawiki/v1/summary.yaml
```diff
@@ -0,0 +1,110 @@
...
+      parameters:
+        - name: title
+          in: path
+          description: The page title.
+          type: string
+          required: true
+      responses:
+        '200':
+          description: >
+            The summary for the given page
```
d00rman: Nit: no need to go into a separate row, it's short enough not to make jscs complain :)

specs/mediawiki/v1/summary.yaml
```diff
@@ -0,0 +1,110 @@
...
+          source:
+            type: string
+            description: Thumbnail image URI
+          width:
+            type: integer
+            description: Thumbnail width
+          height:
+            type: integer
+            description: Thumnail height
+        required: ['source', 'width', 'height']
```
d00rman: Is this supported in swagger as well? I thought this was jsonschema-only.

Pchelolo: @d00rman This part is a JSON-schema :)

config.example.wikimedia.yaml
```diff
@@ -148,6 +149,83 @@ templates:
...
+                x-request-handler:
+                  - get_extracts:
+                      request:
+                        method: post
+                        uri: /{domain}/sys/action/query
+                        body:
+                          prop: 'extracts|pageimages'
+                          redirects: true
+                          exsentences: 5
+                          explaintext: 'true'
```
d00rman: Nit: turn into a Boolean as well (applies to config.test too, ofc)

Pchelolo: @d00rman actually booleans do not work, so replaced with `''`

specs/mediawiki/v1/summary.yaml
gwicke: Swagger *uses* JSON schema. That's partly why I pushed so much for using JSON schema for EventBus, as well.

d00rman:

PR #403, opened 2015-11-09
Travis: Install custom cassandra version

Pchelolo:
I've contacted travis support and asked how can I use the newer version of cassandra in travis. 
This PR is created to test their suggestions. 

The first suggestion was to try setting `group: dev` in the beginning of the travis.yam script, however, it doesn't work - neither in container-based environment, nor with sudo. 

So, this PR is pulling a custom cassandra version, installing it and running the tests against it. We loose nice container-based environment, but it's better not to special-case tests as much as possible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,10 @@ sudo: required
 
 before_install:
   - sudo rm -rf /var/lib/cassandra/*
-  - wget https://archive.apache.org/dist/cassandra/2.1.8/apache-cassandra-2.1.8-bin.tar.gz
-  - tar -xzf apache-cassandra-2.1.8-bin.tar.gz
-  - sudo sh apache-cassandra-2.1.8/bin/cassandra > /dev/null
+  - sudo mkdir /var/lib/cassandra-code
+  - sudo wget -O /var/lib/cassandra-code/cassandra.tar.gz https://archive.apache.org/dist/cassandra/2.1.8/apache-cassandra-2.1.8-bin.tar.gz
+  - sudo tar -xzf /var/lib/cassandra-code/cassandra.tar.gz -C /var/lib/cassandra-code/ --strip-components 1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,10 +4,19 @@ node_js:
...
 
-sudo: false
+sudo: required
 
-services:
-    - cassandra
+before_install:
+  - sudo rm -rf /var/lib/cassandra/*
+  - sudo wget -O /var/lib/cassandra/cassandra.tar.gz https://archive.apache.org/dist/cassandra/2.1.8/apache-cassandra-2.1.8-bin.tar.gz
+  - sudo tar -xzf /var/lib/cassandra/cassandra.tar.gz -C /var/lib/cassandra
```
d00rman: `/var/lib/cassandra` is the default path where Cassandra is putting its data files, so perhaps a different location would be more appropriate? `/var/lib/cassandra-code` ?

Pchelolo:
Tests seem to pass, but a single travis run takes around 10 minutes now...

d00rman:
> Tests seem to pass, but a single travis run takes around 10 minutes now...

Yes, I really think we should disable DTCS for the time being in Travis tests as you suggested earlier today.

Pchelolo:
@d00rman I agree, 10 minutes is too much. Given random test failures it's going to be 20-30 minutes - this is unacceptable. 

This PR was mostly to test suggestions that travis support team gave me on possible ways to resolve this. Looks like the custom cassandra installation is the only working option, so if no-one objects, I'm going to try the 'disable DTCS' path tomorrow morning. @gwicke @eevans what are your thoughts?

gwicke:
@Pchelolo, could we perhaps detect the cassandra version & make the behavior dependent on that? This would be useful for small users as well.

Pchelolo:
Closing the PR as we've decided to take a different path to resolve this.

PR #404, opened 2015-11-10
Support stashing of new page content & empty bodies

gwicke:
- Accept requests with an empty body.
- Fix a bug in rbUtil.cloneRequest() that replaced an empty-string body with
  null.
- Accept stash requests without a version, or with a fake '0' version.

See also: https://phabricator.wikimedia.org/T118070

### DIFF
--- a/lib/rbUtil.js
+++ b/lib/rbUtil.js
@@ -221,7 +221,7 @@ rbUtil.cloneRequest = function(req) {
         method: req.method || 'get',
         headers: req.headers || {},
         query: req.query || {},
-        body: req.body || null,
+        body: req.body !== undefined ? req.body : null,
         params: req.params || {}
     };
 };
--- a/mods/parsoid.js
+++ b/mods/parsoid.js
@@ -709,7 +709,7 @@ PSP.callParsoidTransform = function callParsoidTransform(restbase, req, from, to
         // Fake title to avoid Parsoid error: <400/No title or wikitext was provided>
         parsoidExtras.push('Main_Page');
     }
-    if (rp.revision) {
+    if (rp.revision && rp.revision !== '0') {
         parsoidExtras.push(rp.revision);
     }
     var parsoidExtraPath = parsoidExtras.map(encodeURIComponent).join('/');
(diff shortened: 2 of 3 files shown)


Pchelolo:
[This](https://github.com/wikimedia/restbase/blob/master/specs/mediawiki/v1/content.yaml#L962-L963) doc worths updating now.

mods/parsoid.js
```diff
@@ -784,7 +784,8 @@ PSP.makeTransform = function(from, to) {
 
     return function(restbase, req) {
         var rp = req.params;
-        if (!req.body || !req.body[from]) {
+        if ((!req.body && req.body !== '')
```
Pchelolo: I'm not sure I'm right, but with spec-based request validator in place we might me able to delete this check completely.

lib/rbUtil.js
```diff
@@ -221,7 +221,7 @@ rbUtil.cloneRequest = function(req) {
         method: req.method || 'get',
         headers: req.headers || {},
         query: req.query || {},
-        body: req.body || null,
+        body: req.body || (req.body === '' ? '' : null),
```
d00rman: Nit: we actually want to keep whatever is supplied, if it's defined. So, perhaps it would be better to phrase this as:

```js
body: req.body !== undefined ? req.body : null,
```

mods/parsoid.js
d00rman: Actually, will spec-based checking allow us to supply an empty body at all, @Pchelolo ?

Pchelolo: @d00rman Empty string is allowed as a value for the string parameter. We might want to disallow that, and add support for swaggers custom `allowEmptyValue` parameter, but later, there's already a pending validation PR that needs to be deployed first.

As for `body === undefined` - here it's impossible, as `body[from]` is a required parameter, so body object is also required.

gwicke: Will the spec-based validation also trigger for internal requests?

gwicke: In any case, lets not block an urgent bug fix on this?

Pchelolo: @gwicke ah, actually, no.

But we could add that by adding a `parameters` section to the internal endpoints in the spec.

But anyway, agreed, let's not block. I'll try to play with spec-based thing tomorrow

gwicke:
I addressed the review comments, so this should be ready. Since the underlying issue is still affecting VE users in production, it would be good to get this out to production ASAP.

/cc @wikimedia/services

eevans:
LGTM

PR #409, opened 2015-11-12
key_value mod: Allow users to supply a retention policy

d00rman:
Users of `key_value` may now specify a retention policy to be applied to the underlying bucket table using the `retention_policy` optional stanza. If that's not specified, the *latest* policy is used.

Also included some minor fixes and improvements:
- respect the TID if given for a PUT request
- remove superfluous revision references
- respect the user-supplied ETag header if given

### DIFF
--- a/mods/key_value.js
+++ b/mods/key_value.js
@@ -125,7 +125,7 @@ function returnRevision(req) {
         if (dbResult.body && dbResult.body.items && dbResult.body.items.length) {
             var row = dbResult.body.items[0];
             var headers = {
-                etag: rbUtil.makeETag(row.rev, row.tid),
+                etag: rbUtil.makeETag('0', row.tid),
                 'content-type': row['content-type']
             };
             if (row.headers) {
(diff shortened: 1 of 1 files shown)


d00rman:
cc @wikimedia/services

mods/key_value.js
```diff
@@ -255,7 +266,8 @@ KVBucket.prototype.putRevision = function(restbase, req) {
                 key: rp.key,
                 tid: tid,
                 value: req.body,
-                headers: req.headers
+                headers: req.headers,
+                'content-type': req.headers && req.headers['content-type']
```
Pchelolo: On the previous key_value patch I've intentionally removed this line. This data is already stored in `headers`, so what's the point to duplicate it?

d00rman: Because since we cannot remove the field in the DB, let's populate it - maybe it'll be made useful in a next iteration (categorisation of values, e.g.)

Pchelolo: @d00rman makes sense

mods/key_value.js
```diff
@@ -31,7 +31,7 @@ KVBucket.prototype.getBucketInfo = function(restbase, req, options) {
 };
 
 KVBucket.prototype.makeSchema = function(opts) {
-    opts.schemaVersion = 2;
+    opts.schemaVersion = 3;
     return {
         version: opts.schemaVersion,
         options: {
```
gwicke: We could also add the 'timeseries' update pattern while we are at it.

d00rman: Yeah, wanted to do that but managed to forget it :P Will do in a follow-up

PR #414, opened 2015-11-14
Clean up documentation and small format fix

milimetric:
We would love this to go out with the RESTBase deploy on Monday, since
we're planning on announcing the API then.  I'm a lot happier with the
new documentation and the only code change was to go from YYYYMMDDHH to
YYYYMMDD for the per-article endpoint.  This was needed since we dropped
hourly resolution.  I updated the tests to reflect the change.

### DIFF
--- a/mods/pageviews.js
+++ b/mods/pageviews.js
@@ -174,36 +174,48 @@ var stripOrgFromProject = function(rp) {
 };
 
 
-var validateTimestamp = function(timestamp) {
-    if (!/^[0-9]{10}$/.test(timestamp)) {
+var validateTimestamp = function(timestamp, opts) {
+    opts = opts || {};
+
+    if (timestamp && timestamp.length > 10) {
         return false;
     }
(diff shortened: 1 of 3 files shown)


mods/pageviews.js
```diff
@@ -251,6 +251,12 @@ var validateYearMonthDay = function(rp) {
 PJVS.prototype.pageviewsForArticleFlat = function(restbase, req) {
     var rp = req.params;
 
+    // dates are passed in as YYYYMMDD but we need the HH to match the loaded data
+    // which was originally planned at hourly resolution
+    if (rp.start && rp.end) {
```
d00rman: Instead of working around the fact that you dropped `HH` from the parameters, you should fix the validation logic, since it will still report that `HH` is needed, which will be pretty confusing for users.

milimetric: That's quite fair.  Fixing now.

PR #415, opened 2015-11-14
Spec modules

gwicke:
This patch adds support for loading specs as a special kind of module. Similar
to @mobrovac's PR #413, this makes our config more modular and less verbose.
As with other modules, spec modules can be parametrized with options.

Example config stanza:

```yaml
    /{module:graphoid}:
      x-modules:
        - path: graphoid_test
          type: spec
          options:
            host: http://graphoid.wikimedia.org
            header_whitelist:
              - content-type
              - x-foobar
```

The `options` specified for the module can be referenced in request templates
under `$$.conf`. Example:

```yaml
uri: '{+$$.options.host}/{domain}/...'
headers: '{$$.filter($.request.headers, $$.options.header_whitelist)}'
```
Note that we are passing an array to `$$.filter`, which works as we are
passing around references.

With #417, this can be more cleanly written as:
```yaml
uri: '{{options.host}}/{domain}/...'
headers: '{{ filter(req.headers, options.header_whitelist) }}'
```
Also, `filter` is actually implemented.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "preq": "^0.4.4",
     "restbase-mod-table-cassandra": "^0.8.7",
     "service-runner": "^0.3.0",
-    "swagger-router": "git+https://github.com/gwicke/swagger-router.git#template_fixes",
+    "swagger-router": "^0.2.7",
     "swagger-ui": "git+https://github.com/wikimedia/swagger-ui#master",
     "tassembly": "^0.2.0",
     "json-stable-stringify": "git+https://github.com/wikimedia/json-stable-stringify#master",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "preq": "^0.4.4",
     "restbase-mod-table-cassandra": "^0.8.7",
     "service-runner": "^0.3.0",
-    "swagger-router": "^0.2.5",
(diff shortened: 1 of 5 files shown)


d00rman:
I really like this. It fixes the drawbacks of #413 and enhances it. But, on the other hand, it focuses mostly on the request template. One thing that I really like about #413 is that by using swig we get the added benefit of using blocks, conditionals and loops directly in the spec YAML. So how about having both?

d00rman:
While I like the integration of the configuration into the templating system, I'm quite sure it's not needed. The configuration options are static, so can safely be evaluated only once on start-up. Here, they are evaluated on every request. Also, I find the syntax `{+$$.conf.blah}` rather strange.

package.json
```diff
@@ -41,7 +41,7 @@
     "preq": "^0.4.4",
     "restbase-mod-table-cassandra": "^0.8.7",
     "service-runner": "^0.3.0",
-    "swagger-router": "^0.2.5",
+    "swagger-router": "git+https://github.com/gwicke/swagger-router.git#template_fixes",
```
d00rman: This can now be replaced with `^0.2.7` since that version contains the code pointed to here.

gwicke: Done, but we'll probably have to increment once more.

gwicke:
See also: #417

PR #417, opened 2015-11-16
Spec cleanup

gwicke:
Implement ideas from https://phabricator.wikimedia.org/T118700.

- [x] Un-prefixed dotted path references and function calls.
- [x] `{{foo}}` syntax in all attributes.
- [x] wrap responses in `responses` object
- [x] shorten / alias `request` to `req`

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -35,15 +35,7 @@ templates:
       license:
         name: Apache2
         url: http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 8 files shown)


gwicke:
See also #419 for module option templating support.

d00rman:
> shorten / alias request to req

We agreed we'd continue using `request` so this alias should be removed, IMO.

gwicke:
> We agreed we'd continue using request so this alias should be removed, IMO.

It is removed in #421.

d00rman:

PR #418, opened 2015-11-16
Fix nock to 2.17.0

gwicke:
2.18 seems to have [an ugly bug](https://github.com/pgte/nock/issues/402),
which causes requests to hosts that were previously mocked to time out on node
0.10.

Bug: https://phabricator.wikimedia.org/T118650

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: node_js
 node_js:
+    # Production runs 0.10.29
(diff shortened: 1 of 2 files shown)


gwicke:
@d00rman, added 0.10.29 to the mix in b820f1d.

d00rman:
yup. waiting on travis and we're GTG.

PR #419, opened 2015-11-17
Module option templating

gwicke:
This patch adds the ability to template module options using the familiar
request template syntax. This lets us pass down global options like header
whitelists into individual modules, reducing the need for repetition and copy
& paste.

To accommodate passing in templates as well, an additional 'templates'
property is passed into module constructors verbatim, and the one case where
we pass in an apiRequest template is migrated to it.

In this current implementation, we cheat a little bit by only supporting a
single nesting level, with all option templates being directly evaluated
against the service config. We can consider supporting nested scopes as well,
but will need to clean up the way information is passed down the router trees
first. However, in practice a single module inclusion level is all we have
needed so far, and keeping it that way may be good for our sanity. Lets cross
that bridge when we get to it.

One thing that I noticed while working on this patch is that we are doing some
fairly gross and brittle spec referecing & manipulation in our tests. This
patch should give us tools to get rid of this by moving those variable pieces
into a proper service config.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -122,7 +122,7 @@ templates:
         x-modules:
           - name: action
             type: file
-            options:
+            templates:
               apiRequest:
                 method: post
                 uri: http://{domain}/w/api.php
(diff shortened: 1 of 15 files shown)


d00rman:

PR #420, opened 2015-11-18
MobileApps: Switch to key_value and spec modules

d00rman:
This PR brings many changes to the MobileApps endpoints in RESTBase. Firstly, this a disruptive change for the public API:
- `/page/mobile-html` ceases to exist
- `-html` is being dropped from route names of the other endpoints

Secondly, we now store only the latest render for each title, and do so only for the `mobile-sections` endpoint. Two *key_value* buckets are used for that: `mobileapps.lead` and `mobileapps.remaining`, each storing a part of the response, respectively so that they can be served efficiently, being more popular then the main `mobile-sections` route.

Thirdly, with regards to refreshing the content, only the `mobile-sections` route causes content to be regenerated; sending a `no-cache` request to `mobile-sections-lead` and `mobile-sections-remaining` will cause RESTBase to simply return the latest stored content.

Finally, all of the paths have been simplified and moved out of the configuration file into its own spec module.

This PR also brings some improvements and fixes. Namely:
- the *key_value* module now accepts compression and updates settings
- some `Content-Type` checks in tests have the additional (now-required) `charset` component
- in the handler templating code, added `return_if` to the list of components determining if a request is simple or not
- for route loading, use `hasOwnProperty()` to check if a function for a route/method has been defined

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -3,7 +3,6 @@
 info:
   name: restbase
 
-
 templates:
 
   wmf-content-1.0.0: &wp/content/1.0.0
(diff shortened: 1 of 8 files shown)


d00rman:
a bit late, but cc @wikimedia/services

PR #421, opened 2015-11-19
Handler specs 2.0 & simplified textextracts

gwicke:
- Add a `response` stanza that makes it possible to template response values
  without returning. This is useful to share such templated objects between
  multiple later stanzas, for example to save & return the same content.

  In combination, these improvements make it feasible to define the public API
  spec as well as the backend handlers in a single spec module. The
  textextracts end point illustrates this. This will avoid unnecessary
  indirections in /sys, and will also reduce internal routing overhead.

- A quite significant refactor / rewrite of the handlerTemplate implementation
  introduces a two-phase execution model, which lets us combine parallel
  execution with per-request response templating. We now also support
  returning from one of the parallel branches.

  The new implementation reduces per-request work by minimizing allocations
  and expensive per-request operations. Static closures are used wherever
  possible. Simple wrk runs indicate that even the simple Parsoid HTML end
  point (which only performs one templated sub-request) is sped up by about
  6% (900/s to 950/s). The impact on multi-part handlers will be more
  noticeable.


gwicke:
Todo for textextracts:

- [ ] Check reasoning behind slightly strange use of the revision info end point.
- [ ] Rewrite thumb urls from http to https, by adding a small JS module that injects a small helper into the template globals namespace. To do this, the module can return a 'globals' property in its exports.

mods/restbase/summary_v1.yaml
```diff
@@ -0,0 +1,171 @@
...
+                exsentences: 5
+                explaintext: ''
+                piprop: 'thumbnail'
+                pithumbsize: 320
+                titles: '{{rev_info.body.title}}'
+            response:
+              # Define the response to save & return.
+              headers:
+                content-type: application/json
+                etag: '"{{rev_info.body.rev}}/{{request.params.tid}}"'
```
d00rman: `request.params.tid` does not exists from what I can see.

config.test.yaml
```diff
@@ -159,6 +163,83 @@ templates:
                   options:
                     host: http://appservice.wmflabs.org
 
+      /{module:summary}:
```
d00rman: Why is this here? Looks like this is doing the same thing as `mods/restbase/summary_v1` so why is it needed?

config.test.yaml
```diff
@@ -64,6 +54,27 @@ templates:
...
 
+    x-modules:
+      - name: mobileapps-public
+        path: specs/mediawiki/v1/mobileapps
+        type: spec
+        options:
+          host: http://appservice.wmflabs.org
+
+      - name: graphoid
+        path: mods/restbase/graphoid_v1
```
d00rman: Why `restbase` and not `wikimedia`? Also, IMHO it'd be better to have `mods/wikimedia/v1/graphoid` (idem for summary below).

lib/handlerTemplate.js
```diff
@@ -164,151 +128,337 @@ function validateSpec(spec) {
...
+        }
+
+        var catchPred;
+        if (info.spec.catch) {
+            catchPred = compileCatchFunction(info.spec.catch);
+        }
+
+        var shouldReturn;
+        // Important: `return_if` takes precedence over return, so that a
+        // `return` with `return_if` present behaves like `response`.
```
d00rman: Hm, so if `return_if` does not match, the next step gets the declared `return` instead of the original response? I find this too magical and I'm not sure we need it.

config.test.yaml
d00rman: Ah, this is for no-cache requests. In that case, the back-end module should handle both and the public endpoint should call it. This portion and the public one share a lot of logic, so copy/pasting seems like a bad idea. Also, this should not go into the config, but live in its own specialised module.

lib/handlerTemplate.js
gwicke: The template is actually not applied either if the condition doesn't evaluate to true. Doing so would cause failures for typical try-storage use cases that catch 404s, which don't provide the information expected in the return template. The lack of clear semantics around this is one of the less-than-ideal consequences of separating return_if from return.

An idea we came up with during our conversation this morning is to leverage our new templating powers to support something like

```yaml
return:
  if: '{{status_match(this_request.status, '2xx')}}'
  status: 200
  body: 
   ...
```
or, without returning,
```yaml
response:
  if: '{{status_match(this_request.status, '2xx')}}'
  status: 200
  body: 
   ...
```

config.test.yaml
gwicke: Yeah, I was thinking about moving to a `/v1/` (or `/rest_v1`) vs `/sys/` distinction as well. Less sure about the `wikimedia` prefix. Perhaps no prefix at all?

config.test.yaml
gwicke: I think this came back in the rebase before my push last night. I swear I had deleted it already. In any case, it was dead code & is now removed.

PR #422, opened 2015-11-20
Allow directly loading JSON

milimetric:
We were loading a JSON string instead of actual JSON, and we were
running into encoding problems.  This change upgrades the table schema
to support JSON directly.  We'll fix the load and backfill as much as
needed.

Bug: T118931

### DIFF
--- a/mods/pageviews.js
+++ b/mods/pageviews.js
@@ -78,15 +78,20 @@ var tableSchemas = {
     },
     tops: {
         table: tables.tops,
-        version: 1,
+        version: 2,
         attributes: {
             project: 'string',
             access: 'string',
             year: 'string',
             month: 'string',
             day: 'string',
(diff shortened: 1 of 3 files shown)


mods/pageviews.js
```diff
@@ -365,7 +370,40 @@ PJVS.prototype.pageviewsForTops = function(restbase, req) {
...
 
-    return dataRequest.then(normalizeResponse);
+    return dataRequest.then(normalizeResponse).then(function(res) {
+        if (res.body.items) {
+            res.body.items.forEach(function(item) {
+                // prefer the articlesJSON column if it's loaded
+                if (item.articlesJSON !== null) {
+                    try {
+                        item.articles = item.articlesJSON.map(function(a) {
+                            a.article = decodeURIComponent(a.article);
```
d00rman: Wouldn't it be better to have the URI decoded before storing it in Cassandra, so that it doesn't need to be decoded on each request?

PR #423, opened 2015-11-20
Several fixes to old & new code

gwicke:


lib/handlerTemplate.js
```diff
@@ -421,14 +421,29 @@ function runStep(steps, i, ctx) {
...
+        modelInit = null;
+    }
+
     // Compile all the parallel execution steps into functions.
     var steps = spec.map(function(stepSpec) {
         return makeStep(stepSpec, options);
     });
 
+
+
```
d00rman: Nit: there's already an empty line here, no need for two extra ones ;)

wikimedia/v1/summary_v1.yaml
```diff
@@ -92,9 +94,9 @@ paths:
               uri: /{domain}/sys/action/query
               body:
                 prop: 'extracts|pageimages'
-                redirects: ''
+                redirects: 'true'
```
d00rman: There shouldn't be a need for quoting the value, PHP accepts booleans just fine, AFAIK.

lib/handlerTemplate.js
```diff
@@ -444,7 +459,16 @@ function createHandler(spec, options) {
             _doReturn: false,
             _spec: spec,
         };
-        Object.assign(ctx.model, options.globals);
+        if (modelInit) {
+            var model = ctx.model;
+            // Don't use Object.assign, as we want to give precedence to the
+            // model.
+            Object.keys(modelInit).forEach(function(key) {
```
d00rman: Isn't this the same as:

```js
model = Object.assign({}, modelInit, model);
```

d00rman:
LGTM, minor nits in-lined.

d00rman:
Perhaps it would also be worth switching the `/{domain}/sys/key_value/summary` bucket to //timeseries// given that records will be tid-ordered.

lib/handlerTemplate.js
gwicke: The difference is that one always copies everything, while the other typically does nothing.

wikimedia/v1/summary_v1.yaml
gwicke: Yeah, good point. Was changing this back & forth while hunting down the mysterious missing thumbnail issue, and ended up on the string. Will change it to a boolean.

d00rman:
Also, [mods/parsoid.js:346](https://github.com/wikimedia/restbase/blob/e9c752ed636b2255c544c5f5e50937ffd3fba40c/mods/parsoid.js#L346) still uses `/sys/summary` to refresh the summary info, but that endpoint doesn't exist any more.

gwicke:
@d00rman, re timeseries: It probably makes sense to do that for key_value in general, as the *range* keys will always be tids.

d00rman:
> @d00rman, re timeseries: It probably makes sense to do that for key_value in general, as the range keys will always be tids.

I though about that but in the PR where I introduced the possibility to specify it didn't opt for it not to introduce extra load in Cassandra with migrations. I think it'd be good to supply it here as this is a new CF to be created.

For this PR, there's still the pending issue of:

> Also, [mods/parsoid.js:346](https://github.com/wikimedia/restbase/blob/e9c752ed636b2255c544c5f5e50937ffd3fba40c/mods/parsoid.js#L346) still uses `/sys/summary` to refresh the summary info, but that endpoint doesn't exist any more.

mods/key_value.js
```diff
@@ -77,7 +77,9 @@ KVBucket.prototype.createBucket = function(restbase, req) {
     if (!Array.isArray(opts.compression)) {
         opts.compression = [opts.compression];
     }
-    opts.updates = opts.updates || undefined;
+    opts.updates = {
```
d00rman: Maybe still allow users to specify their own updates instead of forcing timeseries on them?

```js
opts.updates = opts.updates || { pattern: 'timeseries' }
```

gwicke: Do you see a use case for deviating from timeseries? These timeuuids should always be timestamp-ordered.

PR #424, opened 2015-11-22
Mathoid: Use POST only for formula check, GET for renders

d00rman:
The new textvcinfo endpoint in Mathoid is able to check TeX formulae for correctness and normalised them. The endpoint is rather cheap and quick, so call only this one for POST requests. The normalised response is stored into post storage and hashed. As usual, the hash is returned in the `x-resource-location` header, which can then be used to issue GET requests to `/media/math/render/{format}/{hash}` to obtain an actual render.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -72,10 +72,14 @@ templates:
           # One hour, for now.
           response_cache-control: 'max-age: 3600, s-maxage: 3600'
 
+      - name: mathoid
+        path: specs/media/v1/mathoid
+        type: spec
+        options:
+          host: http://mathoid-tester.wmflabs.org
+
     x-subspecs:
       - mediawiki/v1/content
-      - media/v1/mathoid
-    # - mediawiki/v1/revision-scoring
 
   wmf-sys-1.0.0: &wp/sys/1.0.0
     info:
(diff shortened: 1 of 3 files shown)


d00rman:

physikerwelt:
@d00rman I tried to test this on mathosphere.wmflabs.org but I could not get it to work.
restbase dies while searching for
/vagrant/srv/restbase/specs/mediawiki/v1/graphoid.yaml
which is not in the repo here
https://github.com/d00rman/restbase/tree/mathoid/spec-module/specs/mediawiki/v1

physikerwelt:
Ok. Obviously that was my fault, since restbase uses the config.vagrant.yaml.
But for vagrant it did not work with the config.example.yaml as well. Maybe you can have a look, which config is suitable for vagrant?

d00rman:
@physikerwelt `mathosphere` is set up and working, so you can test this patch there.

physikerwelt:
@d00rman Thank you very much. I hope that the vagrant npm process did not create many new gray hear for you.
The patch works perfactly fine. I tested it on http://math.beta.wmflabs.org:7231/mediawiki-vagrant.wmflabs.org/v1/?doc

physikerwelt:
@d00rman @wikimedia/services is there anything left to be done here?

PR #425, opened 2015-11-24
Spec format & router cleanup

gwicke:
- Removed x-subspec and x-subspecs in favor of x-modules.
- Cleaned up [config.test.yaml](https://github.com/gwicke/restbase/blob/mod_cleanup/config.test.yaml) using the yaml merge operator.
- Made x-modules an object rather than an array to facilitate overriding of individual modules / their configs in derived configs, and added support for inline specification of sub-paths.
- Removed spec manipulation in tests.
- Organized all modules (spec, code or both) into `v1/` and `sys/`.
- Factored out example and production configs into project modules, establishing a config interface that should allow us to update the project modules without breaking puppet & third party configs.


gwicke:
Early heads-up to @wikimedia/services.

gwicke:
Simplified configs:

- [example](https://github.com/gwicke/restbase/blob/mod_cleanup/config.example.yaml)
- [wikimedia](https://github.com/gwicke/restbase/blob/mod_cleanup/config.example.wikimedia.yaml)

gwicke:
@wikimedia/services, this is ready to go.

While this is a major change, most of it is 

but at the same time we don't have a lot of major other changes directly affected by this in flight right now. The main ones would be #365 and #388.

d00rman:
Needs a rebase with careful considerations of changes introduced in #424 and #427

gwicke:
@d00rman, #424 is already integrated. #427 looks trivial, let me rebase.

gwicke:
Rebased on current master.

sys/mathoid.yaml
```diff
@@ -0,0 +1,86 @@
+info:
```
d00rman: This file should not be used at all, all of Mathoid's logic is handled in the public spec-module.

gwicke: It isn't used any more (not referenced in the specs), but let me also make sure that it's deleted.

config.example.wikimedia.yaml
```diff
@@ -1,265 +1,68 @@
...
+            username: cassandra
+            password: cassandra
+            defaultConsistency: one # or 'localQuorum' for production
+            storage_groups:
+              - name: default.group.local
+                domains: /./
+          parsoid:
+            host: http://parsoid-lb.eqiad.wikimedia.org
+          action:
+            apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
```
Pchelolo: I'm not sure I completely understand the new templating changes, but why wouldn't this work?
```yaml
apiUriTemplate: https://{{domain}}/w/api.php
```

gwicke: @Pchelolo: Module options are now also templated, so this would expand {{domain}} prematurely.

Pchelolo: @gwicke ah, ok, I see...

gwicke: I explored several options to get around this:

- Have a separate `templates` module parameter; this turned out to be quite ugly & special case, so I removed it again.
- Add a requestTemplate function, and support passing in a compiled request template function from the config (see https://github.com/wikimedia/swagger-router/pull/32#issuecomment-159684767); this works (and is possible with this patch), but exposes more functionality in the client config than we probably want for normal use cases. This area is also still under some discussion, so it's good to keep our options open by not exposing more than necessary.
- Offer a light-weight way to template only the uri, as that's all we needed to change in practice; that's what this patch uses in the configs.

lib/router.js
```diff
@@ -13,6 +13,42 @@ var Template = swaggerRouter.Template;
...
+
+    this.globals = init.globals || null;
+    this.operations = init.operations || null;
+}
+
+ApiScope.prototype.makeChild = function(overrides) {
+    var newScope = new ApiScope(this);
+    Object.keys(overrides).forEach(function(key) {
+        newScope[key] = overrides[key];
+    });
```
Pchelolo: Isn't that equal to `Object.assign(newScope, overrides)`?

gwicke: Yeah, it is. Will amend.

config.example.wikimedia.yaml
d00rman: > Have a separate templates module parameter; this turned out to be quite ugly & special case, so I removed it again.

I'd like us to keep this. I rely on it in [restbase-mod-queue-kafka](https://github.com/d00rman/restbase-mod-queue-kafka/blob/371af36e12383852e6bcdaa187e1b17d6f52e79b/index.js#L14) to load properly the rules. Without that, the whole rule config section would be extremely bloated.

gwicke: @d00rman, that is indeed a use case worth supporting. My main objection against the top-level `templates:` option is the need to forward it everywhere, which complicates a lot of code paths. This could however be mitigated by nesting it *inside* of options, but giving it special case treatment wrt templating. Example:

```yaml
options:
  templated_option: '{{foo}}'
  templates:
    someTemplate: '{{not expanded}}'

Comment (d00rman) [inline config.example.wikimedia.yaml]: Sure, @gwicke, feels more natural too. Ideally, I'd need to specify the name of the unexpanded key as well (since in my concrete use case, I'd be more legible to have `s/templates/rules/`), but that's tricky as that would have to come from the module itself, not the config passed in to it.

Comment (gwicke) [inline config.example.wikimedia.yaml]: Hmm, this got me thinking again.. Here is another idea:

```yaml
options:
  rules:
    - some_rule:
        topic: sometopic
        exec:handler:
          - first_request:
              method: get
              uri:: /{{domain}}//{bar}{/boo}
              headers:
                cache-control: no-cache
            another_request:
              method: get
              uri:: /{{domain}}//{bar}{/boo}
              headers:
                cache-control: no-cache
```

The magic here is in the colon-separated key suffixes:
```yaml
# Basic expression: :=
key :=: templated_value
# Space is optional
key:=: templated_value
# Return value of 'handler' function, applied to the content of the block:
exec :handler:
  the: block
# Shorthand for `uri :uri:`
uri:: http://{{domain}}/{/foo}
```

What I like about this syntax:

- Templating contexts compose, and use yaml syntax throughout.
- There are no hard-coded special cases for specific properties, but yet there is a convenient shorthand for common properties like `uri`.
- The syntax can accommodate other contexts, as well as possible context stacking & parametrization.

Pchelolo: @gwicke I'm not sure I completely get the idea of how would this new syntax solve the problem of premature expansion? 

Also, as I understand, the `request_handler` in the example is the context property which becomes the context for a nested template expansion, right? In that case `::` should be a shorthand for the global context, not the `:uri:`, because short notation templates in uri are taken from `context.request.params`. And do I understand right that `:=:` is referencing the current global context? 

I have a different idea to solve the premature expansion problem: how about introducing a `template escaping` syntax? Something like (note `\`)
```yaml
template: \{{this will not be executed on first expansion}}
```
On first expansion, the escaped template will not be executed, just the escaped symbol removed. This feel quite natural to me, as it's similar to how we write regex in a sting in javascript. Other possibility is to add an escape symbol as a key suffix to allow escaping the whole nested template, not just a single string template.

projects/example.yaml
```diff
@@ -0,0 +1,79 @@
...
+        # points.
+        - path: v1/content.yaml
+
+  /{api:sys}:
+    x-modules:
+      /table:
+        - type: npm
+          name: restbase-mod-table-sqlite
+          version: 1.0.0
+          type: npm
```
Pchelolo: duplicated `type` key

projects/example.yaml
```diff
@@ -0,0 +1,79 @@
...
+          options: 
+            conf: '{{options.table}}'
+      /key_value:
+        - path: sys/key_value.js
```
Pchelolo: Why an array here? How can we set multiple modules under one path?

config.example.wikimedia.yaml
d00rman: Actually, the modules could supply RB with the part of options which should be exempt from expansion. In the module:

```js
// normal module loading, as before
module.exports = function(options, my_special_part) {
  // init module here ...
  return { /* spec, operations, etc. */ };
}
// tell RB which options stanza should not be expanded
module.exports.skipExpand = 'my_special_part';
```

The config would then feel *natural* while allowing the templates not to be expanded:

```yaml
options:
  key1: '{{ expand_me_right_away }}'
  my_special_part:
    key2: '{{ this_will_stay_a_literal_string }}'
```

gwicke: > Also, as I understand, the request_handler in the example is the context property which becomes the context for a nested template expansion, right?

No, it's a function that gets to control the expansion for the nested scope. It normally returns a function, in this case a request_handler. I'm actually wondering if a syntax that's more suggestive of 'returns function' would be more appropriate.

> I have a different idea to solve the premature expansion problem: how about introducing a template escaping syntax? Something like (note \)

That's not that different from what we already have:
```yaml
template: '{{"this will not be executed on first expansion"}}'
```
The issue with this approach is that it would be very ugly for large templates like the change propagation rules. Each expression would need to be individually escaped.

> Actually, the modules could supply RB with the part of options which should be exempt from expansion.

This would only work if the options are specified inline, but won't work if they are passed down from a higher scope.

Pchelolo: >> Also, as I understand, the request_handler in the example is the context property which becomes the context for a nested template expansion, right?

> No, it's a function that gets to control the expansion for the nested scope. It normally returns a function, in this case a request_handler. I'm actually wondering if a syntax that's more suggestive of 'returns function' would be more appropriate.

Emmmmmm... I am confused even more. Where is the function defined? How do I rewrite this in the new syntax? 
```yaml
options:
  templated_option: '{{options.foo}}'
  templates:
    someTemplate: '{{not expanded}}'
```

gwicke: @Pchelolo, the function is registered as a function factory with the templating environment.

The example would look like:
```yaml
options:
  templated_option :=: options.foo
  templates:
    someTemplate :some_handler_template: not expanded
```

There might be a better way to communicate the 'returns function' aspect, possibly involving parentheses.

```yaml
options:
  templated_option =: options.foo
  templates:
    someTemplate = (some_handler_template): not expanded
```

d00rman: > This would only work if the options are specified inline, but won't work if they are passed down from a higher scope.

As in *a js module loaded from a spec-module which is loaded via the config* ? Hm, possibly. In that case, I'd prefer to stay with the hard-coded `templates` stanza in that case.

gwicke: Another option:

```yaml
options:
  rules:
    - some_rule:
        topic: sometopic
        exec(request_handler):
          - first_request:
              method: get
              uri =(): /{{domain}}//{bar}{/boo}
              headers:
                cache-control: no-cache
            another_request:
              method: get
              uri =(): /{{domain}}//{bar}{/boo}
              headers:
                cache-control =: cache-control
```

Legend:
```yaml
# Basic expression: =
key =: templated_value
# Call 'request_handler' factory, returning a function based on 
# the nested block / request spec:
exec(equest_handler):
  foo: bar
# Shorthand for function creation + call. Pseudo code: `uri = new uri(nested_spec)()`
uri =(): http://{{domain}}/{/foo}
```

projects/example.yaml
gwicke: Fixed.

config.example.wikimedia.yaml
Pchelolo: I have one more idea how to solve the premature expansion problem and avoid escaping/special handling of templates passed together with the options. I've tried to explain it on the meeting, but didn't succeed :) I don't think it's extremely good, but it's a possible option.

So, we have a hierarchy of expansion scopes, just like variable scopes in JavaScript. So, what we could do, is expand what we can, and leave unexpandable references untouched. This would produce another template, which we pass to the lower level. Example:

Let's say we have some template like this: 
```yaml
'{{rb_config.protocol}}://{{rb_config.host}}/{{module_config.domain}}/{{some_req_parameter}}'
```
and we pass it to the module. When we start loading the module, we can expand only `rb_config` references, so in module options this becomes:
```yaml
'https://localhost:7231/{{module_config.baseRoute}}/{{some_req_parameter}}'
```
Let's say the module is a 'project module', which loads up more modules. Now, when we pass this template further down, we have access to the `module_config`, so this gets expanded to
```yaml
'https://localhost:7231/rest_api_module/{{some_req_parameter}}'
```
So now the actual module got a correct template.

Pros:
- No special syntax
- We can mix configs from every level of nested templates
- Not to hard to implement

Cons:
- Hard to understand what gets expanded on which stage. But we are able to track variable scopes in JS, and this is very similar.
- The parameter names must not clash. Also viable to avoid that.

gwicke: Lets continue the discussion about future syntax tweaks at https://github.com/wikimedia/swagger-router/pull/32. I moved the previous comment over there.

gwicke: @Pchelolo: To me the "hard to reason about this" bit is crucial. In order to support this, we'd have to change how templates are being expanded to ignore any errors. That would cost us performance and (more importantly) the ability to notice errors quickly.

lib/router.js
```diff
@@ -155,33 +201,156 @@ Router.prototype._loadModule = function(modDef, config) {
     }
 };
 
+Router.prototype._loadModules = function(node, restBaseModules, scope) {
+    var self = this;
+    if (Array.isArray(restBaseModules)) {
```
d00rman: Perhaps to structure it as:

```js
if (Array.isArray(...)) {
  // complain
} else if (typeof restBaseModules !== 'object') {
  return P.resolve();
}

return P.each(/* ... */);
```

Looks cleaner to me that way.

lib/router.js
```diff
@@ -155,33 +201,156 @@ Router.prototype._loadModule = function(modDef, config) {
...
+        throw new Error('Old style module config detected! '
+                + 'New format expects an object, not an array. '
+                + 'Please update your config.');
+    } else if (typeof restBaseModules === 'object') {
+        return P.each(Object.keys(restBaseModules), function(modulePath) {
+            var pathMods = restBaseModules[modulePath];
+            if (!Array.isArray(pathMods)) {
+                throw new Error('Error in module definition for ' + modulePath + '!\n'
+                        + 'Expected an array, got ' + pathMods);
```
d00rman: Nit: `JSON.stringify(pathMods)` might be more correct.

lib/router.js
```diff
@@ -155,33 +201,156 @@ Router.prototype._loadModule = function(modDef, config) {
...
+            }
+            return P.each(pathMods, function(m) {
+                // Share modules
+                return self._loadModule(m, scope.globals)
```
d00rman: I think now sharing modules has become much more trickier because we'd need to check the options passed to it as well to deduce it's the same mod def, shouldn't we?

lib/router.js
```diff
@@ -155,33 +201,156 @@ Router.prototype._loadModule = function(modDef, config) {
...
+                .then(function(module) {
+                    if (!module) {
+                        throw new Error('Null return when loading module ' + modulePath);
+                    }
+                    if (!module.spec) {
```
d00rman: Should the spec really be required? E.g. the kafka module currently does not export any routes, since they are not at all needed for its functioning.

config.example.wikimedia.yaml
Pchelolo: @gwicke that was just one more option on our table. 

To be honest, I don't like any of the options we have now.. Escaping/special-casing is ugly, your syntax I still do not completely understand:), and my idea is easy to mess up and hard to reason about. 

I've been looking at other templating solutions across languages yesterday, tried to get some ideas, but didn't succeed. I'm looking at more stuff now, but nothing seems to match our case.

lib/router.js
gwicke: The risk is that this would explode. Not everything can be JSON.stringified, but `.toString()` always produces *something*.

lib/router.js
```diff
@@ -155,33 +201,156 @@ Router.prototype._loadModule = function(modDef, config) {
...
+    if (pathspec['x-subspec'] || pathspec['x-subspecs']) {
+        throw new Error('x-subspec and x-subspecs is no longer supported! '
+                + 'Use x-modules instead.');
+    }
+
+    // Load sub-spec
+    var loaderPromise = P.resolve();
+    // XXX: find a better way to detect this?
+    if (pathspec.swagger) {
+        if (pathspec.info && pathspec.info['x-is-api-root']) {
```
d00rman: So in order to have `x-is-api-root` detected, both the `swagger` and the `info` stanzas need to be present? Sounds very fragile.

lib/router.js
```diff
@@ -190,57 +359,21 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
...
-                    subSpecPromise = P.resolve(subSpecOrPath);
-                } else {
-                    subSpecPromise = self._readSpec(subSpecOrPath);
-                }
-                return subSpecPromise
-                .then(function(subSpec) {
-                    return self._handleSwaggerSpec(node, subSpec,
-                            operations, specRoot, prefixPath);
-                });
+            // console.log(specRoot, prefixPath, pathspec);
```
d00rman: Nit: let's remove it, please, there are a lot of these in the code :P

lib/router.js
```diff
@@ -254,182 +387,141 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec,
...
+            });
+        } else {
+            // Share the subtree.
+            var origSubtree = subtree;
+            subtree = subtree.clone();
+            subtree.value = value;
+            // Copy over the remaining value properties.
+            Object.assign(subtree.value, origSubtree.value);
+            subtree.value.path = childScope.specRoot.basePath + childScope.prefixPath;
+            // console.log('cloned',
```
d00rman: Nit: yup, remove :)

v1/mobileapps.yaml
```diff
@@ -2,7 +2,7 @@
 
 swagger: 2.0
 paths:
-  /{module:page}/mobile-sections/{title}:
+  /mobile-sections/{title}:
```
d00rman: Why is `/page/` here entirely removed, but on line 92 it isn't?

v1/summary.yaml
```diff
@@ -124,7 +124,7 @@ paths:
               status: 200
               headers:
                 content-type: '{{extract.headers.content-type}}'
-                etag: '{{extract.headers.content-type}}'
+                etag: '{{extract.headers.etag}}'
```
d00rman: Hehe, nice catch!

lib/router.js
d00rman: Touché.

lib/router.js
gwicke: Agreed and changed.

lib/router.js
gwicke: Yeah, or simply remove module caching. With separate project templates the cache didn't trigger any more anyway.

lib/router.js
gwicke: I just kept this behavior. We could change it, but lets perhaps make that a separate PR?

lib/router.js
gwicke: The swagger version is required in any case. The flag to indicate an api root is pathspec.info['x-is-api-root'], so that's one flag. I tried to do without a flag, but didn't come up with anything clean & reliable.

v1/mobileapps.yaml
gwicke: Ohh, good catch! Fix coming.

PR #428, opened 2015-12-04
Update dependencies

Pchelolo:
Updated dependencies. Problem with `nock` we've had is resolved now, so we can update safely, `ajv` got updated quite significantly, even the format of error messages have changed, so I've updated the tests to match new behaviour.

I've run a couple of benchmarks and there's no performance impact of this change. I've also tested this for a while under simulated load on a local machine both with sqlite and cassandra backends and didn't find any issues. However, this obviously needs more thoughtful testing in staging.


test/features/pagecontent/save_api.js
```diff
@@ -78,7 +78,7 @@ describe('page save api', function() {
             throw new Error('Expected an error, but got status: ' + res.status);
         }, function(err) {
             assert.deepEqual(err.status, 400);
-            assert.deepEqual(err.body.detail, 'data.body.csrf_token is a required property');
+            assert.deepEqual(err.body.detail, "data.body should have required property 'csrf_token'");
```
d00rman: `s/should have/is missing the/`

test/features/router/misc.js
```diff
@@ -111,7 +111,7 @@ describe('router - misc', function() {
             throw new Error('Error should be thrown');
         } catch(e) {
             assert.deepEqual(e.constructor.name, 'HTTPError');
-            assert.deepEqual(e.body.detail, 'data.body.testParam is a required property');
+            assert.deepEqual(e.body.detail, "data.body should have required property 'testParam'");
```
d00rman: `s/should have/is missing the/`

test/features/router/misc.js
```diff
@@ -166,7 +166,7 @@ describe('router - misc', function() {
             throw new Error('Error should be thrown');
         } catch(e) {
             assert.deepEqual(e.constructor.name, 'HTTPError');
-            assert.deepEqual(e.body.detail, 'data.body.field2 is a required property');
+            assert.deepEqual(e.body.detail, "data.body should have required property 'field2'");
```
d00rman: `s/should have/is missing the/`

test/features/router/misc.js
Pchelolo: @d00rman this error message doesn't come from our code, it's produced by `ajv`, so we can't change that.

d00rman:
Let's defer the bluebird update for after the holidays.

test/features/router/misc.js
d00rman: Duh, silly me, it's `assert.deepEqual` :P

PR #429, opened 2015-12-04
Make mathoid responses cacheable

gwicke:
Clients will directly hit restbase for SVG renders of mathematical formulas.
The URL contains a hash of the original formula, and the render does not
change very often. For this reason, we can allow Varnish to cache these SVGs
for extended periods of time.

This patch adds a cache-control header to all responses, and makes the actual
value configurable as a module parameter.

Task: https://phabricator.wikimedia.org/T120445

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -24,6 +24,8 @@ default_project: &default_project
             host: http://graphoid.wikimedia.org
           mathoid:
             host: http://mathoid-tester.wmflabs.org
+            # 10 days Varnish caching, one day client-side
+            cache-control: s-maxage=864000, max-age=86400
           mobileapps:
             host: http://appservice.wmflabs.org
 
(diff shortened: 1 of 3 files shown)


physikerwelt:
:+1: ... I think the cache time is reasonable... but I'm not familar with the effect of the $ changes.

config.example.wikimedia.yaml
```diff
@@ -24,6 +24,8 @@ default_project: &default_project
             host: http://graphoid.wikimedia.org
           mathoid:
             host: http://mathoid-tester.wmflabs.org
+            # 10 days Varnish caching, one day client-side
+            cache-control: s-maxage=864000, max-age=86400
```
d00rman: `config.test.yaml` should have the same change. Also, it'd be nice to add this header's check in the `x-amples` section of `v1/mathoid.yaml`

PR #432, opened 2015-12-07
Add a pageviews_proxy.yaml config & subtree / module sharing fixes

gwicke:
The frontend service needs to proxy requests to the AQS service. In the
production config, this is implemented by defining custom /sys/pageviews entry
points directly in the config. In new-style configs. this is now a module
referenced by `projects/wikimedia.org.yaml`.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -50,6 +50,7 @@ ApiScope.prototype.makeChild = function(overrides) {
 function Router(options) {
     this._options = options || {};
     this._nodes = new Map();
+    this._modules = new Map();
     this.router = new SwaggerRouter();
 }
 
@@ -97,9 +98,17 @@ Router.prototype._buildPath = function route(node, path, value) {
 /**
  * Load and initialize a module
  */
-Router.prototype._loadModule = function(modDef, config) {
+Router.prototype._loadModule = function(modDef, globals) {
 
     var self = this;
+
+    // First, check if we have a copy of this module in the cache, so that we
+    // can share it.
+    var cachedModule = self._modules.get(modDef);
+    if (cachedModule && cachedModule._parentGlobals === globals) {
+        return P.resolve(cachedModule);
+    }
+
     var loadPath;
 
     if (modDef.path && !modDef.type) {
@@ -146,7 +155,7 @@ Router.prototype._loadModule = function(modDef, config) {
         // Protect "templates" property from expansion.
         var templates = modDef.options.templates;
         delete modDef.options.templates;
-        options = new Template(modDef.options).expand(config);
+        options = new Template(modDef.options).expand(globals) || {};
         // Add the original "templates" property back.
         options.templates = templates;
     }
@@ -165,7 +174,10 @@ Router.prototype._loadModule = function(modDef, config) {
             var mod = {
                 spec: spec,
                 globals: { options: options, log: options.log },
+                // Needed to check cache validity.
+                _parentGlobals: globals,
             };
+            self._modules.set(modDef, mod);
             return mod;
         });
     } else {
@@ -188,6 +200,9 @@ Router.prototype._loadModule = function(modDef, config) {
             }
             if (!mod.globals) { mod.globals = {}; }
             mod.globals.log = options.log;
+            // Needed to check cache validity.
+            mod._parentGlobals = globals;
+            self._modules.set(modDef, mod);
             return mod;
         });
     }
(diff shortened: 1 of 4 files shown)


Pchelolo:
Great catch. I think it could also worth to change the tests a little bit [here](https://github.com/wikimedia/restbase/blob/master/test/features/pageviews/pageviews.js#L23-L24). Making `s/1970010100/1970010101` would've made our locally run monitoring tests catch this bug, as now it's not caught because of the clash in the mock data inserted by pageviews module tests and an endpoint `x-ample`

gwicke:
> Another thing is, may be we could add an AQS project to /projects to make analytics setup easier?

It is already kind of moved out to the wikimedia.org project, but pageviews isn't very self-contained. The cleanest way would imho be to parametrize the backend request prefix in the v1 config. In aqs, this can then point to /{domain}/sys/, while in the frontend it can be directly pointed to the backend service, without the detour via /sys/.

gwicke:
Self-merging, so that I can test this in staging.

gwicke:
@wikimedia/services, please have another look at my follow-up change above.

@Pchelolo, agreed on your comment re tests. Will follow up on that tomorrow.

lib/router.js
```diff
@@ -97,9 +98,17 @@ Router.prototype._buildPath = function route(node, path, value) {
...
  */
-Router.prototype._loadModule = function(modDef, config) {
+Router.prototype._loadModule = function(modDef, globals) {
 
     var self = this;
+
+    // First, check if we have a copy of this module in the cache, so that we
+    // can share it.
+    var cachedModule = self._modules.get(modDef);
+    if (cachedModule && cachedModule._parentGlobals === globals) {
```
Pchelolo: @gwicke I've tried to put a `console.log` before the resolve call, and this sharing doesn't work. There's a whole bunch of problems with it:
- We don't specify the module type explicitly, but resolve it from the file extension and then modify the `modDef`. So the get call always uses
```javascript
{ path: 'some_path.js' }
```
as a key, while the set call always uses
```javascript
{ path: 'some_path.js', type: 'file' }
```
so the match is never found.
- The map keys equality are determined by the semantics of a `===` operator, so only references to the exact same object instance would be equal. In our case, the `modDef` objects a different instances with the same keys and values, so even if the object are deepEqual, the match wouldn't be found. We need to `JSON.stringify` the key.
- Yet another problem is with globals equality. `===` doesn't work as a deep equality on objects, so the second part of the condition would be almost always false.

Pchelolo:
@gwicke wrote:
> The cleanest way would imho be to parametrize the backend request prefix in the v1 config. In aqs, this can then point to /{domain}/sys/, while in the frontend it can be directly pointed to the backend service, without the detour via /sys/.

I don't think we can do that, as in our setup we need to bind `pageviews_proxy.yaml` and not `pageviews.js`, and in AQS its wise-versa. And we can't template the module path, can we?

lib/router.js
gwicke: > The map keys equality are determined by the semantics of a === operator, so only references to the exact same object instance would be equal. 

This is a conservative default, which shouldn't reduce sharing opportunities in practice, as the globals object identity would likely differ in that case anyway.

> Yet another problem is with globals equality. === doesn't work as a deep equality on objects, so the second part of the condition would be almost always false.

This is on purpose, as complex objects like functions can't be compared otherwise. Again, I'd say this is mostly academic.

gwicke: > We don't specify the module type explicitly, but resolve it from the file extension and then modify the modDef.

Good catch. Fix is coming.

gwicke:
> I don't think we can do that, as in our setup we need to bind pageviews_proxy.yaml and not pageviews.js, and in AQS its wise-versa. And we can't template the module path, can we?

We wouldn't need the proxy module in sys any more. Instead, we'd pass in the backend url to the v1 config.

gwicke:
> I've tried to put a console.log before the resolve call, and this sharing doesn't work. 

This is because the sharing already happens at a higher level (entire project stanza) in our test configs.

Pchelolo:
>>I've tried to put a console.log before the resolve call, and this sharing doesn't work.

>This is because the sharing already happens at a higher level (entire project stanza) in our test configs.

I'm not sure this is true for the following reasons:
1. I've used the project (example) config
2. We could share, for example, `key_value` modules, defined in `wmf_default` and `wikimedia.org` projects on this level, but it doesn't happen.

PR #434, opened 2015-12-08
Tests. Stage 1. Separate generic framework tests from RB specific tests.

Pchelolo:
As a part of the preparation to creating `RESTBase-The-Framework` we need to separate test specific to `RESTBase-The-Product` from generic framework tests, and, more importantly, achieve good coverage for the framework code with generic-only tests. 

This is a big bulk of work, that requires writing new tests and various fixes in completely unrelated parts of the code, so I'm separating it to multiple steps to avoid giant mega-PR.

This part is about moving tests and improving coverage for a request validator, tree building and handler template.

Also, I've found that validator code skips body parameters coercion in JSON posts in an incorrect way.
1. Decision should be made on the spec `in` parameter, not the `name`, as the name can be arbitrary.
2. We should never try to coerce `in: body` parameter, as it's special and can only be specified as an object, so if we have an `in: body` parameter, we must accept only `application/json`
3. The only place where coercion should be conditionally done/skipped, is for `formData` parameters on endpoints, which allow both `application/json` and `form/urlencoded` content-types.

With new generic tests we have 100% coverage for the validator and ~90% coverage for the handler template even without product-specific tests.

Also, I'm duplicating some of the test utils, so that it would be easier to move the framework tests away from product tests.

Step 2 will follow.

Bug: https://phabricator.wikimedia.org/T118404

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -46,10 +46,10 @@ var supportedValidators = ['maximum',
 var Validator = function(parameters) {
     this._ajv = constructAjv();
     this._paramCoercionFunc = this._createTypeCoercionFunc(parameters.filter(function(p) {
-        return p.name !== 'body';
+        return p.in !== 'formData' && p.in !== 'body' && p.type !== 'string';
     }));
     this._bodyCoercionFunc = this._createTypeCoercionFunc(parameters.filter(function(p) {
-        return p.name === 'body';
+        return p.in === 'formData' && p.type !== 'string';
     }));
     this._validatorFunc = this._ajv.compile(this._convertToJsonSchema(parameters));
 };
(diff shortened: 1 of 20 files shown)


lib/validator.js
```diff
@@ -46,10 +46,10 @@ var supportedValidators = ['maximum',
 var Validator = function(parameters) {
     this._ajv = constructAjv();
     this._paramCoercionFunc = this._createTypeCoercionFunc(parameters.filter(function(p) {
-        return p.name !== 'body';
+        return p.in !== 'formData' && p.in !== 'body' && p.type !== 'string';
```
gwicke: Why are you skipping over formData, too?

lib/validator.js
```diff
@@ -78,9 +78,7 @@ Validator.prototype._convertToJsonSchema = function(parameters) {
                 // 'required' array must have at least one element according to json-schema spec,
                 // se we can't preinitialize it.
                 schema.required = schema.required || [];
-                if (schema.required.indexOf(inMapping[param.in]) < 0) {
-                    schema.required.push(inMapping[param.in]);
-                }
+                schema.required.push(inMapping[param.in]);
```
gwicke: Is this relying on each parameter only being defined once? Was the previous deduplication unnecessary?

gwicke:
LGTM overall, only one small question inline. If that looks okay to you I'm :+1: on merging.

lib/validator.js
Pchelolo: @gwicke Yes. The previous code was dead code. We can get here only if `!schema.properties[inMapping[param.in]]` (line 73). And if that condition is true, we immediately set this property (line 74). So uniqueness of the array is already ensured without this if.

PR #436, opened 2015-12-08
Define /sys/ as api roots to scope metrics

gwicke:
API roots define scopes for both the swagger spec and metrics. As we want to
aggregate metrics at the api level (/sys and /v1), we need to define /sys as
an API root as well.

The current logic requires both swagger and info.x-is-api-root attributes to
establish new API spec roots. It feels like we might be able to improve on
this, but the best I can come up with is moving x-is-api-root to the spec
level, and perhaps renaming it. Other ideas?

In any case, we can now tweak this later without affecting third party user's
configs in the normal case.

### DIFF
--- a/projects/example.yaml
+++ b/projects/example.yaml
@@ -51,6 +51,9 @@ paths:
         - path: v1/content.yaml
 
   /{api:sys}:
+    swagger: 2.0
+    info:
+      x-is-api-root: true
     x-modules:
       /table:
         - type: npm
--- a/projects/wikimedia.org.yaml
+++ b/projects/wikimedia.org.yaml
@@ -56,6 +56,9 @@ paths:
         - path: v1/pageviews.yaml
 
   /{api:sys}:
+    swagger: 2.0
+    info:
+      x-is-api-root: true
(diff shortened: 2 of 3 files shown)


gwicke:
Verified the correct metric prefixes with manual requests & metric `type: log`.

Pchelolo:
@gwicke wrote:
> The current logic requires both swagger and info.x-is-api-root attributes to
establish new API spec roots. It feels like we might be able to improve on
this, but the best I can come up with is moving x-is-api-root to the spec
level, and perhaps renaming it. Other ideas?

I was hesitating on the same question today, while writing framework tests. The idea I came up with was a bit harder to implement, but from a semantics POW it feels more straightforward (at least for me). How about detecting an API root on every subtree (the level of the spec), where {api:xxx} is used? And on spec loading, validate that they do not clash (e.g. we can't have `/{api:one}/{api:two}` in a path)

gwicke:
@Pchelolo: As I said on IRC, that sounds like a pretty good option to me, especially as we are already relying on this for security.

PR #437, opened 2015-12-09
Default key_value and key_rev_value to timeseries

gwicke:
Also:

- Fix bucket config & table version combination for key_value and
  key_rev_value buckets by multiplying the bucket schema version by 1000 &
  adding the config version.
- Make retention_policy parameter consistent between key_value and
  key_rev_value buckets.

### DIFF
--- a/sys/key_rev_value.js
+++ b/sys/key_rev_value.js
@@ -30,17 +30,27 @@ KRVBucket.prototype.getBucketInfo = function(restbase, req, options) {
 };
 
 KRVBucket.prototype.makeSchema = function(opts) {
-    var schemaVersionMajor = 1;
+    var schemaVersionMajor = 2;
 
-    var schema =  {
+    return {
+        // Combine option & bucket version into a monotonically increasing
+        // combined schema version. By multiplying the bucket version by 1000,
+        // we increase the chance of catching a reset in the option version.
+        version: schemaVersionMajor * 1000 + (opts.version || 0),
         options: {
-            compression: [
+            compression: opts.compression || [
                 {
                     algorithm: 'deflate',
                     block_size: 256
                 }
-            ]
+            ],
+            updates: opts.updates || {
+                pattern: 'timeseries'
+            },
         },
+        revisionRetentionPolicy: opts.retention_policy
+            // Deprecated version. TODO: Remove eventually.
+            || opts.revisionRetentionPolicy,
         attributes: {
             key: opts.keyType || 'string',
             rev: 'int',
@@ -59,23 +69,10 @@ KRVBucket.prototype.makeSchema = function(opts) {
             { attribute: 'tid', type: 'range', order: 'desc' }
         ]
     };
-
-    if (opts.revisionRetentionPolicy) {
-        schema.revisionRetentionPolicy = opts.revisionRetentionPolicy;
-    }
-    if (opts.version) {
-        schema.version = schemaVersionMajor + opts.version;
-    }
-
-    return schema;
 };
 
 KRVBucket.prototype.createBucket = function(restbase, req) {
-    var opts = req.body;
-    if (!opts.type) { opts.type = 'key_rev_value'; }
-    if (!opts.keyType) { opts.keyType = 'string'; }
-    if (!opts.valueType) { opts.valueType = 'blob'; }
-    var schema = this.makeSchema(opts);
+    var schema = this.makeSchema(req.body || {});
     schema.table = req.params.bucket;
     var rp = req.params;
     var storeRequest = {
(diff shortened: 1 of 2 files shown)


sys/key_rev_value.js
```diff
@@ -30,17 +30,20 @@ KRVBucket.prototype.getBucketInfo = function(restbase, req, options) {
...
-                    block_size: 256
-                }
-            ]
+            compression: opts.compression,
+            updates: opts.updates || {
+                pattern: 'timeseries'
+            },
         },
+        revisionRetentionPolicy: opts.revisionRetentionPolicy
+            || opts.retention_policy,
```
Pchelolo: This code is now duplicated on lines 66 to 68

sys/key_value.js
```diff
@@ -31,12 +31,17 @@ KVBucket.prototype.getBucketInfo = function(restbase, req, options) {
...
+
     return {
-        version: opts.schemaVersion,
+        // 1000 external version bumps per internal version bump ought to be
+        // enough for anybody.
+        version: schemaVersionMajor * 1000 + (opts.version || 0),
         options: {
             compression: opts.compression,
-            updates: opts.updates
+            updates: opts.updates || {
```
d00rman: Not needed, this is explicitly set on line 85 already.

d00rman:
I may be missing something, but it seems this change only duplicates code. Also, I don't understand the point of setting the schema version to 5k.

gwicke:
> I may be missing something, but it seems this change only duplicates code.

@mobrovac, I did indeed miss that createBucket is directly changing the options. Let me move all that into makeSchema for clarity & avoidance of option modifications.

> Also, I don't understand the point of setting the schema version to 5k.

We have two versions to combine into one table schema version: 

a) A version for the bucket options from the request, and
b) a version for how this is mapped to a table schema in the bucket implementation.

The requirement is that versions are monotonically increasing per-table. If we assume that both versions are indeed always increasing, then adding the two versions is enough to guarantee monotonicity. However, if one of them is dropped or reset, there is a good chance to revert to a previous value, which in turn means that the error might be covered up. By multiplying the bucket schema version by 1000, this chance is significantly reduced as long as option versions are relavitely small (< 1000), which I think is a reasonable assumption to make.

sys/key_rev_value.js
d00rman: Given that we have settled on having *snake_case* config options, `retention_policy` should probably be preferred. Also, do we want to set a sane default here in case the option hasn't been provided (which usually isn't) ?

gwicke: I was wondering about the default as well. Most use cases would probably want "latest". The only thing that kept me from setting that is the worry that some uses might be surprised by data being deleted in a revisioned storage bucket.

Pchelolo: @gwicke we are providing a latest-1 rev policy in a key_value storage, this case doesn't seem different

gwicke: I think there is no expectation for long-term versioning in key_value storage. The versioning is an implementation detail in that case. For key_rev_value, there is an expectation of versioning. The question is whether this expectation typically includes the tid, or only the revision.

gwicke: > Given that we have settled on having snake_case config options, retention_policy should probably be preferred.

Changed the order in the amend.

sys/key_rev_value.js
```diff
@@ -30,17 +30,27 @@ KRVBucket.prototype.getBucketInfo = function(restbase, req, options) {
...
+        version: schemaVersionMajor * 1000 + (opts.version || 0),
         options: {
-            compression: [
-                {
-                    algorithm: 'deflate',
-                    block_size: 256
-                }
-            ]
+            compression: opts.compression || [
+            {
```
d00rman: Nit: indentation: either put then together with `[]` (above and below) or indent the contents.

sys/key_value.js
```diff
@@ -31,14 +31,29 @@ KVBucket.prototype.getBucketInfo = function(restbase, req, options) {
...
-        version: opts.schemaVersion,
+        // Combine option & bucket version into a monotonically increasing
+        // combined schema version. By multiplying the bucket version by 1000,
+        // we increase the chance of catching a reset in the option version.
+        version: schemaVersionMajor * 1000 + (opts.version || 0),
         options: {
-            compression: opts.compression,
-            updates: opts.updates
+            compression: opts.compression || [
+            {
```
d00rman: Nit: idem as above.

d00rman:
Minor nit, :+1: otherwise

gwicke:
Rebased & tweaked the indent.

PR #438, opened 2015-12-09
Router refactoring: detect API spec based on the 'api' prefix

Pchelolo:
Several changes and refactoring incorporated in this PR:

- Now we detect the API root by the `api` prefix in the path spec. This is way cleaner than detecting it based on the `swagger` and `x-is-api-root` stanza.
- The `/sys` access protection is done on the basis of `params.api` not by the "second element in the path" rule. This gives framework users ability to specify the internal sys APIs on any level. In future I want to make an `x-internal-api` stanza, or, maybe, use special `api_private` prefix to detect private API roots - this will give even more flexibility, but it's a subject to a different PR
- General refactoring of the request handler code - the function was to big, so I've split it into several pieces with clear responsibilities.
- Fixed a bug in restbase child creation - we were trying to inherit the `_req` property and never assigned one, so it was always `null`.

This is a part of framework-separation work.

Further refactoring is possible, but it's a subject to another PR as more drastic changes needed, and that should be done together with authentication modules refactoring.

IMPORTANT: I've found more bugs in the router, so please treat this as WIP until further notice.

### DIFF
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -25,7 +25,7 @@ function RESTBase(options, req) {
         this._parent = par;
         // Remember the request that led to this child instance at each level, so
         // that we can provide nice error reporting and tracing.
-        this._req = par._req;
+        this._req = req;
         this._recursionDepth = par._recursionDepth + 1;
         this._priv = par._priv;
         this.rb_config = this._priv.options.conf;
@@ -173,54 +173,28 @@ RESTBase.prototype.defaultWebRequestHandler = function(req) {
     return P.resolve(preq(req));
 };
 
-
 RESTBase.prototype._isSysRequest = function(req) {
     return ((req.uri.params && req.uri.params.api === 'sys')
         // TODO: Remove once params.api is reliable
-            || (req.uri.path && req.uri.path.length > 1 && req.uri.path[1] === 'sys'));
+    || (req.uri.path && req.uri.path.length > 1 && req.uri.path[1] === 'sys'));
 };
 
-// Process one request
-RESTBase.prototype.request = function(req) {
-    // Protect the sys api from direct access
-    // Could consider opening this up with a specific permission later.
-    if (this._recursionDepth === 0 && this._isSysRequest(req)) {
-        return P.reject(new HTTPError({
-            status: 403,
-            body: {
-                type: 'access_denied#sys',
-                title: 'Access to the /{domain}/sys/ hierarchy is restricted to system users.'
-            }
-        }));
-    }
-
-    if (req.method) {
-        req.method = req.method.toLowerCase();
-    }
-
-    return this._request(req);
-};
-
-
-// Internal request handler
-RESTBase.prototype._request = function(req) {
-    var self = this;
-
-    // Special handling for https? requests
-    if (req.uri.constructor === String && /^https?:\/\//.test(req.uri)
-            || req.uri.urlObj) {
-        return this.defaultWebRequestHandler(req);
-    }
-
-    var priv = this._priv;
-    if (this._recursionDepth > priv.options.maxDepth) {
+/**
+ * Checks if the maximum recursion depth has been exceeded by the request.
+ * If yes, the 500 error is thrown, othervise this is a no-op
+ *
+ * @param {Object} req - a current request object
+ * @private
+ */
+RESTBase.prototype._checkMaxRecursionDepth = function(req) {
+    if (this._recursionDepth > this._priv.options.maxDepth) {
         var parents = [];
         var rb = this._parent;
         while (rb) {
             parents.push(rb._req);
             rb = rb._parent;
         }
-        return P.resolve({
+        throw new HTTPError({
             status: 500,
             body: {
                 type: 'request_recursion_depth_exceeded',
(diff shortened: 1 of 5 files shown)


Pchelolo:
cc @wikimedia/services

lib/restbase.js
```diff
@@ -173,79 +173,131 @@ RESTBase.prototype.defaultWebRequestHandler = function(req) {
...
-             || (req.uri.path.length > 1 && req.uri.path[1] === 'sys'))) {
-        return P.reject(new HTTPError({
+/**
+ * Protects /sys APIs from the direct access.
+ *
+ * @param {Object} match - a found request handler
+ * @private
+ */
+RESTBase.prototype._checkInternalApiRequest = function(match) {
+    if (this._recursionDepth === 0 && match && match.params && match.params.api === 'sys') {
```
d00rman: I'd be prudent here and still keep the `req.uri.path[1] === 'sys'` check as a fall-back in case `/{api:sys}` is not specified in the URI on the bases that it is not strictly needed to actually put it there, hence preventing RB users to accidentally exposing internal routes to the public.

Pchelolo: Amended the change to return that line back

lib/restbase.js
```diff
@@ -25,7 +25,7 @@ function RESTBase(options, req) {
         this._parent = par;
         // Remember the request that led to this child instance at each level, so
         // that we can provide nice error reporting and tracing.
-        this._req = par._req;
+        this._req = req;
```
gwicke: I think the intention here is to have a reference to the root request, which this won't preserve.

lib/restbase.js
```diff
@@ -173,79 +173,135 @@ RESTBase.prototype.defaultWebRequestHandler = function(req) {
...
+ */
+RESTBase.prototype._checkInternalApiRequest = function(req, match) {
     if (this._recursionDepth === 0 &&
-            ((req.uri.params && req.uri.params.api === 'sys')
-             // TODO: Remove once params.api is reliable
-             || (req.uri.path.length > 1 && req.uri.path[1] === 'sys'))) {
-        return P.reject(new HTTPError({
+            (match && match.params && match.params.api === 'sys'
+            // Fallback to protect against accidentally exposing system routes in RESTBase
+            || req.uri.path.length > 1 && req.uri.path[1] === 'sys')) {
```
gwicke: +1 for double-securing this for now.

lib/restbase.js
```diff
@@ -173,79 +173,135 @@ RESTBase.prototype.defaultWebRequestHandler = function(req) {
...
     var methods = match && match.value && match.value.methods;
     var handler = methods && (
             (self._rootReq && self._rootReq.method === 'head' && methods.head)
-            || methods[childReq.method]
-            || methods.all);
-    if (!handler &&
-            (req.method === 'head'
-            || self._rootReq && self._rootReq.method === 'head')) {
+            || methods[childReq.method] || methods.all);
+    if (!handler && (req.method === 'head' || self._rootReq && self._rootReq.method === 'head')) {
```
gwicke: Perhaps re-add some line breaks for clarity?

lib/router.js
```diff
@@ -339,41 +339,40 @@ Router.prototype._handleSwaggerPathSpec = function(node, pathspec, scope) {
...
+            methods: {},
+            path: specRoot.basePath + '/',
+            globals: node.value.globals,
+        };
+        node.setChild('', listNode);
+
+        loaderPromise = loaderPromise.then(function() {
+            return self._handleSwaggerSpec(node, pathspec, scope);
+        });
+    } else {
```
gwicke: This seems to remove the ability to define inline specs at places in the hierarchy other than the api level.

lib/restbase.js
```diff
@@ -173,79 +173,135 @@ RESTBase.prototype.defaultWebRequestHandler = function(req) {
...
+            return self._authService.checkPermissions(self, childReq)
+            .then(function() { return handlerPromise; });
+        }
+    } else {
+        return handlerPromise;
+    }
+};
 
-// Internal request handler
+// Process one request
```
gwicke: Since this code path it's very hot it would be good to double-check that there is no significant performance penalty from this refactor. It seems unlikely that it would have a big effect, but good to check.

gwicke:
LGTM overall, but a couple of questions & nits inline.

lib/restbase.js
Pchelolo: @gwicke No, to preserve root request we have a special property, `_rootReq` (line 32). Here the intention is to store a request that was the reason to create this particular child instance. Currently it's used to build out the request chains on recursion depth limit errors to put the chain into the error.

lib/router.js
Pchelolo: @gwicke I think you mean `securityDefinitions` stanzas, not `security` stanzas?

For secDef, having them in non-root spec doesn't make any sense since the AuthService looks for the them only in the rootSpec for the current request chain. 

Also, according to the swagger spec schema, the `securityDefinitions` and `definitions` properties can appear only at the top-level, which is represented by a rootSpec in our code. This means, that actually `handleSwaggerSpec` should be called only for the `rootSpec`, which beginning is now defined by an `api` prefix. 

So, I think that the current change is correct, and actually, we need to remove several more `handleSwaggerSpec` calls, but right now it's too risky, as we don't have enough tests for different kinds of specs. I'm currently writing those tests, but they depend on this change.

lib/restbase.js
Pchelolo: @gwicke Indeed there was a problem with adding one more P.try, but actually it's not needed. I've removed an extra `try` and now the performance is intact.

lib/router.js
Pchelolo: @gwicke Hm, actually I've found a configuration, where this doesn't work as expected.. Need some more time

lib/restbase.js
gwicke: Okay, makes sense.

PR #440, opened 2015-12-09
Add a static robots.txt in the wikimedia config

gwicke:
Task: https://phabricator.wikimedia.org/T119786

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -53,6 +53,19 @@ spec_root: &spec_root
     # global domain
     /{domain:wikimedia.org}: *wikimedia.org
 
+    # A robots.txt to make sure that the content isn't indexed.
+    /robots.txt:
+      get:
+        x-request-handler:
+          - static:
+              return:
+                headers:
+                  content-type: text/plain
+                body: |
+                  User-agent: *
+                  Allow: /*/v1/?doc
+                  Disallow: /
+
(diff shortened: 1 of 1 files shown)


config.example.wikimedia.yaml
```diff
@@ -53,6 +53,20 @@ spec_root: &spec_root
     # global domain
     /{domain:wikimedia.org}: *wikimedia.org
 
+    # A robots.txt to make sure that the content isn't indexed.
```
d00rman: Would be nice to add this to `config.test.yaml` as well and have a test for it, either via an `x-amples` stanza or a proper test under `test/features/`

gwicke: I'm hesitant to clutter the config any more, so I think x-amples is out. A separate test to assert that the robots.txt entry point is in the config doesn't seem to add a lot of value. The general handler functionality is tested elsewhere, so the only thing we'd assert is that the config has this configured in the test config, which says nothing about production.

eevans:
LGTM :+1:

d00rman:
So, my main objection here is the divergence between the test and the production config. Granted, this one is really a minor difference. However, if we continue down this path, we'll be running tests locally nad on Travis that do not fully comply with what we have in production, while at the same time paradoxically having a high coverage percentage.

I concede that this stems from our awesome config possibilities, but we need to keep track of these differences to keep the high confidence in tests we currently have.

PR #441, opened 2015-12-10
Add a rerenderBlacklist option for the parsoid module

gwicke:
A few huge & extremely frequently edited pages are using a disproportionate
amount of resources for re-renders in both storage and Parsoid.  While longer
term improvements like https://phabricator.wikimedia.org/T120171 and Parsoid
performance will make it easier to deal with such (ab)use, in the short term
it is useful to disable job queue triggered updates for specific problematic
pages. Regular access to those pages still works, but tends to be rare for
these pages, leading to substantial savings in overall resource usage.

To be clear: This is not meant as a replacement for performance improvements,
but reduces the impact those few pages have right now until efficient
incremental parsing & chunked storage are implemented.

Task: https://phabricator.wikimedia.org/T120971
Task: https://phabricator.wikimedia.org/T120171
Task: https://phabricator.wikimedia.org/T94121
Task: https://phabricator.wikimedia.org/T120972

### DIFF
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -99,6 +99,29 @@ paths:
             #purge:
             #  host: '239.128.0.112'
             #  port: 4827
+
+            # A list of pages that we don't currently want to re-render on
+            # each edit. Most of these are huge bot-edited pages, which are
+            # rarely viewed in any case. 
+            rerenderBlacklist:
+              www.wikidata.org:
+                Q21558717: true
+                Q21505108: true
+                Q21481867: true
+                Q21521425: true
+              en.wikipedia.org:
+                Wikipedia:Administrators'_noticeboard/Incidents: true
+                User:JamesR/AdminStats: true
+                User:Kudpung/Dashboard: true
+              ur.wikipedia.org:
+                نام_مقامات_ایل: true
+                نام_مقامات_ڈی: true
+                نام_مقامات_جے: true
+                نام_مقامات_جی: true
+                نام_مقامات_ایچ: true
+                نام_مقامات_ایم: true
+              commons.wikipedia.org:
+                Commons:Quality_images_candidates/candidate_list: true
       /mobileapps:
         - path: sys/mobileapps.yaml
           options: '{{options.mobileapps}}'
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -44,7 +44,7 @@ var cacheURIs = [
 
 function ParsoidService(options) {
     var self = this;
-    options = options || {};
+    this.options = options = options || {};
     this.log = options.log || function() {};
     this.parsoidHost = options.parsoidHost
         || 'http://parsoid-lb.eqiad.wikimedia.org';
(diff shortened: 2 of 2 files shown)


projects/wmf_default.yaml
```diff
@@ -99,6 +99,28 @@ paths:
...
+              en.wikipedia.org:
+                Wikipedia:Administrators'_noticeboard/Incidents: true
+                User:JamesR/AdminStats: true
+                User:Kudpung/Dashboard: true
+              ur.wikipedia.org:
+                نام_مقامات_ایل: true
+                نام_مقامات_ڈی: true
+                نام_مقامات_جے: true
+                نام_مقامات_جی: true
+                نام_مقامات_ایچ: true
```
subbuss: Also this title .. https://ur.wikipedia.org/w/index.php?title=%D9%86%D8%A7%D9%85_%D9%85%D9%82%D8%A7%D9%85%D8%A7%D8%AA_%D8%A7%DB%8C%D9%85&action=history

Pchelolo:
@gwicke Two comments:
- I'd really like to see this code wrapped in `//TODO: THIS IS A TEMP CODE TO BE REMOVED UNTIL PHUB_XXX IS RESOLVED`
- Are we sure, that the RestbaseUpdateJobs would react to these 403 as we expect (do nothing)?

projects/wmf_default.yaml
gwicke: Yup, added.

sys/parsoid.js
```diff
@@ -380,6 +380,22 @@ PSP.getRevisionInfo = function(restbase, req) {
...
+ * Internal check to see if it's okay to re-render a particular title in
+ * response to a no-cache request.
+ * @param {Request}
+ * @return {boolean} Whether re-rendering this title is okay.
+ */
+PSP._okayToRerender = function(req) {
+    var blackList = this.options.rerenderBlacklist;
+    if (blackList) {
+        return !blackList[req.params.domain]
+            || !blackList[req.params.domain][req.params.title];
```
Pchelolo: May be worths adding one more 'or' clause with `rbUtil.normalizeTitle` to be sure?

d00rman:
I second @Pchelolo's first comment . It could also be nice to add a test for this.

> Are we sure, that the RestbaseUpdateJobs would react to these 403 as we expect (do nothing)?

It will return `false` to the hook caller, but that's not relevant because all of the hooks we are using do not abort anything when `false` is returned, i.e. they are *after-the-fact* hooks.

gwicke:
> Are we sure, that the RestbaseUpdateJobs would react to these 403 as we expect (do nothing)?

Yeah, I made very sure that there are no retries from the job queue. It used to constantly retry some titles forever (as in many months), which clearly was a bad idea.

Also added a comment about the temporary-ness of this.

sys/parsoid.js
gwicke: The titles from the job queue should all be normalized, so I don't think that's needed.

Pchelolo:
@d00rman wrote:
> It will return false to the hook caller, but that's not relevant because all of the hooks we are using do not abort anything when false is returned, i.e. they are after-the-fact hooks.

I was more worried that it will start to retry the request. Probably not, but I'm not 100% sure.

sys/parsoid.js
d00rman: Right, but malicious clients could get passed this check because of this assumption.

gwicke: Malicious clients can also just request every revision of every article, focusing on the most expensive pages. This isn't meant as an effective defense against external requests. It's focused on requests from the job queue.

Pchelolo:
Unless we wanna protect against a ddos, LGTM

sys/parsoid.js
```diff
@@ -417,6 +439,18 @@ PSP.getFormat = function(format, restbase, req) {
     });
 
     if (req.headers && /no-cache/i.test(req.headers['cache-control'])) {
+        if (!self._okayToRerender(req)) {
+            // Still update the revision metadata.
+            self.getRevisionInfo(restbase, req);
+            throw new rbUtil.HTTPError({
```
d00rman: Wouldn't getting revinfo and throwing the error in parallel cancel the revinfo promise given that all of the operations are executed in the `P.try()` context?

gwicke: No, the only thing cancelling the parallel revInfo request would be the process exiting, or some failure inside the revInfo request.

PR #443, opened 2015-12-12
Add cache headers in pageviews responses

gwicke:
Per Nuria, the analitics team is planning to publish a blog post on pageviews
next week, and would like to have caching active before then. They don't have
time to implement this in AQS, so I'm adding those headers in the
pageviews_proxy response.

Task: https://phabricator.wikimedia.org/T119886

### DIFF
--- a/sys/pageviews_proxy.yaml
+++ b/sys/pageviews_proxy.yaml
@@ -10,7 +10,7 @@ paths:
               uri: '{{options.host}}/pageviews/per-article/{+rest}'
             return:
               status: '{{get_from_backend.status}}'
-              headers: '{{merge(get_from_backend.headers, {"cache-control":"s-maxage: 86400, max-age: 86400"})}}'
+              headers: '{{merge(get_from_backend.headers, {"cache-control":"s-maxage: 3600, max-age: 3600"})}}'
               body: '{{get_from_backend.body}}'
 
   /per-project/{+rest}:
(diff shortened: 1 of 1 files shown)


milimetric:
I'd be happy to add it to AQS, just wasn't sure how hard that would be.  Are there examples?

gwicke:
@milimetric, it's just adding a header in the response. You could do so in pageviews.js, for example [here](https://github.com/wikimedia/restbase/blob/master/sys/pageviews.js#L307):

```javascript
res.headers['cache-control'] = 's-maxage=86400, max-age=86400';
```

milimetric:
Thanks, @gwicke, I cached for only 1 hour as some of the data is hourly and we might be adding more granularities later.  So I'm not sure if you still wanted to do this PR, but mine's at https://github.com/wikimedia/restbase/pull/444

milimetric:
Thanks, @gwicke, this works for me.  It makes sense to add these on the front-end then, it looks like the ops config change would delay us otherwise.

d00rman:
:+1: The added benefit here is that if AQS starts emitting `Cache-Control` headers, those will be honoured instead of the ones set by RB.

gwicke:
@d00rman, it turned out that the backend service apparently sends explicit no-cache headers, which made this non-functional. For that reason, I changed the precedence in https://github.com/wikimedia/restbase/pull/448.

PR #444, opened 2015-12-14
Cache pageview responses for 1 hour

milimetric:


gwicke:
@milimetric, it might be worth adding this to the project & top entry points as well.

milimetric:
@gwicke oh, the response from all the methods is passed through normalizeResponse, so all endpoints get the 1 hour cache.

gwicke:
@milimetric: Oh, okay. 1 hour cache time is quite short, but it's easy to change.

PR #448, opened 2015-12-15
Override cache-control headers from the pageviews backend

gwicke:
It looks like AQS is emitting cache-control headers of its own, which override
the front-end's private cache-control heders. This patch switches the
precedence to override the backend headers.


gwicke:
Self-merging, as page view metric caching is currently broken in production.

PR #449, opened 2015-12-15
Graphoid: Change the test graph revision/hash

d00rman:
It seems the hashing algorithm has changed recently, causing RESTBase to ask for an invalid graph hash.


d00rman:
Travis passing, self-merging to do an emergency deploy.

PR #451, opened 2015-12-18
Log the bodyOnly warning headers properly

d00rman:
Bug: [T121852](https://phabricator.wikimedia.org/T121852)

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -926,8 +926,9 @@ PSP.makeTransform = function(from, to) {
             // https://phabricator.wikimedia.org/T114185).
             // Log remaining bodyOnly uses / users
             if (to === 'html' && originalBodyOnly) {
-                self.log('warn/parsoid/bodyonly',
-                    restbase._rootReq && restbase._rootReq.headers);
+                self.log('warn/parsoid/bodyonly', {
+                    req_headers: restbase._rootReq && restbase._rootReq.headers
+                });
             }
             return res;
         });


d00rman:
cc @wikimedia/services

eevans:
LGTM

Pchelolo:
@d00rman BTW, this will still log the `[Object object]` message. Look [here](https://github.com/wikimedia/service-runner/blob/master/lib/logger.js#L178) - `info` is still an object, so it will follow the `'' + info` clause. To fix that we need to add a 'message' property, but I'm not sure we need to do that - the patch OCG is already done, pending deployment, so we will remove these lines soon anyway.

d00rman:
Uh, good point, @Pchelolo . Then we definitely have a bug in `service-runner`, since one possible execution path there is:

```js
msg = (typeof info === 'object') ? '' + info;
```

... which is plain wrong. In the meantime, I'll create a new PR here for adding the needed `message` field.

Pchelolo:
@d00rman I wouldn't say it's wrong. If we don't have a clear message, what do we use as a message? We can't use `JSON.stringify` as it can fail and would produce too long messages.

d00rman:
I would say it's wrong, because at that point we already know that `info` is an object, so `'' + info` will always produce `[Object object]` which isn't helpful at all.

I agree though that it is not easy to set on a generic message is not easy, but I'd personally prefer to see `"No message supplied"` than `"[Object object]"`

Pchelolo:
@d00rman However, if the provided object has a defined toString(), then `'' + info` would produce a proper result. How about amending the logger to do: `(typeof info.toString === 'function') ? '' + info : 'Message not supplied'`

d00rman:
Unfortunately @Pchelolo that wouldn't work because all objects already have a `toString()` method defined:

```
$ nodejs
> var a = {jk: 'blah'};
undefined
> (typeof a.toString === 'function') ? a.toString() : 'no method here';
'[object Object]'
```

Pchelolo:
@d00rman Right, I've realised that on the way to the office, but couldn't do anything. How about  `('' + info' !== '[object Object]') ? '' + info : 'Message not supplied' ?

d00rman:
#452 follows up with setting a message

d00rman:
> Right, I've realised that on the way to the office, but couldn't do anything. How about `('' + info' !== '[object Object]') ? '' + info : 'Message not supplied' ?

That could do it as well, with a small performance penalty as it stringifies the object twice.

Pchelolo:
@d00rman We can avoid double stringification, I've just showed the idea. Let me make a PR.

PR #453, opened 2015-12-18
Configs: make example and test config less different

Pchelolo:
Recently we've had several bugs in production which were not caught by the tests because of the difference between our test and example config. This PR reduces the difference to minimum. Changes:
- Reuse project files in test config
- Reorganise the test_module config so that it could be 'mounted' on the config level, not project level
- Make a separate AQS project
- Separate RB and AQS parts of the test_module
- Reorganise security a bit  - do not check `/sys` routes and mount security one level up.

### DIFF
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -174,14 +174,17 @@ RESTBase.prototype.defaultWebRequestHandler = function(req) {
 };
 
 
+RESTBase.prototype._isSysRequest = function(req) {
+    return ((req.uri.params && req.uri.params.api === 'sys')
+        // TODO: Remove once params.api is reliable
+            || (req.uri.path && req.uri.path.length > 1 && req.uri.path[1] === 'sys'));
+};
+
 // Process one request
 RESTBase.prototype.request = function(req) {
     // Protect the sys api from direct access
     // Could consider opening this up with a specific permission later.
-    if (this._recursionDepth === 0 &&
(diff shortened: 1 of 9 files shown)


Pchelolo:
cc @wikimedia/services

d00rman:
Nice! LGTM :+1:

lib/restbase.js
```diff
@@ -271,7 +274,8 @@ RESTBase.prototype._request = function(req) {
         var childRESTBase = this.makeChild(req);
         childReq.params = match.params;
 
-        var accessRestricted = match.permissions
+        var accessRestricted = !this._isSysRequest(req)
```
gwicke: It might be worth adding a comment here to make it clear that sys requests are in fact still access restricted, but the check happens elsewhere.

test/aqs_test_module.yaml
```diff
@@ -0,0 +1,96 @@
+# Simple test spec
```
gwicke: The aqs stuff actually needs a project like this for its v1 api in any case. Do you think we could use this for the real thing?

Pchelolo: @gwicke see `/projects/aqs_default.yaml` it sets up an example prod aqs. These are just fake endpoints to enter the test data into cassandra.

gwicke:
Very nice overall. A super minor request for a comment & a question are inline. Looks good to go to me.

PR #454, opened 2015-12-21
Tests: Covered swagger-ui usages with tests

Pchelolo:
Just a "get into coding mode on Monday morning" patch with a couple of additional tests, covering how swagger-ui is handled and API and domain listings.

Also, I've found several test-bugs - the tests were passing, but didn't actually test the feature they're supposed to test.

cc @wikimedia/services


Pchelolo:
Rebased on master

PR #456, opened 2015-12-22
Frameworkefication: Split rbUtils to MW-specific, internal and external

Pchelolo:
This is mostly moving code around, without adding any value. The idea is, if we wanna extract the lib folder as a framework, we need to set a set of clear exports from the library. One of the candidates for exports is `rbUtil`, but it contains framework-internal stuff, general-purpose code and MW-specific code. Here I've split that into 3 files:
- `mwUtil` - mediawiki-specific utilities, that wouldn't be extracted into the framework
- `utils` - framework-internal utilities (right now it's almost empty, but more stuff will follow
- `exports` - exported general-purpose utilities. Right now the only important one is `HTTPError`. 
However, there's a nasty `addCSPHeaders` that is placed in the exports too for now, but it's nasty, we'd need to put it somewhere else (probably even convert into a separate CSP-management library)

Also, functions that are related only to one class/file were moved to the appropriate file.

There's still some work to do in terms of refactoring, and unused code removal, but it's better to do that separately from this giant code move.


Pchelolo:
cc @wikimedia/services

eevans:
We could probably bikeshed the naming/namespaces here (exports, exports.misc, utils, etc), but the overall breakup of code seems reasonable, so +1 from me.

Pchelolo:
@eevans The `exports.misc` namespace should go away, it doesn't belong here. I'm thinking about a more general architecture for this - probably we could make/use a separate package for CSP manipulation and introduce a bit of code duplication to get rid of it.

I'd be happy to rename others if you have ideas for better naming.

test/utils/cleandb.sh
```diff
@@ -5,7 +5,7 @@ dropKeyspaces ( ) {
   then
     PATTERN=$1
     echo "looking for keyspaces named '*$PATTERN*'..."
-    for KEYSPACE in `echo 'select keyspace_name from system.schema_keyspaces;' | cqlsh | grep $PATTERN`
+    for KEYSPACE in `echo 'select keyspace_name from system.schema_keyspaces;' | cqlsh `
```
d00rman: Why has the `grep` been removed? Now this is dropping all of the keyspaces in Cassandra.

Pchelolo: @d00rman That's obviously a mistake, returned back.

d00rman:
I second @eevans' comments about names, but given that this is an intermediary step towards greatness, we can differ the bike-shedding for later when it will really matter :)

PR #457, opened 2015-12-28
Bugfix: Wrong comparasion of revision info

Pchelolo:
We're comparing old&new revision info on update to avoid storing identical revision info objects. We treat all falsy values as identical, but we do it in a wrong way - if old is falsy and new is not, we still treat them as equal.

### DIFF
--- a/sys/page_revisions.js
+++ b/sys/page_revisions.js
@@ -190,15 +190,15 @@ PRS.prototype._checkSameRev = function(firstRev, secondRev) {
             var value = rev[key];
             // Ignore the tid attribute
             if (key !== 'tid') {
-                if (!value || (Array.isArray(value) && !value.length)) {
-                    // treat all falsy values, as well as an empty array equally - as null.
-                    result[key] = null;
-                } else if (key === 'timestamp') {
+                if (key === 'timestamp') {
                     // 'timestamp' fields need to be parsed because Cassandra
                     // returns a ISO8601 ts which includes milliseconds, while
                     // the ts returned by MW API does not
                     result[key] = Date.parse(value);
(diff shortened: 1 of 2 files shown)


Pchelolo:
cc @wikimedia/services

eevans:
LGTM!

sys/page_revisions.js
```diff
@@ -183,32 +184,26 @@ PRS.prototype.listTitles = function(restbase, req, options) {
...
-                for (var idx = 0; idx < firstVal.length; idx++) {
-                    if (firstVal[idx] !== secondVal[idx]) {
-                        return true;
-                    }
+    function normalizeRev(rev) {
+        var result = {};
+        Object.keys(rev).forEach(function(key) {
+            var value = rev[key];
+            // Ignore the tid attribute
+            if (key !== 'tid') {
```
gwicke: Possibly slightly clearer: 

```javascript
if (key === 'tid') {
  return;
}
```

sys/page_revisions.js
```diff
@@ -183,32 +184,26 @@ PRS.prototype.listTitles = function(restbase, req, options) {
...
+                if (key === 'timestamp') {
+                    // 'timestamp' fields need to be parsed because Cassandra
+                    // returns a ISO8601 ts which includes milliseconds, while
+                    // the ts returned by MW API does not
+                    result[key] = Date.parse(value);
+                } else if (value && (!Array.isArray(value) || value.length)) {
+                    result[key] = value;
                 }
-                return false;
+                // ignore all falsy values, as well as an empty array
```
gwicke: Perhaps make this more explicit by testing for those things & returning early?

Pchelolo: @gwicke done

PR #458, opened 2015-12-29
More explicitly validate the response status

gwicke:
Explicitly check for 100 <= status < 400, and treat other status values
(including non-numeric status values) as an error.

### DIFF
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -313,7 +313,7 @@ RESTBase.prototype._request = function(req) {
                         req: req
                     }
                 });
-            } else if (res.status >= 400 && !(res instanceof Error)) {
+            } else if (!(res.status >= 100 && res.status < 400) && !(res instanceof Error)) {
                 var err = new HTTPError(res);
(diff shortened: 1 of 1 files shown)


lib/restbase.js
```diff
@@ -313,7 +313,7 @@ RESTBase.prototype._request = function(req) {
                         req: req
                     }
                 });
-            } else if (res.status >= 400 && !(res instanceof Error)) {
+            } else if (!(res.status >= 100 && res.status < 400) && !(res instanceof Error)) {
                 var err = new HTTPError(res);
```
Pchelolo: Shouldn't we set some valid status code in this case?

Pchelolo:
@gwicke this is uncovering a bug in security/action code. [Over here](https://github.com/wikimedia/restbase/blob/master/sys/action.js#L155) we have to set the `res.body` to that value and return `res`. This should also be accounted [here](https://github.com/wikimedia/restbase/blob/master/lib/auth.js#L76).

PR #459, opened 2015-12-29
Fix cache-control header in stored mathoid responses

gwicke:
The previous code did not set a proper cache-control header for stored
responses. This caused clients to unnecessarily reload math renders, and
disallowed Varnish caching altogether.


Pchelolo:
LGTM

PR #461, opened 2015-12-30
Parsoid: Do not query for html in case rerender is disabled

Pchelolo:
Finally found the reason of EMERGENCY 'unhandled error' log messages. When the the serenader is disabled, we were still making a `get` request for html content, but didn't return that promise. But the request was made anyway, so if it failed, we got an unhandled rejection. And the result is ignored anyway, so we should check if the serenader is disabled prior to making the request for html.

cc @wikimedia/services

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -395,17 +395,18 @@ PSP.getRevisionInfo = function(restbase, req) {
  * https://phabricator.wikimedia.org/T120972 are resolved / resource
  * consumption for these articles has been reduced to a reasonable level.
  *
- * @param {Request}
+ * @param {Request} req the request being processed
  * @return {boolean} Whether re-rendering this title is okay.
  */
 PSP._okayToRerender = function(req) {
-    var blackList = this.options.rerenderBlacklist;
-    if (blackList) {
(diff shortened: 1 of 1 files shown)


sys/parsoid.js
```diff
@@ -440,25 +441,25 @@ PSP.getFormat = function(format, restbase, req) {
         }
     }
 
+    if (!self._okayToRerender(req)) {
```
gwicke: This drops the no-cache check, rendering those pages completely unaccessible. Could you reinstate that check?

Pchelolo: @gwicke See line 402

gwicke: I see, missed that. I still think it would be more obvious to avoid splitting the no-cache logic, but it'll work as-is.

Pchelolo: If we don't split the `no-cache` logic, we'll need to duplicate the 
```
var contentReq = restbase.get({
          uri: self.getBucketURI(rp, format, rp.tid)});
});
```
code. And this check is a temporary solution, so I think its better to have all it's logic self-contained.

gwicke:
@Pchelolo: Okay, merging.

PR #463, opened 2016-01-11
Specs: Removed duplicated key from a spec

Pchelolo:
A new version of `js-yaml` package was just released, and it became more strict on duplicated keys in the yaml file - it doesn't merge the objects any more, but fails with an exception. This PR removes dup keys from RESTBase specs.


Pchelolo:
cc @wikimedia/services

gwicke:
:+1: 

Will merge once the tests are done.

PR #464, opened 2016-01-12
API: Added definitions enpoint for wiktionary domains

Pchelolo:
This PR adds a `definition` endpoint for `wiktionary` domains instead of a summary endpoint. For details see https://phabricator.wikimedia.org/T123142

I created the definition response schema based on sale responses from the service, so I'm not completely sure it's 100% correct and covers all the optionals.

Another thing happening here is some refactoring of the x-amples tests:
1. I've removed the `x-default-params` stanza, because it doesn't make sense - default params should be domain-specific, but there's no way to achieve that. Also, there's very little use of that, it's easier to set the domain for each test.
2. To verify all the `x-amples` we need to run tests with different domains. If the domain of the `x-ample` doesn't match currently tested domain, we need to skip that test, bc the test data is domain-dependent.
3. This adds a very tacky thing, but it's the only possible solution I could come up with. We're running tests against the labs parsoid, so the domain of all tests, requiring parsoid, should be changed to labs, while x-amples should still point to en.wiki, as they're used in production. So, to solve this I've added `tags` to x-amples, that kinda define what's being involved in the test, so that we could rewrite the domain of every test that involve parsoid, but not other tests. If there're any other ideas how to achieve the same behaviour, please tell me. 

cc @wikimedia/services

### DIFF
--- a/test/features/specification/monitoring.js
+++ b/test/features/specification/monitoring.js
@@ -50,14 +50,6 @@ function constructTests(spec, options) {
                     return;
                 }
 
-                // This is a hack. We're running tests again parsoid instance in labs
-                // which doesn't have access to en.wikipedia.org, but all of our x-amples
-                // point to en.wiki. So for tests which involve parsoid we have to manually
-                // rewrite the domain.
-                if (ex.tags && ex.tags.indexOf('parsoid') >= 0) {
-                    ex.request.params.domain = 'en.wikipedia.beta.wmflabs.org';
-                }
-
                 ret.push(constructTestCase(
                     ex.title,
                     uri.toString({params: ex.request.params}),
--- a/test/features/specification/monitoring.js
+++ b/test/features/specification/monitoring.js
@@ -25,15 +25,8 @@ function constructTestCase(title, path, method, request, response) {
     };
 }
 
-function constructTests(spec, defParams) {
+function constructTests(spec, options) {
     var paths = spec.paths;
-    if (spec['x-default-params']) {
-        Object.keys(spec['x-default-params']).forEach(function(paramName) {
-           if (!defParams[paramName]) {
-               defParams[paramName] = spec['x-default-params'][paramName];
-           }
-        });
-    }
     var ret = [];
     Object.keys(paths).forEach(function(pathStr) {
         if (!pathStr) return;
@@ -53,9 +46,21 @@ function constructTests(spec, defParams) {
             }
             p['x-amples'].forEach(function(ex) {
                 ex.request = ex.request || {};
+                if (ex.request.params && ex.request.params.domain !== options.domain) {
+                    return;
+                }
+
+                // This is a hack. We're running tests again parsoid instance in labs
+                // which doesn't have access to en.wikipedia.org, but all of our x-amples
+                // point to en.wiki. So for tests which involve parsoid we have to manually
+                // rewrite the domain.
+                if (ex.tags && ex.tags.indexOf('parsoid') >= 0) {
+                    ex.request.params.domain = 'en.wikipedia.beta.wmflabs.org';
+                }
+
                 ret.push(constructTestCase(
                     ex.title,
-                    uri.toString({params: Object.assign({}, defParams, ex.request.params || {})}),
+                    uri.toString({params: ex.request.params}),
                     method,
                     ex.request,
                     ex.response || {}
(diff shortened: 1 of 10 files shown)


d00rman:
I wouldn't be up for completely removing `x-default-params` support. It is true that currently we use it only for the `domain:` stanza, but I can see it working equally well for other stuff.

test/features/specification/monitoring.js
```diff
@@ -53,9 +46,21 @@ function constructTests(spec, defParams) {
             }
             p['x-amples'].forEach(function(ex) {
                 ex.request = ex.request || {};
+                if (ex.request.params && ex.request.params.domain !== options.domain) {
+                    return;
+                }
+
+                // This is a hack. We're running tests again parsoid instance in labs
```
d00rman: Test hacks again? Grrrr :( -1

v1/content.yaml
```diff
@@ -292,8 +291,10 @@ paths:
         #      content-type: /^text\/html.+/
         #    body: /^<!DOCTYPE html>.*/
         - title: Get html by title from storage
+          tags: [ 'content', 'parsoid' ]
```
d00rman: Why the `tags` in the test `x-amples` ?

v1/definition.yaml
```diff
@@ -0,0 +1,172 @@
...
+        # if there's no access), and so that we can use the normalized title &
+        # other info in case we have a storage miss.
+        - rev_info:
+            request:
+              method: get
+              uri: /{domain}/sys/page_revisions/page/{term}
+            response:
+              body: '{{rev_info.body.items[0]}}'
+
+          storage:
```
d00rman: So this will return the definition if it is found in storage. However, an older version might be stored here, which can create inconsistent results. I don't feel too strongly about this though, given that we have the hack in place to refresh this endpoint.

Pchelolo:
@d00rman Hm, how do you expect it to be useful? My problem with it is that's not clear what ends up being in there - it's merged from different files for a spec root, the stuff in there could be overridden in unexpected ways by other files, so for me it looks like it's too bug-prone. It's way more clean to define each x-ample on its own, 

However, I can return the support for it back, just it would be dead code for now until we find some use cases.

test/features/specification/monitoring.js
Pchelolo: Yep, don't like that too.. I'm explaining the need for this in the pr description, and if you have any suggestions how to do it better - I'm both hands up for doing this other way.

v1/definition.yaml
Pchelolo: This is a pretty standard way to do it, we do the same in summary  and, I think, in mobile apps too, so I'm not sure what's the problem here?

test/features/specification/monitoring.js
d00rman: Nah, it's only in the tests, so I can live with it :)

v1/definition.yaml
d00rman: It can introduce short-lived inconsistencies. Just pointing it out.

v1/definition.yaml
```diff
@@ -0,0 +1,172 @@
...
+          in: path
+          description: The term to define
+          type: string
+          required: true
+      responses:
+        '200':
+          description: The definition for the given term
+          schema:
+            $ref: '#/definitions/definitionsResponse'
+        '404':
```
d00rman: Also add 501 for unsupported languages.

v1/definition.yaml
```diff
@@ -0,0 +1,172 @@
...
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+
+      x-setup-handler:
+        # Set up a simple key-value bucket.
+        - init:
+            method: 'put'
+            uri: /{domain}/sys/key_value/definition
```
d00rman: `definition` seems a bit too broad to me. Perhaps something like `term.definition` would be more appropriate?

v1/definition.yaml
gwicke: The quickest fix would be to truncate the summary tables; we don't have that much data in there, and nothing high-volume depends on that content yet.

v1/definition.yaml
gwicke: Or wiktionary_definition.

Pchelolo: I'm more up for term.definition, because the `{domain}` would be resolved to `wiktionary` anyway

v1/definition.yaml
gwicke: @d00rman, I guess your concern is about updates to the definition end point lagging revision metadata updates?

v1/definition.yaml
gwicke: Good point. :+1:

PR #465, opened 2016-01-12
Switch to Node 4.2

d00rman:
We have switched our production hosts to use Node 4.2 LTS. Indicate that
in package.json so that the docker build script builds the node module
dependencies against the correct Node version.

Bug: [T107762](https://phabricator.wikimedia.org/T107762)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-    # Production runs 0.10.29
-    - "0.10.29"
     - "0.10"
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

Pchelolo:
LGTM

d00rman:
Trivial, and we need it for the next repo build, so self-merging

gwicke:
:+1:

PR #466, opened 2016-01-13
Summary: Added lang and direction to the summary endpoint

Pchelolo:
Added `lang` and `dir` properties to the summary endpoint response. The info is requested from the API.

We need to truncate the summary storage after deploying this to invalidate old content.

Bug: https://phabricator.wikimedia.org/T68114

cc @wikimedia/services

### DIFF
--- a/v1/summary.yaml
+++ b/v1/summary.yaml
@@ -93,7 +93,7 @@ paths:
               method: post
               uri: /{domain}/sys/action/query
               body:
-                prop: 'extracts|pageimages'
+                prop: 'info|extracts|pageimages'
                 redirects: true
                 exsentences: 5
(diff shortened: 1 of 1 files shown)


v1/summary.yaml
```diff
@@ -185,4 +187,10 @@ definitions:
             type: integer
             description: Thumnail height
         required: ['source', 'width', 'height']
-    required: ['title', 'extract']
+      lang:
+        type: string
+        description: The page language
```
gwicke: Perhaps: "The page language code; ex: 'en'".

gwicke:
LGTM overall. One minor documentation suggestion inline.

PR #468, opened 2016-01-13
force MediaWiki reads of revision metadata

eevans:
There is currently no way for RESTBase to know when a user has been deleted /
suppressed on a site-wide basis.  Since the `/page/revision/{revision}` endpoint
seems to be the only place that suppressable user information is returned, and
since this endpoint is low volume, this changeset proposes to (temporarily)
address this by explicitly setting the `Cache-Control: no-cache` header, (i.e.
always read revision metadata from MediaWiki).

Bug: https://phabricator.wikimedia.org/T120212

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -101,7 +101,8 @@ paths:
             request:
               uri: /{domain}/sys/page_revisions/page/{title}
               headers:
-                cache-control: '{cache-control}'
+                # FIXME: Temporary work-around (see T120212).
+                cache-control: 'no-cache'
                 if-unmodified-since: '{if-unmodified-since}'
       x-monitor: true
       x-amples:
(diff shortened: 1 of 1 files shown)


gwicke:
:+1: 

/cc @wikimedia/services

gwicke:
LGTM.

v1/content.yaml
```diff
@@ -101,7 +101,7 @@ paths:
             request:
               uri: /{domain}/sys/page_revisions/page/{title}
               headers:
-                cache-control: '{cache-control}'
+                cache-control: 'no-cache'
```
d00rman: A TODO-style comment above would be much appreciated.

v1/content.yaml
```diff
@@ -777,7 +777,7 @@ paths:
             request:
               uri: /{domain}/sys/page_revisions/rev/{revision}
               headers:
-                cache-control: '{cache-control}'
+                cache-control: 'no-cache'
```
d00rman: Idem.

d00rman:
LGTM too :+1: Minor nits in-lined.

PR #469, opened 2016-01-14
Definitions invalidation: Narrow down logging

Pchelolo:
After we've enabled the definitions endpoint, we're receiving lots of error logs for unsupported languages. 
Ignore those errors as it's ok that an underlying service doesn't implement this functionality.

cc @wikimedia/services

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -354,7 +354,7 @@ PSP.generateAndSave = function(restbase, req, format, currentContentRes) {
                     }
                 })
                 .catch(function(e) {
-                    if (e.status !== 501) {
+                    if (e.status !== 501 && e.status !== 404) {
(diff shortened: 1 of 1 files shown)


gwicke:
LGTM. Also learned about `String.prototype.indexOf` in the process.

sys/parsoid.js
```diff
@@ -354,7 +354,10 @@ PSP.generateAndSave = function(restbase, req, format, currentContentRes) {
                     }
                 })
                 .catch(function(e) {
-                    self.log('error/summary', e);
+                    if (e.status !== 501) {
```
d00rman: We might want to ignore 404s as well, since they are not technically errors from RESTBase's point of view.

PR #470, opened 2016-01-14
Disabled test

eevans:
Since we are not caching revision metadata (a temporary work-around), it does
not make sense to test the caching of revision metadata.

Bug: https://phabricator.wikimedia.org/T120212

### DIFF
--- a/test/features/pagecontent/pagecontent.js
+++ b/test/features/pagecontent/pagecontent.js
@@ -345,7 +345,9 @@ describe('page content access', function() {
             assert.deepEqual(res.status, 403);
         });
     });
-    it('should correctly store updated restrictions', function() {
+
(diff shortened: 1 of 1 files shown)


test/features/pagecontent/pagecontent.js
```diff
@@ -345,73 +345,75 @@ describe('page content access', function() {
...
-        }).then(function() {
-            throw new Error('403 should be thrown');
-        }, function(e) {
-            assert.deepEqual(e.status, 403);
-        }).then(function() { api.done(); })
-        .finally(function() { nock.cleanAll(); });
-    });
+
+// FIXME: Test disabled until cached responses are once again used (see: T120212).
+//     it('should correctly store updated restrictions', function() {
```
d00rman: You can use `it.skip` to achieve the same thing without commenting out this huge block of code.

PR #471, opened 2016-01-14
Mobileapps: Enable pre-generation

d00rman:
This PR enables pre-generation for MobileApps public endpoints. Every time RESTBase detects a page's HTML has been changed, it will send a request to Parsoid and once the result is in storage, a request is sent to the Mobile Content Service to regenerate the page's content so that it can be served directly from storage to clients.

Bug: [T121293](https://phabricator.wikimedia.org/T121293)

### DIFF
--- a/sys/mobileapps.yaml
+++ b/sys/mobileapps.yaml
@@ -8,13 +8,12 @@ paths:
             request:
               method: get
               uri: /{domain}/sys/page_revisions/page/{title}
-              headers:
-                cache-control: '{cache-control}'
         - cache_branch:
             request:
               method: post
               uri: /{domain}/sys/mobileapps/v1/handling/content/{route}/{$$.default($.request.headers.cache-control, 'none-given')}/{title}
               headers:
+                cache-control: '{cache-control}'
                 content-type: application/json
               body: '{$.get_rev.body}'
 
@@ -25,18 +24,16 @@ paths:
             request:
               method: get
               uri: '{+$$.options.host}/{domain}/v1/page/mobile-sections/{title}'
-              headers:
-                x-restbase-etag: '{$.request.body.items[0].rev}/{$.request.body.items[0].tid}'
         - store_lead:
             request:
               method: put
-              uri: /{domain}/sys/key_value/mobileapps.lead/{title}/{$.request.body.items[0].tid}
+              uri: /{domain}/sys/key_value/mobileapps.lead/{title}
               headers: '{$.get_mobileapps.headers}'
               body: '{$.get_mobileapps.body.lead}'
-        - store_remaining:
+          store_remaining:
             request:
               method: put
-              uri: /{domain}/sys/key_value/mobileapps.remaining/{title}/{$.request.body.items[0].tid}
+              uri: /{domain}/sys/key_value/mobileapps.remaining/{title}
               headers: '{$.get_mobileapps.headers}'
               body: '{$.get_mobileapps.body.remaining}'
         - return:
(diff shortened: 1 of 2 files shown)


d00rman:

sys/parsoid.js
```diff
@@ -359,6 +359,15 @@ PSP.generateAndSave = function(restbase, req, format, currentContentRes) {
                             'summary' : 'definition', e);
                     }
                 }),
+                // MobileApps pre-generation
+                restbase.get({
+                    uri: new URI([rp.domain, 'v1', 'page', 'mobile-sections', rp.title]),
```
Pchelolo: One option to consider is to go directly to `/sys/movileapps/v1/handling/content/mobile-sections/no-cache/{title}` - like this we would avoid a request to title/revision, which we don't really need, as we've already verified that content is accessible while running the parsoid request. So, a bit less load on cassandra.

d00rman: Oh, right! I wanted to do that but forgotten to change it :P

d00rman:
Hm, so now [this test](https://github.com/wikimedia/restbase/blob/8cb51a6e69fb3f0839959fbcda7c779cd5c9c216/test/features/errors/notfound.js#L83) is failing. And only when using the Cassandra back-end. This smells fishy.

sys/mobileapps.yaml
```diff
@@ -74,13 +65,17 @@ paths:
         - check_storage_lead:
             request:
               method: get
-              uri: /{domain}/sys/key_value/mobileapps.lead/{title}/{$.request.body.items[0].tid}
+              uri: /{domain}/sys/key_value/mobileapps.lead/{title}
+              headers:
+                cache-control: '{cache-control}'
```
Pchelolo: There's no need to send the header to `key_value` here, as we will never get here with a `no-cache` value of that header.

sys/mobileapps.yaml
```diff
@@ -74,13 +65,17 @@ paths:
...
             catch:
               status: 404
         - check_storage_remaining:
             request:
               method: get
-              uri: /{domain}/sys/key_value/mobileapps.remaining/{title}/{$.request.body.items[0].tid}
+              uri: /{domain}/sys/key_value/mobileapps.remaining/{title}
+              headers:
+                cache-control: '{cache-control}'
```
Pchelolo: Idem

sys/mobileapps.yaml
d00rman: Touché!

d00rman:
Updated

Pchelolo:
LGTM

sys/mobileapps.yaml
```diff
@@ -4,39 +4,30 @@ paths:
   /v1/handling/rev/{route}/{title}:
     get:
       x-request-handler:
-        - get_rev:
-            request:
-              method: get
-              uri: /{domain}/sys/page_revisions/page/{title}
-              headers:
-                cache-control: '{cache-control}'
         - cache_branch:
```
gwicke: [meta comment, not specific to this patch]: I'm still having trouble understanding all this handler branching business. Do you think we could simplify this in a future iteration?

d00rman: Hopefully, yes. I'll look it into it in a second pass.

sys/mobileapps.yaml
```diff
@@ -4,39 +4,30 @@ paths:
   /v1/handling/rev/{route}/{title}:
     get:
       x-request-handler:
-        - get_rev:
```
Pchelolo: We need to put this back..

d00rman:
@Pchelolo resurrected the revision call and made it parallel to the content fetch.

Pchelolo:
Ok, now definitely LGTM

PR #472, opened 2016-01-14
Travis: Install custom version of cassandra and run tests against it

Pchelolo:
We've run into the issue with 2.0.9 cassandra, so try installing our own custom cassandra version

cc @wikimedia/services


eevans:
LGTM :+1:

gwicke:
Very nice, @Pchelolo!

PR #473, opened 2016-01-15
Updated bluebird to 3.1.1

Pchelolo:
I've updated RB and all our dependencies to the newest `bluebird@3.1.1`. This needs to be tested in staging heavily before we publish all dependencies, so I've made restbase use all dependencies from brunches of my forks, so in case we decide not to update, rolling back would be less painful. If we decide to proceed, I'll create PRs for all our packages and we'll publish changes.

Quick testing on local machine shows, that throughput is not affected at all, while memory usage under load is a bit lower with new bluebird: 220 megs vs 240 megs for old bluebird.


Pchelolo:
Something is not working on 0.10. Investigating.

d00rman:
Temporarily merging this so we can test it in staging.

PR #476, opened 2016-01-19
Confine wiktionary updates to en.wiktionary

d00rman:
We track all of the wiktionaries and force updates for the
/page/definition end point. However, the back-end currently supports
only the English variant, so confine the updates to it, ignoring the
rest.

Bug: [T124028](https://phabricator.wikimedia.org/T124028)


Pchelolo:
LGTM, but travis seems to be out for lunch

d00rman:
hehe, something like that. I restarted it, so let's see ...

PR #479, opened 2016-01-26
Support streaming responses

gwicke:
Use cases like chunked storage and proxied requests can be sped up by
streaming the response as it comes in.

This patch adds support for returning a `Readable` stream from handlers, and
is targeted at chunked storage buckets. In the longer term, we probably want
to switch to the richer
[ReadableStream](https://streams.spec.whatwg.org/#rs-model) abstraction, but
node implementations for that are still lacking, and this gets us started with
Node's native streams.

Task: https://phabricator.wikimedia.org/T124696

### DIFF
--- a/test/framework/streaming/general.js
+++ b/test/framework/streaming/general.js
@@ -1,0 +1,44 @@
+'use strict';
+
+var assert = require('../utils/assert.js');
+var Server = require('../utils/server.js');
+var preq = require('preq');
+
+// mocha defines to avoid JSHint breakage
+/* global describe, it, before, beforeEach, after, afterEach */
+
+describe('Handler Template', function() {
+    var server = new Server('test/framework/streaming/test_config.yaml');
+    before(function() { return server.start(); });
+
+    it('Basic streaming', function () {
+        return preq.get({ uri: server.hostPort + '/test/hello' })
+        .then(function (res) {
+            assert.deepEqual(res.status, 200);
+            assert.deepEqual(res.headers['content-type'], 'text/html');
+            assert.deepEqual(res.body, 'hello');
+        });
+    });
+
+    it('Buffer streaming', function () {
+        return preq.get({ uri: server.hostPort + '/test/buffer' })
+        .then(function (res) {
+            assert.deepEqual(res.status, 200);
+            assert.deepEqual(res.headers['content-type'], 'text/html');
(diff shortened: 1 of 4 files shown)


eevans:
Could we add a test that exercises this?  LGTM otherwise.

test/framework/streaming/stream_module.js
```diff
@@ -0,0 +1,71 @@
...
+                }
+            }
+        },
+        operations: {
+            hello: hello,
+            buffer: buffer,
+            chunks: chunks,
+        }
+    };
+};
```
Pchelolo: Could be nice to add an async test, that calls `body.end()` from a timeout callback, to verify that async writing also work.

gwicke: Good idea. I made the buffer test async.

Pchelolo:
:+1:

test/framework/streaming/general.js
```diff
@@ -0,0 +1,56 @@
...
+    it('Buffer streaming', function () {
+        return preq.get({ uri: server.hostPort + '/test/buffer' })
+        .then(function (res) {
+            assert.deepEqual(res.status, 200);
+            assert.deepEqual(res.headers['content-type'], 'text/html');
+            assert.deepEqual(res.body, 'hello');
+        });
+    });
+
+    it('Buffer streaming, no compression', function () {
```
d00rman: This test is identical to the one above (except the `gzip: true` bit), so how can you tell / test that gzip wasn't used?

eevans: Would it be enough to test the `content-encoding` response header?

eevans:
Nice tests!  With the exception of @d00rman's comments on gzip vs. w/o gzip responses, :+1:

gwicke:
As far as I can tell request (and thus preq) don't expose content-encoding after transparently decoding the gzip response. I don't think we can thus tell whether gzip was used by looking at the response.

d00rman:
> As far as I can tell request (and thus preq) don't expose content-encoding after transparently decoding the gzip response. I don't think we can thus tell whether gzip was used by looking at the response.

It is possible to get it from //request//, but that would slightly complicate preq and would need intervention there. We can defer that for later.

Note: tests are now failing because `graphoid.wm.org` has been removed from DNS / VCL.

PR #480, opened 2016-01-27
Graphoid: use graphoid-beta.wmflabs.org

d00rman:
graphoid.wikimedia.org has been decommissioned, so switch to
graphoid-beta.wmflabs.org for testing.

Temporarily switch the monitoring point off, as the graphoid Labs
instance needs an update, which is blocked on havig Node 4.2 in Labs.

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -23,7 +23,7 @@ default_project: &default_project
           action:
             apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
           graphoid:
-            host: http://graphoid.wikimedia.org
+            host: http://graphoid-beta.wmflabs.org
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

gwicke:
> which is blocked on havig Node 4.2 in Labs.

Is this related to the relevant node not being on Jessie yet, or is there another issue with 4.2 not being available on Jessie?

d00rman:
> Is this related to the relevant node not being on Jessie yet, or is there another issue with 4.2 not being available on Jessie in labs?

The VMs hosting the services in Beta are on Trusty, so need to be re-imaged.

PR #481, opened 2016-01-28
Added first draft implementation of a chunked bucket

Pchelolo:
This is the first draft implementation of the chunked storage. It's almost full-functional, except minor details that I will address later, like configuration and more tests. 

Also, I've refactored `key_value` and `key_rev_value` buckets, removed unused stuff and cleaned up a little bit. And created tests for them, as they're very similar to chunked bucket tests.

Ticket: https://phabricator.wikimedia.org/T122028


Pchelolo:
cc @wikimedia/services

sys/chunked_bucket.js
```diff
@@ -0,0 +1,433 @@
...
+    return result;
+};
+
+ChunkedBucket.prototype._joinData = function(dataParts) {
+    if (!dataParts || !dataParts.length) {
+        return '';
+    } else if (Buffer.isBuffer(dataParts[0])) {
+        return Buffer.concat(dataParts);
+    } else {
+        return String.prototype.concat.apply(dataParts[0], dataParts.slice(1));
```
gwicke: This is equivalent to '`'.concat(dataParts)`, with the latter avoiding the slice.

sys/chunked_bucket.js
```diff
@@ -0,0 +1,433 @@
...
+            { attribute: 'chunk_id', type: 'hash' },
+            { attribute: 'tid', type: 'range', order: 'desc' }
+        ]
+    };
+};
+
+ChunkedBucket.prototype.createBucket = function(restbase, req) {
+    var metaSchema = this._makeMetaSchema(req.body || {});
+    var dataSchema = this._makeChunksSchema(req.body || {});
+    metaSchema.table = this._metaTableName(req);
```
gwicke: We discussed using an extension attribute ("custom"?) in the table schema to record this information. That information can then be retrieved from the internal meta table using `GET /table/{name}`, which calls [getTableSchema](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/index.js#L145). The advantage of doing it this ways is two less tables & one less keyspace per chunked bucket.

The disadvantage of this approach is the need to change the table storage backends to accept / ignore the custom attribute. But, perhaps we could just hard-code parameters for now, and only start storing them once we actually have a need to support bucket meta options?

Pchelolo: It's hard coded now, I think it's a minor issue that we could address later

gwicke: Ohh, I totally misunderstood what the meta table is about. Sorry.

sys/chunked_bucket.js
```diff
@@ -0,0 +1,433 @@
...
+ChunkedBucket.prototype._makeMetaSchema = function(opts) {
+    var schemaVersionMajor = 1;
+
+    return {
+        version: schemaVersionMajor * 1000 + (opts.version || 0),
+        options: {
+            compression: opts.compression || [
+                {
+                    algorithm: 'deflate',
+                    block_size: 1024
```
gwicke: Such huge blocks are unlikely to make sense for the meta schema. Perhaps go with 128k?

Pchelolo:
Renamed `chunked_bucket` to `key_rev_large_value`. 
There was also an idea to provide the data_size parameter when table is created and let the bucket decide what to use, but that should be per-bucket, so we'd need to store this data somewhere, and we don't support that yet.

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,433 @@
...
+    return result;
+};
+
+ChunkedBucket.prototype._joinData = function(dataParts) {
+    if (!dataParts || !dataParts.length) {
+        return '';
+    } else if (Buffer.isBuffer(dataParts[0])) {
+        return Buffer.concat(dataParts);
+    } else {
+        return ''.concat(dataParts);
```
gwicke: Argh, actually I was wrong, and the correct way would be `dataParts.join('')`:

```javascript
> ''.concat.call(['foo','bar'])
'foo,bar'
> ['foo','bar'].join('')
'foobar'
```
Sorry.

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,433 @@
...
+                    algorithm: 'deflate',
+                    block_size: 128
+                }
+            ]
+        },
+        attributes: {
+            key: opts.keyType || 'string',
+            rev: 'int',
+            tid: 'timeuuid',
+            num: 'int',
```
gwicke: Alternatives: `chunk_count`, `num_chunks`

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,433 @@
...
+        return P.all(range(metadata.num).map(function(chunkId) {
+            return restbase.get({
+                uri: new URI([rp.domain, 'sys', 'table', self._chunksTableName(req), '']),
+                body: {
+                    table: self._chunksTableName(req),
+                    attributes: {
+                        key: rp.key,
+                        chunk_id: chunkId,
+                        tid: {
+                            le: metadata.tid
```
gwicke: Shouldn't this be an exact match for now?

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,433 @@
...
+                return restbase.put({
+                    uri: new URI([req.params.domain, 'sys', 'table', self._metaTableName(req), '']),
+                    body: {
+                        table: self._metaTableName(req),
+                        attributes: Object.assign({}, row, { _ttl: GRACE_TTL })
+                    }
+                })
+                .then(function() {
+                    return P.all(range(row.num).map(function(chunkId) {
+                        return restbase.get({
```
gwicke: It might be worth moving the nested code here to a separate function.

sys/key_rev_large_value.js
gwicke: Actually, I wonder why the tests didn't catch this.

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,433 @@
...
+        },
+        index: [
+            { attribute: 'key', type: 'hash' },
+            { attribute: 'chunk_id', type: 'hash' },
+            { attribute: 'tid', type: 'range', order: 'desc' }
+        ]
+    };
+};
+
+ChunkedBucket.prototype.createBucket = function(restbase, req) {
```
gwicke: It might be nice to check that only compatible / supported key_rev_value options are passed in, so that code relying on others fails quickly.

gwicke:
Looks sane to me overall. Nice work, @Pchelolo!

gwicke:
@wikimedia/services, any objections to landing this after a rebase?

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,449 @@
...
+        return '';
+    } else if (Buffer.isBuffer(dataParts[0])) {
+        return Buffer.concat(dataParts);
+    } else {
+        return dataParts.join('');
+    }
+};
+
+ChunkedBucket.prototype._metaTableName = function(req) {
+    return req.params.bucket + '_meta';
```
d00rman: In the underlying storage implementation, we already use the term *meta*. Maybe it'd be less confusing if here we addressed this as *index* or similar?

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,449 @@
...
+function coerceTid(tidString) {
+    if (uuid.test(tidString)) {
+        return tidString;
+    }
+
+    // Out of luck
+    throw new HTTPError({
+        status: 400,
+        body: {
+            type: 'key_rev_value/invalid_tid',
```
d00rman: Did you mean(TM): `key_rev_large_value` ?

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,449 @@
...
+            title: 'Invalid tid parameter',
+            tid: tidString
+        }
+    });
+}
+
+function parseRevision(rev) {
```
d00rman: This smells like a good candidate to be moved to `mwUtil`. `coerceTid()` too actually.

sys/key_rev_large_value.js
```diff
@@ -0,0 +1,449 @@
...
+                body: {
+                    items: [ metadata ]
+                }
+            };
+        })
+        .then(returnRevision(req, metadata));
+    });
+};
+
+function getLimit(hyper, req) {
```
d00rman: Also move to `mwUtil` ?

test/features/buckets/revisioned_bucket.js
```diff
@@ -0,0 +1,169 @@
+"use strict";
```
d00rman: Did you mean(TM) `test/features/buckets/key_rev_large_bucket.js` ? :)

sys/key_rev_large_value.js
gwicke: Yeah, I proposed something similar in https://github.com/wikimedia/restbase/pull/481#discussion_r51305790.

test/features/buckets/revisioned_bucket.js
Pchelolo: No. This set of tests run both for `key_rev_value` and `key_rev_large_value`.

PR #482, opened 2016-01-28
Graphoid: re-enable the Graphoid spec test

d00rman:
`graphoid-beta.wmflabs.org` is now back and operational, so re-enable the x-ample from the Grapohid spec.


d00rman:
cc @wikimedia/services

Pchelolo:
:+1:

d00rman:
Trivial, so self-merging

PR #484, opened 2016-01-28
Refatored parsoid module and improved test coverage

Pchelolo:
Currently `parsoid` module is a mess. In this PR I've extracted some pieces of functionality to separate functions to improve understandability. Also, I've verified the access checking code paths and limited it to just 2 checks, one of which is actually unnecessary, but that one should be addressed later, it's hard. 
Also added some tests to improve coverage.


sys/parsoid.js
```diff
@@ -593,16 +616,7 @@ PSP._getOriginalContent = function(restbase, req, revision, tid) {
 PSP._getStashedContent = function(restbase, req, etag) {
     var self = this;
     var rp = req.params;
-    if (!rp.title || !rp.revision) {
```
Pchelolo: This is only called from `transfromRevision`, which is only called if we have `title` and `revision`, so this is dead code.

d00rman:
@Pchelolo mind also rebasing?

Pchelolo:
@d00rman yes, yes, there's also some weird test failure, investigating it.

PR #485, opened 2016-01-28
Set up prefixed metric trees

gwicke:
We currently mix metrics for internal and external requests with update
requests. As the performance characteristics of especially update requests
differ significantly, this makes it hard to establish actual performance for
the most important use cases.

This patch uses new child metric producer support in service-runner 1.0 to
separate metrics of primary and derived requests into three trees:

1) external requests,
2) internal update requests, and finally
3) internal service requests.

This is done by setting up two prefixed child metric producers, and passing
those to the handler to use for all metrics reporting when internal / update
requests are identified.

### DIFF
--- a/lib/restbase.js
+++ b/lib/restbase.js
@@ -33,12 +33,13 @@ function cloneRequest(req) {
     };
 }
 
-function RESTBase(options, req) {
+function RESTBase(options, req, parOptions) {
     if (options && options.constructor === RESTBase) {
         // Child instance
         var par = options;
-        this.log = par.log;
-        this.metrics = par.metrics;
+        parOptions = parOptions || {};
+        this.log = parOptions.log || par.log;
+        this.metrics = parOptions.metrics || par.metrics;
         this.reqId = par.reqId ||
             req && req.headers && req.headers['x-request-id'] ||
             utils.generateRequestId();
@@ -98,8 +99,8 @@ RESTBase.prototype.setRequestId = function(req) {
 };
 
 // Make a child instance
-RESTBase.prototype.makeChild = function(req) {
-    return new RESTBase(this, req);
+RESTBase.prototype.makeChild = function(req, options) {
+    return new RESTBase(this, req, options);
 };
 
 // A default listing handler for URIs that end in / and don't have any
@@ -247,11 +248,11 @@ RESTBase.prototype._checkInternalApiRequest = function(req) {
     }
 };
 
-RESTBase.prototype.request = function(req) {
+RESTBase.prototype.request = function(req, options) {
     if (req.method) {
         req.method = req.method.toLowerCase();
     }
-    return this._request(req);
+    return this._request(req, options);
 };
 
 RESTBase.prototype._wrapInMetrics = function(handlerPromise, match, req) {
@@ -317,7 +318,7 @@ RESTBase.prototype._wrapInAccessCheck = function(handlerPromise, match, childReq
 };
 
 // Process one request
-RESTBase.prototype._request = function(req) {
+RESTBase.prototype._request = function(req, options) {
     var self = this;
 
     // Special handling for https? requests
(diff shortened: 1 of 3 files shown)


lib/server.js
```diff
@@ -204,15 +209,25 @@ function handleRequest(opts, req, resp) {
...
+     * We implement this by passing a specialized prefixed metric producer to
+     * the handler. All derived metrics associated with a private / update
+     * request are thus recorded in .internal or .internal_update prefixed
+     * metric trees.
+     */
     if (/^(?:::ffff:)?(?:10|127)\./.test(remoteAddr)) {
-        reqOpts.metrics.increment('requests.private');
-    } else {
-        reqOpts.metrics.increment('requests.public');
+        if (req.headers['cache-control'] && /no-cache/i.test(req.headers['cache-control'])) {
```
d00rman: Should we perhaps use the UA here instead?

lib/server.js
```diff
@@ -301,7 +316,11 @@ function main(options) {
         conf: conf,
         logger: options.logger,
         log: options.logger.log.bind(options.logger),
-        metrics: options.metrics
+        metrics: options.metrics,
+        child_metrics: {
```
d00rman: I fail to see where the child metrics objects are actually used to emit messages, only the main one seems to be ever used.

lib/server.js
gwicke: Do you have a use case in mind where the current system based on standardized IP ranges does not work as intended?

lib/server.js
gwicke: Umm, see a couple lines above (227 and 229).

lib/server.js
d00rman: I'm thinking about correctness here - it is not inconceivable to have an internal client that issues a `no-cache` request. Granted, this will not happen on regular basis, but still...

gwicke:
Rebased on master. @wikimedia/services, I think this is ready to go.

lib/server.js
```diff
@@ -238,7 +243,7 @@ function handleResponse(opts, req, resp, response) {
         }
     } else {
         opts.log('error/request', {
-            req: req,
+            root_req: req,
```
Pchelolo: Since this uses the child logger we've created for this request, `root_req` field will be included by default. By overwriting it here you're bypassing logged headers whitelist we've set up when we created the child logger.

lib/restbase.js
```diff
@@ -33,12 +33,13 @@ function cloneRequest(req) {
     };
 }
 
-function RESTBase(options, req) {
+function RESTBase(options, req, parOptions) {
```
Pchelolo: I have an improvement in mind, but it's a more major refactoring, so we can deal with it as a followup. Instead of calling `request` on a parent `RESTBase` instance and making it create a chile internally, we could pull the child creation process to the server and call the `request` method directly on a child. I think that might me cleaner, as right now inside RESTBase we have access both to the parent and a child, so it's messy and hard to reason about. 

Some applies the the cloned request.

Anyway, this is just an idea to discuss, no need to block the PR on that.

lib/server.js
gwicke: Arguably, in that case the classification as an update request is correct. The definition as I see it is not "update requests only from RESTBaseUpdateJob", but "all internal update requests".

lib/restbase.js
gwicke: I considered that option too, but then decided against it. I realized that we still need to support implicit child creation so that the restbase instance for sub-requests has the proper context. We'd have to pass a flag to indicate that we don't want to create another child in the special created-child-in-server-case, or figure out some other criteria to base the decision on.

lib/server.js
gwicke: Yeah, good catch. However, I think that's actually not new in this patch. Perhaps we can address it in a follow-up?

Pchelolo: @gwicke yep, I thing the same problem existed before. I'm OK with a follow up, just let's not forget about it

PR #488, opened 2016-01-29
Documentation improvements

gwicke:
- Switch all descriptions to the non-folding yaml syntax. This preserves
  paragraphs, making the descriptions easier to read.
- Explain ETag headers and their structure in HTML end point descriptions.


Pchelolo:
:+1:

PR #490, opened 2016-01-30
Mathoid: Open the POST end point to the public

d00rman:
We have changed the POST end point to execute only tex check and sanitisation, which addresses security concern about it. Consequently, open it up to the public.

Bug: [T116147](https://phabricator.wikimedia.org/T116147)

### DIFF
--- a/v1/mathoid.yaml
+++ b/v1/mathoid.yaml
@@ -43,11 +43,6 @@ paths:
           description: Error
           schema:
             $ref: '#/definitions/problem'
-      security:
-        - header_match:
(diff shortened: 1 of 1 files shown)


d00rman:
cc @wikimedia/services @physikerwelt @Stype

physikerwelt:
:+1: I did not have any issues with api.formulasearchengine.com which was hit by several requests
https://www.mediawiki.org/wiki/Extension:Math#Configuration

gwicke:
We should probably add a conservative rate limit on this for requests from outside the cluster.

PR #491, opened 2016-01-30
Projects: Add SQLite variants

d00rman:
There is currently no way to tell RB which table back-end to use via the config, so do an ugly copy of the project YAML files and make them use the restbase-mod-table-sqlite back-end.

Amongst other things, we need this in order to support MW-Vagrant installs.


d00rman:
cc @wikimedia/services

gwicke:
Looking forward to have templating on the module name. Would make this kind of copy & paste unnecessary.

PR #493, opened 2016-02-01
Spec swagger ui cleanup

gwicke:
swaggerUI fixes:
- Consistently match file extensions.
- Don't read content as utf-8, as this corrupts binary files like images and
  fonts.

Spec updates:
- Add a summary property to all end points. This shows up in the un-expanded
  state (next to the URL), and thus provides a quick summary of the available
  functionality.
- Add response schemas to all end points, even if they just return a string.
- Describe response headers.
- Remove listings at html/ and data-parsoid/: These were aliased to title/,
  and it's not clear that we would want to separate them later.
- Only support retrieval of data-parsoid with exact revision/tid. Any
  other access is likely to lead to client bugs & unintentional
  corruption, as the ids used to reference HTML elements are not stable
  between renders. I am also not aware of any legitimate use case for
  retrieving data-parsoid without also retrieving the corresponding HTML.
- Remove the html2html end point until this is actually supported in
  Parsoid. See https://phabricator.wikimedia.org/T114413 for progress.
- Improve JSON schema for listings, as well as spec references to two other
  definitions.

Test fixes to reflect spec changes.

### DIFF
--- a/lib/swaggerUI.js
+++ b/lib/swaggerUI.js
@@ -18,11 +18,12 @@ function staticServe(hyper, req) {
         throw new Error("Invalid path.");
     }
 
-    return fs.readFileAsync(filePath, 'utf8')
+    return fs.readFileAsync(filePath)
     .then(function(body) {
         if (reqPath === '/index.html') {
             // Rewrite the HTML to use a query string
-            body = body.replace(/((?:src|href)=['"])/g, '$1?doc=&path=')
+            body = body.toString()
+                .replace(/((?:src|href)=['"])/g, '$1?doc=&path=')
                 // Some self-promotion
                 .replace(/<a id="logo".*?<\/a>/,
                         '<a id="logo" href="https://www.mediawiki.org/wiki/RESTBase">RESTBase</a>')
@@ -38,14 +39,17 @@ function staticServe(hyper, req) {
         var contentType = 'text/html';
         if (/\.js$/.test(reqPath)) {
             contentType = 'text/javascript';
-            body = body.replace(/underscore\-min\.map/, '?doc=&path=lib/underscore-min.map');
-        } else if (/\.png/.test(reqPath)) {
+            body = body.toString()
+                .replace(/underscore\-min\.map/, '?doc=&path=lib/underscore-min.map');
+        } else if (/\.png$/.test(reqPath)) {
             contentType = 'image/png';
         } else if (/\.map$/.test(reqPath)) {
             contentType = 'application/json';
-        } else if (/\.css/.test(reqPath)) {
+        } else if (/\.ttf$/.test(reqPath)) {
+            contentType = 'application/x-font-ttf';
+        } else if (/\.css$/.test(reqPath)) {
             contentType = 'text/css';
-            body = body.replace(/\.\.\/(images|fonts)\//g, '?doc&path=$1/');
+            body = body.toString().replace(/\.\.\/(images|fonts)\//g, '?doc&path=$1/');
         }
         return P.resolve({
             status: 200,
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -368,7 +368,8 @@ paths:
         '200':
           description: The content was not changed, and no new version was created.
           schema:
-            type: string
+            type: object
+            # FIXME: document
         '201':
           description: A new revision of the page has been created
           headers:
@@ -489,7 +490,11 @@ paths:
             [/transform/html/to/wikitext{/title}{/revision}](#!/Transforms/transform_html_to_wikitext__title___revision__post)
             entry point.
           schema:
-            type: string
+            oneOf:
+              - type: string
+              - type: object
+                  patternProperties:
+                    "^mw\w+$": string
           headers:
             ETag:
               description: |
@@ -908,7 +913,8 @@ paths:
         '201':
           description: A new revision of the page has been created
           schema:
-            type: string
+            type: object
+            # FIXME: document
           headers:
             ETag:
               description: |
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -178,37 +178,37 @@ paths:
                 page: '{page}'
       x-monitor: false
 
-#  /{module:page}/html/:
-#    get:
-#      tags:
-#        - Page content
-#      summary: List titles available as HTML.
-#      description: |
-#        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
-#      produces:
-#        - application/json
-#      parameters:
-#        - name: page
-#          in: query
-#          description: Next page token, provided by _links.next.href property.
-#          type: string
-#          required: false
-#      responses:
-#        '200':
-#          description: A list of page titles.
-#          schema:
-#            $ref: '#/definitions/listing'
-#        default:
-#          description: Error
-#          schema:
-#            $ref: '#/definitions/problem'
-#      x-request-handler:
-#        - get_from_backend:
-#            request:
-#              uri: /{domain}/sys/page_revisions/page/
-#              query:
-#                page: '{page}'
-#      x-monitor: false
+  /{module:page}/html/:
+    get:
+      tags:
+        - Page content
+      summary: List titles available as HTML.
+      description: |
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      produces:
+        - application/json
+      parameters:
+        - name: page
+          in: query
+          description: Next page token, provided by _links.next.href property.
+          type: string
+          required: false
+      responses:
+        '200':
+          description: A list of page titles.
+          schema:
+            $ref: '#/definitions/listing'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_backend:
+            request:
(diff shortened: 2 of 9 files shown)


gwicke:
This patch removes some API end points that were previously marked as "unstable", so we should communicate these changes ahead of time. According to graphite, the only entry point that is actually getting requests is https://en.wikipedia.org/api/rest_v1/?doc#!/Page_content/get_page_data_parsoid_title_revision_tid, and in that case the only change is that the `tid` becomes mandatory.

d00rman:
> Only support retrieval of data-parsoid with exact revision/tid. Any other access is likely to lead to client bugs & unintentional corruption, as the ids used to reference HTML elements are not stable between renders. I am also not aware of any legitimate use case for retrieving data-parsoid without also retrieving the corresponding HTML.

I'm wondering if it would be better to expose the `page-bundle` instead in that case. `data-parsoid` is meaningless without the exact corresponding HTML render.

d00rman:
> According to graphite, the only entry point that is actually getting requests is https://en.wikipedia.org/api/rest_v1/?doc#!/Page_content/get_page_data_parsoid_title_revision_tid, and in that case the only change is that the tid becomes mandatory.

With our without the `tid` ? If with, then there's no real change for the client(s) using it.

v1/content.yaml
```diff
@@ -487,8 +477,30 @@ paths:
...
+            spec](https://www.mediawiki.org/wiki/Parsoid/MediaWiki_DOM_spec) for a
+            description of the MediaWiki-specific semantic markup in this HTML.
+            Note that additional metadata is available in the HTML head.
+
+            This HTML can be edited using arbitrary HTML tools. The modified HTML
+            can be converted back to wikitext using the
+            [/transform/html/to/wikitext{/title}{/revision}](#!/Transforms/transform_html_to_wikitext__title___revision__post)
+            entry point.
+          schema:
+            type: string
```
Pchelolo: This is not 100% precise: in case `sections` property is provided, this returns JSON.

v1/content.yaml
```diff
@@ -366,9 +366,18 @@ paths:
         - application/json
       responses:
         '200':
-          description: The existing revision of the page matches the sent text
+          description: The content was not changed, and no new version was created.
+          schema:
+            type: string
```
Pchelolo: This is not 100% precise: in case sections property is provided, this returns JSON.

gwicke:
> With our without the tid ? If with, then there's no real change for the client(s) using it.

I think with tid, as anything else would run the risk of returning mis-matched data-parsoid. Basically, if they don't pass the tid, it would be a serious bug in the client. Removing the possibility of requesting by revision only avoids that class of client bugs.

v1/content.yaml
```diff
@@ -518,104 +530,114 @@ paths:
...
+#            status: 200
+#            headers:
+#              etag: /.+/
+#              content-type: application/json
+#            body:
+#              counter: /\d+/
+#              ids: /.*/
+#              sectionOffsets: /.*/
 
   /{module:page}/data-parsoid/{title}/:
```
Pchelolo: Since we store `html` and `data-parsoid` together, this endpoint will always have the same result as `html` listing. So may be we could remove it too?

gwicke:
> I'm wondering if it would be better to expose the page-bundle instead in that case. data-parsoid is meaningless without the exact corresponding HTML render.

From a performance perspective the current solution should be better, as

a) the HTML will likely be cached (once we allow caching / start purging), and
b) we don't need to deal with dynamic assembly of large JSON blobs from separately stored fragments. Instead, we can stream out data-parsoid as a compressed buffer, after the current / archival separation is implemented.

v1/content.yaml
```diff
@@ -518,104 +530,114 @@ paths:
...
+#              title: Foobar
+#          response:
+#            status: 200
+#            headers:
+#              etag: /.+/
+#              content-type: application/json
+#            body:
+#              counter: /\d+/
+#              ids: /.*/
+#              sectionOffsets: /.*/
```
Pchelolo: Does it make sense to comment it out and not just remove the code (and also remove the `sys` endpoints it uses? If we have not intention to return it back - why keep dead code?

v1/graphoid.yaml
```diff
@@ -31,6 +32,10 @@ paths:
       responses:
         '200':
           description: The PNG render of the requested graph.
+          schema:
+            type: blob
+          headers:
+            ETag: {}
```
Pchelolo: Is that left empty on purpose?

v1/mobileapps.yaml
```diff
@@ -146,7 +172,16 @@ paths:
           required: true
       responses:
         '200':
-          description: The JSON comprising the HTML and various page attributes for the given page title.
+          description: >
+            JSON comprising the HTML and various page attributes for the given page title.
+          schema:
+            type: object
```
Pchelolo: FIXME?

v1/content.yaml
gwicke: There is still a chance that one of `data-parsoid`, `sections` (and soon `data-mw`) fail, resulting in some metadata without corresponding HTML. We currently also delete html and data-parsoid separately, also creating the chance of a mis-match.

Pchelolo: But should we expose existence of these semi-corrupted content at all? How could it be useful?

v1/graphoid.yaml
gwicke: Yeah, I didn't have much descriptive to say here.

v1/content.yaml
gwicke: I'm happy to remove it if you are all fine with the change.

v1/content.yaml
gwicke: For most clients, I agree that it's not very useful. For debugging, we can also query the db directly.

If there is general agreement I'd be +1 on removing these listings as well.

v1/graphoid.yaml
Pchelolo: @gwicke hm, actually graphoid endpoint doesn't expose an `etag` header at all.

gwicke: FIXME? ;)

d00rman:
> From a performance perspective the current solution should be better, as
> 
> a) the HTML will likely be cached (once we allow caching / start purging), and
> b) we don't need to deal with dynamic assembly of large JSON blobs from separately stored fragments. Instead, we can stream out data-parsoid as a compressed buffer, after the current / archival separation is implemented.

I agree, but the API is about the client's needs first an foremost. If the client is interested in `data-parsoid`, that basically means that they want to put the two together. Parsoid already gives us the page bundle as a whole, so storing that instead of data-parsoid might be a better solution.

gwicke:
> Parsoid already gives us the page bundle as a whole, so storing that instead of data-parsoid might be a better solution.

This would roughly double the storage size, for questionable gain.

The only client currently interested in data-parsoid is Google, and they have no trouble making two requests. data-mw will have more takers, but clients like VE only want to fetch data-mw when they need it. If the page is using Parsoid HTML in the first place they won't fetch that at all, and if it doesn't, they'll want to prioritize the HTML for fast editor startup. Data-mw is only needed for specific functionality like template editing, so can be loaded asynchronously.

v1/content.yaml
d00rman: +1 to removing it instead of commenting these out.

lib/swaggerUI.js
```diff
@@ -38,14 +39,17 @@ function staticServe(hyper, req) {
...
         if (/\.js$/.test(reqPath)) {
             contentType = 'text/javascript';
-            body = body.replace(/underscore\-min\.map/, '?doc=&path=lib/underscore-min.map');
-        } else if (/\.png/.test(reqPath)) {
+            body = body.toString()
+                .replace(/underscore\-min\.map/, '?doc=&path=lib/underscore-min.map');
+        } else if (/\.png$/.test(reqPath)) {
             contentType = 'image/png';
         } else if (/\.map$/.test(reqPath)) {
             contentType = 'application/json';
```
Pchelolo: The `.map` file is actually UTF-8 content, so we might want to `.toString()` it

v1/content.yaml
gwicke: I'm not completely sure how to model this.

gwicke: http://spacetelescope.github.io/understanding-json-schema/reference/combining.html suggests that should work, too. Patch coming.

Pchelolo:
```
schema:
   oneOf:
      - type: string
      - type: object
        patternProperties:
            "^mw\w+$": string
```
Could work I think, but I'm not sure if we want to complicate things that much

v1/content.yaml
gwicke: This is the save end point. Are you sure we return the saved content?

gwicke: Based on https://github.com/wikimedia/restbase/blob/master/sys/page_save.js#L185 and saveWikitext further up I'm pretty sure that this is returning the (massaged) PHP API response, rather than a string.

lib/swaggerUI.js
d00rman: Why?

Pchelolo: hm.. Actually disregard my comment

gwicke:
Here is how the docs look in overview mode now:

image

PR #494, opened 2016-02-01
Logging: Use child request-aware logger whenever possible

Pchelolo:
Right now we provide a logger to modules via constructor options. However, that's actually wrong: modules a singletons, so we're actually providing the root logger, that doesn't have all the useful data about the request we put to the child loggers. Instead, we need to use a logger exposed by a `hyper` instance, provided to the handler. This instance always uses a child, request-aware logger.

This is a WIP to discuss the approach actually. I think we need to stop providing modules the root logger in options completely, so that there's no ambiguity. But in order to achieve that, I need to update storage modules to use a logger from `hyper` object, not the one from options. 

@gwicke @d00rman @eevans What do you think? Should I proceed here?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -137,7 +137,6 @@ function handleResponse(opts, req, resp, response) {
         }
         opts.log(logLevel, {
             message: response.message,
-            root_req: req,
             res: response,
             stack: response.stack
         });
@@ -232,7 +231,6 @@ function handleResponse(opts, req, resp, response) {
         } else {
             if (response.headers['content-length']) {
                 opts.log('warn/response/content-length', {
-                    root_req: req,
                     res: response,
                     reason: new Error('Invalid content-length')
                 });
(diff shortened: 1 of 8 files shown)


lib/server.js
```diff
@@ -137,7 +137,6 @@ function handleResponse(opts, req, resp, response) {
         }
         opts.log(logLevel, {
             message: response.message,
-            root_req: req,
```
Pchelolo: This was actually overwriting the `root_req` property we've set when we created this child logger and bypasses the header whitelist we've set up.

d00rman:
The ideal would be to combine both methods I guess. It is true that having a singleton implies we cannot have additional logging entities. On the other hand, exposing the `hyper` object means that we'd need to pass it around to all internal methods just so that we can log stuff.

Pchelolo:
@d00rman in RESTBase code it's not a problem as you see, but in storage modules code it's quite an issue, so I tend to agree we need to keep both. Luckily this PR doesn't remove the ability use global logger, just fixes what we use in modules in RESTBase

eevans:
> LGTM as is for now :+1: @gwicke @eevans any objections against merging this one?

No objections here.

PR #495, opened 2016-02-02
Load modules relative to the app base path

Pchelolo:
Currently in module/spec loading we rely on the fact that `router` is located in `/lib/router.js`. However, when we move lib directory to the framework, that will not be the case any more. To solve the problem we will get an appBasePath from service-runner and resolve everything from it.

Tests will not pass until https://github.com/wikimedia/service-runner/pull/80 is merged and published.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -52,24 +52,6 @@ function Router(options) {
     this.router = new SwaggerRouter();
 }
 
-// Load & parse a yaml spec from disk
-Router.prototype._readSpec = function(path) {
-    var fsPath = __dirname + '/../';
-    if (/^\//.test(path)) {
-        // absolute path
-        fsPath = path;
-    } else if (/\.yaml$/.test(path)) {
-        fsPath += path;
-    } else {
-        fsPath += 'specs/' + path + '.yaml';
-    }
-    // Let errors bubble up for now. Fail loud & early.
-    return fs.readFileAsync(fsPath)
-    .then(function(yamlString) {
-        return yaml.safeLoad(yamlString);
-    });
-};
-
 // Extend an existing route tree with a new path by walking the existing tree
 // and inserting new subtrees at the desired location.
 Router.prototype._buildPath = function route(node, path, value) {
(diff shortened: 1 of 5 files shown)


test/features/router/buildTree.js
```diff
@@ -20,71 +20,20 @@ var rootSpec = {
     }
 };
 
-// x-subspec and x-subspecs is no longer supported.
-var faultySpec = {
```
Pchelolo: Not really related, just a small cleanup

Pchelolo:
cc @wikimedia/services

d00rman:
Hm, this will not work in prod, because `basePath` there is one level up from the repo's location in the directory tree. The way to get around this is to tell `service-runner` what `basePath` really is, as it depends on the environment it is run in.

lib/router.js
```diff
@@ -558,35 +542,27 @@ Router.prototype.handleResources = function(hyper) {
...
+ * @param {Object} spec specification object
+ * @param {HyperSwitch} hyper HyperSwitch context
+ * @returns {Promise.<Router>} a router with loaded spec
  */
 Router.prototype.loadSpec = function(spec, hyper) {
     var self = this;
     var rootNode = new Node();
-    var specPromise;
-    if (spec && spec.constructor === String) {
-        specPromise = this._readSpec(spec);
```
Pchelolo: This doesn't happen in reality, so it's dead code and can be removed.

Pchelolo:
This would require a config change on next deploy, because now we try to load all modules from the base_path, which in production is in the deploy repo root. So, we need to add an `app_base_path` property to the config.

d00rman:
@Pchelolo I think this deserves a version bump to articulate a slight incompatibility with 0.9.x. Let's go for 0.10.0 ?

Pchelolo:
@d00rman done

d00rman:
@eevans @gwicke any objections against merging this? Mental note: we'll need config changes for this.

PR #497, opened 2016-02-03
Consistently use the child hyperswitch's logger / metrics reporter

gwicke:
We are passing child loggers & metrics reporters into hyperswitch, which are
then used to set up the new child hyperswitch instance. The intention is to
separate logs / metrics for internal, internal update & external requests.

However, we neglected to then consistently use the child hyperswitch's logger
and metrics reporter, causing top-level entry points to all end up in the
default logger / metrics reporter, rather than being separated properly into
classes.

### DIFF
--- a/lib/hyperswitch.js
+++ b/lib/hyperswitch.js
@@ -380,7 +380,7 @@ HyperSwitch.prototype._request = function(req, options) {
 
         reqHandlerPromise = childHyperSwitch._wrapInMetrics(reqHandlerPromise, match, req)
         .then(function(res) {
-            childHyperSwitch.log('trace', {
+            childHyperSwitch.log('trace/hyper/response', {
                 req: req,
                 res: res,
                 request_id: childHyperSwitch.reqId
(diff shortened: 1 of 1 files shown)


Pchelolo:
LGTM

lib/hyperswitch.js
```diff
@@ -378,9 +378,9 @@ HyperSwitch.prototype._request = function(req, options) {
             reqHandlerPromise = P.try(handler, [childHyperSwitch, childReq]);
         }
 
-        reqHandlerPromise = self._wrapInMetrics(reqHandlerPromise, match, req)
+        reqHandlerPromise = childHyperSwitch._wrapInMetrics(reqHandlerPromise, match, req)
         .then(function(res) {
-            self.log('trace', {
+            childHyperSwitch.log('trace', {
```
d00rman: Since we are here, maybe also add a logging component to this log entry: `trace/request` perhaps?

gwicke: I made it trace/hyper/response, since technically what we are tracing here is a successful response.

PR #498, opened 2016-02-03
Metrics naming consistency: Rename 'private' to 'internal'

gwicke:
Should only be merged / deployed in coordination with a rename of the graphite
metrics.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -398,7 +398,7 @@ function main(options) {
         log: options.logger.log.bind(options.logger),
         metrics: options.metrics,
         child_metrics: {
-            internal: options.metrics.makeChild('private'),
+            internal: options.metrics.makeChild('internal'),
             internal_update: options.metrics.makeChild('internal_update'),
         }
     };


Pchelolo:
@gwicke I'm wondering if it would be more consistent if external requests have an `external` prefix? Right now it's not really obvious that non-prefixed metrics are external.

gwicke:
@Pchelolo, we could consider changing that too, but the trade-offs involved in that change are different. We'd have to update a lot of dashboards to use the new 'external' prefix. On the plus side, summing all requests (internal, internal_update, external) would be simpler that way.

d00rman:
> I'm wondering if it would be more consistent if external requests have an external prefix? Right now it's not really obvious that non-prefixed metrics are external.

+1

> we could consider changing that too, but the trade-offs involved in that change are different. We'd have to update a lot of dashboards to use the new 'external' prefix.

That's true, but yesterday @eevans started updating some of the dashboards, and it seems to me now that we might want to add some new dashboards / rows for different prefixes (e.g. parsoid performance is mostly relevant to internal-update requests, while storage 2xx latencies are important for external users mostly). I have a feeling we'll need to change quite a lot of stuff regardless of the `external` prefix.

> On the plus side, summing all requests (internal, internal_update, external) would be simpler that way.

Yup, that's a bit messy / clumsy in the current state.

gwicke:
Okay, so we'll need to deploy this in coordination with an ops, mostly likely Filippo, so that we can rename the metrics in sync with the deploy to minimize metric loss.

I'm +1 for adding the external/ prefix as well, we'll just need to spend some time doing those dashboard fix-ups.

d00rman:
cc @filippog for context and background :)

filippog:
:+1:

I'll rename `private` to `internal` just before restbase restarts, that should minimize the loss

d00rman:
Great, cheers @filippog ! @gwicke how about getting that `external` prefix in as well?

gwicke:
Seperseded by https://github.com/wikimedia/hyperswitch/pull/6, which also includes the `external` prefix change.

PR #499, opened 2016-02-03
Remove deprecated scrubWikitext parameter

Pchelolo:
We've deprecated `scrubWikitext` parameter long ago. Recently I've added the logging of remaining users. After it's been deployed, there wasn't a single usage of a deprecated parameter for several days, so it's safe to remove it.

Additionally, we've switched to v3 parsoid API, that officially supports `scrub_wikitext` and not `scrubWikitext`. So stop using old format and switch to new (see https://www.mediawiki.org/wiki/Parsoid/API) 

cc @wikimedia/services

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -751,13 +751,7 @@ PSP.transformRevision = function(hyper, req, from, to) {
             body2[from] = req.body[from];
         }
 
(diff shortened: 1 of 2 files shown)


gwicke:
LGTM.

PR #501, opened 2016-02-04
Description update to use aqs_default in prod

milimetric:
Also added content-type headers to all responses because hyperswitch
doesn't add any by default (it shouldn't).  See the mailing list:

https://lists.wikimedia.org/pipermail/analytics/2016-February/004908.html

Bug: T122249

### DIFF
--- a/projects/aqs_default.yaml
+++ b/projects/aqs_default.yaml
@@ -7,7 +7,7 @@ paths:
       version: 1.0.0-beta
       title: Analytics REST API
       description: >
-          Sample project for AQS RESTBase setup
+          AQS RESTBase setup
       x-is-api-root: true
 
     x-host-basePath: /api/rest_v1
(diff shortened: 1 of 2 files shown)


gwicke:
Strictly speaking UTF-8 is [implied by application/json](https://www.ietf.org/rfc/rfc4627.txt):

> JSON text SHALL be encoded in Unicode.  The default encoding is UTF-8.

So, one could argue that Chrome should default to that when displaying JSON. However, it's only adding a couple bytes, and if it helps read the JSON more comfortably in Chrome.. why not.

gwicke:
:+1: overall, but waiting for travis.

d00rman:
> overall, but waiting for travis.

yeah, me too... :)

milimetric:
i was a bit surprised by this bug too.  Encoding problems ftl :/

d00rman:
Travis seems to be out. I checked it locally and all's good (as expected), so merging.

PR #503, opened 2016-02-05
Cleanup: Removed unused capability to provide TID as timestamp

Pchelolo:
We've had ability to pass TID to buckets as timestamp strings, but we've never used this capability, so I'm removing it.

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -50,27 +50,20 @@ mwUtil.parseETag = function(etag) {
     }
 };
 
-/**
- * Creates a deterministic version 1 UUID from a given date
- * @param {Date|string|Number} date a Date
- * @returns {string} a deterministic v1 UUID
- */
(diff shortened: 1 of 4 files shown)


d00rman:
Hm, it is true that we never explicitly use it inside the code, but we provide the ability to users to supply human-friendly variants. AFAICT, this change disallows that, which constitutes a breaking API change.

Pchelolo:
@d00rman Actually, allowing that is bad and can lead to misunderstandings. When we insert the data, we use `TimeUuid.now()` function, which is taking random `nodeId` and has sub-millisecond resolution. So if users knows the timestamp and what's to use it as tid 99.999% of situations he will get 404, because in `tidFromDate` we use a fixed `nodeId`, so you need to be extremely lucky to match the one which was randomly chosen on insertion.

d00rman:
Ah, yes, good point. We're doing exact queries, not range queries.

PR #504, opened 2016-02-08
Fix new AQS config

milimetric:


Pchelolo:
@milimetric I'm quite sure you're need to update the page views tests and remove the `metrics` here :   https://github.com/wikimedia/restbase/blob/master/test/features/pageviews/pageviews.js#L17-L25

Pchelolo:
LGTM @milimetric

PR #505, opened 2016-02-09
Chunked storage response streaming

gwicke:
This patch changes the behavior of read requests to stream body chunks as they
become available, rather than collecting & concatenating all chunks before
returning a response.

This patch also bumps up the default chunk size to 1.8mb, in line with the
benchmark results from https://phabricator.wikimedia.org/T122028. This won't
provide special compression efficiency with deflate, but will provide a decent
combination of compression ratio and scalability using brotli compression.

### DIFF
--- a/sys/key_rev_large_value.js
+++ b/sys/key_rev_large_value.js
@@ -6,6 +6,7 @@ var mwUtil = require('../lib/mwUtil');
 var HTTPError = require('../lib/exports').HTTPError;
 var URI = require('swagger-router').URI;
 var preq = require('preq');
+var stream = require('stream');
 
 var yaml = require('js-yaml');
 var fs = require('fs');
@@ -16,7 +17,7 @@ var spec = yaml.safeLoad(fs.readFileSync(__dirname + '/key_rev_value.yaml'));
  * @type {number}
  * @const
  */
-var CHUNK_SIZE = 31000;
+var CHUNK_SIZE = 1800000;
 
 /**
  * The grace_ttl parameter for a revision policy
@@ -75,7 +76,7 @@ ChunkedBucket.prototype._makeMetaSchema = function(opts) {
             compression: opts.compression || [
                 {
                     algorithm: 'deflate',
-                    block_size: 128
+                    block_size: 256
                 }
             ]
         },
@@ -103,7 +104,7 @@ ChunkedBucket.prototype._makeChunksSchema = function(opts) {
             compression: opts.compression || [
                 {
                     algorithm: 'deflate',
-                    block_size: 1024
+                    block_size: 256
                 }
             ]
         },
@@ -157,11 +158,6 @@ function returnRevision(req, metadata) {
     return function(dbResult) {
         if (dbResult.body && dbResult.body.items && dbResult.body.items.length) {
             var row = dbResult.body.items[0];
-            return {
-                status: 200,
-                headers: metadata.headers,
-                body: row.value
-            };
         } else {
             throw new HTTPError({
                 status: 404,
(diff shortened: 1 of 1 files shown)


sys/key_rev_large_value.js
```diff
@@ -157,11 +158,6 @@ function returnRevision(req, metadata) {
     return function(dbResult) {
         if (dbResult.body && dbResult.body.items && dbResult.body.items.length) {
             var row = dbResult.body.items[0];
-            return {
```
d00rman: euh, so this function returns nothing any more?

sys/key_rev_large_value.js
```diff
@@ -232,19 +212,18 @@ ChunkedBucket.prototype.getRevision = function(hyper, req) {
...
-            metadata.value = self._joinData(sections);
-            return {
-                status: 200,
-                body: {
-                    items: [ metadata ]
-                }
-            };
         })
-        .then(returnRevision(req, metadata));
+        .then(function() { return byteStream.end(); });
```
Pchelolo: We need to add a `catch` here to avoid unhandled exceptions

Pchelolo:
Oh, already needs a rebase..

sys/key_rev_large_value.js
```diff
@@ -218,13 +195,16 @@ ChunkedBucket.prototype._getMetadata = function(hyper, req) {
 ChunkedBucket.prototype.getRevision = function(hyper, req) {
     var rp = req.params;
     var self = this;
+    var chunksTable = self._chunksTableName(req);
+    var chunkURI = new URI([rp.domain, 'sys', 'table', chunksTable, '']);
     return this._getMetadata(hyper, req)
     .then(function(metadata) {
-        return P.all(range(metadata.num_chunks).map(function(chunkId) {
+        var byteStream = new stream.PassThrough();
+        P.each(range(metadata.num_chunks), function(chunkId) {
```
Pchelolo: A comment about `missing return on purpose` could be nice.

sys/key_rev_large_value.js
```diff
@@ -232,19 +212,24 @@ ChunkedBucket.prototype.getRevision = function(hyper, req) {
...
-            };
         })
-        .then(returnRevision(req, metadata));
+        .then(function() { byteStream.end(); })
+        .catch(function(e) {
+            hyper.log('error/key_rev_large_value/body', e);
+            // FIXME: Figure out a way to abort the response, but don't signal
+            // success via .end().
+        });
+
```
d00rman: Nit: too many empty lines here.

sys/key_rev_large_value.js
d00rman: +1

d00rman:
As it currently stands, we can only ever return a 200. I'm not sure that's good, tbh. I'd prefer for us to find a way to communicate errors in a sane way before embarking on this *streamisation* of the code base.

gwicke:
We can (and do) return errors if the metadata is not found (requested item does not exist). With streaming, we inherently can't signal a different HTTP response code if something goes wrong while streaming a body part. It would be nice, however, to be able to force-close the connection in a way that is guaranteed to be picked up as an error by the client, rather than letting the connection time out.

d00rman:
If the metada is found, but the chunks aren't there, we'd return a 200 with an empty body here.

gwicke:
> If the metada is found, but the chunks aren't there, we'd return a 200 with an empty body here.

More precisely, the client response would time out.

gwicke:
Closing / archiving this, as we found that per-chunk overheads dominated at small chunk sizes. We could revive this if we had a need to store & stream really large chunks (such a media).

PR #508, opened 2016-02-09
Deploy build: Fix the node version to 4.2.4

d00rman:
The deploy-repo script's nvm uses the latest 4.2.x when '4.2' is
specified, so fix the version to 4.2.4, which is what we use exactly in
production.


d00rman:
cc @wikimedia/services

Pchelolo:
:+1:

PR #509, opened 2016-02-09
Returned paging tokens encoding/decoding back to RESTBase from Hyper

Pchelolo:
The `encodeToken` and `decodeToken` functions were in `HyperSwitch` object only because they need to access the `salt`, however they don't really belong neither in `HyperSwitch` object, nor in the framework itself, because JWT signing is RESTBase-specific way of obtaining paging tokens.

Pulled back to RESTBase codebase.

### DIFF
--- a/sys/page_revisions.js
+++ b/sys/page_revisions.js
@@ -143,7 +143,7 @@ PRS.prototype.listTitles = function(hyper, req, options) {
     };
 
     if (req.query.page) {
-        Object.assign(listReq.body, hyper.decodeToken(req.query.page));
+        Object.assign(listReq.body, mwUtil.decodePagingToken(hyper, req.query.page));
     }
 
     return hyper.get(listReq)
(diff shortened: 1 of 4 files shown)


sys/page_revisions.js
```diff
@@ -130,6 +130,7 @@ PRS.prototype._checkRevReturn = function(item) {
 
 // /page/
 PRS.prototype.listTitles = function(hyper, req, options) {
+    console.log(hyper.config.salt);
```
d00rman: Hehe

Pchelolo: Ups..

d00rman:
As is, I don't think it's actually correct - throwing 500s back to the client because of our own fault seems to be the wrong thing to do(TM). However, that said, I'm ok with merging this if we implement this in a much saner way, say via *start-up HyperSwitch modules* which could check for the missing parameter.

@eevans @gwicke any objections against merging?

d00rman:
I put down some ideas in [T126417](https://phabricator.wikimedia.org/T126417) regarding the start-up procedure of HyperSwitch which would allow us to validate the `salt` config option.

gwicke:
I'd be in favor of merging this for now. There are quite a few other options we don't validate on startup (including various backend host names); while I agree that it's a lot easier to validate a salt, it's (IMHO) not the end of the world if it throws a run-time error until we get around to do this in a nicer way.

PR #511, opened 2016-02-11
Enable basic Varnish caching for mobileapps end points

gwicke:
- Allow Varnish caching for one hour by default.
- Simplify mobileapps backend code, and fix one edge case where a 404
  for the lead section would be ignored if the remainder existed.

### DIFF
--- a/v1/mobileapps.yaml
+++ b/v1/mobileapps.yaml
@@ -40,12 +40,22 @@ paths:
           schema:
             $ref: '#/definitions/problem'
       x-request-handler:
-        - get_from_backend:
+        - access_check:
             request:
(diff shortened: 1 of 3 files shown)


gwicke:
See https://phabricator.wikimedia.org/T113591.

v1/mobileapps.yaml
```diff
@@ -193,10 +219,16 @@ paths:
           schema:
             $ref: '#/definitions/problem'
       x-request-handler:
-        - get_from_backend:
+        - from_backend:
```
Pchelolo: We've lost the access check here

PR #512, opened 2016-02-11
Use new config format to make modules export valid swagger spec

Pchelolo:
As we've discussed, current config situation is bad - all of the configs are invalid swagger specs. I've made a change to `HyperSwitch` to change the config format one more last time. Now every module exports a valid swagger spec, modules can be nested in whatever way user wants, and, what's most important, the API is very simple and straightforward: You have a valid swagger spec, and on a path you can either specify handlers, or load a set of modules which will be bound to the current path. Modules can also be inlined if you don't want to have too many files.

This is still work in progress, but I've created this PR to give a glance on how our configs will look like when it's all finished.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -4,43 +4,40 @@
 # in the root_spec further down.
 default_project: &default_project
   x-modules:
-    /:
-      - path: projects/wmf_default.yaml
-        options: &default_options
-          table:
-            hosts: [localhost]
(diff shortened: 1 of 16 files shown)


gwicke:
The changes look good to me. This depends on https://github.com/wikimedia/hyperswitch/pull/8.

d00rman:
Published as [v0.11.0](https://github.com/wikimedia/restbase/tree/v0.11.0)

PR #513, opened 2016-02-12
MobileApps: propagate the headers received from the service

d00rman:
In PR #511 we added the caching headers to the mobile-sections end points' responses. However, these were overwriting the headers returned by the service itself (such as ETag). This brings them back and merges them with our cache-control response header.

Bug: [T113591](https://phabricator.wikimedia.org/T113591)


Pchelolo:
LGTM

PR #514, opened 2016-02-12
Completely removed parsoid-lb endpoint from the codebase

Pchelolo:
We've switched all tests to beta parsoid long time ago, but still used `parsoid-lb` endpoint for manual local testing. Now it's being decommissioned, so I'm removing it completely from the codebase. 

The main reason for keeping it was that beta parsoid was not configured to work with en.wikipedia, but now that blocker is fixed, so beta parsoid can be used for manual local testing too.

Bug: https://phabricator.wikimedia.org/T110474

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -17,7 +17,7 @@ default_project: &default_project
               - name: default.group.local
                 domains: /./
           parsoid:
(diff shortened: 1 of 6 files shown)


gwicke:
LGTM. Will merge once the tests have run.

PR #515, opened 2016-02-13
Added purging for mobileapps endpoints

Pchelolo:
Created a basic event service. For now it will directly purge the caches, but in future it we will emit the event to change propagation service from there.

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -6,6 +6,11 @@ default_project: &default_project
   x-modules:
     /:
       - path: test/test_module.yaml
+        options:
+          events:
+            purge:
+              host: 127.0.0.1
+              port: 4321
       - path: projects/wmf_default.yaml
         options: &default_options
           table:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -30,6 +30,9 @@ default_project: &default_project
             cache-control: s-maxage=864000, max-age=86400
           mobileapps:
             host: http://appservice.wmflabs.org
+          purge:
+            host: 127.0.0.1
+            port: 4321
 
 labs_project: &labs_project
   x-modules:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -30,9 +30,10 @@ default_project: &default_project
             cache-control: s-maxage=864000, max-age=86400
           mobileapps:
             host: http://appservice.wmflabs.org
-          purge:
-            host: 127.0.0.1
-            port: 4321
+          events:
+            purge:
+              host: 127.0.0.1
+              port: 4321
 
 labs_project: &labs_project
   x-modules:
--- a/projects/wikimedia.org_sqlite.yaml
+++ b/projects/wikimedia.org_sqlite.yaml
@@ -67,3 +67,6 @@ paths:
       /pageviews:
         - path: sys/pageviews_proxy.yaml
           options: '{{options.pageviews}}'
+      /events:
+        - path: sys/events.js
+          options: '{{options.purge}}'
--- a/projects/wikimedia.org_sqlite.yaml
+++ b/projects/wikimedia.org_sqlite.yaml
@@ -69,4 +69,4 @@ paths:
           options: '{{options.pageviews}}'
       /events:
         - path: sys/events.js
-          options: '{{options.purge}}'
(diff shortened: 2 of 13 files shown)


config.test.yaml
```diff
@@ -30,6 +30,9 @@ default_project: &default_project
             cache-control: s-maxage=864000, max-age=86400
           mobileapps:
             host: http://appservice.wmflabs.org
+          purge:
```
gwicke: For consistency, perhaps call this 'event'?

projects/wikimedia.org_sqlite.yaml
```diff
@@ -68,3 +68,6 @@ paths:
       /pageviews:
         - path: sys/pageviews_proxy.yaml
           options: '{{options.pageviews}}'
+      /events:
+        - path: sys/events.js
+          options: '{{options.purge}}'
```
gwicke: .events?

sys/events.js
```diff
@@ -0,0 +1,44 @@
...
+    this.options = options;
+
+    if (options && options.host && options.port) {
+        this.purger = new Purger({ routes: [ options ] });
+    }
+};
+
+EventService.prototype.emitEvent = function(hyper, req) {
+    var self = this;
+    if (this.purger) {
```
gwicke: Perhaps also validate the body:
```javascript
if (!Array.isArray(req.body)) {
  throw new HTTPError(...);
}
```

sys/events.js
```diff
@@ -0,0 +1,44 @@
...
+        self.purger.purge(req.body.map(function(event) {
+            return event.uri.toString();
```
gwicke: Some validation here would be nice too.

sys/events.js
```diff
@@ -0,0 +1,44 @@
...
+        }))
+        .catch(function(e) {
+            hyper.log('error/purge', e);
```
gwicke: error/events/purge ?

gwicke:
Nice and clean overall!

sys/mobileapps.yaml
```diff
@@ -51,7 +51,15 @@ paths:
...
               body: '{{from_service.body.remaining}}'
-        - return:
+        - purge_caches:
+            request:
+              method: post
+              uri: /{domain}/sys/events/
+              body:
+                - uri: //{{request.params.domain}}/api/rest_v1/page/mobile-sections/{request.params.title}
+                - uri: //{{request.params.domain}}/api/rest_v1/page/mobile-sections-lead/{request.params.title}
+                - uri: //{{request.params.domain}}/api/rest_v1/page/mobile-sections-remaining/{request.params.title}
```
gwicke: I kind of wished we had gone with `req.` instead of `request.`, but.. too late.

Pchelolo: @gwicke we have `request` everywhere in the code, could replace all at once later for consistency

sys/mobileapps.yaml
```diff
@@ -51,7 +51,15 @@ paths:
               uri: /{domain}/sys/key_value/mobileapps.remaining/{title}
               headers: '{{from_service.headers}}'
               body: '{{from_service.body.remaining}}'
-        - return:
+        - purge_caches:
```
gwicke: Minor: Perhaps call this `purge_and_return`?

sys/mobileapps.yaml
gwicke: Yeah, definitely not urgent. Just a bit of lamenting on the side.

gwicke:
LGTM. Proper escaping depends on https://github.com/wikimedia/swagger-router/pull/38, so lets merge this in coordination with that patch.

Pchelolo:
Note: for this to work we need to deploy a config change.

Pchelolo:
LGTM

d00rman:
Why require the purge body request to contain objects with only the `uri` property? It would be cleaner to just have an array with strings containing the URIs, wouldn't it?

Pchelolo:
@d00rman The `events` module will be a general event-emitting service, se we a being general right from the beginning.

gwicke:
Yeah, lets not make this purge-specific.

PR #516, opened 2016-02-13
Replaced single braces with doble braces everywhere

Pchelolo:
We're moving to the semantics when single braces in templates would be URIencoded all the time, so we need to replace them with double braces throughout the codebase.

cc @wikimedia/services

### DIFF
--- a/sys/mobileapps.yaml
+++ b/sys/mobileapps.yaml
@@ -98,8 +98,8 @@ paths:
               uri: /{domain}/sys/mobileapps/mobile-sections/{title}
             return:
               status: 200
-              headers: '{$.get_sections.headers}'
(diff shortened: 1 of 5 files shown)


gwicke:
nit: s/doble/double/ in the commit message.

gwicke:
:+1:, happy to merge once Travis concurs.

Pchelolo:
@gwicke let's wait until https://github.com/wikimedia/swagger-router/pull/38 is released and rerun the tests

gwicke:
There is no harm in merging this now, I think. Worst case would be that we missed something, but we'll notice once we have released https://github.com/wikimedia/swagger-router/pull/38.

PR #519, opened 2016-02-17
Update Wiktionary definition end point documentation

gwicke:
- Add a summary.
- Remove mention of external no-cache use to avoid encouraging its use.
- Provide more detail on limitations and link to wiki page providing
  background.


Pchelolo:
:+1:

mdholloway:
:+1: 

Thanks for the heads-up, @gwicke!

PR #520, opened 2016-02-18
Improve purging & caching

gwicke:
- Introduce a shared `purged_cache_control` config option, which is used to
  configure the `cache-control` headers returned by all purged entry points.
- Rename per-module cache header configs to response_cache_control.
- Add purging and caching to summary and parsoid HTML end points.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -27,6 +27,8 @@ default_project: &default_project
           cache-control: s-maxage=864000, max-age=86400
         mobileapps:
           host: http://appservice.wmflabs.org
+        # 10 days Varnish caching, one day client-side
+        purged_cache_control: s-maxage=864000, max-age=86400
 
 # A different project template, sharing configuration options.
 wikimedia.org: &wikimedia.org
@@ -99,7 +101,7 @@ logging:
   #  port: <%= @logstash_port %>
 
 metrics:
-  type: statsd
-  host: localhost
-  port: 8125
-  batch: true
+  #type: txstatsd
+  #host: localhost
+  #port: 8125
+  #batch: true
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -101,7 +101,7 @@ logging:
   #  port: <%= @logstash_port %>
 
 metrics:
-  #type: txstatsd
-  #host: localhost
-  #port: 8125
-  #batch: true
+  type: statsd
+  host: localhost
+  port: 8125
+  batch: true
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -52,7 +52,8 @@ paths:
               x-modules:
                 - path: v1/content.yaml
                 - path: v1/mobileapps.yaml
-                  options: '{{options.mobileapps}}'
+                  options: '{{merge({"response_cache_control": options.purged_cache_control}, 
+                                options.mobileapps)}}'
                 - path: v1/graphoid.yaml
                   options: '{{options.graphoid}}'
                 - path: v1/summary.js
@@ -137,7 +138,8 @@ paths:
             /mobileapps:
               x-modules:
                 - path: sys/mobileapps.yaml
-                  options: '{{options.mobileapps}}'
+                  options: '{{merge({"response_cache_control": options.purged_cache_control}, 
+                                options.mobileapps)}}'
             /events:
               x-modules:
                 - path: sys/events.js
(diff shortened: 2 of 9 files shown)


config.example.wikimedia.yaml
```diff
@@ -27,6 +27,8 @@ default_project: &default_project
           cache-control: s-maxage=864000, max-age=86400
         mobileapps:
           host: http://appservice.wmflabs.org
+        # 10 days Varnish caching, one day client-side
+        purged_cache_control: s-maxage=864000, max-age=86400
```
Pchelolo: Should we enable it for definitions right away too?

gwicke: Yeah, added purging and longer-term caching for that in ec00898138d331c687d11c1f6447c332950117e0.

projects/wmf_default.yaml
```diff
@@ -52,12 +52,13 @@ paths:
...
                 - path: v1/mobileapps.yaml
-                  options: '{{options.mobileapps}}'
+                  options: '{{merge({"response_cache_control": options.purged_cache_control}, 
+                                options.mobileapps)}}'
                 - path: v1/graphoid.yaml
                   options: '{{options.graphoid}}'
                 - path: v1/summary.js
                   options:
-                    response_cache-control: 'max-age: 3600, s-maxage: 3600'
+                    response_cache_control: '{{options.purged_cache_control}}'
```
Pchelolo: For `mathoid` we call this option just `cache-control`. Should we change it either there or here to be consistent

projects/wmf_sqlite.yaml
```diff
@@ -140,9 +137,10 @@ paths:
             /mobileapps:
               x-modules:
                 - path: sys/mobileapps.yaml
-                  options: '{{options.mobileapps}}'
+                  options: '{{merge({"response_cache_control": options.purged_cache_control},
```
Pchelolo: Does `sys` mobile apps module need this option at all?

projects/wmf_default.yaml
```diff
@@ -141,9 +138,10 @@ paths:
             /mobileapps:
               x-modules:
                 - path: sys/mobileapps.yaml
-                  options: '{{options.mobileapps}}'
+                  options: '{{merge({"response_cache_control": options.purged_cache_control},
```
Pchelolo: Why do we allow overriding it for mobile apps and not for parsoid and others? Lets be consistent.

Pchelolo:
Left some comments in https://github.com/wikimedia/restbase/pull/520/files#diff-fea0bd07c4904c3627ee7de8e58c0049 However the same applies to all projects. 

Also, we can't release this until https://phabricator.wikimedia.org/T127370 is fixed.

projects/wmf_default.yaml
gwicke: I was considering to generally change it to `purged_cache_control`, as some end points won't have purging.

projects/wmf_default.yaml
gwicke: This is using the default global setting for mobileapps, and merges it into the other mobileapps settings. The alternative is to define the cache-control setting twice in the config.

Pchelolo: Oh, indeed, completely misread the line..

v1/definition.yaml
```diff
@@ -99,18 +99,26 @@ paths:
...
               headers: '{{extract.headers}}'
               body: '{{extract.body}}'
+        - emit_change_event:
+            request:
+              method: post
+              uri: /{domain}/sys/events/
+              body:
+                - meta:
+                    uri: //{request.params.domain}/api/rest_v1/page/definition/{request.params.title}
+          response:
```
Pchelolo: Nit: replace single curly braces with double for the domain parameter.

v1/summary.yaml
```diff
@@ -118,7 +116,14 @@ paths:
...
 
-        - store_and_return:
+        - emit_change_event:
+            request:
+              method: post
+              uri: /{domain}/sys/events/
+              body:
+                - meta:
+                    uri: //{request.params.domain}/api/rest_v1/page/summary/{request.params.title}
+          store_and_return:
```
Pchelolo: Nit: single braces -> double for domain parameter.

gwicke: This is only the domain, which ought to not contain any special characters.

v1/definition.yaml
gwicke: Single braces ought to be safer here.

Pchelolo:
LGTM, pending required varnish fixes

gwicke:
A puppet config change setting up the `purged_cache_control` variable is available at https://gerrit.wikimedia.org/r/#/c/273974/. It can be safely deployed before this PR starts using the config variable.

PR #521, opened 2016-02-18
Events: Optionally prepend the protocol to the event URI

Pchelolo:
We need to send http protocol to varnish so that it works, so prepend it if protocol-relative uri was used in the event.

### DIFF
--- a/sys/events.js
+++ b/sys/events.js
@@ -27,7 +27,7 @@ EventService.prototype.emitEvent = function(hyper, req) {
             }
 
             return self.purger.purge(req.body.map(function(event) {
-                if (!event.meta || !event.meta.uri) {
+                if (!event.meta || !event.meta.uri || /^\/\/:/.test(event.meta.uri)) {
                     hyper.log('error/events/purge', {
                         message: 'Invalid event URI',
                         event: event
--- a/sys/events.js
+++ b/sys/events.js
@@ -27,13 +27,13 @@ EventService.prototype.emitEvent = function(hyper, req) {
             }
 
             return self.purger.purge(req.body.map(function(event) {
-                if (!event.meta || !event.meta.uri || !/^\/\//.test(event.meta.uri.toString())) {
+                if (!event.meta || !event.meta.uri || !/^\/\//.test(event.meta.uri)) {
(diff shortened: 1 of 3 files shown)


test/features/events/events.js
```diff
@@ -22,9 +22,9 @@ describe('Change event emitting', function() {
                         return message.slice(22, 22 + message.readInt16BE(20)).toString();
                     }).sort();
                     assert.deepEqual(receivedMessages,
-                    [ '//en.wikipedia.org/api/rest_v1/page/mobile-sections-lead/User%3APchelolo',
-                        '//en.wikipedia.org/api/rest_v1/page/mobile-sections-remaining/User%3APchelolo',
-                        '//en.wikipedia.org/api/rest_v1/page/mobile-sections/User%3APchelolo' ]);
+                    [ 'http://en.wikipedia.org/api/rest_v1/page/mobile-sections-lead/User%3APchelolo',
```
gwicke: Since this is an internal request, could you omit the protocol here?

gwicke:
LGTM, just waiting for travis.

sys/events.js
```diff
@@ -27,13 +27,17 @@ EventService.prototype.emitEvent = function(hyper, req) {
...
-                if (!event.meta || !/^\/\/[^\/]+/.test(event.meta.uri)) {
+                if (!event.meta || !event.meta.uri) {
                     hyper.log('error/events/purge', {
                         message: 'Invalid event URI',
                         event: event
                     });
                 } else {
-                    return event.meta.uri.toString();
+                    var uri = event.meta.uri.toString();
+                    if (!/^https?:/.test(uri)) {
```
d00rman: Nit-pick: while we have agreed to send proto-relative URIs to `/sys/event`, maybe it would be prudent to check for `//` as well. As it currently stands, if `en.wikipedia.org` is passed in, we'd end up with `http:en.wikipedia.org` which isn't what we expect to happen.

gwicke:
:+1:

d00rman:
Published as [v0.11.1](https://github.com/wikimedia/restbase/tree/v0.11.1)

PR #523, opened 2016-02-22
Pageviews: Simplify config and pass through AQS response intact

Pchelolo:
As we've got separate repo for AQS now we don't need to have a trick with `pageviews_proxy.yaml` module and can call AQS directly from v1 module. 

Also, from now on we will not modify AQS response headers and leave full control over the response to AQS.

cc @wikimedia/services @milimetric


milimetric:
any estimate on when this can be merged and deployed?  I'm only slightly nervous that usage is starting to increase and I'd like to be able to control the caching better to respond.

PR #524, opened 2016-02-23
Access Checks: Implemented restrictions table

Pchelolo:
This PR creates a new `restrictions` table that would be used for lightweight access checks in the `title_value` and `title_rev_value` buckets, and will be used for latest/archival data separation. 

The `restrictions` table contains only items for restricted revisions and deleted pages, so the table will be extra small, which gives us fast access checks and ability to replicate it to POPs.

Also this fixes a problem that historic revisions of deleted pages were still accessible.

Also, in the previous round of access checks test refactoring I've introduced some bugs in tests, which are fixed now.

This code contains some TODOs:
1. I've found a bug in SQLite, fix will be done soon, but for the time of a review I've just implemented a workaround.
2. We can't switch to using the new restrictions table right now, because we need to fill in the data first. So the new access checking code is commented-out and we still rely on old access checks.
3. While we are filling the `restrictions` table we need to store it's content regardless of whether the revision was changed or not.

cc @wikimedia/services


lib/mwUtil.js
```diff
@@ -102,5 +102,58 @@ mwUtil.getLimit = function(hyper, req) {
...
+ *
+ * @param {Object} item the item to check restriction on
+ * @param {string} [targetTid] the item tid.
+ *                             Used in case item is a restriction to indicate the content tid.
+ */
+mwUtil.applyAccessChecks = function(item, targetTid) {
+    // Page was deleted - new version of tracking
+    if (item.page_deleted) {
+        var deletionTime = uuid.fromString(item.page_deleted).getDate();
+        var revisionTime = uuid.fromString(targetTid || item.tid).getDate();
```
gwicke: I'm not completely sure that this is actually getting the timestamp corresponding to the original revision, rather than the time at which we saved the revision metadata / restrictions.

gwicke: I think we are using the current time when we save the revision metadata: https://github.com/Pchelolo/restbase/blob/access_check/sys/page_revisions.js#L384

sys/page_revisions.js
```diff
@@ -80,6 +83,152 @@ PRS.prototype.getTableSchema = function() {
...
+PRS.prototype.restrictionsTableURI = function(domain) {
+    return new URI([domain, 'sys', 'table', this.restrictionsTableName, '']);
+};
+
+PRS.prototype.getRestriction = function(hyper, req) {
+    var self = this;
+    var rp = req.params;
+    var attributes = { title: rp.title };
+    if (rp.revision) {
+        attributes.rev = rp.revision;
```
gwicke: If we select `ge` and order by rev asc, then we should be able to get the restriction for the exact revision requested, *or any newer*. If we always store a row on page deletion corresponding to the last revision of that page, this would mean that we don't need to perform a second request on 404, as

- either the page is deleted, in which case we'll get a result (not necessarily for the exact revision we asked for), or
- nothing is found, which means there are no restrictions on either the revision or page.

sys/page_revisions.js
```diff
@@ -88,34 +237,11 @@ PRS.prototype.getTableSchema = function() {
  *  - userhidden: remove user information from response
  *
  * @param item Object the revision item
- * @return true
  * @throws HTTPError if access to the revision should be denied
  */
 PRS.prototype._checkRevReturn = function(item) {
+    mwUtil.applyAccessChecks(item);
```
gwicke: I actually wonder if applyAccessChecks would be best located in this module. Do you expect other modules to use it manually?

lib/mwUtil.js
gwicke: Example cqlsh session:
```sql
cqlsh:local_group_default_T_title__revisions> select rev from data where "_domain"='en.wikipedia.org' and title = 'Foobar' and rev >= 698187533;

 rev
-----------
 702883545
 701285702
 698187533

(3 rows)
cqlsh:local_group_default_T_title__revisions> select rev from data where "_domain"='en.wikipedia.org' and title = 'Foobar' and rev >= 698187533 order by rev asc limit 1;

 rev
-----------
 698187533

(1 rows)
```

sys/parsoid.js
```diff
@@ -158,6 +158,24 @@ PSP._dependenciesUpdate = function(hyper, req) {
  * @private
  */
 PSP._wrapInAccessCheck = function(hyper, req, promise) {
+    // TODO: Enable new checking style when the restrictions table is filled
```
gwicke: Also todo: Move to filter ;)

gwicke:
LGTM overall. A couple of nits / thoughts inline.

sys/page_revisions.js
Pchelolo: @gwicke yes. You can look at the usage example in the commented-out code section in the parsoid module.

gwicke: Won't that move to a filter, anyway?

gwicke:
@Pchelolo, would you prefer to merge this before moving towards filters, or should we tackle the filterification right away?

Pchelolo:
@gwicke We need to rebase it anyway, so let's move to filters right away

sys/page_revisions.js
Pchelolo: @gwicke so, the check was moved to a special filter now. However, it's not used yet, because we need to fill the restrictions table first.

lib/access_check_filter.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+var mwUtil = require('./mwUtil');
+var P = require('bluebird');
+var URI = require('hyperswitch').URI;
+
```
gwicke: Very nice and clean!

gwicke:
I think this will need some updating to account for the revision-table based access checks, and the refactor in parsoid.js.

gwicke:
Update: Fixed the issue in labs, and pushed a new PR at #599.

Pchelolo:
This was implemented in another PR

PR #526, opened 2016-02-26
Docs: Added summary to math and pageview endpoints

Pchelolo:
The summary is a short description of the API shown on the right side of collapsed entry point doc in the [swagger docs](https://en.wikipedia.org/api/rest_v1/?doc=). We have them everywhere except mathoid and pageviews.

cc @wikimedia/services @milimetric

### DIFF
--- a/v1/mathoid.yaml
+++ b/v1/mathoid.yaml
@@ -6,6 +6,7 @@ paths:
   /math/check/{type}:
     post:
       tags: ['Math']
+      summary: Check and normalize a TeX formula.
       description: |
         Checks the supplied TeX formula for correctness and returns the
         normalised formula representation as well as information about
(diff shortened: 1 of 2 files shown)


v1/pageviews.yaml
```diff
@@ -181,6 +184,7 @@ paths:
     get:
       tags:
         - Pageviews data
+      summary: Get most viewed articles for a project.
```
gwicke: Minor: "Get the most viewed articles for a project." reads slightly more natural.

gwicke:
LGTM overall.

v1/mathoid.yaml
```diff
@@ -94,6 +95,7 @@ paths:
   /math/render/{format}/{hash}:
     get:
       tags: ['Math']
+      summary: Get rendered formula in a given format.
```
d00rman: `s/a/the/`

Pchelolo: @d00rman done.

PR #527, opened 2016-02-29
Buckets: First implementation of archived bucket

Pchelolo:
Under our ongoing effort to separate latest content and archival content I've implemented a drop-in replacement bucket for archival data. It has the same spec as `key_rev_value`. In case we use it everywhere instead of `key_rev_value`, all the tests pass and all seems to work fine. Basically, we could change the PR a bit to set the `archival` part name same as the overall bucket name, and then start using the new bucket right away, with no additional mangling of in-storage data.

This PR doesn't begin using the bucket anywhere, just adds it to the repo and adds tests for it.

cc @wikimedia/services


sys/key_rev_archival_value.js
```diff
@@ -0,0 +1,149 @@
...
+
+function ArchivalBucket(options) {
+}
+
+ArchivalBucket.prototype._latestName = function(bucket) {
+    return bucket + '.latest';
+};
+
+ArchivalBucket.prototype._archiveName = function(bucket) {
+    return bucket + '.archive';
```
gwicke: Perhaps we could support some exceptions / aliases in the module options to simplify the transition:

```yaml
options:
  bucket_aliases:
    parsoid.html.archive: parsoid.html
```

Pchelolo: Or may be we could just state that archival table name will be the same as for `key_rev_value`. Then we would be able to switch this feature on/off..

sys/key_rev_archival_value.js
```diff
@@ -0,0 +1,149 @@
...
+        latestRequestPath.push(rp.tid);
+    }
+    return hyper.get({
+        uri: new URI(latestRequestPath),
+        headers: req.headers
+    })
+    .then(function(res) {
+        if (rp.revision) {
+            var etagInfo = mwUtil.parseETag(res.headers.etag);
+            if (Number.parseInt(etagInfo.rev) !== Number.parseInt(rp.revision)
```
gwicke: Very minor: `parseInt` is global, so the `Number` prefix shouldn't be needed here.

sys/key_rev_archival_value.js
gwicke: Yeah, as long as we don't create any new buckets it works out the same. With the explicit alias setup, new tables could start using the new suffixes right away, saving a few keyspace renames with possibly associated downtime later.

sys/key_rev_archival_value.js
```diff
@@ -0,0 +1,149 @@
...
+    headers.etag = headers.etag || mwUtil.makeETag(rev, tid);
+
+    return hyper.get({
+        uri: new URI([rp.domain, 'sys', 'key_value', self._latestName(rp.bucket), rp.key])
+    })
+    .catch({ status: 404 }, function() { /* Ignore */ })
+    .then(function(res) {
+        if (res) {
+            var etagInfo = mwUtil.parseETag(res.headers.etag);
+            if (Number.parseInt(etagInfo.rev) > Number.parseInt(rp.revision)) {
```
gwicke: Omitting `Number.` from `parseInt` should be fine here as well.

test/features/buckets/revisioned_bucket.js
```diff
@@ -6,6 +6,8 @@ var server = require('../../utils/server.js');
 var uuid = require('cassandra-uuid').TimeUuid;
 var P = require('bluebird');
 
+var mwUtils = require('../../../lib/mwUtil');
```
gwicke: Those relative paths are a bit awkward.. One interesting pattern I have seen recently is to link a module into its own node_modules, making it easier to require (as `restbase/mwUtil`). Perhaps worth considering when / if we start using babel & a build step.

Pchelolo:
Depends on https://phabricator.wikimedia.org/T128419

Pchelolo:
Refactored the patch, with `latest_hash` revision policy and use of `key_rev_value` for latest content storage it's now way simpler.

Pchelolo:
To store already gzipped content in the latest content bucket we need some support from hyperswitch https://github.com/wikimedia/hyperswitch/pull/21

Pchelolo:
I've made the latest contents bucket store already encrypted content and spit it out as encrypted buffers right away. This has giant impact on performance for large articles: Barack_Obama read throughput increased from 100 req/s to 300 req/s on my machine. However, there're some caveats: for sections and transforms we might need to decompress the data on the fly. 

Current version of the patch doesn't enable new bucket right away, but it's a drop-in replacement for a key_rev_value bucket. Simple `s/key_rev_value/key_rev_latest_value` in parsoid.js will switch it on.

sys/key_rev_latest_value.js
```diff
@@ -0,0 +1,144 @@
...
+    return bucket;
+};
+
+ArchivalBucket.prototype.createBucket = function(hyper, req) {
+    var self = this;
+    var rp = req.params;
+    var latestConfig = Object.assign({}, req.body, {
+        revisionRetentionPolicy: { type: 'latest_hash' }
+    });
+    if (latestConfig.valueType !== 'json') {
```
gwicke: Can we avoid the special case for JSON?

sys/key_rev_latest_value.js
```diff
@@ -0,0 +1,144 @@
...
+        latestConfig.options = latestConfig.options || {};
+        latestConfig.options.compression = [];
+    }
+    return P.join(
+        hyper.put({
+            uri: new URI([rp.domain, 'sys', 'key_rev_value', self._latestName(rp.bucket)]),
```
gwicke: The table backing this should not have any compression enabled (or at most something really fast like lzo).

sys/key_rev_latest_value.js
```diff
@@ -0,0 +1,144 @@
...
+        uri: new URI([rp.domain, 'sys', 'key_rev_value', self._archiveName(rp.bucket), rp.key, '']),
+        query: req.query
+    });
+};
+
+ArchivalBucket.prototype.putRevision = function(hyper, req) {
+    var self = this;
+    var rp = req.params;
+    var prepare;
+    if (/^application\/json/.test(req.headers['content-type'])) {
```
gwicke: We should really store JSON in compressed form as well. The main question is how to avoid breaking existing consumers expecting decoded JSON. For now, we can decompress & decode manually here. Longer term, we should figure out a way to defer that until it's really needed, for example by moving towards a `fetch` like response API with a `json()` method.

sys/key_rev_latest_value.js
```diff
@@ -0,0 +1,144 @@
...
+    var rp = req.params;
+    return hyper.get({
+        uri: new URI([rp.domain, 'sys', 'key_rev_value', self._archiveName(rp.bucket), rp.key, '']),
+        query: req.query
+    });
+};
+
+ArchivalBucket.prototype.putRevision = function(hyper, req) {
+    var self = this;
+    var rp = req.params;
```
gwicke: We should make sure that we have a tid at this level. The tids of archival & latest versions should agree.

sys/parsoid.js
```diff
@@ -182,6 +183,24 @@ PSP._wrapInAccessCheck = function(hyper, req, promise) {
     .then(function(responses) { return responses.content; });
 };
 
+function bodyToString(contentResponse) {
```
gwicke: This is equivalent to the .text() method in [the fetch spec](https://fetch.spec.whatwg.org/).

gwicke:
LGTM overall. My main questions are about:

- Handling of JSON, and
- representation of possibly-compressed response data in response.body.

Both are closely related to https://phabricator.wikimedia.org/T118195. Perhaps it is time to seriously consider how we can gradually adopt accessors on the response, similar to `fetch`?

sys/key_rev_latest_value.js
Pchelolo: Since right now we will need to decompress it in the bucket all the time, it's faster to store uncompressed JSON in the latest bucket.

gwicke:
As discussed on IRC, it would be nice if we could store current JSON as compressed blobs as well.

gwicke:
This is ready for testing in staging. We should hold off on deploying this until we have given this a good work-out in staging. This week, we can deploy #552. Alternatively / next week, we can temporarily switch the parsoid module back to key_rev_value before the deploy, which makes this code a no-op.

PR #529, opened 2016-03-02
Added If-Match header to spec and log remaining users

Pchelolo:
As sending the `If-Match` header is a canonical way of setting the render tid, we should mention it in the spec and allow users send it from the swagger sandbox. 

Also, log all users that don't send the header yet and rely on mw:TimeUuid fallback.

Related to https://phabricator.wikimedia.org/T128525

### DIFF
--- a/v1/transform.yaml
+++ b/v1/transform.yaml
@@ -51,8 +51,8 @@ paths:
         - name: if-match
           in: header
           description: |
-            An `ETag` header of the original render indicating it's revision and timeuuid.
-            Required if the `title` and `revision` parameters are present.
+            The `ETag` header of the original render indicating it's revision and timeuuid.
+            Required if both `title` and `revision` parameters are present.
           type: string
           required: false
       responses:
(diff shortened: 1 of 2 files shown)


v1/transform.yaml
```diff
@@ -48,6 +48,13 @@ paths:
           description: Normalise the DOM to yield cleaner wikitext?
           type: boolean
           required: false
+        - name: if-match
+          in: header
+          description: |
+            An `ETag` header of the original render indicating it's revision and timeuuid.
```
gwicke: Minor: "The `ETag` header of.."

v1/transform.yaml
```diff
@@ -48,6 +48,13 @@ paths:
...
+            Required if the `title` and `revision` parameters are present.
```
gwicke: Perhaps more precisely, to emphasize that it's not an `or`: "Required if both the `title` and `revision` parameters are present."

gwicke:
Great idea! The logging will be very helpful for determining when we can remove the TimeUuid.

Pchelolo:
@gwicke amended the PR with improved wording according to your comments

PR #530, opened 2016-03-02
Move to node 4.3

d00rman:
We are now running node 4.3.0 in production, so instruct the build script and Travis use it.


d00rman:
cc @wikimedia/services

PR #532, opened 2016-03-02
Parsoid: Ensure res.headers is set before setting cache-control

d00rman:
When setting the `cache-control` header, the code only checked if
`self.options.response_cache_control` is set. However, some execution
paths do not yield a `res.headers` object, causing TypeError exceptions
when setting the `cache-control` header. To circumvent the issue, place
a call to `HyperSwitch.misc.addCSPHeaders()` first, as it creates the
object if hasn't been defined, thus ensuring `cache-control` can be
safely set afterwards.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "0.11.2",
+  "version": "0.11.3",
   "description": "REST storage and service dispatcher",
   "main": "lib/server.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

d00rman:
Self-merging in the interest of completing the deploy.

d00rman:
Published as [v0.11.3](https://github.com/wikimedia/restbase/tree/v0.11.3)

PR #533, opened 2016-03-03
API: Initial implementation for a cached related pages endpoint

Pchelolo:
Primary implementation for the related pages API. Right now it's not storing the result, just setting a `cache-control` headers and proxying the request to MW API. 

The main issue is thumbnail size, but clients could mangle the thumb uri.

Bug: https://phabricator.wikimedia.org/T125983
cc @wikimedia/services

### DIFF
--- a/sys/action.js
+++ b/sys/action.js
@@ -140,8 +140,8 @@ function buildQueryResponse(apiReq, res) {
         throw new HTTPError({
             status: 404,
             body: {
-                type: 'not_found#page_revisions',
-                description: 'Page or revision not found.',
+                type: 'not_found',
+                description: 'Requested resource is not found.',
                 apiRequest: apiReq
             }
         });
--- a/sys/action.js
+++ b/sys/action.js
@@ -128,7 +128,7 @@ function ActionService(options) {
     }
 }
 
-function buildQueryResponse(res) {
+function buildQueryResponse(apiReq, res) {
     if (res.status !== 200) {
         throw apiError({
             info: 'Unexpected response status ('
@@ -137,7 +137,14 @@ function buildQueryResponse(res) {
     } else if (!res.body || res.body.error) {
         throw apiError((res.body || {}).error);
     } else if (!res.body.query || (!res.body.query.pages && !res.body.query.userinfo)) {
-        throw apiError({ info: 'Missing query pages from the PHP action API response.' });
+        throw new HTTPError({
+            status: 404,
+            body: {
+                type: 'not_found#page_revisions',
+                description: 'Page or revision not found.',
+                apiRequest: apiReq
+            }
+        });
     }
 
     if (res.body.query.pages) {
@@ -161,7 +168,7 @@ function buildQueryResponse(res) {
     }
 }
 
-function buildEditResponse(res) {
+function buildEditResponse(apiReq, res) {
     if (res.status !== 200) {
         throw apiError({
             info: 'Unexpected response status ('
@@ -187,7 +194,7 @@ ActionService.prototype._doRequest = function(hyper, req, defBody, cont) {
     if (!apiRequest.body.hasOwnProperty('continue') && apiRequest.action === 'query') {
         apiRequest.body.continue = '';
     }
-    return hyper.request(apiRequest).then(cont);
+    return hyper.request(apiRequest).then(cont.bind(null, apiRequest));
 };
 
 ActionService.prototype.query = function(hyper, req) {
--- a/v1/related.yaml
+++ b/v1/related.yaml
@@ -1,0 +1,133 @@
+swagger: '2.0'
+info:
+  version: '1.0.0-beta'
+  title: MediaWiki Related Pages API
+  description: Related pages API
+  termsofservice: https://github.com/wikimedia/restbase#restbase
+  contact:
+    name: Services
+    email: services@lists.wikimedia.org
+    url: https://www.mediawiki.org/wiki/Services
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /related/{title}:
+    get:
+      tags:
+        - Page content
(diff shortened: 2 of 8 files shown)


gwicke:
LGTM. /cc @mdholloway

ebernhardson:
mostly just for reference, we added caching at the mediawiki level for these too. The hit rate is reasonable but not amazing, looking at 65-75% on a 24hr TTL. It certainly makes a difference though: https://grafana.wikimedia.org/dashboard/db/elasticsearch-percentiles?panelId=55&fullscreen

sys/action.js
```diff
@@ -137,7 +137,14 @@ function buildQueryResponse(res) {
     } else if (!res.body || res.body.error) {
         throw apiError((res.body || {}).error);
     } else if (!res.body.query || (!res.body.query.pages && !res.body.query.userinfo)) {
-        throw apiError({ info: 'Missing query pages from the PHP action API response.' });
+        throw new HTTPError({
+            status: 404,
+            body: {
+                type: 'not_found#page_revisions',
```
d00rman: This looks wrong. The query module has *a lot* of sub-modules, so emitting revinfo 404s here doesn't look good to me.

v1/related.yaml
```diff
@@ -0,0 +1,133 @@
...
+                prop: 'pageterms|pageimages|pageprops'
+                ppprop: 'mainpage|disambiguation'
+                wbptterms: 'description'
+                generator: 'search'
+                gsrsearch: 'morelike:{request.params.title}'
+                gsrnamespace: 0
+                gsrwhat: 'text'
+                gsrinfo: ''
+                gsrprop: 'redirecttitle'
+                gsrlimit: 5
```
d00rman: I believe we need 20 here to accommodate iOS

v1/related.yaml
```diff
@@ -0,0 +1,133 @@
...
+                piprop: 'thumbnail'
+                pithumbsize: 640
```
mdholloway: Could we make this an optional param and default to 640 rather than hard-coding the value?  In this implementation it looks like there's no way to request a larger images.  Currently, we request 800 or 1024 for larger devices.  This is important as it's for the article's lead image.

gwicke: This is the old issue of not having an official API that lets clients select the thumb size. We are pushing for progress on https://phabricator.wikimedia.org/T66214, but in the meantime we'll have to work around this. Options:

1) Clients rewrite /\/320(px-[^\/])$/ to `/<desiredWidth>px-....jpg`. To accurately calculate resulting sizes, original dimensions or a very accurate aspect ratio should be provided in the response.
2) We provide several size options in the standard response.
3) Fragment on thumb size. While easy, this will seriously limit our hit rates, which is bad for latency and backend load.

sys/action.js
gwicke: Yeah, the `#page_revisions` part is not entirely accurate. Perhaps we could make it something like `#action_query_pages`?

v1/related.yaml
mdholloway: I see.  We could work with either 1) or 2).  2) seems cleaner, but it doesn't look like the Action API will emit multiple thumbnails in a single request, so it would likely take multiple requests on the backend to compose a response.  In that case, it's probably better for clients to rewrite the dimension (which it appears can be rewritten arbitrarily -- is this the case?) as described in 1) than to introduce that additional latency.

gwicke: We could possibly implement 2) by performing the scaling computations server-side, without hitting the action API multiple times. I'm not sure that it's worth it, though. No set of sizes will be right for every use case, and extra sizes for each entry do bloat the response size.

If all you need is an image with a given width, then rewriting in the client isn't really harder than a regexp replace. Arbitrary sizes are supported. The main issue with 1) is that the URL syntax is not actually designed to be a public API. It is a stable internal API, so it's safe to use for now. Hopefully we can clean this up & document a public API soon.

mdholloway: It seems like 1) is our best option, then.  In that case, I think the patch looks good as-is -- especially since we are requesting 640px images in most cases (and, as I read their code[1], so is the iOS app). 

[1] https://github.com/wikimedia/wikipedia-ios/blob/0735c777c8d8efee5b63e6f1f13ce00c758cee70/Wikipedia/Code/UIScreen%2BWMFImageWidth.m

mdholloway:
Looks good to me from a client perspective.  :+1:

sys/action.js
gwicke: The question seems to be if there can be other possible reasons for having no query.pages. If so, then we should keep the old behavior, and let apiError handle the overall error. In this particular case we seem to be getting a generic 500 though, and I agree with @Pchelolo that matching the error string is rather hacky.

Could we perhaps pass a status to apiError, or improve apiError to handle this case better?

Pchelolo: Hm.. Actually for normal `query` requests the `pages` property is always there, but the missing articles are marked with `missing` property in the results. However for search generator the pages is not in the response, which indicates a not found. So, I think it's semantically OK to throw a 404 here, just change the wording a bit

gwicke:
LGTM, but lets wait until @d00rman had a chance to check it over as well.

d00rman:
LGTM. As per https://phabricator.wikimedia.org/T125983#2087991, let's give the rest of the Reading team some more time to comment. Otherwise I'll merge/deploy this on Monday.

v1/related.yaml
berndsi: The Android app only uses 320px images for related pages and similar thumbnails (search, history, saved pages). 

See WikipediaApp.java:
PREFERRED_THUMB_SIZE = 320;

berndsi: Correction. It should use 320px but for some reason it uses 640px, as Michael mentioned earlier. I'm not sure why because the thumbnails in read more seem to have the same size as for search, history, saved pages.

PR #534, opened 2016-03-03
Disallow caching for non-canonical requests

gwicke:
This is a temporary fix to avoid caching non-canonical requests. The better
longer-term solution will be to send a redirect, and thus leverage CDN
caching.

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -430,7 +430,12 @@ PSP.getFormat = function(format, hyper, req) {
                     // Re-try to retrieve from storage with the
                     // normalized title & revision
                     rp.title = revInfo.title;
-                    return self.getFormat(format, hyper, req);
+                    return self.getFormat(format, hyper, req)
+                    .then(function(res) {
+                        // Disable caching for non-canonical requests, as
+                        // those are not purged.
+                        res.headers['cache-control'] = 'no-cache, must-revalidate';
+                    });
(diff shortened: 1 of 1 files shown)


Pchelolo:
:+1:

sys/parsoid.js
```diff
@@ -430,7 +430,12 @@ PSP.getFormat = function(format, hyper, req) {
                     // Re-try to retrieve from storage with the
                     // normalized title & revision
                     rp.title = revInfo.title;
-                    return self.getFormat(format, hyper, req);
+                    return self.getFormat(format, hyper, req)
+                    .then(function(res) {
+                        // Disable caching for non-canonical requests, as
+                        // those are not purged.
+                        res.headers['cache-control'] = 'no-cache, must-revalidate';
```
d00rman: Why both `no-cache` **and** `must-revalidate` when the latter applies to a subset of the former?

gwicke: Functionally, it doesn't really matter. This is a temporary patch, so the idea was to get it out quickly, to avoid non-canonical requests receiving outdated cached content. The entire retry / normalization logic can be removed once we redirect.

Pchelolo:
The same effect is achieved by https://github.com/wikimedia/restbase/pull/535

gwicke:
Closing in favor of #535.

PR #535, opened 2016-03-04
Title normalization: added a redirecting filter

Pchelolo:
This is a WIP PR, it still lacks tests and thoughtful testing. Created for reference. 

cc @wikimedia/services

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -32,12 +32,11 @@ module.exports = function(hyper, req, next, options, specInfo) {
                 return normalize(rp.title, siteInfo);
             })
             .catch(function(e) {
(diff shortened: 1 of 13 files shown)


d00rman:
wikimedia/hyperswitch@v0.3.1 has been published, so you'll need to adjust the code accordingly (`x-route-filters`, etc)

Pchelolo:
Just checked with vagrant that titles we get from RestBaseUpdateJobs extension are already normalised, so we should have everything normalised in the storage.

d00rman:
This PR contained a rather nasty bug, cf. wikimedia/restbase#537. I wonder how why/how tests didn't catch that right away...

PR #537, opened 2016-03-07
Action module: Bug fix: define the correct returner function

d00rman:
wikimedia/restbase@18799139 introduced the `ActionService.siteinfo()` method which relies on `ActionService._doRequest()` to perform the request. However, the continuation function passed to it had the wrong signature. This PR fixes the issue.


d00rman:
cc @wikimedia/services

PR #538, opened 2016-03-07
Sample normalization logging with a 5% chance

gwicke:
The log volumes from normalizations are higher than expected, as big consumers
are still sending un-normalized requests. To avoid sending an excessive volume
of log events, this patch samples logs with a 5% chance.

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -64,9 +64,14 @@ module.exports = function(hyper, req, next, options, specInfo) {
                             location: contentLocation
                         }
                     });*/
-                    hyper.log('warn/normalize', {
-                        msg: 'Normalized requested title from ' + rp.title + ' to ' + result
-                    });
+
+
+                    // 5% chance of logging to bound volume.
+                    if (Math.random() < 0.05) {
+                        hyper.log('warn/normalize', {
+                            msg: 'Normalized requested title from ' + rp.title + ' to ' + result
+                        });
+                    }
                     return next(hyper, req)
                     .then(function(res) {
                         if (res) {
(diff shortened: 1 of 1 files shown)


lib/normalize_title_filter.js
```diff
@@ -64,9 +64,14 @@ module.exports = function(hyper, req, next, options, specInfo) {
...
                     });*/
-                    hyper.log('warn/normalize', {
-                        msg: 'Normalized requested title from ' + rp.title + ' to ' + result
-                    });
+
+
+                    // 5% chance of logging to bound volume.
+                    if (Math.random() < 0.05) {
+                        hyper.log('warn/normalize', {
+                            msg: 'Normalized requested title from ' + rp.title + ' to ' + result
```
d00rman: It would be really useful to have `from` and `to` fields as well, as it makes it much easier to inspect the logs (structured vs text)

lib/normalize_title_filter.js
```diff
@@ -64,9 +64,14 @@ module.exports = function(hyper, req, next, options, specInfo) {
                             location: contentLocation
                         }
                     });*/
-                    hyper.log('warn/normalize', {
-                        msg: 'Normalized requested title from ' + rp.title + ' to ' + result
-                    });
+
+
+                    // 5% chance of logging to bound volume.
```
d00rman: Could you please add a comment stating this is a tem hack:

```js
/** TEMP TEMP TEMP
```

or the like.

gwicke: The entire logging is temporary, and there is a comment to that effect on line 45.

PR #541, opened 2016-03-08
Don't redirect for user and user_talk namespaces

Pchelolo:
Bug: https://phabricator.wikimedia.org/T129135

cc @wikimedia/services


Pchelolo:
Note: when this get's merged we will start storing a lot of content under different titles. This means that content should be either restored or deleted.

PR #542, opened 2016-03-08
API cleanup

gwicke:
- Remove experimental /page/html/ and /page/data-parsoid/ listings. These are
  unused, and the same information is available at /page/title/.
- Make the `tid` parameter in the data-parsoid entry point mandatory. This
  makes sure that clients get the data-parsoid matching the HTML retrieved
  earlier.

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -178,38 +178,6 @@ paths:
                 page: '{{page}}'
       x-monitor: false
 
-  /html/:
-    get:
-      tags:
-        - Page content
-      summary: List titles available as HTML.
-      description: |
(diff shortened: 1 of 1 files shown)


gwicke:
I announced this for next week: https://lists.wikimedia.org/pipermail/mediawiki-api-announce/2016-March/000106.html

Lets wait with merging until then.

Pchelolo:
Unless some issues arise on the list I'm OK with this. :+1:

gwicke:
Google has updated their crawler to send tids, so this should be good to go now.

d00rman:
I'll take care of this tomorrow.

PR #545, opened 2016-03-09
Don't fail on coveralls issues

gwicke:


.travis.yml
```diff
@@ -19,4 +19,4 @@ before_install:
   - tar -xzf ../apache-cassandra-2.1.12-bin.tar.gz -C ../
   - sh ../apache-cassandra-2.1.12/bin/cassandra > /dev/null
 
-script: sh test/utils/run_tests.sh coverage all && (npm run-script coveralls || exit 0)
+script: (sh test/utils/run_tests.sh coverage all && npm run-script coveralls || exit 0)
```
Pchelolo: Em, what're we trying to achieve here? 

I thought the previous behaviour was 'not to fail on coveralls issue, but fail unconditionally if the tests failed'. Now it seems the exit code would be 0 in any case, even if the tests actually failed. (1 && 1 || 0 === 0)

d00rman: The way I read it, this will always return 0 no matter what. Is that what we want? :)

gwicke: Good point. 

I could swear that there have been several cases recently where the return status was non-zero despite all tests passing, but can't seem to find the output any more. I'll look more closely the next time this happens.

PR #547, opened 2016-03-10
WIP: binary png support

physikerwelt:
needs more work as discussed with @d00rman 
> the problem is that the binary format is not understood by restbase it returns something like {"type":"Buffer","data":[137,80,78,71... 
and is not interpreted as a png by the browser

### DIFF
--- a/v1/mathoid.yaml
+++ b/v1/mathoid.yaml
@@ -116,6 +116,7 @@ paths:
           enum:
             - svg
             - mml
+            - png
         - name: hash
           in: path
           description: The hash string of the previous POST data
@@ -144,6 +145,11 @@ paths:
             body:
               keyType: string
               valueType: string
+        - init_png:
+            uri: /wikimedia.org/sys/key_value/mathoid.png
+            body:
+              keyType: string
+              valueType: blob
       x-request-handler:
         - check_storage:
             request:


physikerwelt:
I have no idea why the node 0.10 tests fail and the rest passes https://travis-ci.org/wikimedia/restbase/builds/115086087
This depends on https://gerrit.wikimedia.org/r/#/c/276495/

Pchelolo:
@physikerwelt 0.10 test fixed after restarting. Some intermittent cassandra issue

v1/mathoid.yaml
```diff
@@ -116,6 +116,7 @@ paths:
           enum:
             - svg
             - mml
+            - png
```
Pchelolo: You need to add png to the list on line 109.

v1/mathoid.yaml
```diff
@@ -144,6 +145,11 @@ paths:
             body:
               keyType: string
               valueType: string
+        - init_png:
+            uri: /wikimedia.org/sys/key_value/mathoid.png
+            body:
+              keyType: string
+              valueType: blob
```
Pchelolo: We never use this bucket (never store content). see lines 178-190

d00rman:
Superseded by #552

PR #549, opened 2016-03-10
Don't redirect POST requests

Pchelolo:
It doesn't make sense to redirect POST requests even if they're done with unnormalised title.
It's not cached, so we can normalise a title and continue processing.

cc @wikimedia/services

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -43,7 +43,8 @@ module.exports = function(hyper, req, next, options, specInfo) {
             .then(function(result) {
                 var resultText = result.getPrefixedDBKey();
                 if (resultText !== rp.title) {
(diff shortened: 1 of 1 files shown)


lib/normalize_title_filter.js
```diff
@@ -43,7 +43,8 @@ module.exports = function(hyper, req, next, options, specInfo) {
             .then(function(result) {
                 var resultText = result.getPrefixedDBKey();
                 if (resultText !== rp.title) {
-                    if (result.getNamespace().isUser() || result.getNamespace().isUserTalk()) {
+                    if (result.getNamespace().isUser()
+                            || result.getNamespace().isUserTalk()) {
```
gwicke: Can we fold POST into this? Explicitly disabling caching for that case would be a good idea as well.

Pchelolo: Ok

PR #550, opened 2016-03-11
Mathoid: store texvcinfo check results and reuse them

d00rman:
Mathoid's `/texvcinfo` endpoint check the correctness of a formula and normalises it. Because of the normalisation, we weren't reusing its result, since different formulae can have different TeX representations but the same render. Unfortunately, that implied that Mathoid had to be called on each request for a formula check. This PR introduces the `mathoid.js` sys module which de-duplicates such cases. First, storage is checked for the texvcinfo result of the provided formula. If one cannot be found, an indirection record is looked up in a special, hash-mapping table. In case an indirection exists, that hash is used to complete the request (the texvcinfo table is queried again). Only if all of the above fails, Mathoid is called.

Also:
- add a route to the `post_data` module which calculates the hash
- fix a minor edge case in testing utils
- add tests for the `mathoid` module
- release v0.11.4

Bug: [T128593](https://phabricator.wikimedia.org/T128593)

### DIFF
--- a/sys/mathoid.js
+++ b/sys/mathoid.js
@@ -89,17 +89,18 @@ MathoidService.prototype.checkInput = function(hyper, req) {
                 'content-type': 'application/json',
                 'x-resource-location': hash
             };
-            return P.all([
+            return P.join(
                 hyper.put({
                     uri: new URI([rp.domain, 'sys', 'key_value', 'mathoid.check', hash]),
                     headers: checkRes.headers,
                     body: checkRes.body
                 }),
-                indirectionP
-            ]);
-        }).then(function() {
-            checkRes.headers['cache-control'] = self.options['cache-control'];
-            return checkRes;
+                indirectionP,
+                function() {
+                    checkRes.headers['cache-control'] = self.options['cache-control'];
+                    return checkRes;
+                }
+            );
         });
     });
 
--- a/sys/mathoid.js
+++ b/sys/mathoid.js
@@ -48,10 +48,6 @@ MathoidService.prototype.checkInput = function(hyper, req) {
                 });
             });
         });
-    }).then(function(res) {
-        // we have a record, return that
-        res.headers['cache-control'] = self.options['cache-control'];
-        return res;
     }).catch({ status: 404 }, function() {
         // if we are here, it means this is a new input formula
         // so call mathoid
@@ -87,6 +83,7 @@ MathoidService.prototype.checkInput = function(hyper, req) {
             // store the result
             checkRes.headers = {
                 'content-type': 'application/json',
+                'cache-control': 'no-cache',
                 'x-resource-location': hash
             };
             return P.join(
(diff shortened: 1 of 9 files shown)


d00rman:

sys/mathoid.js
```diff
@@ -0,0 +1,140 @@
...
+    var checkRes;
+
+    // start by calculating the hash
+    return hyper.post({
+        uri: new URI([rp.domain, 'sys', 'post_data', 'mathoid.input', 'hash']),
+        body: { q: req.body.q, type: rp.type }
+    }).then(function(res) {
+        hash = origHash = res.body;
+        // short-circuit if it's a no-cache request
+        if (req.headers && /no-cache/.test(req.headers['cache-control'])) {
```
Pchelolo: We could've done that even before calculating the hash

sys/mathoid.js
```diff
@@ -0,0 +1,140 @@
...
+            uri: self.options.host + '/texvcinfo',
+            headers: { 'content-type': 'application/json' },
+            body: {
+                q: req.body.q,
+                type: rp.type
+            }
+        }).then(function(res) {
+            checkRes = res;
+            // store the normalised version
+            return hyper.put({
```
Pchelolo: Where do we actually use the data stored inside the post storage table?

sys/mathoid.js
```diff
@@ -0,0 +1,140 @@
...
+                    headers: { 'content-type': 'text/plain' },
+                    body: hash
+                });
+            }
+            // store the result
+            checkRes.headers = {
+                'content-type': 'application/json',
+                'x-resource-location': hash
+            };
+            return P.all([
```
Pchelolo: P.join is faster

sys/mathoid.js
d00rman: We still need the original hash later at L80 to know whether we need to store an indirection record or not.

sys/mathoid.js
```diff
@@ -0,0 +1,141 @@
...
+            headers: { 'content-type': 'application/json' },
+            body: {
+                q: req.body.q,
+                type: rp.type
+            }
+        }).then(function(res) {
+            checkRes = res;
+            // store the normalised version
+            return hyper.put({
+                uri: new URI([rp.domain, 'sys', 'post_data', 'mathoid.input', '']),
```
Pchelolo: Hm. I'm wondering why are we storing the input at all? Shouldn't we store only the checked/normalized formula?

d00rman: `mathoid.input` is used in the [`/render/{format}/{hash}` endpoint](https://github.com/d00rman/restbase/blob/61c4fb4e6b224b68d8944c8321dfd5cd3f2901fc/v1/mathoid.yaml#L147) when a render cannot be found. It is true that we could use `mathoid.check`'s `body.checked` field, but Mathoid payloads are usually very small, so we can afford to store it in order not to parse the larger `mathoid.check` blob.

Pchelolo:
@d00rman LGTM, however needs changes to address failing tests.

sys/mathoid.js
d00rman: Also, note that the actual input is not stored in `mathoid.input`, but rather the normalised version (cf. L72)

d00rman:
Published as [v0.11.4](https://github.com/wikimedia/restbase/tree/v0.11.4)

PR #552, opened 2016-03-12
Mathoid: Enable PNG support

d00rman:
Mathoid sends the PNG in binary form to the caller. However, when wrapping the PNG response body in a JSON for its `/complete` endpoint, it gets serialised to a plain `{type: 'Buffer', data: [...]}` object. To circumvent that and be able to still make only one request for all of the renders, requesting the renders from Mathoid and storing them in separate buckets has been moved to the mathoid module, where the PNG render gets converted back into a Buffer and stored (and served) properly.

Note: cannot be merged and deployed before https://gerrit.wikimedia.org/r/#/c/276734/ is merged.

Bug: [T71702](https://phabricator.wikimedia.org/T71702)

### DIFF
--- a/sys/mathoid.js
+++ b/sys/mathoid.js
@@ -7,6 +7,9 @@ var URI = HyperSwitch.URI;
 var HTTPError = HyperSwitch.HTTPError;
 
 
+var FORMATS = ['mml', 'svg', 'png'];
+
+
 function MathoidService(options) {
 
     this.options = options;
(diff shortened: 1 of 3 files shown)


d00rman:

sys/mathoid.js
```diff
@@ -103,6 +106,69 @@ MathoidService.prototype.checkInput = function(hyper, req) {
...
+            uri: new URI([domain, 'sys', 'key_value', 'mathoid.' + format, hash]),
+            headers: Object.assign(
+                completeBody[format].headers, { 'x-resource-location': hash }),
+            body: completeBody[format].body
+        };
+        if (format === 'png' && reqObj.body && reqObj.body.type === 'Buffer') {
+            // for png, we need to convert the encoded data manually
+            // because we are receiving it wrapped inside a JSON
+            reqObj.body = new Buffer(reqObj.body.data);
+            completeBody[format].body = new Buffer(reqObj.body);
```
Pchelolo: Why are we calling `new Buffer` twice?

d00rman: Duh, old version left behind.

d00rman:
@Pchelolo rebased on master and updated the small Buffer issue.

Pchelolo:
:+1: lgtm

sys/mathoid.js
```diff
@@ -103,6 +106,69 @@ MathoidService.prototype.checkInput = function(hyper, req) {
 };
 
 
+MathoidService.prototype._storeRenders = function(hyper, domain, hash, completeBody) {
+
+    var idx;
+    var len = FORMATS.length;
+    var reqs = new Array(len);
+
+    for (idx = 0; idx < len; idx++) {
```
gwicke: Minor: Since this is a small number of entries, FORMATS.forEach(..) should perform well enough, and has the benefit of being clearer.

sys/mathoid.js
```diff
@@ -103,6 +106,69 @@ MathoidService.prototype.checkInput = function(hyper, req) {
...
+        var format = FORMATS[idx];
+        // ensure that we have a proper response for a given format
+        if (!completeBody[format] || !completeBody[format].headers || !completeBody[format].body) {
+            return P.reject(new HTTPError({
```
gwicke: A plain `throw` should work as well. This is all implicitly wrapped in Promises.

sys/mathoid.js
```diff
@@ -103,6 +106,69 @@ MathoidService.prototype.checkInput = function(hyper, req) {
...
+    return P.all(reqs).then(function() { return completeBody; });
+
+};
+
+
+MathoidService.prototype.requestAndStore = function(hyper, req) {
+
+    var self = this;
+    var rp = req.params;
+    var hash = req.headers['x-resource-location'];
```
gwicke: To me it looks as if this hash is actually controlled by the client, as forwarded at https://github.com/wikimedia/restbase/pull/552/files#diff-e5918fd3a44fe7806e66e07b48b078a7R157.

Wouldn't this let an attacker save mathoid content at an arbitrary hash, possibly overwriting other formulae?

d00rman: The client is only supplying the hash that has been previously received. If they supply an invalid hash, the request will fail because the fetch from the `post_data` bucket will fail - https://github.com/wikimedia/restbase/pull/552/files#diff-e5918fd3a44fe7806e66e07b48b078a7R148

gwicke: Makes sense. Perhaps worth a comment at https://github.com/wikimedia/restbase/pull/552/files#diff-e5918fd3a44fe7806e66e07b48b078a7R157 ?

gwicke:
LGTM overall.

gwicke:
Merging before #527, so that we can easily deploy this separately.

PR #553, opened 2016-03-14
Temporarily log & don't cache remaining requests from rest.wikimedia.org

gwicke:
Before shutting down rest.wikimedia.org, we need to gauge remaining use, and
possibly alert remaining users.

This patch

- sets up logging for rest.wikimedia.org requests, and
- makes sure that none of those responses are cached, as they aren't purged
  either.

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -7,6 +7,31 @@ var Title = require('mediawiki-title').Title;
 
 module.exports = function(hyper, req, next, options, specInfo) {
     var rp = req.params;
+    /**
+     * Temporary logging & handling for rest.wikimedia.org requests.
+     */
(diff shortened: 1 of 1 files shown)


gwicke:
This reminds me.. we could consider adding the rewrite from /api/rest_v1/ in this handler as well.

Pchelolo:
LGTM. As for rewriting the /api/rest_v1 - let's not mix multiple separate things in a single filter. Also, the rewrite should happen before routing, and this filter is called after routing.

gwicke:
> Also, the rewrite should happen before routing, and this filter is called after routing.

Good point, hadn't considered that.

gwicke:
It would be good to get this out before further ramping up TTLs in https://gerrit.wikimedia.org/r/#/c/277112/.

d00rman:
Deploying this tomorrow too.

PR #555, opened 2016-03-15
Data-parsoid and page title cleanup

gwicke:
- Remove data-parsoid listing & by-title retrieval. We missed these entry
  points in the last pass.
- Explicitly mention the use of underscores in the documentation for all
  `title` parameters.

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -76,7 +76,7 @@ paths:
       parameters:
         - name: title
           in: path
-          description: Page title.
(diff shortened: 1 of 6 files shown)


gwicke:
/cc @wikimedia/services 

We should probably merge & deploy this before announcing the redirect changes for next week.

Pchelolo:
May be we could improve the docs of other `title` parameters right away, in other endpoints like summary etc

Pchelolo:
:+1:

PR #556, opened 2016-03-16
Redirect file pages to commons if they were not found in local wiki

Pchelolo:
In case the page was not found, and it's in `File` namespace, try to redirect the client to commons, if it's set up for this particular wiki.

Bug: https://phabricator.wikimedia.org/T118306


Pchelolo:

berndsi:
Thank you!

PR #557, opened 2016-03-17
Emit events to the eventlogging service from RESTBase

Pchelolo:
Without a config change it's a no-op, but in order to start going to the change propagation direction we need to start emitting the `resource_change` event from restbase, so that we could monitor what's going on with it, check how EventBus and EventLogging service respond to it and so on, so that when change propagation is there we can be confident that emitting part works well.

Considerations:
1. Before making a config change in prod verify that topics are set up in event logging. (cc @ottomata)
2. EventLogging service is already supported in vagrant, so a config change in vagrant is also needed.
3. In't not clear yet what to do with third party installs as we don't have a lightweight kafka replacement yet.

Bug: https://phabricator.wikimedia.org/T126571

cc @wikimedia/services @ottomata

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -13,7 +13,7 @@ default_project: &default_project
         events_emit:
           eventlogging_service:
             uri: http://127.0.0.1:8085/v1/events
-            topic: wmf.resource_change
+            topic: resource_change
     - path: projects/wmf_default.yaml
       options: &default_options
         table:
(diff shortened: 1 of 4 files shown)


gwicke:
The code change looks good to me. As mentioned in the summary, we need to set up the topic before merging this.

Pchelolo:
@gwicke I've added event tagging with `restbase` tag. Also verified with @ottomata that we can start producing these events.

PR #558, opened 2016-03-17
Option for the bucket type backing the parsoid module

gwicke:
This patch makes the bucket type used to back the parsoid module configurable.
The default type is changed to key_rev_value, which means that this
effectively disables the key_rev_latest_value change for initial deployment.

Once we are satisfied with test results, we can enable key_rev_latest_value
with a config change.


Pchelolo:
:+1:

gwicke:
We should get this in before the next deploy, to avoid switching bucket types just yet.

Pchelolo:
I think we can merge this right now

PR #560, opened 2016-03-17
Added a table module to switch storage with config option

Pchelolo:
Introduced a simple wrapper for storage modules that allows to select backend implementation with a config option. After this is landed, and vagrant is updated, we can remove the `wmf_sqlite.yaml` and `wikimedia.org.sqlite.yaml` files. Also, we don't need to do the magic with temp files when running tests in sqlite mode.

cc @wikimedia/services

### DIFF
--- a/sys/table.js
+++ b/sys/table.js
@@ -1,5 +1,9 @@
 "use strict";
 
+/*
+ * A simple wrapper module over storage modules which allows to switch between storage
+ * implementation using a config option.
+ */
 
 module.exports = function(options) {
     options.conf.backend = options.conf.backend || 'cassandra';
--- a/sys/table.js
+++ b/sys/table.js
@@ -1,0 +1,13 @@
+"use strict";
+
+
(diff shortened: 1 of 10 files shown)


sys/table.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
```
gwicke: A small comment explaining what this does would be useful.

gwicke:
Nice!

Pchelolo:
Created a patch for vagrant to use new config option: https://gerrit.wikimedia.org/r/278764

d00rman:
This also needs a prod config change

d00rman:
> This also needs a prod config change

Ah no, actually it doesn't strictly need it due to the set default. Nevertheless, it's good to have the complete config, so I put up https://gerrit.wikimedia.org/r/#/c/278893/ for today's PuppetSWAT.

PR #561, opened 2016-03-21
Enable redirects for non-normalized titles

Pchelolo:
It's time to start redirecting non-normalised titles. 

I've monitored the logs, and we're quite good - most of the clients send requests in proper format.
The only leftovers is some client from google cloud, but there's no way of finding out who is that. Also, a german search engine fireball.de requests summaries for some strange titles sometimes. We might consider contacting them, but it's also not a major client.

cc @wikimedia/services

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -60,12 +60,12 @@ module.exports = function(hyper, req, next, options, specInfo) {
         .then(function(result) {
             var resultText = result.getPrefixedDBKey();
             if (resultText !== rp.title) {
+                rp.title = resultText;
                 if (req.method === 'post' // Don't redirect POSTs as it's not cached anyway
-                || result.getNamespace().isUser()
-                || result.getNamespace().isUserTalk()) {
+                        || result.getNamespace().isUser()
+                        || result.getNamespace().isUserTalk()) {
                     // Due to gender variations of User and User_Talk namespaces in some langs
                     // use canonical name for storage, but don't redirect. Don't cache either
-                    rp.title = resultText;
                     return next(hyper, req)
                     .then(function(res) {
                         if (res) {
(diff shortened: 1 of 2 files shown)


lib/normalize_title_filter.js
```diff
@@ -74,58 +75,30 @@ module.exports = function(hyper, req, next, options, specInfo) {
...
-                            from: rp.title,
-                            to: resultText,
-                        });
-                    }
-                    nextRequest = next(hyper, req)
-                    .then(function(res) {
-                        if (res) {
-                            res.headers = res.headers || {};
-                            res.headers['cache-control'] = 'no-cache';
+                    rp.title = resultText;
```
gwicke: Minor: This assignment is now shared between the branches, so we could pull it out.

lib/normalize_title_filter.js
```diff
@@ -74,58 +75,30 @@ module.exports = function(hyper, req, next, options, specInfo) {
...
+                    var pathBeforeTitle = specInfo.path
+                        .substring(0, specInfo.path.indexOf('{title}'));
+                    pathBeforeTitle = new URI(pathBeforeTitle, rp, true).toString();
+                    var pathAfterTitle = req.uri.toString().replace(pathBeforeTitle, '');
+                    var backCount = (pathAfterTitle.match(/\//g) || []).length;
```
gwicke: Alternative naming idea: 'pathSuffixCount' and `pathSuffix`

lib/normalize_title_filter.js
```diff
@@ -60,8 +61,8 @@ module.exports = function(hyper, req, next, options, specInfo) {
             var resultText = result.getPrefixedDBKey();
             if (resultText !== rp.title) {
                 if (req.method === 'post' // Don't redirect POSTs as it's not cached anyway
-                    || result.getNamespace().isUser()
-                    || result.getNamespace().isUserTalk()) {
+                || result.getNamespace().isUser()
```
gwicke: The indented version looks a bit cleaner to me, tbh.

lib/normalize_title_filter.js
```diff
@@ -74,58 +75,29 @@ module.exports = function(hyper, req, next, options, specInfo) {
...
+                    }).join('');
+                    var pathPatternAfterTitle = specInfo.path
+                        .substring(specInfo.path.indexOf('{title}') - 1);
+                    var contentLocation = backString
+                        + new URI(pathPatternAfterTitle, rp, true).toString().substr(1);
+                    return P.resolve({
+                        status: 301,
+                        headers: {
+                            location: contentLocation,
+                            'cache-control': options.redirect_cache_control
```
gwicke: Should we provide a default, such as `|| 'no-cache'` ?

gwicke:
This looks good to me, lets get this on the deploy train & test it properly in staging.

PR #564, opened 2016-03-22
Parsoid: Always use key_rev_value for stashed content

d00rman:
PR #558 introduced the possibility to specify which bucket type to use for regular Parsoid content. However, stashed content always needs to use the `key_rev_value` bucket type, given that there is no notion of 'latest' there. As both code paths use `getBucketURI()` to determine the URI of the underlying table to use, fix the value of the bucket type when requesting stashed content to `key_rev_value`.

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -257,8 +257,9 @@ PSP.wrapContentReq = function(hyper, req, promise, format, tid) {
     });
 };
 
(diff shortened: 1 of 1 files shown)


d00rman:
cc @wikimedia/services

d00rman:
Bah, Mathoid tests failing due to some problems with it in Labs...

Pchelolo:
:+1:

PR #565, opened 2016-03-22
Mathoid: Use the BetaCluster instance for tests

d00rman:


d00rman:
cc @wikimedia/services

PR #567, opened 2016-03-22
Removed non-needed require

Pchelolo:
A leftover `preq` require from development is unused, but in production preq is not installed as a dependency, so restbase startup fails. 

cc @wikimedia/services


gwicke:
LGTM, just waiting for the tests.

PR #568, opened 2016-03-23
Strip out the domain when calculating redirect location

Pchelolo:
When nodes are shared the domain part in the spec path can be wrong, so strip it out before calculating the redirect location. This is needed to make current deploy repo work.

cc @wikimedia/services

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -79,9 +79,9 @@ module.exports = function(hyper, req, next, options, specInfo) {
                         .substring(0, specInfo.path.indexOf('{title}'));
                     pathBeforeTitle = new URI(pathBeforeTitle, rp, true).toString();
                     // Omit the domain prefix as it could be wrong for node shared between domains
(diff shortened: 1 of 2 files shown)


lib/normalize_title_filter.js
```diff
@@ -78,7 +78,11 @@ module.exports = function(hyper, req, next, options, specInfo) {
                     var pathBeforeTitle = specInfo.path
                         .substring(0, specInfo.path.indexOf('{title}'));
                     pathBeforeTitle = new URI(pathBeforeTitle, rp, true).toString();
-                    var pathSuffix = req.uri.toString().replace(pathBeforeTitle, '');
+                    // Omit the domain prefix as it could be wrong for node shared between domains
+                    pathBeforeTitle = pathBeforeTitle.replace(/\/[^\/]+\//, '');
```
d00rman: Perhaps anchor the regex to the beginning of the string.

gwicke:
Looks good to go to me, just waiting for the tests.

lib/normalize_title_filter.js
Pchelolo: @d00rman done

PR #569, opened 2016-03-23
Use 302 redirect for commons pages

Pchelolo:
301 is not correct here as local description page might be created instead of commons page. 

cc @wikimedia/services


jdforrester:
This'd be for any ForeignFileRepo page, not just Commons, I assume? Just thinking of third parties here. :-)

Pchelolo:
@jdforrester Yep, the logic is the following: we attempt to get content locally, then check if it's a file namespace, if it is - make a redirect to the shared file repo fetched from MW API.

PR #571, opened 2016-03-23
Added a script to dump URIs used for testing

Pchelolo:
Created a script that would scrub meaningful URIs we use for testing. The script scrubs only GET uris, and excludes uris examining endpoints which we make accessible only for testing. Also, those depending on nock are excluded. 

We don't include POST endpoints, as it's better to just look at the spec for them. 

The RB code total coverage achieved by requesting each of the produced URIs is about 55%. The problem is that many of the code paths are executed only if the content is not in the storage, which wouldn't be true for the fuzzer.

Bug: https://phabricator.wikimedia.org/T129926

cc @wikimedia/services

### DIFF
--- a/test/utils/link_creator.sh
+++ b/test/utils/link_creator.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+DIR="$( cd "$( dirname "$0" )" && pwd )"
+cd ${DIR}/../../
 
 rm -f uri_dump;
 
 # Enable logging in config
-sed -i.bak 's/dump_test_uris:\ false/dump_test_uris:\ true/' config.test.yaml
+sed 's/dump_test_uris:\ false/dump_test_uris:\ true/' config.test.yaml > temp.config.test.yaml
+export RB_TEST_CONFIG=${DIR}/../../temp.config.test.yaml
 
 # Run tests
 sh $DIR/run_tests.sh test sqlite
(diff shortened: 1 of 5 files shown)


test/utils/link_creator.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
d00rman: Make this file executable, otherwise it fails if run with anything else other than bash

test/utils/link_creator.sh
```diff
@@ -0,0 +1,32 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
d00rman: Better: `s/${BASH_SOURCE[0]}/$0` to make it sh-neutral

test/utils/link_creator.sh
```diff
@@ -0,0 +1,32 @@
...
+
+rm -f uri_dump;
```
d00rman: No no no. `s/uri_dump/${DIR}/../../uri_dump` - don't assume the user's CWD.

d00rman: Hmm, the whole script assumes the user's CWD is the restbase repo's root. No bueno. Better for the script to `cd` into it before doing anything: `cd ${DIR}/../..`

test/utils/link_creator.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Enable logging in config
+sed -i.bak 's/dump_test_uris:\ false/dump_test_uris:\ true/' config.test.yaml
```
d00rman: I don't like this. If something goes wrong during the script's execution, we end up with a modified `config.test.yaml` and will need manual intervention.

Pchelolo:
@d00rman Addressed all your comments.

PR #578, opened 2016-03-29
Rename pageview to metrics and add unique devices

jobar:
Rename pageview definition to metrics in order to have only one file defining metrics redirections to AQS.
Add unique devices entry point to this file.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -39,7 +39,7 @@ wikimedia.org: &wikimedia.org
     - path: projects/wikimedia.org.yaml
       options:
         <<: *default_options
-        pageviews:
+        metrics:
           host: https://wikimedia.org/api/rest_v1/metrics
 
 
@@ -98,7 +98,7 @@ info:
 services:
   - name: restbase
     module: hyperswitch
-    conf: 
+    conf:
       port: 7231
       spec: *spec_root
       salt: secret
--- a/v1/metrics.yaml
+++ b/v1/metrics.yaml
@@ -1,8 +1,8 @@
 swagger: '2.0'
 info:
   version: '1.0.0-beta'
-  title: Analytics Pageview API
-  description: Analytics Pageview API
+  title: Analytics Metrics API
+  description: Analytics Metrics API
   termsofservice: https://github.com/wikimedia/restbase#restbase
   contact:
     name: Analytics
@@ -122,9 +122,9 @@ paths:
           required: true
         - name: agent
           in: path
-          description: If you want to filter by agent type, use one of user, bot or spider. If you are interested in pageviews regardless of agent type, use all-agents
+          description: If you want to filter by agent type, use one of user or spider. If you are interested in pageviews regardless of agent type, use all-agents
           type: string
-          enum: ['all-agents', 'user', 'spider', 'bot']
+          enum: ['all-agents', 'user', 'spider']
           required: true
         - name: granularity
           in: path
@@ -134,12 +134,12 @@ paths:
           required: true
         - name: start
           in: path
-          description: The timestamp of the first day to include, in YYYYMMDDHH format
+          description: The timestamp of the first hour/day/month to include, in YYYYMMDDHH format
           type: string
           required: true
         - name: end
           in: path
-          description: The timestamp of the first day to include, in YYYYMMDDHH format
+          description: The timestamp of the last hour/day/month to include, in YYYYMMDDHH format
           type: string
           required: true
       responses:
@@ -186,7 +186,7 @@ paths:
         - Pageviews data
       summary: Get the most viewed articles for a project.
       description: |
-        Lists the 1000 most viewed articles for a given project and timespan (year, month or day). You can filter by access method
+        Lists the 1000 most viewed articles for a given project and timespan (month or day). You can filter by access method
         Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
       produces:
         - application/json
@@ -204,7 +204,7 @@ paths:
           required: true
         - name: year
           in: path
-          description: The year of the date for which to retrieve top articles, in YYYY format. If you want to get the top articles of a whole year, the month and day parameters should be all-months and all-days respectively
+          description: The year of the date for which to retrieve top articles, in YYYY format.
           type: string
           required: true
         - name: month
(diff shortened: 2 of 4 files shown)


config.example.wikimedia.yaml
```diff
@@ -39,7 +39,7 @@ wikimedia.org: &wikimedia.org
     - path: projects/wikimedia.org.yaml
       options:
         <<: *default_options
-        pageviews:
+        metrics:
```
d00rman: For the time being, I think we should leave the name as-is, since changing the name requires us to do a RB config change which slows down the deployment unnecessarily.

milimetric: I disagree, I think we don't really have a choice since we're adding a new endpoint to AQS.  Don't we need an RB config change for that anyway?  But even if we don't, it doesn't make too much sense to point something called pageviews to AQS in general.

d00rman: No, that's not true @milimetric . All this does is it renames an internal options object's property name. I'm not saying we shouldn't do this, I'm saying going through a config change procedure just to change a name that is seen nowhere and affects nothing is futile and a waste of time. Once we have some other config changes to do to RB, I'd be more than happy to include this small change too.

milimetric: Ok, but what I mean is, don't we have to change RB anyway because we're
adding a new endpoint for /metrics/unique-devices next to
/metrics/pageviews ?  So while we do that we might as well rename this?

d00rman: Yes, but the change is happening in `/projects/wikimedia.org` which is located in this repo, so all is needed here is a deploy of RESTBase. To go the extra step of renaming `options.pageviews` to `options.metrics` we'd need a coordinated config and repo deploy of the front-end RB, which involves a patch to `ops/puppet` and coordination with Ops.

jobar:
@d00rman : While I understand the concern of deploy synchronization, the only question I have is:
if we don't do it now, will it ever happen? Meaning, if you have synchronized deploys once every now and then, we can easily have a ready-to-be-merged PR on the parameter name while not waiting for this big change. If synchronized deploys don't happen often, then maybe it's better to have it done this time make sure it's done and not forgotten? What do you think is best?

Pchelolo:
@jobar It's better to separate the renaming to a separate PR and leave that PR pending.

Normally we mark such PRs with a `config change` on github, accumulate a number of them and do a synchronised code-config deploy.

v1/metrics.yaml
```diff
@@ -232,6 +232,59 @@ paths:
...
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      produces:
+        - application/json
+      parameters:
+        - name: project
+          in: path
+          description: The name of any Wikimedia project formatted like {language code}.{project name}, for example en.wikipedia. You may pass en.wikipedia.org and the .org will be stripped off. For projects like commons without language codes, use commons.wikimedia
+          type: string
+          required: true
+        - name: access-site
```
Pchelolo: Why is this parameter called `access-site` while a parameter with almost the same semantics in other endpoints is called just `access`?

v1/metrics.yaml
```diff
@@ -232,6 +232,59 @@ paths:
               uri: '{{options.host}}/pageviews/top/{project}/{access}/{year}/{month}/{day}'
       x-monitor: false
 
+  /unique-devices/{project}/{access-site}/{granularity}/{start}/{end}:
```
Pchelolo: If we wanted to be in line with other endpoints, I'd say that this should be named `/metrics/pageviews/unique/{other_parameters}`. However, taking into account the https://phabricator.wikimedia.org/T126947 bug, I understand why this doesn't include the `pageviews` word. 

However, as I understand we want to gradually migrate `/metrics/pageviews/` to `/metrics/views` or something similar? May be we should expose the new endpoint in line with the intended future name for other, older endpoints? So, finish bike shedding on migration first and use a new layout right away?

milimetric: Hm, but uniques is a completely different metric.  It has nothing to do with pageviews.  We're not putting it higher up because of the ad-block bug, it really is separate.  Hence the desire to rename above.  So, just to be clear, pageviews and unique_devices are coming from two totally different data pipelines that have nothing to do with each other besides the basic varnishkafka -> kafka -> camus that all our webrequest data comes through.

v1/metrics.yaml
milimetric: it's not the same semantics, we're trying to keep names short but as descriptive as possible.  `access` is the client access method (desktop, mobile, etc.).  And `access-site` is the version of our site that the client accesses (en.wikipedia vs. en.m.wikipedia).  If the docs aren't clear on this, we should change them.

v1/metrics.yaml
Pchelolo: @milimetric Ah, sorry for misunderstanding. I thought it's like "vies of the site by unique devises"..

v1/metrics.yaml
Pchelolo: @milimetric So, if I'm going from a mobile device to the desktop version of the site the `access` would be `mobile`, but the `access-site` will be `desktop-site`, right?

milimetric: Sorry, I wrote this quickly and it's more confusing than that.

 * `access` is the way our content was accessed from the client point of view.  So it does include the desktop site (`en.wikipedia.org`), the mobile site (`en.m.wikipedia.org`), but also the mobile app (however that works, we track mw api requests, restbase, etc.)
 * `access-site` doesn't know anything about the client, and so it can only be `en.wikipedia.org` which we call desktop, or `en.m.wikipedia.org` which we call mobile

This is due to how the cookie that powers the `unique_devices` data works, and I agree now that the two parameters have almost the same semantics, which is why we chose a similar name.  But we didn't want the same exact name because of the mobile-app difference and potentially other differences as we expand each metric (like we may have facebook, iOS, or google knowledge graph choices for the `access` parameter, but we'll be restricted to our own domains for the unique devices dataset).

Pchelolo: Ah, ok, thank you for explanation.

d00rman:
+1 to what @Pchelolo said. @jobar the main point here is that renaming `options.pageviews` to `options.metrics` gives us nothing but pain in the current scheme of things. The only thing we gain by doing it now is to have config consistency, but given that this is only an internal thing to the front-end RB instance, I don't it's worth going through a coordinated deploy. From my side, I can also say that going through this procedure is not an option for me in a time when (a) the end of the quarter is a couple of days away; and (b) I am going to the Hackathon and hence possibly not be fully available to deal with potential consequences.

FTR, when we do change this, we'll need to change it in MW-Vagrant too, not only in labs/staging/prod. TL;DR too much pain for literally no gain.

d00rman:
oh, and @jobar could you please rebase this on current master? The travis failures witnessed here are not related to this PR, so let's have that on record too ;)

jobar:
Dropping this PR, creating two new ones.

PR #580, opened 2016-03-29
Only wrapContentReq after saving the Parsoid result

gwicke:
One no-cache test with sections showed a failure after merging content-type
fixes. This patch fixes those failures.

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -372,10 +372,10 @@ PSP.generateAndSave = function(hyper, req, format, currentContentRes) {
                 body: res.body[format].body
             };
             resp.headers.etag = mwUtil.makeETag(rp.revision, tid);
-            var responsePromise = self.wrapContentReq(hyper, req,
-                    P.resolve(resp), format, tid);
             return self.saveParsoidResult(hyper, req, format, tid, res)
             .then(function() {
+                var responsePromise = self.wrapContentReq(hyper, req,
+                        P.resolve(resp), format, tid);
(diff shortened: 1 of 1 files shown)


sys/parsoid.js
```diff
@@ -372,10 +372,10 @@ PSP.generateAndSave = function(hyper, req, format, currentContentRes) {
                 body: res.body[format].body
             };
             resp.headers.etag = mwUtil.makeETag(rp.revision, tid);
-            var responsePromise = self.wrapContentReq(hyper, req,
-                    P.resolve(resp), format, tid);
             return self.saveParsoidResult(hyper, req, format, tid, res)
             .then(function() {
+                var responsePromise = self.wrapContentReq(hyper, req,
+                        P.resolve(resp), format, tid);
```
Pchelolo: So, if I understand correctly we have to query section offsets only after we've said new offsets, right?

gwicke: Yes, I think that's right.

gwicke:
Merged to un-break tests on master.

sys/parsoid.js
Pchelolo: @gwicke Just it's not 100% clear to me what's happening here and why was this bug exposed only after content-type version bump. Could you point me to a failed test?

gwicke: @Pchelolo, it's possible that this failure hid between all the content-type failures, and thus managed to sneak through. I restarted https://travis-ci.org/wikimedia/restbase/builds/119320343, which should show the failure fixed by this patch.

PR #581, opened 2016-03-29
Added tests for pageviews

Pchelolo:
We've recently accidentally broken pageviews API and didn't notice that neither using tests nor monitoring. Add monitoring for page views endpoints to avoid such situations in future.

/cc @wikimedia/services @milimetric 

@milimetric Do I assume correctly that numbers used in the response validation are not supposed to change ever?

### DIFF
--- a/test/features/metrics.js
+++ b/test/features/metrics.js
@@ -1,0 +1,47 @@
+"use strict";
+
+var assert = require('../utils/assert.js');
+var server = require('../utils/server.js');
+var preq   = require('preq');
+
+describe('Metrics', function() {
+    this.timeout(20000);
+
+    before(function () { return server.start(); });
+
+    it('Should get page views per page', function() {
+        return preq.get({
+            uri: server.config.globalURL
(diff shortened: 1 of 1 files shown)


d00rman:
It might be wise to wait for #578 to land and then pick one end-point under `/pageviews/` and the other under `/unique-devices/`.

Pchelolo:
@d00rman Might be wise to wait, I agree, however I'd like to leave monitoring for every page views endpoint. We don't have any tests for them, so they're really easy to break. However, I could convert some monitoring checks to normal mocha tests if you think monitoring each of them is too much.

Pchelolo:
Ups, accidentally clicked a wrong button

d00rman:
> I could convert some monitoring checks to normal mocha tests if you think monitoring each of them is too much.

:+1: to converting. The monitoring tests are run in prod every minute on every host, so we should keep them to a minimum.

Pchelolo:
@d00rman converted

PR #583, opened 2016-03-30
Rename pageview to metrics and add unique devices

jobar:
Rename pageview definition to metrics in order to have only one file defining metrics redirections to AQS.
Add unique devices entry point to this file.

PR #584, opened 2016-03-30
Modify pageviews host option

jobar:
To reflect change in frontend for analytics service, pageviews are renamed metrics (since we now serve more than pageviews only - uniques devices to be precise).

This PR has a dependency on https://github.com/wikimedia/restbase/pull/583.


d00rman:
The file rename happened at some point, so I think it's safe to abandon this PR. OK with you @jobar ?

jobar:
Ok with me :)
There is a name leftover in here: https://github.com/wikimedia/restbase/blob/master/projects/wikimedia.org.yaml#L71

I'll care that in the near future :)

PR #585, opened 2016-03-31
Update service-runner and hyperswitch dependencies

gwicke:


Pchelolo:
Oh, travis is too slow, should be fine, merging

PR #586, opened 2016-04-01
Add query string to redirects

Pchelolo:
We need to preserve the query string on redirects to avoid losing `sections` part.

cc @wikimedia/services


gwicke:
:+1: 

Waiting for tests to run.

lib/normalize_title_filter.js
```diff
@@ -21,6 +22,13 @@ function normalizeTitle(title, siteInfo) {
     }
 }
 
+function getQueryString(req) {
+    if (Object.keys(req.query).length) {
```
d00rman: Do we guarantee `req.query` will never be `null`/`undefined` ?

Pchelolo: Yes we do.

PR #587, opened 2016-04-04
Drop custom title normalizations and rely on the filter

Pchelolo:
After mediawiki-title was integrated all the titles incoming to any of the endpoint request are guaranteed to be in normalised form. The only source of non-normalised titles left is MW API. It returns the title in the text form with spaces, not the DBKey format with underscores, so we need to apply title normalisation to the result as well. This also ensures that we're storing gender-neutral version for gendered namespaces.

If the result title doesn't match the originally requested one - some non-needed normalisation happened, so we need to throw 404 as per https://phabricator.wikimedia.org/T131726

Also, some of our tests started to fail after this fix, but that's good, it exposed bugs in tests

cc @wikimedia/services

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -112,8 +112,9 @@ module.exports = function(hyper, req, next, options, specInfo) {
                                 'cache-control': options.redirect_cache_control || 'no-cache'
                             }
                         };
+                    } else {
+                        throw e;
                     }
-                    throw e;
                 });
             });
         }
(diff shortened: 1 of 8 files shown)


lib/normalize_title_filter.js
```diff
@@ -107,26 +90,31 @@ module.exports = function(hyper, req, next, options, specInfo) {
...
+                        return {
+                            status: 302,
+                            headers: {
+                                location: redirectPath,
+                                'cache-control': options.redirect_cache_control || 'no-cache'
+                            }
+                        };
                     }
-                };
+                    throw e;
```
gwicke: [Minor / matter of taste]: I personally prefer to use an else branch in this kind of situation to make the exclusivity clearer. I had to scan back up to figure out that the early return prevents this throw for commons redirects, while I wouldn't have done the same with an else.

gwicke:
:+1:

PR #588, opened 2016-04-04
Reenable dependency updates for all namespaces

Pchelolo:
Re https://phabricator.wikimedia.org/T131554

Also I've refactored the event emitting test a bit to make it more reliable.


gwicke:
:+1:

PR #590, opened 2016-04-05
Simple content-type enforcement filter

gwicke:
This patch adds a simple content-type check & enforcement filter, which

1) compares the returned content-type with the first entry of the `produces`
   array in the route spec, and
2) repeats the request with `no-cache` header set if the content-type
   mis-matches.

### DIFF
--- a/lib/ensure_content_type.js
+++ b/lib/ensure_content_type.js
@@ -26,51 +26,55 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
     .then(function(res) {
         // Do not check or re-render if the response was only rendered within
         // the last two minutes.
-      //  if (res.headers.etag
-      //      && Date.now() - mwUtil.extractDateFromEtag(res.headers.etag) < 120000) {
-      //      return res;
-      //  }
+        if (res.headers && res.headers.etag
+            && Date.now() - mwUtil.extractDateFromEtag(res.headers.etag) < 120000) {
+            return res;
+        }
+
+        function updateSpecWarning() {
+            // Returned content type version is higher than what the
+            // spec promises. Log a warning about the need to update the spec.
+            hyper.log('warn/content-type/spec_outdated', {
+                msg: 'Spec needs update to reflect latest content type.',
+                expected: expectedContentType,
+                actual: res.headers['content-type']
+            });
+            // Don't fail the request.
+            return res;
+        }
 
         if (res.headers && res.headers['content-type'] !== expectedContentType) {
             // Parse the expected & response content type, and compare profiles.
             var expectedProfile = cType.parse(expectedContentType).parameters.profile;
             var actualProfile = cType.parse(res.headers['content-type']).parameters.profile;
 
-            if (actualProfile && expectedProfile && actualProfile !== expectedProfile) {
+            if (actualProfile && actualProfile !== expectedProfile) {
+                if (!expectedProfile) {
+                    return updateSpecWarning();
+                }
                 // Check if actual content type is newer than the spec
                 var actualProfileParts = splitProfile(actualProfile);
                 var expectedProfileParts = splitProfile(expectedProfile);
                 if (actualProfileParts.path === expectedProfileParts.path
                         && actualProfileParts.version > expectedProfileParts.version) {
-                    // Returned content type version is higher than what the
-                    // spec promises. Log a warning about the need to update the spec.
-                    hyper.log('warn/content-type/spec_outdated', {
-                        msg: 'Spec needs update to reflect latest content type.',
-                        expected: expectedContentType,
-                        actual: res.headers['content-type']
-                    });
-                    // Don't fail the request.
-                    return res;
+                    return updateSpecWarning();
                 }
             }
 
-            // The old content might not have any profile at all
-            if (expectedProfile && actualProfile !== expectedProfile) {
-                // Re-try request with no-cache header
-                if (!mwUtil.isNoCacheRequest(req)) {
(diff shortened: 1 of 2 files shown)


gwicke:
TODO:

[ ] Tests

Pchelolo:
This will unlikely work for `/html/{title}` endpoint as it can return `application/json` in case the sections attribute is provided.

gwicke:
Testing the upgrade path of this filter looks a bit tricky. We'd need to find a way to get older content into the database (or fake a different content type, perhaps in another test-only filter?). For now, testing the normal code path (which does not trigger updates) looks fairly decent already. Once we add it to the mobile apps default config this will be tested as part of the existing tests.

Pchelolo:
My concern re "This will unlikely work for /html/{title} endpoint as it can return application/json in case the sections attribute is provide" is still valid. I understand that we wanna enable it on mobile endpoints at first, but for generality we could try to address that issue too.

As for the tests - we could use `nock` to test the upgrade codepath.

gwicke:
> My concern re "This will unlikely work for /html/{title} endpoint as it can return application/json in case the sections attribute is provide" is still valid. I understand that we wanna enable it on mobile endpoints at first, but for generality we could try to address that issue too.

For parsoid, we'll need more sophisticated pagebundle2pagebundle and format negotiation support, which will be fairly parsoid-specific. I don't think it would make sense to use this general-purpose filter for that.

lib/ensure_content_type.js
```diff
@@ -0,0 +1,89 @@
...
+ *   match the latest version.
+ */
+
+function checkContentType(hyper, req, next, expectedContentType, responsePromise) {
+    return responsePromise
+    .then(function(res) {
+        if (res.headers && res.headers['content-type'] !== expectedContentType) {
+            // Parse the expected & response content type, and compare
+            // profiles.
+            var expectedProfile = cType.parse(expectedContentType).parameters.profile;
```
Pchelolo: When I was experimenting with the `content-type` module, I've found that `parse` is pretty slow. Since we don't expect to have a lot of different content-types, it could be cool to cache the results and key them on the `content-type` string.

lib/ensure_content_type.js
```diff
@@ -0,0 +1,89 @@
...
+
+                // Do not re-render if the response was only rendered within
+                // the last two minutes.
+                if (res.headers.etag
+                        && Date.now() - mwUtil.extractDateFromEtag(res.headers.etag) < 120000) {
+                    return res;
+                }
+
+                // Re-try request with no-cache header
+                if (!/no-cache/.test(req.headers['cache-control'])) {
```
Pchelolo: There's an `mwUtils.isNoCacheRequest` utility function for that

lib/ensure_content_type.js
```diff
@@ -0,0 +1,89 @@
...
+                    req.headers['cache-control'] = 'no-cache';
+                    return checkContentType(hyper, req, next,
+                            expectedContentType, next(hyper, req));
+                } else {
+                    // Log issue
+                    hyper.log('warn/content-type/upgrade_failed', {
```
Pchelolo: If we've failed to provide the requested version of the content, may be we should not varnish-cache it to avoid clients getting stuck with outdated content?

gwicke: Good point. At the very least, we should reduce the Varnish cache ttl significantly.

lib/ensure_content_type.js
gwicke: Yeah, a utility wrapper for that could be nice. We'd need to make sure that we don't mutate the result, but perhaps we can use Object.freeze() for that.

lib/ensure_content_type.js
gwicke: Switched to that now.

gwicke:
@Pchelolo, anything blocking this?

Pchelolo:
@gwicke all travis tests are failing...

Pchelolo:
@gwicke I've rebased your PR and modified a filter a bit to accommodate summary migration (it didn't have any profile before), please have a look. For me this LGTM

lib/ensure_content_type.js
```diff
@@ -45,19 +52,15 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
...
 
-                // Do not re-render if the response was only rendered within
-                // the last two minutes.
-                if (res.headers.etag
-                        && Date.now() - mwUtil.extractDateFromEtag(res.headers.etag) < 120000) {
-                    return res;
-                }
-
+            // The old content might not have any profile at all
+            if (expectedProfile && actualProfile !== expectedProfile) {
```
gwicke: Why is this needed? Undefined would differ from an expected profile, so I would expect this case to be handled already.

Pchelolo: Em, didn't understand your question

gwicke: Could you explain the thinking behind adding this condition? Why is the simple content-type equality check not sufficient?

Pchelolo: @gwicke In case the spec doesn't specify the profile but the response does we would rerender, while we actually shouldn't

gwicke: I was assuming that it does not make sense to configure this content-type versioning filter if the spec does not have any versioned content type. If we want to protect against such a misconfiguration, perhaps we could just fatal on startup?

Pchelolo: We can't fatal right away since there's no code that's executed on startup, and we don't wanna fatal on request.

Pchelolo:
LGTM

PR #591, opened 2016-04-05
Redirect filter

Pchelolo:
First iteration of the redirect support in RESTBase.

Note: before merging this we need to temporarily disable caching for redirect=no requests, ask mobile apps and VE to supply redirect=no query param. After that is merged we can implement a varnish optimisation so that VE shared the cache. Also we will be able to remove the redirect handling from mobile service and make them rely on RESTBase.

Further performance optimisations will be achieved with restrictions table.

Bug: https://phabricator.wikimedia.org/T131570

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -4,7 +4,6 @@ var HyperSwitch = require('hyperswitch');
 var mwUtil = require('./mwUtil');
 var URI = HyperSwitch.URI;
 var P = require('bluebird');
-var querystring = require('querystring');
 
 function getQueryString(req) {
     if (Object.keys(req.query).length) {
@@ -62,28 +61,10 @@ module.exports = function(hyper, req, next, options, specInfo) {
                     return res;
                 });
             } else {
-                var pathBeforeTitle = specInfo.path
-                    .substring(0, specInfo.path.indexOf('{title}'));
-                pathBeforeTitle = new URI(pathBeforeTitle, rp, true).toString();
-                // Omit the domain prefix as it could be wrong for node shared between domains
-                pathBeforeTitle = pathBeforeTitle.replace(/^\/[^\/]+\//, '');
-                var pathSuffix = req.uri.toString()
-                    .replace(/^\/[^\/]+\//, '')
-                    .replace(pathBeforeTitle, '');
-                var pathSuffixCount = (pathSuffix.match(/\//g) || []).length;
-                var backString = Array.apply(null, { length: pathSuffixCount }).map(function() {
-                    return '../';
-                }).join('');
-                var pathPatternAfterTitle = specInfo.path
-                    .substring(specInfo.path.indexOf('{title}') - 1);
-                var contentLocation = backString
-                    + new URI(pathPatternAfterTitle, rp, true).toString().substr(1)
-                    + getQueryString(req);
-
                 return P.resolve({
                     status: 301,
                     headers: {
-                        location: contentLocation,
+                        location: mwUtil.createRelativeTitleRedirect(specInfo.path, req),
                         'cache-control': options.redirect_cache_control || 'no-cache'
                     }
                 });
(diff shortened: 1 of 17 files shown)


lib/normalize_title_filter.js
```diff
@@ -94,7 +66,8 @@ module.exports = function(hyper, req, next, options, specInfo) {
                 // until https://phabricator.wikimedia.org/T130757 (VE
                 // redirect support) is resolved.
                 // TODO: Remove.
-                && /^WikipediaApp\//.test(req.headers['user-agent'])) {
+                && /^WikipediaApp\//.test(req.headers['user-agent'])
+                && req.query.redirect !== 'no') {
```
gwicke: Lets declare this as an official parameter, and rely on the coercion code in hyperswitch to normalize this (see https://github.com/wikimedia/hyperswitch/pull/33).

lib/revision_table_access_check_filter.js
```diff
@@ -0,0 +1,55 @@
...
+    var checkURIParts = [rp.domain, 'sys', 'page_revisions', 'page', rp[titleName]];
+    if (rp.revision) {
+        checkURIParts.push('' + rp.revision);
+    }
+
+    return hyper.get({ uri: new URI(checkURIParts) })
+    .then(function(res) {
+        var revision = res.body.items[0];
+
+        if (revision.redirect && req.query.redirect !== 'no') {
```
gwicke: `&& ! req.query.redirect`, with the updated coercion from hyperswitch.

lib/revision_table_access_check_filter.js
```diff
@@ -0,0 +1,55 @@
+"use strict";
+
+// TODO: This is a temp solution that will
+// be eventually replaced by the restrictions table.
+
+var URI = require('hyperswitch').URI;
+var mwUtil = require('./mwUtil');
+var entities = require('entities');
+
+var redirectRegEx = /<link rel="mw:PageProp\/redirect" href="\.\/([^"#]+)"/;
```
gwicke: This won't match redirects to sections (with a literal `#`).

Pchelolo: This will remove the sections from a redirect. And that's what we need as RB doesn't support sections anyway

gwicke: If there is a literal hash, then I think the redirect won't be matched at all. I will never get past the hash.

projects/example.yaml
```diff
@@ -34,6 +34,8 @@ paths:
             /page:
               x-modules:
                 - path: v1/content.yaml
+                  options:
+                    response_cache_control: '{{options.purged_cache_control}}'
```
gwicke: `Response` is fairly generic. Some of our responses aren't purged, and have considerably shorter TTLs.

What do you think about keeping the name `purged_cache_control`?

lib/revision_table_access_check_filter.js
```diff
@@ -0,0 +1,55 @@
...
+            return hyper.get({ uri: new URI(htmlPath) })
+            .then(function(res) {
+                var html = res.body;
+                var redirectMatch = redirectRegEx.exec(html);
+                if (redirectMatch) {
+                    var newParams = Object.assign({}, rp);
+                    newParams[titleName] = decodeURIComponent(entities.decodeXML(redirectMatch[1]));
+                    var location = mwUtil.createRelativeTitleRedirect(specInfo.path, req,
+                        newParams, titleName);
+                    return next(hyper, req).then(function(res) {
```
gwicke: You could kick off the request a bit earlier by calling next() before the `hyper.get()` on line 29.

sys/page_revisions.js
```diff
@@ -301,11 +301,14 @@ PRS.prototype.fetchAndStoreMWRevision = function(hyper, req) {
...
         .then(function(res) {
-            var sameRev = res && res.body.items
-                    && res.body.items.length > 0
-                    && self._checkSameRev(revision, res.body.items[0]);
-            if (!sameRev) {
-                throw new HTTPError({ status: 404 });
+            if (res && res.body.items && res.body.items.length > 0) {
+                var storedRev = res.body.items[0];
+                // The redirect in MW API is based on the latest revision,
+                // so for older revisions it must never be updated.
```
gwicke: Perhaps add a TODO about this not being accurate for old revisions (might miss redirects if latest version is no longer a redirect).

gwicke:
:+1: overall, only a few nits inline.

sys/page_revisions.js
Pchelolo: Done

gwicke:
I think this is ready to go, but we need to wait for VisualEditor and the mobile content service to pass `?redirect=false` before we can merge & deploy this.

PR #593, opened 2016-04-08
Fix tests

gwicke:


Pchelolo:
@gwicke Fixed now, but let's squash these commits when merging :)

PR #594, opened 2016-04-08
Update metrics endpoints documentation for project

jobar:
AQS now strips trailing www. and ending .org from project names, this patch
updates the documentation accordingly.

### DIFF
--- a/v1/metrics.yaml
+++ b/v1/metrics.yaml
@@ -47,7 +47,7 @@ paths:
       parameters:
         - name: project
           in: path
-          description: The name of any Wikimedia project formatted like {language code}.{project name}, for example en.wikipedia. You may pass en.wikipedia.org and the .org will be stripped off.  For projects like commons without language codes, use commons.wikimedia
+          description: If you want to filter by project, use the name of any Wikimedia project formatted like {language code}.{project name}, for example en.wikipedia. You may pass en.wikipedia.org and the .org will be stripped off. For projects like commons without language codes, use commons.wikimedia. For projects like www.mediawiki.org, you can use that full string, or just use mediawiki or mediawiki.org. If you are interested in all pageviews regardless of project, use all-projects.
           type: string
           required: true
         - name: access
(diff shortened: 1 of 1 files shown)


v1/metrics.yaml
```diff
@@ -47,7 +47,7 @@ paths:
       parameters:
         - name: project
           in: path
-          description: The name of any Wikimedia project formatted like {language code}.{project name}, for example en.wikipedia. You may pass en.wikipedia.org and the .org will be stripped off.  For projects like commons without language codes, use commons.wikimedia
+          description: If you want to filter by project, use the name of any Wikimedia project formatted like {language code}.{project name}, for example en.wikipedia. You may pass en.wikipedia.org and the .org will be stripped off. For projects like commons without language codes, use commons.wikimedia. For projects like www.mediawiki.org, you can use that full string, or just use mediawiki or mediawiki.org. If you are interested in all pageviews regardless of project, use all-projects.
           type: string
```
Pchelolo: Could be nice to wrap examples in ` to make them stand out of the text like `{language code}.{project name}`

gwicke:
I am not a huge fan of offering many different ways of requesting the same project. It fragments caches even further, and complicates the documentation. But, that ship has sailed..

v1/metrics.yaml
d00rman: :+1: I second the idea.

d00rman:
> I am not a huge fan of offering many different ways of requesting the same project. It fragments caches even further, and complicates the documentation. But, that ship has sailed..

That's https://phabricator.wikimedia.org/T119094 . Out of scope for this patch, but let's start bike-shedding about it there.

gwicke:
> That's https://phabricator.wikimedia.org/T119094 . Out of scope for this patch, but let's start bike-shedding about it there.

This is a cached end point, so I think caching and issues of fragmentation are in scope here as well. This particular patch actually adds a lot of verbiage to describe all the different ways of selecting projects, and at the same time makes those variants part of the public API.

jobar:
While I understand the various points, this specific PR is just about documentation.
The change of striping www and .org from projects has been asked long time ago and by enhough of our users for to implement it.
I agree, it's not optimal for caching and clarity (multiple ways to access the same data). However for some users "www.mediawiki.org makes more sense than "mediawiki" in the project field.
I view three major roads to move forward:
- We keep the functionality but don't advertise it (reject this PR)
- We keep the functionality and advertise it (merge this PR)
- We remove the functionality altogether (developement involved on AQS + user management).
How do you want us to sync on that?

gwicke:
@jobar, I would propose to encourage users to (ideally) use a single variant in the documentation, and gradually phasing out / deprecating the others. 

To me, the reduction of variants is more important than which exact variant is picked. That said, my personal preference would be to use the full domain ("www.mediawiki.org"), as that will also play well with per-domain entry points. People know the domains, so using them everywhere promotes consistency.

milimetric:
I agree with encouraging people to use the full domain now that we're stripping properly.  It just wasn't possible before.

jobar:
Closing this one to reopen a [new one](https://github.com/wikimedia/restbase/pull/596) with doc updated according to this discussion.

PR #595, opened 2016-04-12
Optimize html fetching for redirect check, enable mobile service tests

Pchelolo:
1. We only need to attach body to redirects where it makes sense - for HTML and data-parsoid.
2. Enable some mobile-related tests
3. Optimise content fetch in a filter - don't fetch html twice

cc @wikimedia/services


d00rman:
LGTM. Looking at this PR, we need to find a way to properly document all the possible configs for filters.

Pchelolo:
@d00rman Either JSDoc on the filter method with explanation of `options` parameter, or we could create a new page in /doc folder and list all the filters over there

d00rman:
> Either JSDoc on the filter method with explanation of options parameter, or we could create a new page in /doc folder and list all the filters over there

Ideally, I'd like to see both, but let's differ that to another PR. Out of scope here.

PR #597, opened 2016-04-15
Sunset rest.wikimedia.org

gwicke:
Start returning errors, but provide a friendly hint in the error details.

### DIFF
--- a/lib/normalize_title_filter.js
+++ b/lib/normalize_title_filter.js
@@ -1,33 +1,34 @@
 "use strict";
 
+var HyperSwitch = require('hyperswitch');
+var HTTPError = HyperSwitch.HTTPError;
 var mwUtil = require('./mwUtil');
 var P = require('bluebird');
 
 module.exports = function(hyper, req, next, options, specInfo) {
     var rp = req.params;
     /**
-     * Temporary logging & handling for rest.wikimedia.org requests.
+     * Temporary hint response for rest.wikimedia.org requests.
      */
     if (req.headers.host && req.headers.host === 'rest.wikimedia.org') {
-        // 1:10 sampled logging until we know that the rate is really low.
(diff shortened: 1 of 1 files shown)


lib/normalize_title_filter.js
```diff
@@ -1,33 +1,32 @@
...
-            return origNext(hyper, req)
-            .then(function(res) {
-                if (res) {
-                    res.headers = res.headers || {};
-                    res.headers['cache-control'] = 'no-cache';
-                }
-                return res;
-            });
-        };
+        hyper.log('warn/normalize/rest_wikimedia_org', {
```
Pchelolo: I think it's better to still sample the logs 1:10

Pchelolo:
:sunrise: :wave: rest.wikimedia.org 👍

lib/normalize_title_filter.js
d00rman: :+1: at least until we know that the volume is low

lib/normalize_title_filter.js
```diff
@@ -1,33 +1,32 @@
...
+            msg: 'Request to rest.wikimedia.org',
+        });
+
+        // Satisfy jscs's 100 char limit
+        var mURL = 'https://lists.wikimedia.org/pipermail/mediawiki-api/2016-January/003691.html';
+        throw new HTTPError({
+            status: 410,
+            body: {
+                type: 'not_found',
```
d00rman: `s/not_found/gone/` ?

gwicke: I originally used not_found, as we already have an error page for that at https://www.mediawiki.org/wiki/HyperSwitch/errors/not_found. I agree that `gone` is more precise though, so I went ahead & created https://www.mediawiki.org/wiki/HyperSwitch/errors/gone / changed the type.

PR #598, opened 2016-04-22
Update README.md

physikerwelt:
Add Cassandra version requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,8 @@ node server
 The defaults without a config file should work for a local Cassandra
 installation with the default passwords. Restbase has been tested with
 [Cassandra 2.1.13](http://wiki.apache.org/cassandra/DebianPackaging).
-To customize RESTBase's behavior,
-copy the example config to its default location:
+To customize RESTBase's behavior, copy the example config to its
+default location:
 
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +54,9 @@ node server
 ```
 
 The defaults without a config file should work for a local Cassandra
-installation with the default passwords. To customize RESTBase's behavior,
+installation with the default passwords. Restbase has been tested with
+[Cassandra 2.1.13](http://wiki.apache.org/cassandra/DebianPackaging).
```
d00rman: The link doesn't point to the exact Cassandra version, so why include it?

README.md
```diff
@@ -54,7 +54,9 @@ node server
...
+To customize RESTBase's behavior,
```
d00rman: Nit: reformat with the line below.

README.md
physikerwelt: It's the link to the installation guide and might help people to install cassandra. It says
> You will want to replace 21x by the series you want to use: 20x for the 2.0.x series, 12x for the 1.2.x series, etc... You will not automatically get major version updates unless you change the series, but that is a feature.

PR #599, opened 2016-04-26
Access check redux

gwicke:
Rebase & amend of #524.


gwicke:
Next steps:

0) rebase on current master,
1) create a process / script for back-filling restrictions, and
2) properly supporting page un-deletions. This will require some work in EventBus, to supply the revision a page should be un-deleted to.

sys/parsoid.js
```diff
@@ -159,9 +159,8 @@ PSP._dependenciesUpdate = function(hyper, req) {
     });
 };
 
-PSP.getBucketURI = function(rp, format, tid, useKeyRevValue) {
-    var bucket = useKeyRevValue ? 'key_rev_value' : this.options.bucket_type;
-    var path = [rp.domain, 'sys', bucket, 'parsoid.' + format, rp.title];
+PSP.getBucketURI = function(rp, format, tid) {
+    var path = [rp.domain, 'sys', this.options.bucket_type, 'parsoid.' + format, rp.title];
```
Pchelolo: Why do we need to change this?

sys/page_revisions.js
```diff
@@ -373,16 +536,19 @@ PRS.prototype.getTitleRevision = function(hyper, req) {
...
-                    result.restrictions.push('page_deleted');
-                    return hyper.put({
-                        uri: self.tableURI(rp.domain),
-                        body: {
-                            table: self.tableName,
-                            attributes: result
-                        }
-                    }).throw(e);
+                    result.tid = TimeUuid.now().toString();
+                    result.page_deleted = result.rev;
```
Pchelolo: When we switch completely to the new access checks model we'd be able to remove this. TODO might be nice.

sys/parsoid.js
gwicke: Not sure where this came from. Reverted it. 

Thanks for caching this!

sys/page_revisions.js
gwicke: Yup, added a todo about handling deletions / un-deletions with targeted events.

sys/page_revisions.js
```diff
@@ -80,6 +83,190 @@ PRS.prototype.getTableSchema = function() {
...
+
+PRS.prototype.storeRestrictions = function(hyper, req, revision) {
+    var self = this;
+    var rp = req.params;
+    // Do not even define attributes we don't want to overwrite.
+    var restrictionObject = {};
+    if (revision.restrictions && revision.restrictions.length) {
+        restrictionObject.restrictions = revision.restrictions;
+    }
+    if (revision.redirect !== true && revision.redirect) {
```
Pchelolo: What's is this supposed to mean? the `revision.redirect` is a boolean, how can it be trythy and not a `true`?

sys/page_revisions.js
```diff
@@ -80,6 +83,190 @@ PRS.prototype.getTableSchema = function() {
...
+                            // TODO: Reset page_deleted to actual start
+                            // revision!
+                            page_deleted: null,
+                        }
+                    }
+                });
+            }
+            return P.resolve({ status: 200 });
+        })
+        .catch({ status: 404 }, function() {
```
Pchelolo: Minor. Could use `catchReturn` here

v1/content.yaml
```diff
@@ -180,7 +180,11 @@ paths:
 
   /html/{title}:
     x-route-filters:
-      - path: ./lib/revision_table_access_check_filter.js
+      #- path: ./lib/revision_table_access_check_filter.js
+      #  options:
+      #    redirect_cache_control: '{{options.purged_cache_control}}'
+      #    attach_body_to_redirect: true
+      - path: ./lib/access_check_filter.js
```
Pchelolo: Why are we updating only these 2 endpoints? I thought we would replace the filter everywhere.

sys/page_revisions.js
gwicke: The case we are interested in is it being a `string`, the redirect target. The `true` case is the old revision object from the API, which we want to ignore.

v1/content.yaml
gwicke: See the commit summary. I enabled this for these two end points so that we can test this, but am proposing to actually revert to revision_table_access_check before merging this PR, so that we can gradually phase in the revision table / back-fill restrictions.

sys/page_revisions.js
Pchelolo: Oh, so the `revision` parameter here is not the revision we get from MW API.. A bit misleading.

gwicke: It is in some cases, but not others.

Pchelolo:
I general this 👍 to me, but indeed let's revert switching to new filter for now. 

Also, we need some deployment plan, don't really want this to interleave with change-propagation rollout

sys/page_revisions.js
gwicke: {{done}}.

gwicke:
@Pchelolo Re deployment: I think the access check stuff needs more time in any case, to prepare page deletion & back-fill script. So, I'd propose to do changeprop summary first.

sys/parsoid.js
```diff
@@ -264,18 +264,44 @@ PSP.generateAndSave = function(hyper, req, format, currentContentRes) {
...
+            }
+
             resp.headers.etag = mwUtil.makeETag(rp.revision, tid);
             return self.saveParsoidResult(hyper, req, format, tid, res)
             .then(function() {
                 var dependencyUpdate = self._dependenciesUpdate(hyper, req);
                 if (mwUtil.isNoCacheRequest(req)) {
                     // Finish background updates before returning
-                    return dependencyUpdate.thenReturn(resp);
+                    return P.all([redirectUpdate, dependencyUpdate])
```
Pchelolo: I'm wondering if we need to update redirects sync before dependencies. The dependency update might use the redirect information (to not access mobile service to generate useless redirect content)

Although it's a minor issue as we will remove dependency updates soon enough

sys/page_revisions.js
```diff
@@ -290,26 +454,26 @@ PRS.prototype.fetchAndStoreMWRevision = function(hyper, req) {
...
+                    attributes: {
+                        title: revision.title,
+                        rev: revision.rev
+                    }
                 }
+            }),
+            // TODO: Before we fill in the restrictions table we need
+            // to store the restriction regardless of the revision change
+
+            self.storeRestrictions(hyper, req, revision))
```
Pchelolo: `return self.storeRestrictions`

Pchelolo:
I've done one more pass over this, and the latest version LGTM

gwicke:
Rebased on master once more. There were a couple of conflicts, but nothing serious.

Pchelolo:
@wikimedia/services I've rebased and reviewed this PR again and I would like to proceed with this. 

Step 1: Remove config-changes prior to merging, deploy to staging, run a dump
Step 2: Analyse how the restrictions are being filled up
Step 3. Deploy prod
Step 4: Backfill restrictions
Step 5: Enable the filter in production and get rid of the old filter.

What do you think? Way too long we've been neglecting this, let's finally make some progress

gwicke:
Thanks for picking this up again, @Pchelolo! The deployment plan also sounds good to me. I'll give @d00rman and @eevans time to have a look as well, but will merge this later this week if there are no objections.

Pchelolo:
This can be merged now

PR #600, opened 2016-04-27
Don't include revisions and tids to a redirect target

Pchelolo:
Bug: when constructing redirect targets we take into account the revision number and tid etc. But the redirect target page doesn't have that revision, so redirected requests with a revision number always result in 404.

cc @wikimedia/services

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -252,10 +252,7 @@ mwUtil.createRelativeTitleRedirect = function(path, req, newReqParams, titlePara
     var backString = Array.apply(null, { length: pathSuffixCount }).map(function() {
         return '../';
     }).join('');
-    var pathPatternAfterTitle = path.substring(path.indexOf('{title}') - 1);
-    return backString
-        + new URI(pathPatternAfterTitle, newReqParams, true).toString().substr(1)
-        + mwUtil.getQueryString(req);
+    return backString + encodeURIComponent(newReqParams.title) + mwUtil.getQueryString(req);
 };
 
 module.exports = mwUtil;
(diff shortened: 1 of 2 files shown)


d00rman:
While this fix should be deployed ASAP, I wonder if we should point the client to *the page at the time the original revision*. What I mean is, a revision ID is like a time stamp and if the client specified one, we should honour it. Now, they'd just be pointed to the latest version. But, consider this edge case:

A points to B for rev1. In rev2, A starts pointing to C. If a client asks for `A/rev1` then I reckon they should be pointed to B, not C.

That said, this is probably a bug that will be hit on at most a dozen pages, while it requires considerable work. Is it worth it?

Pchelolo:
> A points to B for rev1. In rev2, A starts pointing to C. If a client asks for A/rev1 then I reckon they should be pointed to B, not C.

Right now it will still point to B, the redirect target is parsed out of the html for rev1, just we point to the latest revision of B, not the revision of B that A was pointing in the time of rev1.

In general, I think it's impossible to find out the revision number of B which A was pointing to, because B might have been changed many times while A was pointing to it, so there might be a lot of revisions of B that we might select.

d00rman:
@Pchelolo yeah, good point, I had it vice-versa (re B vs C), but still wonder if there's a way to get to the "correct" version of B if we loosen up the definition of "correct" enough :)

Pchelolo:
@d00rman We'd need an index of revision by time for that, I don't think it worths it..

d00rman:
@Pchelolo yeah, don't take it seriously, I'm rambling.

PR #601, opened 2016-04-29
Don't redirect for update requests

Pchelolo:
We generically don't want to redirect for no-cache requests to avoid change prop following those redirects and updating what it actually should not. Also moved redirect tests around to make them all end up in one group. Removed duplicated tests.

cc @wikimedia/services

### DIFF
--- a/test/features/pagecontent/pagecontent.js
+++ b/test/features/pagecontent/pagecontent.js
@@ -310,40 +310,6 @@ describe('item requests', function() {
     //        assert.deepEqual(res.status, 200);
     //    });
     //});
-
-    it('should redirect to commons for missing file pages', function() {
-        return preq.get({
-            uri: server.config.bucketURL + '/html/File:ThinkingMan_Rodin.jpg'
-        })
-        .then(function(res) {
-            assert.deepEqual(res.status, 200);
-            assert.deepEqual(res.headers['content-location'],
-                'https://commons.wikimedia.org/api/rest_v1/page/html/File%3AThinkingMan_Rodin.jpg');
-        });
-    });
-
-    it('should not redirect if file is missing on commons', function() {
-        return preq.get({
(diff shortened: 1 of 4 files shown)


test/features/pagecontent/pagecontent.js
```diff
@@ -310,40 +310,6 @@ describe('item requests', function() {
...
-            uri: server.config.bucketURL + '/html/File:ThinkingMan_Rodin.jpg'
-        })
-        .then(function(res) {
-            assert.deepEqual(res.status, 200);
-            assert.deepEqual(res.headers['content-location'],
-                'https://commons.wikimedia.org/api/rest_v1/page/html/File%3AThinkingMan_Rodin.jpg');
-        });
-    });
-
-    it('should not redirect if file is missing on commons', function() {
```
gwicke: Should we preserve these commons redirect tests?

Pchelolo: They were duplicated in the redirects file. I've copied them but forgot to remove

gwicke: Ah, I see. Thanks for the clarification.

berndsi:
I must be missing something here, but I fail to see how caching should have any impact on redirecting to another page.

Pchelolo:
@berndsi The `cache-control: no-cache` request headers have a very special meaning in RESTBase. It indicates the need to serenader the content stored under that title and used internally for updates. Frontend Varnishes don't actually pass those headers to RESTBase, so this only affects internal update jobs. Obviously, if we want to rerender a redirect page, for example, we don't want to also rerender a redirect target page, so the update requests should generally not follow redirects. It's easier to just make restbase not return the redirects then making sure that we don't follow them in every client.

berndsi:
@Pchelolo Thank you for the explanation that this is only for internal requests. That makes sense now.

PR #604, opened 2016-05-05
Bug fix: Don't propagate the cache-control header on redirect detection

d00rman:
cc @wikimedia/services

### DIFF
--- a/lib/revision_table_access_check_filter.js
+++ b/lib/revision_table_access_check_filter.js
@@ -34,7 +34,13 @@ module.exports = function(hyper, req, next, options, specInfo) {
             }
             return hyper.get({
                 uri: new URI(htmlPath),
-                headers: req.headers
+                /* TODO FIXME TEMP
+                 * This removes the cache-control header when requesting the HTML
+                 * for redirect inspection. However, this is a temporary hack in
+                 * order to avoid re-renders for no-cache requests. We need to
+                 * implement a proper header inspection routine.
+                 */
+                headers: Object.assign({}, req.headers, { 'cache-control': undefined })
             })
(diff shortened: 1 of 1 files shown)


d00rman:
Self-merging so as to deploy right away. This is, of course, a temporary fix, though. We need to filter the header properly.

Note: the 0.10 test is failing only on travis.

lib/revision_table_access_check_filter.js
```diff
@@ -34,7 +34,13 @@ module.exports = function(hyper, req, next, options, specInfo) {
...
             return hyper.get({
                 uri: new URI(htmlPath),
-                headers: req.headers
+                /* TODO FIXME TEMP
+                 * This removes the cache-control header when requesting the HTML
+                 * for redirect inspection. However, this is a temporary hack in
+                 * order to avoid re-renders for no-cache requests. We need to
+                 * implement a proper header inspection routine.
+                 */
+                headers: Object.assign({}, req.headers, { 'cache-control': undefined })
```
gwicke: Actually, is there any reason for forwarding headers here?

d00rman: I was wondering the same thing. Probably the only header we need is the cookie one, in case it's present, but given this was a hot fix, didn't want to poke the devil.

gwicke: There is actually no need to forward the cookie header explicitly. For internal requests it is completely unused, and for external requests in private wikis [it is transparently injected](https://github.com/wikimedia/hyperswitch/blob/87e12a62cf408ba848aa68b247f0f3d7e4d96a7e/lib/filters/http.js#L40-L47).

PR #605, opened 2016-05-06
Don't forward request headers in redirect check

gwicke:
Request headers are not needed for this request. Follow-up from #604.


Pchelolo:
👍

PR #606, opened 2016-05-09
By default, preserve all parameters in relative redirects

gwicke:
Relative redirects are used for different use cases, some of which
require the preservation of all parameters (ex: title normalization), and
others do not. Before eec5180, we defaulted to full preservation, and have
since switched to no preservation. As a result, normalized redirects for
anything but wikitext redirects are currently not fully preserving all
parameters.

This patch makes the path stripping optional with a boolean option. To
generally reduce the number of parameters, an `options` object is introduced.

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -244,10 +244,10 @@ mwUtil.getQueryString = function(req) {
  * @param {string} path, the path pattern from specInfo.
  * @param {hyper.request} req, the request
  * @param {object} options, with possible parameters:
- *    - {object} newReqParams, use these parameters instead of the original
- *    request parameters.
- *    - {string} titleParamName, the name of the title parameter.
- *    - {boolean} dropPathAfterTitle, indicating that the redirect should drop
+ * @param {object} options.newReqParams, use these parameters instead of the
+ * original request parameters.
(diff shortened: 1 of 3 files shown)


lib/mwUtil.js
```diff
@@ -238,9 +238,24 @@ mwUtil.getQueryString = function(req) {
...
+ * @param {hyper.request} req, the request
+ * @param {object} options, with possible parameters:
+ *    - {object} newReqParams, use these parameters instead of the original
+ *    request parameters.
+ *    - {string} titleParamName, the name of the title parameter.
+ *    - {boolean} dropPathAfterTitle, indicating that the redirect should drop
+ *    the path after the title parameter. Typically used for redirects to
+ *    another title, where revision & tid would no longer match.
+ * @return {string} Location header value containing a relative redirect path.
+ */
```
Pchelolo: JSDoc has special syntax for describing parameter properties: http://usejsdoc.org/tags-param.html#parameters-with-properties

Pchelolo:
Could you add a test for these cases?

lib/mwUtil.js
gwicke: Good to know, thanks for the pointer! Changed the syntax to use the dotted-name syntax.

PR #607, opened 2016-05-09
Do not emit an event if it will cause a loop

Pchelolo:
After [an outage](https://phabricator.wikimedia.org/T134537) we've discovered, that when change-prop is introduced, some bugs playing together might cause rerender loops in RESTBase, when change-prop serenaders some content, and it creates the same event that caused a rerender to happen. 

This PR adds some protection against it - change-prop will add the `x-rerender-reason` header (name subject to a bike shed) to a RESTBase request, containing an original event URI (it's enough as RB only emits `resource_change` events). If we are about to emit the same event causing a cycle - log an error and drop the event.

Questions:
1. What to use as an event hash. I think URI is enough, but suggestions are welcome
2. This doesn't prevent multiple-step loops (event1 causes rerender1 causes event2 causes rerender2 causes event1). Do we want to support cases like that? As a generic way of preventing multi-step loops, we've had an idea to add something like an XFF header to the event meta to indicate the whole sequence of events that caused some event to fire. This could be useful for debugging as well as to prevent multi-step loops

PS: need to add this header in change-propagation.

cc @wikimedia/services

### DIFF
--- a/sys/events.js
+++ b/sys/events.js
@@ -86,10 +86,10 @@ EventService.prototype._emit = function(hyper, req) {
             })
             .filter(function(event) { return !!event; });
 
-            // Change-propagation will set up the x-rerender-reason header, indicating
+            // Change-propagation will set up the x-triggered-by header, indicating
             // the event which caused the rerender. In case RESTBase is about to emit
             // the same event, it will cause a rerender loop. So, log an error and skip
-            // an event.
+            // the event.
             var triggeredBy = req.headers && req.headers['x-triggered-by']
                 || hyper._rootReq && hyper._rootReq['x-triggered-by'];
             if (triggeredBy) {
--- a/sys/events.js
+++ b/sys/events.js
@@ -90,11 +90,13 @@ EventService.prototype._emit = function(hyper, req) {
             // the event which caused the rerender. In case RESTBase is about to emit
             // the same event, it will cause a rerender loop. So, log an error and skip
             // an event.
-            var reRenderReason = req.headers && req.headers['x-rerender-reason']
-                || hyper._rootReq && hyper._rootReq['x-rerender-reason'];
-            if (reRenderReason) {
+            var triggeredBy = req.headers && req.headers['x-triggered-by']
+                || hyper._rootReq && hyper._rootReq['x-triggered-by'];
+            if (triggeredBy) {
+                triggeredBy = triggeredBy.replace(/https?:/g, '');
                 events = events.filter(function(event) {
-                    if (reRenderReason.replace(/^https?:/, '') === event.meta.uri.replace(/^https?:/, '')) {
+                    var eventId = event.meta.topic + ':' + event.meta.uri.replace(/^https?:/, '');
+                    if (triggeredBy.indexOf(eventId) !== -1) {
                         hyper.log('error/events/rerender_loop', {
                             message: 'Rerender loop detected',
                             event: event
--- a/sys/events.js
+++ b/sys/events.js
@@ -85,6 +85,26 @@ EventService.prototype._emit = function(hyper, req) {
                 return event;
             })
             .filter(function(event) { return !!event; });
+
+            // Change-propagation will set up the x-rerender-reason header, indicating
+            // the event which caused the rerender. In case RESTBase is about to emit
+            // the same event, it will cause a rerender loop. So, log an error and skip
+            // an event.
+            var reRenderReason = req.headers && req.headers['x-rerender-reason']
+                || hyper._rootReq && hyper._rootReq['x-rerender-reason'];
+            if (reRenderReason) {
+                events = events.filter(function(event) {
+                    if (reRenderReason.replace(/^https?:/, '') === event.meta.uri.replace(/^https?:/, '')) {
+                        hyper.log('error/events/rerender_loop', {
+                            message: 'Rerender loop detected',
+                            event: event
+                        });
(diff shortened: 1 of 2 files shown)


gwicke:
> 1. What to use as an event hash. I think URI is enough, but suggestions are welcome

Perhaps `{topic}:{url}`, as in `resource_change://en.wikipedia.org/v1/page/html/Foobar` ?

> 2. This doesn't prevent multiple-step loops (event1 causes rerender1 causes event2 causes rerender2 causes event1). Do we want to support cases like that? As a generic way of preventing multi-step loops, we've had an idea to add something like an XFF header to the event meta to indicate the whole sequence of events that caused some event to fire. This could be useful for debugging as well as to prevent multi-step loops

Yeah, I think this would be really useful. Even in the absence of loops, it's really useful to see where an event came from.

sys/events.js
```diff
@@ -85,6 +85,26 @@ EventService.prototype._emit = function(hyper, req) {
                 return event;
             })
             .filter(function(event) { return !!event; });
+
+            // Change-propagation will set up the x-rerender-reason header, indicating
+            // the event which caused the rerender. In case RESTBase is about to emit
+            // the same event, it will cause a rerender loop. So, log an error and skip
+            // an event.
+            var reRenderReason = req.headers && req.headers['x-rerender-reason']
+                || hyper._rootReq && hyper._rootReq['x-rerender-reason'];
```
gwicke: [minor, bikeshed, alternative]: `x-triggers` or `x-events`?

sys/events.js
```diff
@@ -85,6 +85,26 @@ EventService.prototype._emit = function(hyper, req) {
...
+            if (reRenderReason) {
+                events = events.filter(function(event) {
+                    if (reRenderReason.replace(/^https?:/, '') === event.meta.uri.replace(/^https?:/, '')) {
```
gwicke: It would be good to include the topic as well. Perhaps pre-process the trigger / reason header into an object, and then match on that?

Also, why do we need to replace `https:/` here? I thought our internal events are all proto-relative, is that not the case?

sys/events.js
d00rman: :+1: for `x-triggered-by`

sys/events.js
```diff
@@ -85,6 +85,26 @@ EventService.prototype._emit = function(hyper, req) {
                 return event;
             })
             .filter(function(event) { return !!event; });
+
+            // Change-propagation will set up the x-rerender-reason header, indicating
+            // the event which caused the rerender. In case RESTBase is about to emit
+            // the same event, it will cause a rerender loop. So, log an error and skip
+            // an event.
+            var reRenderReason = req.headers && req.headers['x-rerender-reason']
```
d00rman: Minor nit: `reRenderReason` looks ... ugh, unnatural to me. `origReason` perhaps?

sys/events.js
d00rman: > It would be good to include the topic as well. Perhaps pre-process the trigger / reason header into an object, and then match on that?

I agree. The combo is what makes the event unique.

> Also, why do we need to replace https:/ here? I thought our internal events are all proto-relative, is that not the case?

I'm reading this as making sure it's proto-relative and as such I think it makes for more solid code.`

Pchelolo: So, as a conclusion, we all agree we need to get the optional XFF-like field in meta in all events. 

As for the regex - basically we're making sure that a loose `s` in https doesn't get us in trouble. Events in the extension are emitted the https, while restbase exists them with http.

sys/events.js
Pchelolo: `x-restbase-triggered-by`?

sys/events.js
Pchelolo: `origEvent`?

sys/events.js
d00rman: Yup for XFF-like header, but IMHO this can be a separate PR.

sys/events.js
d00rman: :+1:

sys/events.js
d00rman: I think we may want to make it independent of RESTBase. We don't want to re-emit an event for the same URI/topic combo regardless of the original source.

sys/events.js
gwicke: triggers?

d00rman: Gotta love bike-shedding :) `trigger`(`s`) LGTM :+1:

Pchelolo:
Updated to match the changes in https://github.com/wikimedia/change-propagation/pull/32

sys/events.js
```diff
@@ -85,6 +85,28 @@ EventService.prototype._emit = function(hyper, req) {
                 return event;
             })
             .filter(function(event) { return !!event; });
+
+            // Change-propagation will set up the x-rerender-reason header, indicating
```
d00rman: x-rerender-reason?

sys/events.js
```diff
@@ -85,6 +85,28 @@ EventService.prototype._emit = function(hyper, req) {
...
+            // the event which caused the rerender. In case RESTBase is about to emit
+            // the same event, it will cause a rerender loop. So, log an error and skip
+            // an event.
```
d00rman: s/an/the/

d00rman:
Minor nits, LGTM otherwise

Pchelolo:
@d00rman updated

PR #608, opened 2016-05-09
Add a sanity check for ETags passed into transform end points

gwicke:
To rule out T134593 being caused by clients passing incorrect ETag headers,
this patch adds a sanity check comparing the passed-in timeuuid with the value
of mw:TimeUuid in the HTML content. If those two don't match, an error is
logged, but the request is not aborted (for now).

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -589,12 +589,21 @@ PSP.transformRevision = function(hyper, req, from, to) {
         if (etag) {
             // Prefer the If-Match header
             tid = etag.tid;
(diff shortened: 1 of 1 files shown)


gwicke:
@wikimedia/services 

@d00rman, it would be great to deploy this tomorrow, so that we can use this to narrow down the cause of https://phabricator.wikimedia.org/T134593.

Pchelolo:
👍 

@d00rman if you don't feel like deploying tomorrow, I can deploy it after our weekly meeting

PR #611, opened 2016-05-11
More robust parsoid change detection

gwicke:
- Always compare re-rendered content to the stored version of the same
  revision, and only do so when the original content response status was 200.
- Move revision metadata fetching into `generateAndSave()``.

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -451,6 +451,12 @@ PSP.getFormat = function(format, hyper, req) {
         if (self.options.response_cache_control) {
             res.headers['cache-control'] = self.options.response_cache_control;
         }
+        if (/^null$/.test(res.headers.etag)) {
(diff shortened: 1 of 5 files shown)


sys/parsoid.js
```diff
@@ -237,52 +236,79 @@ PSP.generateAndSave = function(hyper, req, format, currentContentRes) {
...
+    .then(function(revInfo) {
+        rp.revision = revInfo.rev + '';
+        if (reqRevision !== rp.revision) {
+            // Try to fetch the HTML corresponding to the requested revision,
+            // so that the change detection makes sense.
+            return hyper.get({
+                uri: self.getBucketURI(rp, format, rp.tid)
+            })
+            .then(function(contentRes) {
+                    console.log('res', contentRes.status);
```
d00rman: Superfluous. Also, too much indentation.

PR #613, opened 2016-05-13
Start testing rate limiting on transform end points

gwicke:
Since March, we have a basic rate limit filter in the hyperswitch repository.
This is currently only logging when rates are exceeded, so that we can start
by verifying the basic operation without a risk of aborting requests
unnecessarily.

This patch enables this filter with conservative limits on the un-cached
`transform` end points. By deploying this change (and a puppet change to
configure the rate limiter), we can move forward by testing rate limiting in
staging & production.

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -118,3 +118,10 @@ metrics:
   host: localhost
   port: 8125
   batch: true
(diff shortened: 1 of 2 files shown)


Pchelolo:
👍

PR #614, opened 2016-05-16
Limit the maximum rate of requests per IP

milimetric:
Configure the metrics endpoints to filter out requests that happen more
than X times per second from the same IP to the same endpoint.  The
exact number is not decided yet but I'm tentatively setting it at 100.

Right now, the filter is just logging if the limit is exceeded so we can
get a feel for how often this happens.  We may decide to make it a real
filter as part of this PR or later.

https://phabricator.wikimedia.org/T135240
Bug: T135240

### DIFF
--- a/v1/metrics.yaml
+++ b/v1/metrics.yaml
@@ -92,10 +92,22 @@ paths:
           description: Error
           schema:
             $ref: '#/definitions/problem'
+      x-route-filters:
+        - type: default
+          name: ratelimit_route
+          options:
+            limits:
+              internal: 100
+              external: 100
+            log_only: true
       x-request-handler:
         - get_from_backend:
             request:
               uri: '{{options.host}}/pageviews/per-article/{project}/{access}/{agent}/{article}/{granularity}/{start}/{end}'
+              headers:
+                cache-control: '{{cache-control}}'
+                if-unmodified-since: '{{if-unmodified-since}}'
+                x-client-ip: '{{x-client-ip}}'
       x-monitor: false
 
   /pageviews/aggregate/{project}/{access}/{agent}/{granularity}/{start}/{end}:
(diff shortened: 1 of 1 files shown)


milimetric:
Let us know if this is ok, but we should probably talk tomorrow to see if others agree with doing logOnly at first.

v1/metrics.yaml
```diff
@@ -151,10 +163,22 @@ paths:
...
+              internal: 100
+              external: 100
+            log_only: true
       x-request-handler:
         - get_from_backend:
             request:
               uri: '{{options.host}}/pageviews/aggregate/{project}/{access}/{agent}/{granularity}/{start}/{end}'
+              headers:
+                cache-control: '{{cache-control}}'
+                if-unmodified-since: '{{if-unmodified-since}}'
```
Pchelolo: Out of curiosity, how would AQS use the `cache-control` and `if-unmodified-since` headers?

v1/metrics.yaml
```diff
@@ -92,10 +92,22 @@ paths:
...
+            limits:
+              internal: 100
+              external: 100
+            log_only: true
       x-request-handler:
         - get_from_backend:
             request:
               uri: '{{options.host}}/pageviews/per-article/{project}/{access}/{agent}/{article}/{granularity}/{start}/{end}'
+              headers:
+                cache-control: '{{cache-control}}'
```
gwicke: This is unlikely to do much for external requests, as Varnish strips cache-control headers.

gwicke:
LGTM overall. Some of the forwarded headers (cache-control and if-unmodified-since) probably do nothing for you right now.

v1/metrics.yaml
milimetric: I was just thinking they'd be convenient to have in general, but maybe you're right, maybe it doesn't make sense at our layer.

milimetric:
I agree, I'll get those out, since there's no immediate use I can think of.

v1/metrics.yaml
Pchelolo: In general we use `cache-control` to force a rerender, and `if-unmodified-since` to skip a forced rerender if requests come out of order. Since AQS data is practically immutable, you can't rerender, so unlikely you need them.

PR #620, opened 2016-05-25
Remove MobileApps background updates

d00rman:
We are starting to use Change Prop as the canonical source of updates, so remove them from RESTBase.


d00rman:
cc @wikimedia/services

PR #621, opened 2016-05-25
Remove Varnish purging and release v0.14.0

d00rman:
We will start emitting Varnish purge requests from Change Prop, so remove the hacky cruft code for that from RESTBase.

Note: cannot be deployed before [Gerrit 290748](https://gerrit.wikimedia.org/r/#/c/290748/)

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -6,14 +6,9 @@ default_project: &default_project
   x-modules:
     - path: test/test_module.yaml
       options:
-        events_purge:
-          purge:
-            host: 127.0.0.1
-            port: 4321
-        events_emit:
-          eventlogging_service:
-            uri: http://127.0.0.1:8085/v1/events
-            topic: resource_change
+        events:
+          uri: http://127.0.0.1:8085/v1/events
+          topic: resource_change
     - path: projects/wmf_default.yaml
       options: &default_options
         table:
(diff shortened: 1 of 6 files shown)


d00rman:
cc @wikimedia/services

d00rman:
The prod config change is in [Gerrit 290786](https://gerrit.wikimedia.org/r/#/c/290786/) so this PR should be deployed together with it, even though that's not strictly needed as this PR is compatible with old configs too.

config.test.yaml
```diff
@@ -40,9 +35,8 @@ default_project: &default_project
         mobileapps:
           host: http://appservice.wmflabs.org
         events:
-          purge:
-            host: 127.0.0.1
-            port: 4321
+          uri: http://127.0.0.1:8085/v1/events
+          topic: resource_change
```
Pchelolo: We didn't add the events config here before, because
- it's not strictly needed for testing
- not adding it verified that RB doesn't crash if the config is not there
- it makes tests slower a little bit since we make some additional requests

I don't have a strong opinion whether we should remove these 2 lines or leave them, up to you

sys/events.js
```diff
@@ -1,133 +1,83 @@
...
-            'EventLogging options should provide a uri');
-    }
-
-    if (options && options.eventlogging_service
-        && !options.eventlogging_service.topic) {
-        throw new Error('Incorrect configuration of events module. ' +
-            'EventLogging options should provide a topic');
+    if (options && options.eventlogging_service) {
+        // keep compat with old config
+        options = options.eventlogging_service;
```
Pchelolo: May be add a 'temp/todo/remove' comment?

Pchelolo:
A couple of nits inline (I'm being really picky here, it's not important at all), otherwise 👍

d00rman:
Published as [v0.14.0](https://github.com/wikimedia/restbase/tree/v0.14.0)

PR #622, opened 2016-05-26
Document rate limits for pageview API end points

gwicke:
Add a bullet point clearly documenting the per-entrypoint limits before
enforcing them.


Pchelolo:
👍

PR #624, opened 2016-05-26
Remove the IPA from an extract

Pchelolo:
I've looked through several articles and came up with these regexes, but suggestions on improvement are very welcome.

cc @wikimedia/services 

Bug: https://phabricator.wikimedia.org/T135272

### DIFF
--- a/v1/summary.js
+++ b/v1/summary.js
@@ -15,6 +15,14 @@ module.exports = function(options, templates) {
                     thumb.source = thumb.source.replace(/^http:/, 'https:');
                 }
                 return thumb;
+            },
+            removeIPA: function(text) {
+                if (!text) {
+                    return text;
+                }
+                return text.replace(/\/[^/]+\/;?/, '')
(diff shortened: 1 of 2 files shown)


gwicke:
Doing this regexp-based looks dangerous, especially considering the lack of syntactical guarantees we have on PHP parser output. This is all past-sanitizer, so a mistake in regexp massaging could create XSS issues. It would be more reliable to use template information from Parsoid to reliably identify & strip IPA content on a DOM, ideally in the content service.

I think the plan was to move the summary backend to MCS longer term anyway, so perhaps this would be a good reason to look into that move sooner?

Pchelolo:
@gwicke I completely agree that's dangerous, I've started this PR mostly to begin the conversation about this issue.

berndsi:
> I think the plan was to move the summary backend to MCS longer term anyway

I was not aware of that but that's an interesting thought.

Pchelolo:
Since we've decided not to do it this way, I'm closing this PR

PR #625, opened 2016-05-27
Add Wikidata description to the summary endpoint response

mdholloway:
Many of the app feed cards require a page title, thumbnail, extract, and
Wikidata description.  Adding Wikidata description to the summary endpoint
output will allow us to get the card content in a single request rather
than having to make a separate MediaWiki Action API call.

### DIFF
--- a/v1/summary.js
+++ b/v1/summary.js
@@ -21,6 +21,11 @@ module.exports = function(options, templates) {
                     return revItems[0];
                 }
                 return {};
+            },
+            extractDescription: function(terms) {
+                if (terms && terms.description && terms.description.length) {
+                    return terms.description[0];
+                }
             }
         }
     };
--- a/v1/summary.yaml
+++ b/v1/summary.yaml
@@ -97,13 +97,14 @@ paths:
               method: post
               uri: /{domain}/sys/action/query
               body:
-                prop: 'info|extracts|pageimages|revisions'
+                prop: 'info|extracts|pageimages|revisions|pageterms'
                 exsentences: 5
                 explaintext: true
                 piprop: 'thumbnail'
                 pithumbsize: 320
                 rvprop: 'timestamp'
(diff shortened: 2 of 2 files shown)


Pchelolo:
Although this part looks ok, I think it's too early to include that, because right now we have no way to purge this piece of content. Let's figure out purging before adding it..

mdholloway:
Ah, I see.  Is there a Phab ticket I should follow on issues around purging this?  I was reading through some tickets over the weekend and saw some of the past problems with purging and related problems with VCL encoding/decoding (and found the ticket for this particular issue, which I'd forgotten existed, sorry!) but didn't get a great sense of what the current status is.

Pchelolo:
@mdholloway There's a ticket: https://phabricator.wikimedia.org/T128894

I've written down a very naive way to purge this, but as @d00rman points out, the whole update propagation in wikidata is not perfect now, so we should probably address that first

v1/summary.js
```diff
@@ -15,6 +15,12 @@ module.exports = function(options, templates) {
                     thumb.source = thumb.source.replace(/^http:/, 'https:');
                 }
                 return thumb;
+            },
+
+            extractDescription: function(terms) {
+                if (terms && terms.description) {
```
Pchelolo: Perhaps also check for `terms.description.length`?

Pchelolo:
@mdholloway If you rebase this I will merge/deploy it right after we enable https://github.com/wikimedia/change-propagation/pull/48 (early next week)

v1/summary.js
mdholloway: Done

Pchelolo:
Ok, LGTM. Once we set up purging for this I will merge and deploy (next week)

Pchelolo:
CP is updating summary when wikidata item changes now, so merging.

PR #626, opened 2016-05-31
Start enforcing rate limits

d00rman:
Also, document the rate limits for transform endpoints.


d00rman:
cc @wikimedia/services

Pchelolo:
👍

PR #627, opened 2016-06-01
Do access checks sequentially

Pchelolo:
When we do content request in parallel with an access check, it's obviously good for performance, but it's too risky. API returns very different results for our queries if the access is not granted, and non of our code is ready for that. So, we will very likely hit a ton of 500 errors, may corrupt the storage, or get some completely unpredictable consequences. 

Making all the code in RESTBase behave correctly if access is not granted is a big task, and right now I'm not sure we want to follow that path. Access check is fairly quick, so we're not winning a lot in latency.
If we ever decide we want to speed up private wikis, we can reconsider, but now I don't think it worths the effort.

cc @wikimedia/services

### DIFF
--- a/lib/mediawiki_auth_filter.js
+++ b/lib/mediawiki_auth_filter.js
@@ -65,16 +65,8 @@ module.exports = function(hyper, req, next, options) {
         return next(hyper, req);
     }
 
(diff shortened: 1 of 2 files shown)


d00rman:
:+1: I think this slight delay is acceptable for private wikis.

PR #628, opened 2016-06-01
Mathoid: Add purging and the /math/formula/{hash} endpoint

d00rman:
This commit adds a new endpoint - /media/math/formula/{hash} - which is
useful for retrieving input formulae when users report problems with
specific renders.

More importantly, `resource_change` events are now being emitted for
changed or new content (for both formulae and their renders), which will
be picked up by Change Propagation and purged from Varnish.

Bug: [T136205](https://phabricator.wikimedia.org/T136205)

### DIFF
--- a/sys/mathoid.js
+++ b/sys/mathoid.js
@@ -17,6 +17,58 @@ function MathoidService(options) {
 }
 
 
+MathoidService.prototype._invalidateCache = function(hyper, hash) {
+
+    var routes = [];
+    var uri = '//wikimedia.org/api/rest_v1/media/math/';
+
+    routes.push(uri + 'formula/' + hash);
+
+    FORMATS.forEach(function(fmt) {
+        routes.push(uri + 'render/' + fmt + '/' + hash);
+    });
+
+    return hyper.post({
+        uri: new URI(['wikimedia.org', 'sys', 'events', '']),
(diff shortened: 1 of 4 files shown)


d00rman:

sys/mathoid.js
```diff
@@ -17,37 +17,89 @@ function MathoidService(options) {
...
+    var origHash = {};
     var checkRes;
 
     // start by calculating the hash
     return hyper.post({
         uri: new URI([rp.domain, 'sys', 'post_data', 'mathoid.input', 'hash']),
         body: { q: req.body.q, type: rp.type }
     }).then(function(res) {
-        hash = origHash = res.body;
+        hash.id = origHash.id = res.body;
```
Pchelolo: What's the point in making it an object? I don't see where any other properties are added or any difference from what happened.

d00rman: Since we actually change the reference to the hash string object, better to make sure we still have the good reference by making it an object whose ref will not change.

Pchelolo: Emm.. I still do not understand how's the new thing better then the old one

d00rman: Nm, I was thinking in C++ but coding in JS... Will remove.

PR #630, opened 2016-06-13
Skip sending resource_change events if so wanted

d00rman:
RESTBase automatically sends resource_change events when the HTML of a page has changed, which allows us to track dependant updates. However, there are circumstances a user might not want them to be sent. There are two cases where that is true:
- third-party support: these events make sense only when third-party users users have the complete EventBus system installed (HTTP Proxy + Kafka + ChangeProp), which make the events virtually useless for small installs; and
- staging: staging hosts are part of production, which means they emit events to the production Kafka Cluster, and is undesirable when, e.g., dumps are running there.

This patch adds the `skip_updates` configuration stanza that bypasses the Parsoid and Events modules' sending of resource_change events when set to `true` (the default value is `false`).

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -102,10 +102,6 @@ PSP._dependenciesUpdate = function(hyper, req) {
     var rp = req.params;
     var publicURI = '//' + rp.domain + '/api/rest_v1/page/html/' + encodeURIComponent(rp.title);
 
-    if (this.options.skip_updates) {
-        return P.resolve();
-    }
-
     return hyper.post({
         uri: new URI([rp.domain, 'sys', 'events', '']),
         body: [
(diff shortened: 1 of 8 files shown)


Pchelolo:
May be we should just completely shut off all the events in the `events` module in this case? Sending 'resource_change' for `summary` doesn't make any sense for small install either.

d00rman:
> May be we should just completely shut off all the events in the events module in this case? Sending 'resource_change' for summary doesn't make any sense for small install either.

Yup, I realised that while I was writing the commit message. Will update.

d00rman:
@Pchelolo updated

sys/parsoid.js
```diff
@@ -102,6 +102,10 @@ PSP._dependenciesUpdate = function(hyper, req) {
     var rp = req.params;
     var publicURI = '//' + rp.domain + '/api/rest_v1/page/html/' + encodeURIComponent(rp.title);
 
+    if (this.options.skip_updates) {
+        return P.resolve();
+    }
```
Pchelolo: Why do we need to duplicate the check in `parsoid.js` - we can just make it a no-op in `events.js` and don't change anything here - will simplify the config.

d00rman:
Removed the check from the Parsoid mod. Self-merging to unblock dumps in staging.

PR #631, opened 2016-06-14
data-* attributes are only valid in the pagebundle path

arlolra:
Please verify the assumptions in https://gerrit.wikimedia.org/r/#/c/293351/ before merging.

/cc @wikimedia/parsing @wikimedia/services

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -621,9 +621,7 @@ PSP.transformRevision = function(hyper, req, from, to) {
             original: original
         };
         if (from === 'sections') {
-            body2.html = {
-                body: replaceSections(original, parseSections(req))
-            };
+            body2.html = replaceSections(original, parseSections(req));
             from = 'html';
         } else {
             body2[from] = req.body[from];
@@ -707,7 +705,10 @@ PSP.callParsoidTransform = function callParsoidTransform(hyper, req, from, to) {
         // Retrieve pagebundle whenever we want HTML
         parsoidTo = 'pagebundle';
     }
-
+    var parsoidFrom = from;
+    if (from === 'html' && req.body.original && req.body['data-parsoid']) {
+        parsoidFrom = 'pagebundle';
+    }
     var parsoidExtras = [];
     if (rp.title) {
         parsoidExtras.push(rp.title);
@@ -723,7 +724,7 @@ PSP.callParsoidTransform = function callParsoidTransform(hyper, req, from, to) {
 
     var parsoidReq = {
         uri: this.parsoidHost + '/' + rp.domain + '/v3/transform/'
-            + from + '/to/' + parsoidTo + parsoidExtraPath,
+            + parsoidFrom + '/to/' + parsoidTo + parsoidExtraPath,
         headers: {
             'content-type': 'application/json',
             'user-agent': req['user-agent'],
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ Dockerfile
 
 # sqlite file (used in testing)
 test.db.sqlite3
+test.db.sqlite3-journal
 sqlite.db
 db.sqlite3
 


gwicke:
@arlolra, is the pagebundle end point prepared to perform wt2html conversions?

arlolra:
The failures are 403s from mobile frontend, which seem unrelated to this patch.

sys/parsoid.js
```diff
@@ -707,6 +705,10 @@ PSP.callParsoidTransform = function callParsoidTransform(hyper, req, from, to) {
         // Retrieve pagebundle whenever we want HTML
         parsoidTo = 'pagebundle';
     }
+    var parsoidFrom = from;
+    if (from === 'html' && (req.body.original || req.body.previous)) {
```
arlolra: hmm, what this really want to be is if data-* attrs are present

arlolra:
@gwicke Updated the patch here, and in gerrit.  Let me know what you think.

sys/parsoid.js
```diff
@@ -621,9 +621,7 @@ PSP.transformRevision = function(hyper, req, from, to) {
             original: original
         };
         if (from === 'sections') {
-            body2.html = {
-                body: replaceSections(original, parseSections(req))
-            };
+            body2.html = replaceSections(original, parseSections(req));
```
gwicke: This would remove the per-element header / body framing, which I don't think we ever discussed. Unless there are strong reasons for dropping the generality & extensibility of the mime-like structure, I would recommend to not change the pagebundle format in incompatible ways.

gwicke:
The move to the pagebundle end point looks sane to me. I left a comment about the change in the pagebundle structure, which I'm not sure whether it was accidental or intended.

sys/parsoid.js
arlolra: I did this for consistency with L627

`req.body[from]` is not framed.

gwicke: Ah, you are right. `original` is framed as a pagebundle, but we historically included html as a non-pagebundle in the request. We need to change the section case to be consistent with the non-section case, and your change does just that. Evidently we don't have a test covering this. It would be really awesome if you added one as part of this PR. It should go into [transform.js](https://github.com/wikimedia/restbase/blob/master/test/features/parsoid/transform.js).

With the move towards separate data-mw and possibly other metadata in the future, I do wonder if the current ad-hoc format still makes sense, or whether we should convert the `modified` pagebundle to a proper pagebundle as well. This would make `original` and `modified` page structurally identical, and would afford both the same framing / header functionality. It would also be somewhat ironic if the pagebundle2pagebundle end point did not actually accept pagebundles.

In any case, we can address that question in a follow up. This change looks okay to me.

arlolra: Ok, I'll add a test to this PR, and leave the larger refactor to you all.

gwicke: Thank you, @arlolra!

arlolra: @gwicke Sorry, can you clarify what you want me to test? Section transforms already have tests ("sections2wt, replace", "sections2wt, append", etc.). They continue to pass because Parsoid accepts the both framed and unframed html. The change was really about consistency, not correctness, per se. You don't seem to have any other unit tests for ParsoidService, and it isn't even exported to the point where that can be attempted. What this patch changes is the value being sent by RESTBase to the Parsoid HTTP API, so I imagine what you want is to mock the Parsoid HTTP API and confirm the expectations about what's being posted. Please confirm.

gwicke: @arlolra, the bit I was missing was this:

> They continue to pass because Parsoid accepts the both framed and unframed html.

This makes sense now.

As a general feedback, it might be helpful to provide a bit more context in the commit message, so that the intention & background of the changes is clearer during review. You were (again) very patient with me, but I think it could save us both some time if things were clearer up-front.

arlolra: @subbuss would sympathize with your sentiment.  I'm working on getting better at that.

arlolra:
Thanks! Please let me know when it gets deployed.

d00rman:
@arlolra it was deployed yesterday EU morning

PR #632, opened 2016-06-22
Use the canonical file namespace whe redirecting to commons

Pchelolo:
Commons doesn't understand localised File namespace names, so replace it with a canonical name.

Bug: https://phabricator.wikimedia.org/T138081
cc @wikimedia/services

### DIFF
--- a/test/features/pagecontent/redirects.js
+++ b/test/features/pagecontent/redirects.js
@@ -68,6 +68,17 @@ describe('Redirects', function() {
         });
     });
 
+    it('should redirect to commons for missing file pages, dewiki', function() {
+        return preq.get({
+            uri: server.config.hostPort + '/de.wikipedia.org/v1/page/html/Datei:Name.jpg'
+        })
+        .then(function(res) {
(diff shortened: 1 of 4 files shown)


test/features/pagecontent/redirects.js
```diff
@@ -68,6 +68,17 @@ describe('Redirects', function() {
         });
     });
 
+    it('should redirect to commons for missing file pages, dewiki', function() {
+        return preq.get({
+            uri: server.config.hostPort + '/de.wikipedia.org/v1/page/html/Datei:Name.jpg'
+        })
+        .then(function(res) {
+            assert.deepEqual(res.status, 200);
```
d00rman: Why is the return code 200 if it's a redirect and we haven't specified `redirect=false` in the request?

Pchelolo: `preq` is following the redirect, so it's a result for the redirected page. The target of the redirect is checked using the `content-location` header.

PR #633, opened 2016-06-24
Block User:Pentjuuu!.!/sandbox updates

d00rman:
The page is huge and has a ton of null edits which time-out on updates,
so skip it.


Pchelolo:
👍

PR #634, opened 2016-07-05
Promote mobile-sections* endpoints' stability to unstable

d00rman:
The `mobile-sections*` endpoints are now being used by the Android
native mobile app, so promote their stability from experimental to
unstable since they will be kept around, but their output format may
still slightly change. Once the output format becomes final, the
stability should be promoted to stable.


d00rman:

berndsi:
+1

mdholloway:
👍

PR #635, opened 2016-07-06
Switch to Node v4.4.6

d00rman:
We now have the new LTS version in the repo so it's time to switch to
it.

Bug: [T138561](https://phabricator.wikimedia.org/T138561)


d00rman:
cc @wikimedia/services

PR #636, opened 2016-07-08
Add feed and random public endpoints; release v0.14.2

d00rman:
`/feed/featured` is the aggregate endpoint used by the next version of the native Adroid app. The endpoint forwards the request to the Mobile Content Service and returns its response. There is no storage for it at the moment; it should be added in a second iteration (once the endpoint's output format stabilises) and it should be modelled after mobile-sections since it returns the aggregate response for multiple smaller endpoints (they will be added once we start storing this endpoint's response). Varnish is instructed to cache the response for one minute.

`/page/random/{format}` gets a radnom page title from MCS' `page/random/title` endpoint and returns a redirect to the client based on the chosen format. Valid formats are title, html, summary, related, mobile-sections and mobile-sections lead. Since different random titles should be returned as often as possible, the redirect is cached for only 2 seconds.

This PR also makes a couple of improvements to the specification monitoring test utilities. Namely:

- if the example request does not specify the domain, assume it's the domain we are currently testing for
- don't set a field value to '' if it is false-y, but only if it's null or undefined
- if the example's response contains an array with only one element, assume we want to check each item of the returned array against it

Bug: [T136960](https://phabricator.wikimedia.org/T136960)

### DIFF
--- a/v1/feed.yaml
+++ b/v1/feed.yaml
@@ -81,7 +81,8 @@ paths:
                     pageid: /.+/
                     normalizedtitle: /.+/
               random:
-                title: /.+/
+                items:
+                  - title: /.+/
               news:
                 - story: /.+/
                   links:
--- a/v1/feed.yaml
+++ b/v1/feed.yaml
@@ -1,0 +1,105 @@
+# Feed endpoints powered by the MCS
+
+swagger: 2.0
+paths:
+  /featured/{yyyy}/{mm}/{dd}:
+    get:
+      tags:
+        - Feed
+      summary: Aggregated featured content
+      description: |
+        Provides the aggregated feed content for the given date. The endpoint returns
+        the featured article of the day, most read articles for the previous day, news
+        content, a random article, and the featured image and video of the day.
+
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      produces:
+        - application/json
+      parameters:
+        - name: yyyy
+          in: path
+          description: 'Year the aggregated content is requested for'
+          type: string
+          required: true
+        - name: mm
+          in: path
+          description: 'Month the aggregated content is requested for'
+          type: string
+          required: true
+        - name: dd
+          in: path
+          description: 'Day of the month the aggregated content is requested for'
+          type: string
+          required: true
+      responses:
+        '200':
+          description: JSON containing all of the featured content
+          schema:
+            type: object
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - from_mobileapps:
+            request:
+              method: get
+              uri: '{{options.host}}/{domain}/v1/feed/featured/{yyyy}/{mm}/{dd}'
+            return: 
+              status: '{{from_mobileapps.status}}'
+              headers: '{{ merge({"cache-control": options.feed_cache_control},
+                            from_mobileapps.headers) }}'
+              body: '{{from_mobileapps.body}}'
+      x-monitor: true
+      x-amples:
+        - title: Retrieve aggregated feed content for April 29, 2016
+          request:
+            params:
+              domain: en.wikipedia.org
+              yyyy: '2016'
+              mm: '04'
+              dd: '29'
+          response:
+            status: 200
+            headers:
+              content-type: /application\/json/
+            body:
+              tfa:
+                title: /.+/
+                description: /.+/
+                extract: /.+/
+                thumbnail:
+                  source: /.+/
+                  width: /.+/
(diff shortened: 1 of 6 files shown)


d00rman:

d00rman:
Note that this should not be deployed yet. When requesting a random title from MCS, it does not return the MW-style-encoded title, but its normalised version, which would result in more redirects than necessary.

MCS currently returns `{"title": "Barack Obama"}` when it really should return `{"title": "Barack_Obama"}`

v1/feed.yaml
```diff
@@ -0,0 +1,105 @@
...
+        - name: mm
+          in: path
+          description: 'Month the aggregated content is requested for'
+          type: string
+          required: true
+        - name: dd
+          in: path
+          description: 'Day of the month the aggregated content is requested for'
+          type: string
+          required: true
```
Pchelolo: We might wanna set `min` and `max` to these at least. Even is MCS does proper validation, we could save some work by doing some basic validation here.

v1/feed.yaml
```diff
@@ -0,0 +1,105 @@
...
+            type: object
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - from_mobileapps:
+            request:
+              method: get
+              uri: '{{options.host}}/{domain}/v1/feed/featured/{yyyy}/{mm}/{dd}'
```
Pchelolo: So, after all we've decided that MCS will return the whole content, not RB will assemble it from pieces?

v1/feed.yaml
d00rman: Not possible. `min` and `max` apply to integers only, and these must be strings because of zero-padding.

v1/feed.yaml
d00rman: There are two ways to go about this: have an SLA between MCS and RB for the format from which RB will assemble the content, or have MCS return the whole aggregated result and then store each piece separately (as this endpoint aggregates multiple ones), just as we do for `mobile-sections`. This version doesn't store anything, it merely exposes publicly the the endpoint and its format.

We are still free to go either way independently of this PR. I'm not sure which of the two options is better, to be honest, because of the `random` component in the response...

Pchelolo: Sounds reasonable, just wanted to be clear on it.

Pchelolo:
LGTM

v1/feed.yaml
berndsi: I think the aggregated endpoint will be the most requested feed endpoint. 'In the news' is likely updated most frequently (after random, of course).
So, for most of the endpoints pulling results from the aggregated endpoint makes sense to me.

The `random` component is an exception. Storage for this one doesn't seem necessary. IMO that'll probably just cause unnecessary churn for cassandra without much benefit. The 1-2 second caching should take most of the brunt anyways. If there are some clients getting different random results than others that's actually a good thing, and we wouldn't mind that at all.

mdholloway:
Let's merge this before enabling the public endpoints: https://gerrit.wikimedia.org/r/#/c/298024/

d00rman:
> Note that this should not be deployed yet. When requesting a random title from MCS, it does not return the MW-style-encoded title, but its normalised version, which would result in more redirects than necessary.
> 
> MCS currently returns {"title": "Barack Obama"} when it really should return {"title": "Barack_Obama"}

This has been fix in https://gerrit.wikimedia.org/r/#/c/298408/ . Once https://gerrit.wikimedia.org/r/#/c/298243/ lands, we should be good to go.

d00rman:
PR updated with the new format for `page/random/title` returned by MCS (not the one actually returned by RB to the external client)

v1/random.yaml
```diff
@@ -37,7 +37,7 @@ paths:
               status: 303
               headers:
                 cache_control: '{{options.random_cache_control}}'
-                location: '../../{request.params.format}/{title_from_mobileapps.body.title}'
+                location: '../../{request.params.format}/{title_from_mobileapps.body.items[0].title}'
```
berndsi: If you'd like I can revert the change in MCS back to what it was. Now that the app is back to using summary it doesn't need to be in items[0] anymore. Up to you. Either way is fine with me.

d00rman: Even though I too find in superfluous, let's keep it as-is now for sanity's sake. Also, it needs to stay as-is if we want to get it out tomorrow.

berndsi: Sounds good to me. Thanks!

d00rman:
Published as [v0.14.2](https://github.com/wikimedia/restbase/tree/v0.14.2)

PR #638, opened 2016-07-15
Add /page to random redirect title path

mdholloway:
Currently the random title and redirect URLs are missing the /page segment
from their path and therefore they aren't working.

### DIFF
--- a/v1/random.yaml
+++ b/v1/random.yaml
@@ -37,7 +37,7 @@ paths:
               status: 303
               headers:
                 cache_control: '{{options.random_cache_control}}'
-                location: '../../{request.params.format}/{title_from_mobileapps.body.items[0].title}'
+                location: '../{request.params.format}/{title_from_mobileapps.body.items[0].title}'
       x-monitor: true
       x-amples:
         - title: Random title redirect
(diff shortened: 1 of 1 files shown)


v1/random.yaml
```diff
@@ -37,7 +37,7 @@ paths:
               status: 303
               headers:
                 cache_control: '{{options.random_cache_control}}'
-                location: '../../{request.params.format}/{title_from_mobileapps.body.items[0].title}'
+                location: '../../page/{request.params.format}/{title_from_mobileapps.body.items[0].title}'
```
d00rman: Since `/page/` is effectively the root here, would be better to just remove one `../` instead.

v1/random.yaml
```diff
@@ -48,5 +48,5 @@ paths:
             status: 303
             headers:
               cache-control: /.+/
-              location: /^..\/..\/title\/[^\/]+$/
+              location: /^..\/..\/page\/title\/[^\/]+$/
```
d00rman: Idem.

d00rman:
I can't believe this slipped through the cracks. Oddly enough, it worked just fine when I was testing it for some reason (but now it's not working).

Thnx @mdholloway !

PR #639, opened 2016-07-21
Refetch siteInfo in case it failed

Pchelolo:
We're fetching the `siteInfo` of the first request to the domain and then caching the resolved promise to reuse it later. If the fetch was failed for any reason - delete the cache entry so that the next request would retry fetching the information.

cc @wikimedia/services


d00rman:
:+1: 

cc @berndsi @mdholloway Could you please also implement this improvement in MCS?

Pchelolo:
There's also similar code in Change-Prop that needs update, but we're not using it yet

PR #644, opened 2016-08-01
Summary: Fix timestamp extraction logic

Pchelolo:
Under certain circumstances the revision timestamp logic might result in a 500 error. If the page was quickly deleted, but that info wasn't propagated to RESTBase yet, the access check filter allows access to the summary. The MW API call results in an object with a `missing` property, and absent `revisions` property, so template execution fails with 500.

cc @wikimedia/services

### DIFF
--- a/v1/summary.js
+++ b/v1/summary.js
@@ -16,8 +16,11 @@ module.exports = function(options, templates) {
                 }
                 return thumb;
             },
-            getTimestamp: function(items) {
-                return items && items[0] && items[0].timestamp
+            getRevision: function(revItems) {
+                if (Array.isArray(revItems) && revItems.length) {
+                    return revItems[0];
+                }
+                return {};
             }
         }
     };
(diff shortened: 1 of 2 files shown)


v1/summary.js
```diff
@@ -15,6 +15,9 @@ module.exports = function(options, templates) {
                     thumb.source = thumb.source.replace(/^http:/, 'https:');
                 }
                 return thumb;
+            },
+            getTimestamp: function(items) {
+                return items && items[0] && items[0].timestamp
```
d00rman: jscs complains about a missing semicolon at the end of the statement.

v1/summary.js
```diff
@@ -15,6 +15,9 @@ module.exports = function(options, templates) {
                     thumb.source = thumb.source.replace(/^http:/, 'https:');
                 }
                 return thumb;
+            },
+            getTimestamp: function(items) {
```
d00rman: I wonder if it would be better to expose a more generalised function like:

```js
getRevisions: function(revItems) {
  if (Array.isArray(revItems) && revItems.length) {
    return revItems;
  }
  return [{}];
}
```

Pchelolo: Done, but I've added the 'getRevision' function returning an object directly instead, looks nicer imho, and we don't need generality here since it's limited to summary endpoint config

PR #645, opened 2016-08-02
Added postpublish hook that would tag the repo in git

Pchelolo:
We forget to tag the release in git sometimes, so let's do it generically in the prepublish script.

It assumes that your upstream is called `upstream` which, I assume, is true for all of us..

cc @wikimedia/services

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "REST storage and service dispatcher",
   "main": "lib/server.js",
   "scripts": {
-    "postpublish": "git tag -a \"v${npm_package_version}\" -m \"${npm_package_name}@${npm_package_version} release\" && git push upstream --tags",
+    "postpublish": "git tag -a \"v${npm_package_version}\" -m \"${npm_package_name}@${npm_package_version} release\" && git push upstream \"v${npm_package_version}\"",
     "start": "service-runner",
     "test": "sh test/utils/run_tests.sh test",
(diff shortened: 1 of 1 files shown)


Pchelolo:
Hm, apparently `prepublish` is run before `npm install` too and it's by design.. So we can't really do  this, closing the PR.

https://github.com/npm/npm/issues/3059

gwicke:
@Pchelolo: A great idea! Do you think moving this to the `postpublish` hook would avoid the install issue?

d00rman:
:+1: for this idea and for `postpublish` too. I have a local script that does:

```sh
npm publish && git tag -a ${version} -m "${name} ${version} release" && git push ${remote} ${version}
```

so it's essentially a post-publish action.

Pchelolo:
Oh, how did I not think about `postpublish`. Lemme update the PR

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "REST storage and service dispatcher",
   "main": "lib/server.js",
   "scripts": {
+    "postpublish": "git tag -a \"v${npm_package_version}\" -m \"${npm_package_name}@${npm_package_version} release\" && git push upstream --tags",
```
d00rman: Perhaps it'd be safer to push just the tag we've just created, so as to avoid ridiculous situations.

PR #648, opened 2016-08-09
Re-use local variable

edg2s:


Pchelolo:
LGTM, waiting for travis and merging

d00rman:
I must be missing the point, but how is this better? The `sections` variable is still there and is being used in the other code paths ... Besides, that var just holds a reference to `req.body.sections`.

berndsi:
I don't think the point of this patch was to get rid of that variable, just to reuse it instead of traversing req.body.sections again.

PR #651, opened 2016-08-18
Blacklist some articles

Pchelolo:
That page is crazy: https://ceb.wikipedia.org/wiki/Gumagamit:Lsjbot/Anomalier-PRIVAT
And it's edited by a bot every minute or two, 15G of HTML in storage..


eevans:
LGTM (waiting on Travis because reasons).

PR #654, opened 2016-08-22
Re-export hyperswitch as the main package entry point; release v0.15.0

d00rman:
We usually use RESTBase as a service, not a dependency for other projects. However, when it is used as dependency it can't be loaded as `lib/server.js` doesn't exist any more. Hence, re-export hyperswitch as the module's main entry point.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,5 @@
+'use strict';
+
+
+module.exports = require('hyperswitch');
+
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

gwicke:
:+1: 

This should un-break https://github.com/wikimedia/mediawiki-node-services for latest RB.

d00rman:
> This should un-break https://github.com/wikimedia/mediawiki-node-services for latest RB.

Not yet, not yet, but I'm working on it :)

PR #656, opened 2016-08-27
Added explanations to tags

Pchelolo:
Swagger supports providing descriptions for tags, I think we should add something. 

However, I need help choosing the working here, these descriptions are really more like a first attempt to do something...

Depends on https://github.com/wikimedia/hyperswitch/pull/58

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -13,7 +13,7 @@ info:
     url: https://www.apache.org/licenses/LICENSE-2.0
 tags:
   - name: Page content
-    description: various representation of page content
+    description: page content in different formats
 paths:
   /:
     get:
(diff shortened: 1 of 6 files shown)


v1/mobileapps.yaml
```diff
@@ -1,6 +1,7 @@
 # MobileApps page-manipulation service
-
-swagger: 2.0
+tags:
+  - name: Mobile
+    description: page content optimized for mobile apps
```
d00rman: Better: *Mobile-friendly page content*

v1/content.yaml
```diff
@@ -11,6 +11,9 @@ info:
   license:
     name: Apache licence, v2
     url: https://www.apache.org/licenses/LICENSE-2.0
+tags:
+  - name: Page content
+    description: various representation of page content
```
d00rman: *page content in different formats* perhaps?

PR #662, opened 2016-09-06
Feeds: Split the aggregated feed request and cache it

d00rman:
Instead of requesting the whole response of the aggregated feed endpoint
in MCS, it is much faster to request its individual pieces and then
assemble them in RESTBase. The process happens in parallel, so different
MCS workers get to work on a part of the response.

Moreover, in order to decrease the amount of requests to MCS, store the
response using the `ttl` revision retention policy. Currently, we begin
with the conservative ttl of 10 minutes, but it is likely we can
increase it in the future. The random component needs to change more
frequently than that, so whenever we have a cache hit, we re-request a
random title from MCS and incorporate it in the new response. We also
tell Varnish to cache the response for 30 seconds. This means that the
number of requests will be:
- 1 request for the random title every 30 seconds
- 5 requests for the various response components every 10 minutes

Note: depends on the deployment of [Gerrit 308589](https://gerrit.wikimedia.org/r/#/c/308589/)

Bug: [T143912](https://phabricator.wikimedia.org/T143912)

### DIFF
--- a/v1/feed.js
+++ b/v1/feed.js
@@ -1,0 +1,167 @@
+'use strict';
+
+
+var P = require('bluebird');
+var HyperSwitch = require('hyperswitch');
+var uuid = require('cassandra-uuid').TimeUuid;
+
+var URI = HyperSwitch.URI;
+var HTTPError = HyperSwitch.HTTPError;
+
+var spec = HyperSwitch.utils.loadSpec(__dirname + '/feed.yaml');
+
+
+var DEFAULT_TTL = 3600;
+
+var FEED_URIS = {
+    tfa: { uri: ['v1', 'page', 'featured'], date: true },
+    mostread: { uri: ['v1', 'page', 'most-read'], date: true },
+    random: { uri: ['v1', 'page', 'random', 'title'], date: false },
+    image: { uri: ['v1', 'media', 'image', 'featured'], date: true },
+    news: { uri: ['v1', 'page', 'news'], date: false }
+};
+
+
+function Feed(options) {
+
+    this.options = options;
+
+}
+
+
+Feed.prototype._makeFeedRequests = function(parts, hyper, rp, dateArr) {
+
+    var self = this;
+    var props = {};
+
+    parts.forEach(function(part) {
+        var def = FEED_URIS[part];
+        var uriArray = [self.options.host, rp.domain].concat(def.uri);
+        if (def.date) {
+            Array.prototype.push.apply(uriArray, dateArr);
+        }
+        props[part] = hyper.get({
+            uri: uriArray.join('/'),
+            query: { aggregated: true }
+        });
+    });
+
+    return P.props(props);
+
+};
+
+
+Feed.prototype.aggregated = function(hyper, req) {
+
+    var self = this;
+    var rp = req.params;
+    var date;
+    var dateArr;
+
+    // key the records on the date in the format YYYY-MM-DD
+    try {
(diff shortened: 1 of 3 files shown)


d00rman:

v1/feed.js
```diff
@@ -0,0 +1,167 @@
...
+    var date;
+    var dateArr;
+
+    // key the records on the date in the format YYYY-MM-DD
+    try {
+        date = new Date(Date.UTC(rp.yyyy, rp.mm - 1, rp.dd));
+        date = date.toISOString().split('T').shift();
+        dateArr = date.split('-');
+    } catch (err) {
+        return P.reject(new HTTPError({
```
Pchelolo: You can just throw here, no need to wrap in a promise, it's already inside P.try

v1/feed.js
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    // check if we have a record in Cassandra already
+    return hyper.get({
+        uri: new URI([rp.domain, 'sys', 'key_value', 'feed.aggregated', date])
+    }).then(function(res) {
+        // we've got a cache hit, so we just need to request
+        // the random component and return the bundle
+        if (!res.body) {
+            return P.reject(new HTTPError({
```
Pchelolo: Em, is that possible? Also, no need to explicitly wrap in P.reject

v1/feed.js
```diff
@@ -0,0 +1,167 @@
...
+                    // TODO: need a way to dynamically derive this
+                    'content-type': 'application/json; charset=utf-8; ' +
+                        'profile="https://www.mediawiki.org/wiki/Specs/aggregated-feed/0.5.0"'
+                },
+                body: {}
+            };
+            // populate its body
+            Object.keys(result).forEach(function(key) {
+                finalResult.body[key] = result[key].body;
+            });
```
Pchelolo: `Object.assign(finalResult.body, result.body)`.. Or, actually, why do we need to copy it?

v1/feed.js
```diff
@@ -0,0 +1,167 @@
...
+            // assemble the final response to be returned
+            var finalResult = {
+                status: 200,
+                headers: {
+                    'cache-control': self.options.feed_cache_control,
+                    // mimic MCS' ETag value
+                    etag: dateArr.join('') + '/' + uuid.now().toString(),
+                    // TODO: need a way to dynamically derive this
+                    'content-type': 'application/json; charset=utf-8; ' +
+                        'profile="https://www.mediawiki.org/wiki/Specs/aggregated-feed/0.5.0"'
```
Pchelolo: Would be nice to actually create this page, and to set the same content-type in the public spec

v1/feed.js
d00rman: Because we receive 5 HTTP responses, and we need to combine their bodies. Note that we are not copying here, just assigning refs.

Pchelolo: Oh, indeed, stupid me..

v1/feed.js
d00rman: It's more explicit this way, IMHO

Pchelolo: We don't do it anywhere else though..

v1/feed.js
d00rman: Happened to me during testing. I agree that this can happen only in an extreme circumstance, but it doesn't hurt to have it here.

d00rman:
@Pchelolo removed the `P.reject()` constructs and replaced them with `throw`.

Pchelolo:
LGTM

PR #663, opened 2016-09-06
Switch labs tests to https

Pchelolo:
Labs switched to enforcing https, so use https in tests

cc @wikimedia/services


gwicke:
Travis still has trouble connecting to archive.apache.org, so the tests won't be fixed immediately.

PR #664, opened 2016-09-06
Allow providing both sectionId and html

Pchelolo:
Actually providing both id and HTML might be useful, support that.

cc @wikimedia/services


gwicke:
:+1:, waiting for Travis.

Pchelolo:
@gwicke no need to wait for travis, I know it will fail because of https://github.com/wikimedia/restbase/pull/663

PR #665, opened 2016-09-06
Add a security response header filter

gwicke:
Move REST API specific security response header handling from hyperswitch to
restbase.

### DIFF
--- a/lib/security_response_header_filter.js
+++ b/lib/security_response_header_filter.js
@@ -1,0 +1,108 @@
+"use strict";
+const P = require('bluebird');
+
+/**
+ * Filter adding security-relevant response headers like CSP.
+ */
+
+function isHtmlSvgContent(res) {
+    return res.headers
+    && /^(?:text\/html|image\/svg)/i.test(res.headers['content-type']);
+}
+
+var cspHeaderVariants = [
+    'content-security-policy',
+    // For IE 10 & 11
+    'x-content-security-policy',
+    // For Chrome <= v25 (<<1% traffic; todo: revisit support)
+    'x-webkit-csp',
+];
+
+/**
+ * Replaces subdomain with a wildcard
+ *
+ * @param domain a full domain name (e.g. en.wikipedia.org)
+ * @returns {string} wildcard version (e.g. *.wikipedia.org)
+ */
+function wildcardSubdomain(domain) {
+    if ((domain.match(/\./g) || []).length >= 2) {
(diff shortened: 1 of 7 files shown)


gwicke:
See https://github.com/wikimedia/hyperswitch/pull/60 for the corresponding hyperswitch change.

test/features/parsoid/ondemand/ondemand.js
```diff
@@ -222,6 +222,7 @@ describe('on-demand generation of html and data-parsoid', function() {
             uri: pageUrl + '/html/' + title
         })
         .then(function (res) {
+            console.log(res);
```
Pchelolo: oh

lib/security_response_header_filter.js
```diff
@@ -0,0 +1,108 @@
+"use strict";
+const P = require('bluebird');
```
Pchelolo: We don't use this yet in RB

Pchelolo:
LGTM, though needs a bump in `hyperswitch` dependency version.

lib/security_response_header_filter.js
gwicke: {{done}}

Pchelolo:
And travis is still unhappy

gwicke:
Corresponding puppet patch at https://gerrit.wikimedia.org/r/#/c/309486/.

Pchelolo:
@gwicke travis yet again..

d00rman:
FWIW, I think it would have been better to leave the capacity in HyperSwitch, but remove only the RB-specific parts and transfer them here.

gwicke:
@d00rman: The policies on security headers are all quite specific to the API / use case, which is why I moved this to RESTBase. For example, the ServiceWorker proxy needs to allow inline scripts, while the content API is keen on disallowing those.

d00rman:
> The policies on security headers are all quite specific to the API / use case, which is why I moved this to RESTBase. For example, the ServiceWorker proxy needs to allow inline scripts, while the content API is keen on disallowing those.

Sure, sure. I meant that IMHO HyperSwitch should have kept an `addCSP` or similar method, when clients would just then give it the values, and for any value not given HyperSwitch would have a sane default.

gwicke:
@d00rman: If the filter proves useful for other use cases (which seems not that unlikely), then we could move it to a separate package, or even start a collection of bundled filters in hyperswitch itself. Right now, all API-specific filters live in RB, which is why I moved this one over.

PR #666, opened 2016-09-07
Travis: Test with Node 6

d00rman:


d00rman:
cc @wikimedia/services

PR #668, opened 2016-09-08
Populate feed with summaries from RESTBase

Pchelolo:
I'll add more tests in a followup PR.

Bug: https://phabricator.wikimedia.org/T143933

### DIFF
--- a/v1/feed.js
+++ b/v1/feed.js
@@ -161,9 +161,9 @@ Feed.prototype.aggregated = function(hyper, req) {
             feed.random.items.forEach(function(article) { requestTitle(article.title); });
         }
         if (feed.news) {
-            feed.news.forEach(function(story) {
-                if (story.links) {
-                    story.links.forEach(function(article) {
+            feed.news.forEach(function(newsItem) {
+                if (newsItem.links) {
+                    newsItem.links.forEach(function(article) {
                         requestTitle(article);
                     });
                 }
@@ -197,8 +197,8 @@ Feed.prototype.aggregated = function(hyper, req) {
             }
 
             if (feed.news) {
-                feed.news.forEach(function(story) {
-                    story.links = assignAllSummaries(story.links);
+                feed.news.forEach(function(newsItem) {
+                    newsItem.links = assignAllSummaries(newsItem.links);
                 });
             }
             return res;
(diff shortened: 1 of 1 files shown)


v1/feed.js
```diff
@@ -133,8 +133,77 @@ Feed.prototype.aggregated = function(hyper, req) {
...
+                }
+            });
+        }
+
+        return P.props(summaries)
+        .then(function(summaries) {
+            function assignSummary(article) {
+                if (summaries[article.title]) {
+                    // MCS expects the title to be a DB Key
+                    delete summaries[article.title].title;
```
gwicke: If we are serializing this to JSON, then setting to `undefined` ought to work as well:
```javascript
> JSON.stringify({a: undefined})
'{}'
```

v1/feed.js
```diff
@@ -133,8 +133,77 @@ Feed.prototype.aggregated = function(hyper, req) {
...
+                    return Object.assign(article, summaries[article.title]);
+                }
+            }
 
+            function assignAllSummaries(articles) {
+                if (articles) {
+                    return articles.map(assignSummary)
+                    .filter(function(article) { return !!article; });
```
gwicke: The double negation shouldn't be needed.

v1/feed.js
Pchelolo: `Object.assign({a: 'Test'}, {a: undefined }) === {a: undefined}` so if we set it to undefined we would delete the title from the result.

gwicke: Ah, good point. 

Too bad about the inconsistency between title properties in summary & feed end points. But, probably too late since the app already depends on it.

gwicke:
LGTM overall.

v1/feed.js
```diff
@@ -133,8 +133,77 @@ Feed.prototype.aggregated = function(hyper, req) {
...
+                feed.mostread.articles = assignAllSummaries(feed.mostread.articles);
+            }
+
+            if (feed.random) {
+                feed.random.items = assignAllSummaries(feed.random.items);
+            }
+
+            if (feed.news) {
+                feed.news.forEach(function(story) {
+                    story.links = assignAllSummaries(story.links);
```
d00rman: This is confusing because a `feed.news` item contains the `story` **and** `links` properties. Can the iterator here be replaced by a different name? `newsItem` perhaps?

d00rman:
LGTM, one minor nit though.

v1/feed.js
Pchelolo: @d00rman done

PR #669, opened 2016-09-08
Added spec for the feed endpoint

Pchelolo:
We've been talking about adding a schema to the feed endpoint so that some tests could validate agains it - so here's the first iteration of the schema. Depends on https://github.com/wikimedia/restbase/pull/668, after that's merged I will create a test that would validate responses agains this schema and make some minor improvements to the schema itself.

Bug: https://phabricator.wikimedia.org/T141943

### DIFF
--- a/test/features/schema_tests.js
+++ b/test/features/schema_tests.js
@@ -1,0 +1,55 @@
+"use strict";
+
+// mocha defines to avoid JSHint breakage
+/* global describe, it, before, beforeEach, after, afterEach */
+
+var assert = require('../utils/assert.js');
+var server = require('../utils/server.js');
+var preq   = require('preq');
+var Ajv = require('ajv');
+
+describe('Responces should conform to the provided JSON schema of the responce', function() {
+
+    var spec;
+    var ajv;
+
+    before(function() {
+        return server.start()
+        .then(() => {
+            return preq.get({ uri: server.config.baseURL + '/?spec' });
+        })
+        .then((res) => {
+            ajv = new Ajv();
+            Object.keys(res.body.definitions).forEach(function(defName) {
+                ajv.addSchema(res.body.definitions[defName], '#/definitions/' + defName);
+            });
+        });
+    });
+
+    it('/feed/featured should conform schema', () => {
+        return preq.get({
+            uri: server.config.baseURL + '/feed/featured/2016/09/08'
+        })
+        .then(function(res) {
+            if (!ajv.validate('#/definitions/feed', res.body)) {
+                throw new assert.AssertionError({
+                    message: ajv.errorsText()
+                });
+            }
+        });
+    });
+
+    it('/page/summary/{title} should conform schema', () => {
+        return preq.get({
+            uri: server.config.baseURL + '/page/summary/Tank'
+        })
+        .then(function(res) {
+            if (!ajv.validate('#/definitions/summary', res.body)) {
+                throw new assert.AssertionError({
+                    message: ajv.errorsText()
+                });
+            }
+        });
+    });
+
+});
--- a/test/features/schema_tests.js
+++ b/test/features/schema_tests.js
@@ -32,6 +32,18 @@ describe('Responses should conform to the provided JSON schema of the responce',
         });
     });
 
+    it('/feed/featured should conform schema, ruwiki', function() {
+        return preq.get({ uri: server.config.hostPort + '/ru.wikipedia.org/v1/feed/featured/2016/09/08' })
+        .then(function(res) {
+            if (!ajv.validate('#/definitions/feed', res.body)) {
+                throw new assert.AssertionError({
+                    message: ajv.errorsText()
+                });
+            }
+        });
+    });
+
+
     it('/page/summary/{title} should conform schema', function() {
         return preq.get({ uri: server.config.baseURL + '/page/summary/Tank' })
         .then(function(res) {
--- a/test/features/schema_tests.js
+++ b/test/features/schema_tests.js
@@ -9,27 +9,20 @@ var preq   = require('preq');
 var Ajv = require('ajv');
 
 describe('Responces should conform to the provided JSON schema of the responce', function() {
-
-    var spec;
-    var ajv;
+    var ajv = new Ajv({});
 
     before(function() {
         return server.start()
-        .then(() => {
-            return preq.get({ uri: server.config.baseURL + '/?spec' });
-        })
-        .then((res) => {
-            ajv = new Ajv();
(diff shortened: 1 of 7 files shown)


v1/feed.yaml
```diff
@@ -97,3 +97,141 @@ paths:
...
+      normalizedtitle:
+        type: string
+        description: Article title as it should be presented to the user
+      views:
+        type: integer
+        description: Number of views on the requested day
+      rank:
+        type: integer
+        description: Position in the list of most viewed articles
+      fromencoded:
```
Pchelolo: Q for MCS - what's that?

v1/feed.yaml
```diff
@@ -97,3 +97,182 @@ paths:
...
+    required:
+      - title
+
+  mostread_article:
+    type: object
+    properties:
+      title:
+        type: string
+        description: Article title in a form of DB key
+      normalizedtitle:
```
gwicke: Is this already in the current output / expected by the app?

Pchelolo: This is in the current output, but I don't know how the app uses these

test/features/schema_tests.js
```diff
@@ -0,0 +1,45 @@
...
+
+// mocha defines to avoid JSHint breakage
+/* global describe, it, before, beforeEach, after, afterEach */
+
+var assert = require('../utils/assert.js');
+var server = require('../utils/server.js');
+var preq   = require('preq');
+var Ajv = require('ajv');
+
+describe('Responces should conform to the provided JSON schema of the responce', function() {
```
berndsi: typo in response

v1/feed.yaml
```diff
@@ -97,3 +97,182 @@ paths:
                   width: /.+/
                   height: /.+/
 
+definitions:
+  # A https://tools.ietf.org/html/draft-nottingham-http-problem
+  problem:
```
berndsi: So, the 'problem' definition is for error responses?

v1/feed.yaml
berndsi: We use the normalized title in the UI when we display the page title. Users don't want to look at titles with underscores, etc.

v1/feed.yaml
```diff
@@ -97,3 +97,182 @@ paths:
...
+        description: Lead image for the article
+        $ref: '#/definitions/thumbnail'
+      description:
+        type: string
+        description: Wikidata description of the article
+      extract:
+        type: string
+        description: First several sentences of an article in plain text
+    required:
+      - title
```
berndsi: Actually, normalizedtitle is also required.

v1/feed.yaml
berndsi: Good question. I had not seen this until now. I cannot find the string `fromencoded` in our code base. I guess this is coming in from the PageView API. cc: @mdholloway

v1/feed.yaml
```diff
@@ -97,3 +97,182 @@ paths:
...
+    type: array
+    items:
+      $ref: '#/definitions/news_item'
+
+  random:
+    type: object
+    properties:
+      items:
+        type: array
+        items:
```
berndsi: The two levels of `items` looks suspicious.

v1/feed.yaml
Pchelolo: Yep, it's the same everywhere in the API

v1/feed.yaml
Pchelolo: The first `items` is the property name, second `items` is a keyword in JSON schema.

Right now random portion is returned in the form of
```
random:
  items:
    - title: 'ads'
      
```

why is that? Why don't we change it to
```
random:
  title: 'sdcasdc'
```
?

v1/feed.yaml
berndsi: @mdholloway is not around today. Maybe @milimetric recognizes the `fromencoded` property and what it represents? I'm leaning towards leaving this out from the spec for now unless one of them says that it's important.

v1/feed.yaml
berndsi: I was leaning towards skipping the items array as well. It was just introduced to match the RB format of the /page/title/{title} endpoint response. The one point that puts me back on the fence about this would be that maybe in the future we might want to return multiple random articles.

Pchelolo: @berndsi good point about multiple random articles...

Pchelolo:
Updated according to comments, rebased.

v1/feed.yaml
gwicke: We could just use a different key for different kinds of random titles. My personal preference would be to keep things flat, if that's possible without breaking existing clients.

The main reason why RB is using `items` in listing responses is that this sets up a top-level object, which allows adding continuations & other attributes later.

Pchelolo: @gwicke using items here would be in line with other RB endpoints and would allow adding more random articles without paint. 

If you use different keys, how would you add 5 random articles instead of 1? `random`, `random2`, `random3` ... ? Kinda ugly. Since it doesn't really matter let's leave it as is?

v1/feed.yaml
Pchelolo: done

v1/feed.yaml
Pchelolo: @gwicke Ok, converted `random` portion to an array

gwicke:
LGTM.

v1/feed.js
```diff
@@ -193,7 +197,7 @@ Feed.prototype.aggregated = function(hyper, req) {
             }
 
             if (feed.random) {
-                feed.random.items = assignAllSummaries(feed.random.items);
+                feed.random = assignAllSummaries(feed.random.items);
```
d00rman: Uh, the Android app expects to have `feed.random.items`. Why this change in the interface?

gwicke: I asked the same question, but it seems that the Android app is not using this property yet, and IOS will be the first to start using this soon. Would be good to have confirmation from @berndsi, though.

berndsi: That is correct. The Android app is not using this property yet, and IOS will be the first to start using this.

v1/feed.yaml
```diff
@@ -97,3 +96,178 @@ paths:
...
+        description: Data about the featured article of the day
+        $ref: '#/definitions/article_summary'
+      mostread:
+        description: Data about most viewed articles
+        $ref: '#/definitions/mostread'
+      news:
+        description: Data about in the news section
+        $ref: '#/definitions/news'
+      random:
+        description: A single random article
```
d00rman: Aren't there multiple ones?

Pchelolo: @d00rman Right now there's only one

PR #670, opened 2016-09-14
Use eslint

Pchelolo:
ESLint is the new generation of static code analysis tools coming to replace JSCS and JSHint. 

I imagine the eslint config would become an endless source of bike shedding, because what I wanna achieve is to settle on the common eslint config for all node services we maintain :)

The current config was generated by analysing our current RESTBase code-style

cc @wikimedia/services

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,0 +1,234 @@
+env:
+  node: true
+extends: 'eslint:recommended'
+rules:
+  accessor-pairs: error
+  array-bracket-spacing: 'off'
+  array-callback-return: error
+  arrow-body-style: error
+  arrow-parens: error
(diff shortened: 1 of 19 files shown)


d00rman:
The es config seems to be missing here.

Pchelolo:
Ups, forgot to `git add` it..

Pchelolo:
And here we are, on the crossroads. eslint itself is node 4+

d00rman:
Yup, @Pchelolo . But I don't think that's bad per se, we want to drop node 0.x anyway. Let's wait a bit with this PR, though.

Pchelolo:
Same functionality implemented in #681. Closing

PR #673, opened 2016-09-19
Undelete wasn't working with new restrictions table.

Pchelolo:
No need to refetch stuff that we've just saved.

cc @wikimedia/services

### DIFF
--- a/sys/page_revisions.js
+++ b/sys/page_revisions.js
@@ -477,6 +477,11 @@ PRS.prototype.fetchAndStoreMWRevision = function(hyper, req) {
             }
         })
         .catch({ status: 404 }, function() {
+            if (!revision.page_deleted) {
+                // Clear up page_deleted
+                revision.page_deleted = null;
+            }
+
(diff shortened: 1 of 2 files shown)


sys/page_revisions.js
```diff
@@ -679,10 +681,18 @@ PRS.prototype.getRevision = function(hyper, req) {
             }
         });
     }
+
+    function fetchFromMW() {
```
gwicke: The naming suggests that this is only fetching, but it is really saving as well. Perhaps something like fetchStoreAndCheckMWRevision could be better? Or, actually, is there a call to fetchAndStoreMWRevision that doesn't want the checking part?

Pchelolo: Yep, when we're using it from `getTitleRevison` we don't wanna check because it's already checking afterwards.

gwicke:
LGTM. Waiting for travis before merging.

d00rman:
:100:

PR #674, opened 2016-09-19
Emit transclusion-related events to the separate topic

Pchelolo:
Bug: https://phabricator.wikimedia.org/T145804
cc @wikimedia/services

### DIFF
--- a/sys/events.js
+++ b/sys/events.js
@@ -24,6 +24,17 @@ EventService.prototype.emitEvent = function(hyper, req) {
         return { status: 200 };
     }
     return P.try(function() {
+        // Change-propagation will set up the x-triggered-by header, indicating
+        // the event which caused the rerender. In case RESTBase is about to emit
+        // the same event, it will cause a rerender loop. So, log an error and skip
+        // the event.
+        var triggeredBy = req.headers && req.headers['x-triggered-by']
+            || hyper._rootReq && hyper._rootReq['x-triggered-by'];
+        var topic = self.options.topic;
+        if (triggeredBy && /transcludes/.test(triggeredBy)) {
+            topic = self.options.transcludes_topic;
+        }
+
         var events = req.body.map(function(event) {
             if (!event.meta || !event.meta.uri || !/^\/\//.test(event.meta.uri)) {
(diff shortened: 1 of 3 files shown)


Pchelolo:
Needed to be deployed together with https://gerrit.wikimedia.org/r/311594

Pchelolo:
Should be deployed after https://github.com/wikimedia/change-propagation/pull/107

sys/events.js
```diff
@@ -24,6 +24,17 @@ EventService.prototype.emitEvent = function(hyper, req) {
...
     return P.try(function() {
+        // Change-propagation will set up the x-triggered-by header, indicating
+        // the event which caused the rerender. In case RESTBase is about to emit
+        // the same event, it will cause a rerender loop. So, log an error and skip
+        // the event.
+        var triggeredBy = req.headers && req.headers['x-triggered-by']
+            || hyper._rootReq && hyper._rootReq['x-triggered-by'];
+        var topic = self.options.topic;
+        if (triggeredBy && /transcludes/.test(triggeredBy)) {
+            topic = self.options.transcludes_topic;
```
d00rman: Since this is a new configuration option, we should either provide a default or fail on start-up if it hasn't been provided. Perhaps the former is more appropriate here.

Pchelolo: We've used to simply ignore the call if `topic` isn't configured, so do the same for `transcludes_topic` now

d00rman: But wouldn't the natural fallback here be to emit an event for `self.options.topic` if it specified? That way we keep the backwards compatibility with the previous *modus operandi*. The user should not be required to set up a new topic if they are not facing the same challenges the WMF is.

Pchelolo: Good idea, done.

Pchelolo:
@d00rman We don't need the new topic, it already exist.

d00rman:
> We don't need the new topic, it already exist: https://github.com/wikimedia/mediawiki-event-schemas/blob/master/config/eventbus-topics.yaml#L113-L114

Duh, indeed, forgot that it's already in use. LGTM, :+1: but will wait with merging the PR until it's safe to deploy it.

PR #675, opened 2016-09-20
Remove no-longer-needed fields from feed spec

mdholloway:
See MCS patch: https://gerrit.wikimedia.org/r/#/c/311733/

Note: Spec-related unit tests will fail if the MCS patch is merged before
this is.

Bug: T146041

### DIFF
--- a/v1/feed.yaml
+++ b/v1/feed.yaml
@@ -79,9 +79,7 @@ paths:
               news:
                 - story: /.+/
                   links:
-                    - pageid: /.+/
-                      ns: /.+/
-                      normalizedtitle: /.+/
+                    - normalizedtitle: /.+/
                       title: /.+/
               image:
                 title: /.+/


v1/feed.yaml
```diff
@@ -71,18 +71,13 @@ paths:
...
-                    normalizedtitle: /.+/
               random:
                 items:
                   - title: /.+/
               news:
                 - story: /.+/
                   links:
-                    - pageid: /.+/
-                      ns: /.+/
-                      normalizedtitle: /.+/
```
Pchelolo: I thought that `normalizedtitle` was used in the app? No?

mdholloway: Yeah, we do.  Good point, probably better to keep checking for it.  I'm going to do a little more testing and maybe withdraw this PR if we can get back with the spec as-is.

mdholloway: All we really need to do is remove the pageid and ns fields from the expected response for news links.  These aren't used in the app.

mdholloway:
The MCS patch has been merged and deployed.  

I think you'll now see the test "Monitoring routes, en.wikipedia.org domain Retrieve aggregated feed content for April 29, 2016" fail with " AssertionError: Body field pageid not found in response!" until this fix to the spec is made.

PR #676, opened 2016-09-20
Don't propagate 302 and location header

Pchelolo:
Fixes the current weird problem with TypeError

Not quite sure why didn't I catch it in labs, perhaps because I didn't use any special characters in titles. This is like the 15th time I'm making this mental note to use special characters in titles in testing...

cc @wikimedia/services

### DIFF
--- a/lib/revision_table_access_check_filter.js
+++ b/lib/revision_table_access_check_filter.js
@@ -81,6 +81,13 @@ module.exports = function(hyper, req, next, options, specInfo) {
                 }
             });
         }
-        return next(hyper, req);
+        return next(hyper, req)
+        .then(function(res) {
+            if (req.query.redirect === false || mwUtil.isNoCacheRequest(req)) {
+                res.status = 200;
+                delete res.headers.location;
+            }
+            return res;
+        });
     });
 };
--- a/lib/revision_table_access_check_filter.js
+++ b/lib/revision_table_access_check_filter.js
@@ -83,7 +83,8 @@ module.exports = function(hyper, req, next, options, specInfo) {
         }
         return next(hyper, req)
         .then(function(res) {
-            if (req.query.redirect === false || mwUtil.isNoCacheRequest(req)) {
+            if (req.status === 302
+                    && (req.query.redirect === false || mwUtil.isNoCacheRequest(req))) {
                 res.status = 200;
                 delete res.headers.location;
             }
--- a/lib/revision_table_access_check_filter.js
+++ b/lib/revision_table_access_check_filter.js
@@ -83,7 +83,7 @@ module.exports = function(hyper, req, next, options, specInfo) {
         }
         return next(hyper, req)
         .then(function(res) {
-            if (req.status === 302
+            if (res.status === 302
                     && (req.query.redirect === false || mwUtil.isNoCacheRequest(req))) {
                 res.status = 200;
                 delete res.headers.location;
--- a/lib/access_check_filter.js
+++ b/lib/access_check_filter.js
@@ -28,8 +28,7 @@ module.exports = function(hyper, req, next, options, specInfo) {
 
             // Use redirect target from restrictions table or content response.
             var redirectTarget = restriction && restriction.body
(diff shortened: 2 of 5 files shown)


lib/revision_table_access_check_filter.js
```diff
@@ -81,6 +81,13 @@ module.exports = function(hyper, req, next, options, specInfo) {
                 }
             });
         }
-        return next(hyper, req);
+        return next(hyper, req)
+        .then(function(res) {
+            if (req.query.redirect === false || mwUtil.isNoCacheRequest(req)) {
+                res.status = 200;
+                delete res.headers.location;
```
d00rman: Why removing the location?

Pchelolo: @d00rman because we don't want it - `?redirect=false`

Pchelolo: Also this code is temporary, we would switch to the `access_check_filter` later

d00rman: Right, but even with `redirect=false` the SLA is to return the location header. And actually Varnish counts on that, it just changes the return code based on the `redirect` flag, so this represents a breaking change. It should be harmless leaving it in now that the location is URI-encoded, shouldn't it?

lib/revision_table_access_check_filter.js
```diff
@@ -81,6 +81,13 @@ module.exports = function(hyper, req, next, options, specInfo) {
                 }
             });
         }
-        return next(hyper, req);
+        return next(hyper, req)
+        .then(function(res) {
+            if (req.query.redirect === false || mwUtil.isNoCacheRequest(req)) {
```
gwicke: Perhaps also test for res.headers.location?

lib/revision_table_access_check_filter.js
```diff
@@ -81,6 +81,13 @@ module.exports = function(hyper, req, next, options, specInfo) {
...
+                res.status = 200;
```
gwicke: It might be slightly safer to only overwrite this if the status was 301 or 302.

lib/revision_table_access_check_filter.js
Pchelolo: @d00rman No, varnish unsets `redirect=false` part, so we never see it for requests going through varnish: https://github.com/wikimedia/operations-puppet/blob/c53f1476ab6cddad6150ce85f1441646ba92444b/modules/varnish/templates/text-frontend.inc.vcl.erb#L110-L134

d00rman: Oh right. I still think it's in our best interest not to change such small details as we have a lot of moving parts. Do you see a compelling reason for removing it (given the usage of `encodeURIComponent` on it) ?

gwicke: @d00rman: Consistency with the Varnish response is a compelling reason, imho. We want a request directly to RB to work the same way as it does through Varnish. This matters for third party users or Vagrant.

d00rman: @gwicke right, and I would take your argument as a +1 for leaving the location header in the response, but somehow I have a feeling you think we'd better remove it :)

gwicke: [Varnish removes it](https://github.com/wikimedia/operations-puppet/blob/c53f1476ab6cddad6150ce85f1441646ba92444b/modules/varnish/templates/text-frontend.inc.vcl.erb#L232-L235), and so should we.

d00rman: Good catch @gwicke ! My arg is mute then :)

lib/revision_table_access_check_filter.js
d00rman: +1 for checking the status.

lib/revision_table_access_check_filter.js
Pchelolo: Right. Actually only for 302, 301 still has to be redirected.

gwicke:
LGTM, waiting for travis.

PR #681, opened 2016-09-30
Syntax to es6

Pchelolo:
Also added ESLint right away. Tests pass, performance unchanged, all seems correct.

cc @wikimedia/services


d00rman:
@Pchelolo travis build fails for both node 4 and 6

Pchelolo:
@d00rman hm.... Actually with the newest dependencies master fails too! Investigating

Pchelolo:
That's the reason: https://github.com/wikimedia/hyperswitch/pull/65

PR #683, opened 2016-10-03
RESTBase should set the 'triggered_by' field on the resource-change

Pchelolo:
Not sure why wasn't it the case, but it would be useful to propagate the `triggered_by` field further.

cc @wikimedia/services

### DIFF
--- a/test/features/events/events.js
+++ b/test/features/events/events.js
@@ -21,7 +21,7 @@ describe('Change event emitting', function() {
         });
     });
 
-    function createEventLogging(done, topic, uri) {
+    function createEventLogging(done, topic, uri, trigger) {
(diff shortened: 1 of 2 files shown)


test/features/events/events.js
```diff
@@ -21,7 +21,7 @@ describe('Change event emitting', function() {
         });
     });
 
-    function createEventLogging(done, topic, uri) {
+    function createEventLogging(done, topic, uri, trigger) {
```
gwicke: Do you think it would make sense to introduce an "options" or "eventdata" object instead, to avoid the parameter list from growing too much when we add more optional bits of information in the future?

Pchelolo: Done

PR #686, opened 2016-10-06
Feed: Increase Varnish maxage to 5 mins and RB TTL to 2h

d00rman:
Bug: [T147463](https://phabricator.wikimedia.org/T147463)


d00rman:

PR #687, opened 2016-10-06
Feed: Increase the table schema version

d00rman:
Since we have removed the random portion, `rb-mod-table` complains that the table schemas differ, so bump the version so that it knows it has been done so intentionally.


d00rman:
cc @wikimedia/services

PR #689, opened 2016-10-14
Store historic feed content in a separate bucket

Pchelolo:
The goal is to have historic content for the news section (and for other feed peices). The solution is to have a separate bucket for historic content. So if the date is 'today, or future' - work as before - have separate bucket with TTL. If it's yesterday - take content from a historic bucket without TTL.

Open questions:
 - Right now the latest state of the current bucket at a particular date would be forever persisted as a history - is that OK? What if some clever vandal changes the featured title at 23.59.59 UTC?
 - It's possible to get some parts of featured content for the older days. Should we attempt to do it? How long ago can we fetch the featured content? Month? Year?

Bug: https://phabricator.wikimedia.org/T139481

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -411,7 +411,7 @@ mwUtil.hydrateResponse = (response, fetch) => {
         }
 
         if (Array.isArray(node)) {
-            for (let i = 0; i < node.length; i++) {
+            for (let i = node.length - 1; i >= 0; i--) {
                 _traverse(node[i], () => node.splice(i, 1));
             }
         } else if (typeof node === 'object') {
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -411,6 +411,10 @@ mwUtil.hydrateResponse = (response, fetch) => {
         }
 
         if (Array.isArray(node)) {
+            // If the item is not available we need to delete it from the array
+            // using the callback with splice. The callbacks are executed in the same
+            // order as they're created here, so reverse the iteration order to
+            // make removal of multiple elements work correctly.
             for (let i = node.length - 1; i >= 0; i--) {
                 _traverse(node[i], () => node.splice(i, 1));
             }
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -390,11 +390,10 @@ mwUtil.isSelfRedirect = (req, location) => {
  * Traverses through a request body and replaces the given keys with
  * the content fetched using the 'fetch' callback
  *
- * @param responce
+ * @param response
  * @param fetch
- * @param mergeKey
  */
-mwUtil.populate = (responce, fetch, mergeKey) => {
+mwUtil.hydrateResponse = (response, fetch) => {
     const requests = {};
     const setters = [];
 
@@ -407,11 +406,7 @@ mwUtil.populate = (responce, fetch, mergeKey) => {
                 } else if (removeFromParent) {
                     removeFromParent();
                 }
-                // TODO: We would normally delete it anyway,
-                // this check is here until we remove a hack with populating $merge internally
-                if (mergeKey.indexOf('$') !== -1) {
-                    delete node[mergeKey];
-                }
+                delete node.$merge;
             });
         }
 
@@ -420,20 +415,18 @@ mwUtil.populate = (responce, fetch, mergeKey) => {
                 _traverse(node[i], () => node.splice(i, 1));
             }
         } else if (typeof node === 'object') {
-            if (Array.isArray(node[mergeKey])) {
-                node[mergeKey].forEach(requestResource);
-            } else if (node[mergeKey]) {
-                requestResource(node[mergeKey]);
+            if (Array.isArray(node.$merge)) {
+                node.$merge.forEach(requestResource);
             } else {
                 Object.keys(node).forEach((key) => _traverse(node[key], () => delete node[key]));
             }
         }
     }
-    _traverse(responce.body);
+    _traverse(response.body);
 
     return P.props(requests)
     .then((content) => setters.forEach((setter) => setter(content)))
-    .thenReturn(responce);
+    .thenReturn(response);
 };
 
 module.exports = mwUtil;
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -386,4 +386,54 @@ mwUtil.isSelfRedirect = (req, location) => {
     return true;
 };
 
+/**
+ * Traverses through a request body and replaces the given keys with
+ * the content fetched using the 'fetch' callback
+ *
+ * @param responce
+ * @param fetch
+ * @param mergeKey
+ */
+mwUtil.populate = (responce, fetch, mergeKey) => {
+    const requests = {};
+    const setters = [];
+
+    function _traverse(node, removeFromParent) {
+        function requestResource(resource) {
+            requests[resource] = requests[resource] || fetch(resource);
+            setters.push((content) => {
+                if (content[resource]) {
+                    Object.assign(node, content[resource]);
+                } else if (removeFromParent) {
+                    removeFromParent();
+                }
+                // TODO: We would normally delete it anyway,
+                // this check is here until we remove a hack with populating $merge internally
+                if (mergeKey.indexOf('$') !== -1) {
+                    delete node[mergeKey];
+                }
+            });
+        }
+
+        if (Array.isArray(node)) {
+            for (let i = 0; i < node.length; i++) {
+                _traverse(node[i], () => node.splice(i, 1));
+            }
+        } else if (typeof node === 'object') {
+            if (Array.isArray(node[mergeKey])) {
+                node[mergeKey].forEach(requestResource);
+            } else if (node[mergeKey]) {
+                requestResource(node[mergeKey]);
+            } else {
+                Object.keys(node).forEach((key) => _traverse(node[key], () => delete node[key]));
+            }
+        }
+    }
+    _traverse(responce.body);
+
+    return P.props(requests)
+    .then((content) => setters.forEach((setter) => setter(content)))
+    .thenReturn(responce);
+};
+
 module.exports = mwUtil;
--- a/test/features/mwutils.js
+++ b/test/features/mwutils.js
@@ -20,6 +20,9 @@ describe('Utils.hydrateResponse', () => {
                         $merge: ['you_shall_not_pass']
                     },
                     {
+                        $merge: ['you_shall_not_pass']
+                    },
+                    {
                         prop: 'this_will_be_overwritten',
                         $merge: ['prop_contained_here']
                     }
--- a/test/features/mwutils.js
+++ b/test/features/mwutils.js
@@ -1,0 +1,60 @@
+'use strict';
+
+/**
+ * Unit tests for util methods
+ */
+
+const P = require('bluebird');
+const mwUtil = require('../../lib/mwUtil');
+const assert = require('../utils/assert');
+
+describe('Utils.populate', () => {
+    it('Should support $merge', () => {
+        const responce = {
+            body: {
+                non_existent: {
+                    $merge: ['you_shall_not_pass']
+                },
+                array: [
+                    {
+                        $merge: ['you_shall_not_pass']
+                    },
+                    {
+                        prop: 'this_will_be_overwritten',
+                        $merge: ['prop_contained_here']
+                    }
+                ],
+                object: {
+                    some_other_prop: 'hello',
+                    $merge: ['prop_contained_here']
+                }
+            }
+        };
+
+        return mwUtil.populate(responce, (uri) => {
+            switch (uri) {
+                case 'you_shall_not_pass':
+                    return P.resolve(undefined);
+                case 'prop_contained_here':
+                    return P.resolve({
+                        prop: 'prop_value'
+                    });
+                default:
+                    return P.reject(new Error('What?'));
+            }
+        }, '$merge')
+        .then((responce) => {
+            assert.deepEqual(responce, {
+                body: {
+                    array: [{
+                        prop: 'prop_value'
+                    }],
+                    object: {
+                        some_other_prop: 'hello',
+                        prop: 'prop_value'
+                    }
+                }
+            });
+        });
+    });
+});
--- a/test/features/mwutils.js
+++ b/test/features/mwutils.js
@@ -8,7 +8,7 @@ const P = require('bluebird');
 const mwUtil = require('../../lib/mwUtil');
 const assert = require('../utils/assert');
 
-describe('Utils.populate', () => {
+describe('Utils.hydrateResponse', () => {
     it('Should support $merge', () => {
         const responce = {
             body: {
@@ -31,7 +31,7 @@ describe('Utils.populate', () => {
             }
         };
 
-        return mwUtil.populate(responce, (uri) => {
+        return mwUtil.hydrateResponse(responce, (uri) => {
             switch (uri) {
                 case 'you_shall_not_pass':
                     return P.resolve(undefined);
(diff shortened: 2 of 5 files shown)


berndsi:
I like it. 

For your first question could we have  something like a cronjob that would update the historic records for the previous day at 0.00.00 UTC or slightly later? For that one case we would have to make an exception to allow updating the historic bucket -- something like an override or force flag. Just some idea. If there are simpler ways of achieving a good strategy to handle the vandalism case then that would be better of course.

We would like to go as far back as 1/1/2016. If a historic entry is missing and MCS is able to calculate it then RB should fill that data into the historic record. If there's already a historic record then we wouldn't need to recalculate other parts of the aggregated feed again, similarly to the news portion. This is as long as we employ the strategy in this PR and as long as question 1 is fully addressed and the cutoff cases are well-tested. Again, having some sort of override flag would be good to have here again if we come to a point where something went terribly wrong or we feel we really need to update the response schema by adding new properties.

Pchelolo:
>> For your first question could we have something like a cronjob that would update the historic records for the previous day at 0.00.00 UTC or slightly later? 

I guess if we have that, we don't need to write to the jistoric bucket every time, just rely on the cron job.

>> `something like an override or force flag. Just some idea.` 

We have this, it's a `no-cache` header.

So, it seems like the algo would be the following:
1. Check if the date is today or later. If yes - proceed with current bucket (with TTL) and do exactly what's going on now.
2. If no - check if date is before 1/1/2016. If yes - 404.
3. Check historick bucket - if content is found - return
4. If content not found - go to MCS and request everything EXCEPT news. Store in historic bucket and return.

Also, support `no-cache` header and do one of the following: write to historic bucket together with every write to notmal bucket (a lot of writes, simple) or set up a cron job to update historic bucket at midnight UTC.

berndsi:
The plan sounds good to me.

d00rman:
I think we should allow going back for as long as we have historical records in RB. The reason for that are https://phabricator.wikimedia.org/T139481 and https://phabricator.wikimedia.org/T147952 . Whether we provide content from the beginning of 2016 or today onwards doesn't really make a difference in the grand scheme of things, while it allows us to concentrate on more important things.

As for vandalism, I think we can work on those on a per-case basis manually, as hitting vandalised content on the last retrieval of the UTC day is not likely to occur often.

Finally, for the historic content itself, we should either store it in its entirety, or reference the exact revision/render of the summary as it was when the last request of the day was issued. Showing the article of the day from a year ago with today's content sounds disputable to me, doesn't it?

Pchelolo:
> Finally, for the historic content itself, we should either store it in its entirety, or reference the exact revision/render of the summary as it was when the last request of the day was issued. Showing the article of the day from a year ago with today's content sounds disputable to me, doesn't it?

I'm on the edge here. On one hand it greatly increases the probability of storing vandalism forever, but on the other hand it makes perfect sense. I guess lets store everything and see if we get troubles with vandalised content..

Pchelolo:
Ready for review!

> @d00rman wrote:
> I think we should allow going back for as long as we have historical records in RB.

I think we can generate for older days id we can, and for featured article, pageviews and featured image we can retrofit the content for older days.

berndsi:
@d00rman wrote:
I think we should allow going back for as long as we have historical records in RB.

I proposed 1/1/2016 as an earliest date. For the `news` component we wouldn't have the data going back further than the time we start collecting it. As @Pchelolo mentioned, for `tfa`, `image`, and `mostread` we can easily go back to 1/1/2016. I've seen PageView API examples for Nov 2015.

Pchelolo:
Things left to do here:
 - Decide on the earliest date
 - Support `no-cache` header to be able to quickly and nicely purge vandalised content.

I propose to add these pieces in a separate PR

gwicke:
As @mobrovac, I would recommend avoiding to store full historic summaries. Those old summaries are not performance sensitive, and in any case the number of summaries is quite small. Storing a reference to revision / id (or, more generally, the summary resource including its revision) avoids those issues.

As we see the need to dynamically assemble / hydrate content in quite a few places now, perhaps it would be worth generalizing the hydration. We could then have a generic `hydrate` method that looks for certain constructs & replaces the references with the referenced resources.

Strawman: 
```javascript
{
  '$resource': 'https://en.wikipedia.org/api/rest_v1/page/summary/SomePage/12345'
}
```

Pchelolo:
> @gwicke wrote:
> As @mobrovac, I would recommend avoiding to store full historic summaries. Those old summaries are not performance sensitive, and in any case the number of summaries is quite small. Storing a reference to revision / id (or, more generally, the summary resource including its revision) avoids those issues.

We don't expose summary-by-revision. It's possible to create an endpoint like that, but I don't think it's extremely useful..

> As we see the need to dynamically assemble / hydrate content in quite a few places now, perhaps it would be worth generalizing the hydration. We could then have a generic hydrate method that looks for certain constructs & replaces the references with the referenced resources.

This is the only place right now. Where's the others?

gwicke:
> We don't expose summary-by-revision. It's possible to create an endpoint like that, but I don't think it's extremely useful..

It would be useful if we want to preserve the ability to show the summary as it was at the time. If we don't need that, we can just use a reference to the latest version (https://en.wikipedia.org/api/rest_v1/page/summary/SomePage).

>> As we see the need to dynamically assemble / hydrate content in quite a few places now, perhaps it would be worth generalizing the hydration. We could then have a generic hydrate method that looks for certain constructs & replaces the references with the referenced resources.
> This is the only place right now. Where's the others?

- Feeds
- Various search end points we have been talking about

Basically, anything that includes a page summary.

Pchelolo:
>> We don't expose summary-by-revision. It's possible to create an endpoint like that, but I don't think it's extremely useful..

> It would be useful if we want to preserve the ability to show the summary as it was at the time. If we don't need that, we can just use a reference to the latest version (https://en.wikipedia.org/api/rest_v1/page/summary/SomePage).

That's exactly the question: do we want that or not? Do you think it's useful? I don't see any real use-cases for that, do you?

Regarding the question about what to store for historic feed, we have 3 options:
 * Store complete snapshot with summaries: already implemented, preserves the summaries as they were at the time, faster access to historic data, no need to introduce the revision-based summary endpoint
 * Store only titles and substitute with current summary - simple. Kinda better for the android app use-case: when user scrolls the feed to the older featured content, he will see new summaries, that might have been improved from the time the article was featured, vandalism can't get in there
 * Store only titles, populate with summaries based on revision - complicated to implement, gain is not clear to me.

What should we do?

gwicke:
I personally don't care that much about versioning summaries for these archives. I would propose to start with storing generic `$ref`s to the latest version only (https://en.wikipedia.org/api/rest_v1/page/summary/SomePage), and re-hydrating those dynamically. If we changed our mind about versioning later, we could add the versioned summary end point & start referencing that from feeds.

berndsi:
I don't see the apps wanting to access the summary as it was back then. The latest and greatest summary is usually what we want. I favor the 2nd option from @Pchelolo's list. I don't think there's need to explore option 3 further at this time.

Pchelolo:
Updated the PR to go with option 2. 

Follow-up work on next PRs:
- Support `no-cache` header
- Make the summary-filling more generic and support @gwicke's idea about $reference.

Pchelolo:
I've implemented a generalized population method. The idea is that the `$merge` key would contain an public URIs of pieces of content that should be merged into the current object, for example `[ 'https://en.wikipedia.org/api/rest_v1/page/summary/Test' ]`. That way we get a really generic way of populating content pieces. Also we can reuse cached content from varnish making the whole thing faster.

Right now MCS doesn't provide the `$merge` array to us, so we populate it manually, but when MCS implements this idea we can remove a couple of hacks.

test/features/schema_tests.js
```diff
@@ -33,7 +38,7 @@ describe('Responses should conform to the provided JSON schema of the responce',
     });
 
     it('/feed/featured should conform schema, ruwiki', function() {
-        return preq.get({ uri: server.config.hostPort + '/ru.wikipedia.org/v1/feed/featured/2016/09/08' })
+        return preq.get({ uri: `${server.config.hostPort}/ru.wikipedia.org/v1/feed/featured/${getToday()}` })
```
gwicke: Minor: Could use an array for the uri instead of a string. This saves the need to parse from the string.

gwicke: Good point, nevermind.

lib/mwUtil.js
```diff
@@ -386,4 +386,54 @@ mwUtil.isSelfRedirect = (req, location) => {
...
 
+/**
+ * Traverses through a request body and replaces the given keys with
+ * the content fetched using the 'fetch' callback
+ *
+ * @param responce
+ * @param fetch
+ * @param mergeKey
+ */
+mwUtil.populate = (responce, fetch, mergeKey) => {
```
gwicke: Alternative naming ideas: 

- `hydrateResources`
- `hydrateJSON`

My main gripe with `populate` is that it's fairly generic. Hydration is more commonly used for the process of expanding ids / references to actual content, for example in the context of a twitter feed.

lib/mwUtil.js
```diff
@@ -386,4 +386,54 @@ mwUtil.isSelfRedirect = (req, location) => {
...
+        }
+
+        if (Array.isArray(node)) {
+            for (let i = 0; i < node.length; i++) {
+                _traverse(node[i], () => node.splice(i, 1));
+            }
+        } else if (typeof node === 'object') {
+            if (Array.isArray(node[mergeKey])) {
+                node[mergeKey].forEach(requestResource);
+            } else if (node[mergeKey]) {
```
gwicke: It looks like this would allow plain strings. Do we want to do that, or limit ourselves to arrays only, for simplicity and consistency?

lib/mwUtil.js
```diff
@@ -386,4 +386,54 @@ mwUtil.isSelfRedirect = (req, location) => {
...
+                requestResource(node[mergeKey]);
+            } else {
+                Object.keys(node).forEach((key) => _traverse(node[key], () => delete node[key]));
+            }
+        }
+    }
+    _traverse(responce.body);
+
+    return P.props(requests)
+    .then((content) => setters.forEach((setter) => setter(content)))
```
gwicke: Nice trick to use a global setter array & P.props. Was wondering how the order is preserved earlier, but this does it nicely!

test/features/mwutils.js
```diff
@@ -0,0 +1,60 @@
...
+                    }
+                ],
+                object: {
+                    some_other_prop: 'hello',
+                    $merge: ['prop_contained_here']
+                }
+            }
+        };
+
+        return mwUtil.populate(responce, (uri) => {
```
gwicke: Minor: `response`

lib/mwUtil.js
gwicke: Hmm, is the `title` attribute even an array?

Pchelolo:
@gwicke Updated according to your comments.

gwicke:
:+1:

Pchelolo:
Created a task for MCS to adapt to our `$merge` idea https://phabricator.wikimedia.org/T148798

lib/mwUtil.js
```diff
@@ -386,4 +386,47 @@ mwUtil.isSelfRedirect = (req, location) => {
...
+                } else if (removeFromParent) {
+                    removeFromParent();
+                }
+                delete node.$merge;
+            });
+        }
+
+        if (Array.isArray(node)) {
+            for (let i = 0; i < node.length; i++) {
+                _traverse(node[i], () => node.splice(i, 1));
```
d00rman: This doesn't traverse the whole array, it skips every second element, because `splice()` removes the current element, and then `i` is increased:

```js
> let a = [1,2,3,4,5];
> for(let i = 0; i < a.length; i++) { console.log(i, a[i]); a.splice(i, 1); }
0 1
1 3
2 5
[ 5 ]
> a
[ 2, 4 ]
```

v1/feed.js
```diff
@@ -42,127 +87,111 @@ class Feed {
...
-                            requestTitle(article.title);
-                        });
-                    }
-                });
+        const date = getDateSafe(rp);
+        const dateKey = toKey(date);
+        const dateArr = dateKey.split('-');
+        const populateSummaries = (res) => {
+            function fetchSummary(uri) {
+                return hyper.get({ uri })
```
d00rman: Wouldn't it be good to regex-test for the RB part in the URI, and if it matches, issue direct internal requests instead of going to Varnish and friends?

v1/feed.js
```diff
@@ -42,127 +87,111 @@ class Feed {
...
+            }
+            _traverse(response);
+            return response;
+        };
+        const getContent = (bucket) => hyper.get({
+            uri: new URI([rp.domain, 'sys', 'key_value', 'feed.aggregated', dateKey])
         });
+        const storeContent = (res, bucket) => {
+            return hyper.put({
+                uri: new URI([rp.domain, 'sys', 'key_value', bucket, date]),
```
d00rman: `s/date/dateKey`

v1/feed.js
gwicke: Lets keep an eye on the latency data and error rates after deploy. If using Varnish is a win we can keep it as is, and if it is not, then we can easily (and generically) remap the URL to a direct RB request inside our http handler filter chain.

v1/feed.yaml
```diff
@@ -73,11 +73,6 @@ paths:
                     title: /.+/
                     pageid: /.+/
                     normalizedtitle: /.+/
-              news:
-                - story: /.+/
-                  links:
-                    - normalizedtitle: /.+/
-                      title: /.+/
```
berndsi: Why is this not needed anymore?

berndsi: Never mind. Just noticed your answer to d00rman's question

v1/feed.yaml
```diff
@@ -73,11 +73,6 @@ paths:
                     title: /.+/
                     pageid: /.+/
                     normalizedtitle: /.+/
-              news:
```
berndsi: So, this is not the public facing Swagger spec?

v1/feed.js
```diff
@@ -42,127 +87,117 @@ class Feed {
...
+            status: 200,
+            headers: {
+                'cache-control': this.options.feed_cache_control,
+                // mimic MCS' ETag value
+                etag: `${dateArr.join('')}/${uuid.now().toString()}`,
+                'content-type': CONTENT_TYPE
+            },
+            body: {}
+        };
+        // hydrateResponse its body
```
Pchelolo: @berndsi Ye, that was quite misleading. Refactored a bit.

Pchelolo:
So, I think I've addressed all the comments on this one, and it's ready for a final pass and merge.

test/features/schema_tests.js
```diff
@@ -11,6 +11,11 @@ var Ajv = require('ajv');
 describe('Responses should conform to the provided JSON schema of the responce', function() {
     var ajv = new Ajv({});
 
+    function getToday() {
+        const now = new Date();
+        return `${now.getUTCFullYear()}/${now.getUTCMonth() + 1}/${now.getUTCDate()}`;
```
gwicke: I'd propose to use ISO timestamps instead. Those sort lexicographically, and are easier to parse.

Pchelolo: @gwicke This is a part of public API (`/feed/featured/{yyyy}/{mm}/{dd}`) so we can't change it any more.

gwicke: I see, good point. Should we actually accept /2016/9/1/ rather than /2016/09/01/ ?

v1/feed.js
```diff
@@ -20,6 +19,52 @@ const FEED_URIS = {
...
+    }
+}
+
+/**
+ * Converts the date to the bucket key, key the records in the format YYYY-MM-DD
+ *
+ * @param {Date} date the date to convert
+ * @return {string}
+ */
+function toKey(date) {
```
gwicke: Ahhh, this is essentially what I was proposing, just in the wrong spot.

PR #690, opened 2016-10-18
Updated CORS headers

Pchelolo:
Bug: https://phabricator.wikimedia.org/T148368

Also updated one test page location, it's been deleted on wikipedia, moved it to labs.

cc @wikimedia/services

### DIFF
--- a/lib/security_response_header_filter.js
+++ b/lib/security_response_header_filter.js
@@ -60,7 +60,7 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
         rh['access-control-allow-origin'] = '*';
         rh['access-control-allow-methods'] = 'GET';
         rh['access-control-allow-headers'] =
-            'accept, accept-encoding, origin, accept-language, content-type';
+            'accept, accept-encoding, accept-language, content-type';
         rh['access-control-expose-headers'] = 'etag';
 
         // Set up security headers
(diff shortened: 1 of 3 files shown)


berndsi:
This change should also be made to the node service template, so it can be picked up by other services, like MCS.

Pchelolo:
@berndsi Yep, that's the plan after we discuss this PR

lib/security_response_header_filter.js
```diff
@@ -59,7 +59,8 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
         // Set up basic CORS headers
         rh['access-control-allow-origin'] = '*';
         rh['access-control-allow-methods'] = 'GET';
-        rh['access-control-allow-headers'] = 'accept, content-type';
+        rh['access-control-allow-headers'] =
+            'accept, accept-encoding, origin, accept-language, content-type';
```
gwicke: It's actually impossible to override `origin` in clients, as it is part of verifying the domain comes from as part of [the CORS system](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS). Adding it here would just be confusing.

Pchelolo: @gwicke Safari 9 sends it with an OPTIONS request. As I understand, normally you need to send back all the headers from OPTIONS preflight request to allow it and I couldn't find anything about `Origin` header being somehow specific

PR #692, opened 2016-10-21
Added more allowed headers to account for a Safari 9 bug

Pchelolo:
Tested by life-hacking RESTBase in beta cluster

Bug: https://phabricator.wikimedia.org/T148368

### DIFF
--- a/lib/security_response_header_filter.js
+++ b/lib/security_response_header_filter.js
@@ -60,7 +60,7 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
         rh['access-control-allow-origin'] = '*';
         rh['access-control-allow-methods'] = 'GET';
         rh['access-control-allow-headers'] = 'accept, content-type, ' +
-            'if-match, if-modified-since, if-none-match, ' +
+            'api-user-agent, if-match, if-modified-since, if-none-match, ' +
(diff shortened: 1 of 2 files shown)


lib/security_response_header_filter.js
```diff
@@ -59,7 +59,10 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
         // Set up basic CORS headers
         rh['access-control-allow-origin'] = '*';
         rh['access-control-allow-methods'] = 'GET';
-        rh['access-control-allow-headers'] = 'accept, content-type';
+        rh['access-control-allow-headers'] = 'accept, content-type, ' +
+            'api-user-agent, if-match, if-modified-since, if-none-match, ' +
+            // There's a bug in Safari 9 that makes it require these as allowed headers
+            'cache-control, dnt, accept-encoding, accept-language';
```
gwicke: `cache-control` and `accept-language` are actually user-controllable, and useful to allow. Lets move those to the first part of the list, so that only dnt and accept-encoding are marked as Safari work-arounds?

gwicke:
LGTM overall.

lib/security_response_header_filter.js
Pchelolo: Done

PR #696, opened 2016-10-25
Added PDF endpoint

Pchelolo:
I've already almost lost this code once in my local branches, creating a PR to avoid loosing it again.

We have 2 issues:
1. Not clear how to generate an etag since we don't know the current article revision
2. swagger-ui download link is not perfect, I'll create an upstream PR to fix it. There were major issues with swagger-ui for this use-case before, but it has been fixed recently, and I've updated our swagger-ui forks to pull in the fixes, so it works now.

cc @wikimedia/services


Pchelolo:
Create a PR for swagger-ui to show the download link nicer

v1/pdf.yaml
```diff
@@ -0,0 +1,84 @@
...
+          schema:
+            $ref: '#/definitions/problem'
+
+      x-request-handler:
+        - get_pdf_from_backend:
+            request:
+              method: get
+              uri: 'https://pdf-electron.wmflabs.org/pdf?accessKey=secret&url=https://{{domain}}/wiki/{title}'
+        - return_response:
+            return:
```
d00rman: No need for a new step, the `return` block can be put directly in the previous block.

v1/pdf.yaml
```diff
@@ -0,0 +1,84 @@
...
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+
+      x-request-handler:
+        - get_pdf_from_backend:
+            request:
+              method: get
+              uri: 'https://pdf-electron.wmflabs.org/pdf?accessKey=secret&url=https://{{domain}}/wiki/{title}'
```
d00rman: It would probably be better to contact the instance present in BetaCluster (`deployment-pdfrender.deployment-prep.eqiad.wmflabs`) since that one will match the one that will be in production.

d00rman:
Minor nits and comments, otherwise LGTM overall, but we probably don't want to deploy this before the service hits production.

Pchelolo:
Ye, for sure don't deploy yet, wait until it's deployed in prod. I've created a PR to avoid loosing the code

Also, I wanna o

v1/pdf.yaml
Pchelolo: Done

v1/pdf.yaml
d00rman: Actually, scratch my earlier comment. We need to parametrise two things here:
- the service's URI (much like we do with Parsoid and all the others)
- the `accessKey` query attribute, as it changes from one environment to the other; it also needs to be `encodeURIComponent`-ed since in production it contains unsafe characters

v1/pdf.yaml
d00rman: Hm, you say it's done, but I still see this as being a two-step procedure - one to fetch the pdf, the other to return it.

v1/pdf.yaml
```diff
@@ -0,0 +1,84 @@
...
+          in: path
+          description: "Page title. Use underscores instead of spaces. Example: `Main_Page`."
+          type: string
+          required: true
+      responses:
+        '200':
+          description: The PDF render of an article
+          schema:
+            type: file
+          # TODO: Add an etag somehow. How???
```
d00rman: I think that doing another round trip to Cassandra is not such a big penalty, given that we don't expect much traffic here and the amount of time it takes to actually render a page in pdf.

d00rman:
> Also, I wanna continue the discussion about adding an optional? {/size} parameter with an enum of page sizes which to print on. Create a task for it perhaps?

The service can accept a window height and width, so this would be possible. The defaults in prod are set to 1024x768, so yeah, we could discuss both the default options and the need for `size`, even though this would probably mean cache fragmentation.

d00rman:
LGTM in general, modulo minor comments (in-lined).

Pchelolo:
@d00rman Updated to address your comments.

config.test.yaml
```diff
@@ -38,6 +38,11 @@ default_project: &default_project
           host: http://appservice.wmflabs.org
         events: {}
         purged_cache_control: test_purged_cache_control
+        pdf:
+          # Cache PDF for 5 minutes since it's not purged
+          cache_control: s-maxage=600, max-age=600
+          uri: https://pdf-electron.wmflabs.org/pdf
```
d00rman: We have kind of gotten into the habit of putting only the `protocol://host:port` part in the parameters, so let's keep the consistency with the other config options. Additionally, if we ever decide to use the other formats supported by the service (png, jpg), we won't need to change the options in the config.

d00rman:
Puppet config change patch is [here](https://gerrit.wikimedia.org/r/#/c/323548/)

Pchelolo:
@d00rman Done!

d00rman:
:+1:

PR #698, opened 2016-10-26
Blacklist some large, high-traffic, bot-generated pages

eevans:
As noted in the ticket, these entries come from https://docs.google.com/spreadsheets/d/1Cn2z-DPbmjigTHTW3kmuPhn5Tre-yMU6v4jnTMIv6Mo, which in turn was taken from this week's wide partition report (emailed to services@wikimedia.org).  They represent some of the widest partitions from last week, and are the ones most obviously (to me) machine-generated lists.

Bug: https://phabricator.wikimedia.org/T149203

### DIFF
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -147,6 +147,13 @@ paths:
                         Commons:Quality_images_candidates/candidate_list: true
                       ceb.wikipedia.org:
                         Gumagamit:Lsjbot/Anomalier-PRIVAT: true
+                      sv.wikipedia.org:
+                        Användare:Lsjbot/Anomalier-PRIVAT: true
+                        Användare:Lsjbot/Namnkonflikter-PRIVAT: true
+                      fr.wikipedia.org:
+                        Utilisateur:ZéroBot/Log/Erreurs: true
+                      ca.wikipedia.org:
+                        Usuari:TronaBot/log:Activitat_reversors_per_hores: true
             /mobileapps:
               x-modules:
                 - path: sys/mobileapps.yaml


Pchelolo:
+1

Pchelolo:
Although, I have a question - should we maybe move this blacklist to change-propagation?  If we do that we'd solve the wide partition issue as we wouldn't rerender the pages any more, but will still allow clients to request these pages is they really want to, so VE will be usable on this pages, just super slow. @gwicke @eevans what do you think?

eevans:
> Although, I have a question - should we maybe move this blacklist to change-propagation? If we do that we'd solve the wide partition issue as we wouldn't rerender the pages any more, but will still allow clients to request these pages is they really want to, so VE will be usable on this pages, just super slow. @gwicke @eevans what do you think?

Perhaps I'm missing a step here, but without the update, wouldn't they be getting an old copy?

Pchelolo:
@eevans The revision number is included in the VE request, but as for the transcluded templates - yep, that will be a stale copy. Actually my concern about VE is not really valid, it wont really work either way. Just interested if Change-Prop would be a more natural place to put this - it seems better to 'not update at all' then 'request and update, but refuse to do it'.

Also, currently the blacklist only works for parsoid, but every request to parsoid-related endpoints is protected by a query to the revisions table, so there should be some extremely wide partitions in the `title_revisions` table too for these titles (it terms of number of entries, not size).

Anyway, it's an unrelated discussion, I'll create a separate task for it

eevans:
> @eevans The revision number is included in the VE request, but as for the transcluded templates - yep, that will be a stale copy. Actually my concern about VE is not really valid, it wont really work either way. Just interested if Change-Prop would be a more natural place to put this - it seems better to 'not update at all' then 'request and update, but refuse to do it'.

I guess a proper blacklist would also propagate up to mediawiki, and disable VE as an option on that title altogether.

PR #699, opened 2016-10-27
Set up citoid proxy endpoint

Pchelolo:
As the first step and the beginning of the discussion, set up a simple proxying Citoid endpoint in RESTBase.

Bug: https://phabricator.wikimedia.org/T108646
cc @wikimedia/services

### DIFF
--- a/v1/citoid.yaml
+++ b/v1/citoid.yaml
@@ -43,6 +43,14 @@ paths:
             parameter manually, it will be done by the docs engine.
           type: string
           required: true
+        - name: basefields
+          in: query
+          type: boolean
+          required: false
+          description: >
+            Optional boolean argument. If it is set to true (or 1), it will return the base
+            fields for the data in addition to the mapped fields. For example, it will return
+            `publicationTitle` in addition to `bookTitle` or `websiteTitle`
         - name: Accept-Language
           in: header
           type: string
--- a/v1/citoid.yaml
+++ b/v1/citoid.yaml
@@ -1,0 +1,103 @@
+swagger: '2.0'
+info:
+  version: '1.0.0-beta'
+  title: MediaWiki Citation API
+  description: Citation API based on the Citoid services
+  termsOfService: https://github.com/wikimedia/restbase#restbase
+  contact:
+    name: Services
+    email: services@lists.wikimedia.org
+    url: https://www.mediawiki.org/wiki/Services
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /citation/{format}/{query}:
+    get:
+      tags:
+        - Citation
+      summary: Get citation data given an article identifyer.
+      description: |
+        Generates citation data given a URL, DOI, PMID, or PMCID.
+
+        See more at [Citoid service documentation](https://www.mediawiki.org/wiki/Citoid)
+
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      produces:
+        - application/json; charset=utf-8;
+        - application/x-bibtex; charset=utf-8
+      parameters:
+        - name: format
+          in: path
+          description: "The format to use for the resulting citation data"
+          enum:
+            - mediawiki
+            - bibtex
+            - zotero
+          required: true
+        - name: query
+          in: path
+          description: >
+            URL of an article, DOI, PMCID or PMID in the URL-encoded format.
+            Note that on the Swagger-UI doc page you don't need to URI-encode the
+            parameter manually, it will be done by the docs engine.
+          type: string
+          required: true
+        - name: Accept-Language
+          in: header
+          type: string
+          required: false
+          description: >
+            For some articles the result depends on the `Accept-Language` header, so
+            provide it if localized content is required.
+      responses:
+        '200':
+          description: The citation data in the requested format
+        '404':
+          description: Citation data was not found.
+          schema:
+            $ref: '#/definitions/problem'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+
+      x-request-handler:
+        - get_citation_from_backend:
+            request:
+              method: get
+              uri: '{{options.host}}/api?format={format}&search={query}'
+              headers:
+                accept-language: '{{accept-language}}'
+      x-monitor: true
+      x-amples:
+        - title: Get citation for Darth Vader
+          request:
+            params:
+              domain: en.wikipedia.org
+              format: bibtex
+              query: 'https://en.wikipedia.org/wiki/Darth_Vader'
+            responce:
+              status: 200
+              headers:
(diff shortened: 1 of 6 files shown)


mvolz:
One possible hiccup that I'm not sure affects this change in a bug upstream in citoid master: bibtex is currently getting returned to as a Buffer for some reason and not plain text. In tests, I get Buffer from bibtex requests. Requests behaves fine with curl. The browser doesn't  seem to handle it well. Because of this bug, restbase might end up storing Buffer, not plain text. 

Also, there seems to be some missing a query params, or does this all go into /{query}/ somehow? There's an additional param called basefields, documented here: https://www.mediawiki.org/wiki/Citoid/API#Arguments

Pchelolo:
@mvolz Just tested the buffer issue, it seems to work fine. Also added the `basefields` parameter.

Also, fyi, in the current version of the patch nothing is stored in RESTBase, the requests are just proxied.

gwicke:
I would hold off on adding the baseFields parameter until we understand how it's used, and whether it could be folded into `format`. It is essentially requesting a different format after all.

mvolz:
Yeah, actually we could just roll it into format; basefields is only a valid param for mediawiki format, so we could proxy basefields = 1 or basefields = true into something like mediawikibasefields and then have basefields = 0 or basefields = false into just mediawiki.

gwicke:
Naming ideas for the `basefields=1` format: `mediawiki_extended`, `mediawiki_complete`

mvolz:
The addition of the base fields is meant to be temporary for backwards compatibility issues. In the future, it should be *only* base fields replacing the type specific fields. There's not really a completeness element to this.

v1/citoid.yaml
```diff
@@ -0,0 +1,111 @@
...
+          schema:
+            $ref: '#/definitions/problem'
+
+      x-request-handler:
+        - get_citation_from_backend:
+            request:
+              method: get
+              uri: '{{options.host}}/api?format={format}&search={query}'
+              headers:
+                accept-language: '{{accept-language}}'
```
d00rman: If the user doesn't provide the header, we should default to the assumed language based on the domain (`en`, `fr`, etc)

Pchelolo: Hm, that's not super easy actually.. For simple stuff like `en` of `fr` that's trivial, but there're a whole bunch of exceptions, like `simple` or `mediawiki.org`. 

So ideally we would want to fetch the wiki language from the [MediaWiki SiteInfo](https://en.wikipedia.org/w/api.php?action=query&meta=siteinfo&siprop=general%7Cnamespaces%7Cnamespacealiases%7Cstatistics) api `general.lang` property - it won't be a huge overhead since all of the `siteInfo` requests are cached in memory, so we normally have all of the config locally. However, `siteInfo` fetch returns a promise, and we don't support promises in the functions we call from the template. So, I will need to add support for promises there before we can proceed with this feature.

Another option seem a bit hackier but way simpler to me - create a `/sys` endpoint in RESTBase that would expose the siteInfo by domain, and then fetch it in a step prior to the execution of the actual request and transclude the default from there. 

I'm on the edge here - adding promise support is hard and it might have perfomance implications on the `Template` code which would affect how fast everything works, adding a `/sys` endpoint is quite hacky, but certainly easier.

berndsi: Why is the other option hackier? I think an endpoint like this might be useful for other services, too. I doubt it would need all of `siteInfo`. If we start with the minimal properties required that would be enough for a start. More could be added later if needed.

Pchelolo: @berndsi That would be an internal RESTBase-only endpoint, we have no intension to expose it publicly or to other services (right now, we may change the opinion if that feels useful)

Why is that more hacky is because normally client would set the `accept-language` header, so the information obtained from that endpoint would be useful only on a small percentage of requests, but our current tech powering the request templates would require us to fetch it regardless of whether the `accept-language` was provided or not.

d00rman: +1 for exposing the site info via `/sys`. Agreed that it's a bit hacky, but it will certainly be useful and much easier than the alternative.

d00rman:
> Naming ideas for the basefields=1 format: mediawiki_extended, mediawiki_complete

Perhaps `mw-complete` or `mw-extended` - shorter and nicer. In any case, +1 on putting basefields as a different format

gwicke:
Another option: `mediawiki++`. Just kidding.

I have to say that to an outsider, the purposes of all these very similar format options appear rather vague. The docs really need to explain why you would use format a, b, or c, and what the trade-offs are. If we struggle to find a strong reason for using one of those formats over the others, then that could be a strong hint for deprecating or even outright removing it.

Pchelolo:
> The docs really need to explain why you would use format a, b, or c, and what the trade-offs are.

I don't have any experience with these formats, so some help from @mvolz on the docs side would be much appreciated.

mvolz:
There is some documentation here:
https://www.mediawiki.org/wiki/Citoid/API#Requests

Zotero and bibtex are standard formats used by other people and software.
 mediawiki, which as you can tell by its name, is designed specifically for
mediawiki, to be used with templateData. The VE team decided together (Me,
jdforrester, trevor et al. ) on a format that would work given the limits
of TemplateData. We would have just used the Zotero format directly except
it had structural properties which made using it with TD difficult.

Mediawiki uses the same field names as Zotero for convenience. Zotero has
"basefields" which are basically super field names for fields and are in
parentheses in the typemap list. For example, if you look at
https://aurimasv.github.io/z2csl/typeMap.xml#map-bookSection, the basefield
for 'bookTitle' is 'publicationTitle'. The basefields parameter was
intended to replace the field with the parenthesis field name where
present. However, currently for backwards compatibility, it merely includes
them as well without replacing them.

MWDeprecated was a very early format that we discarded once we made changes
to the TD extension. I do know some user gadgets were made using it and I
don't know if all of them switched over. We should definitely have a plan
to eliminate this entirely.

There is actually an even larger set of formats to export from would be
trivial for us to allow, see:
https://www.zotero.org/support/dev/data_formats.

Pchelolo:
I've updated the PR trying to resolve all the comments. Because the `mediawiki-baseields` format is rewritten to the `format=mediawiki&basefields=true` we need a JS module for it. Later we can update Citoid to allow the `mediawiki-basefields` format, then we'd be able to get rid of the JS module.

d00rman:
Prod config patchset is [here](https://gerrit.wikimedia.org/r/#/c/324911/).

v1/citoid.yaml
```diff
@@ -0,0 +1,105 @@
...
+    url: https://www.mediawiki.org/wiki/Services
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /citation/{format}/{query}:
+    get:
+      tags:
+        - Citation
+      summary: Get citation data given an article identifyer.
```
mvolz: Identifier is misspelled :).

PR #700, opened 2016-10-28
Add content-location header to all responses

Pchelolo:
The implementation is a bit hacky (we use the knowledge that the 'domain' parameter will be the first in the path, followed by the `v1` parameter), but this allows us to set this filter as a 'request' filter, meaning it's executed only once for the root request, and never executed for subsequent requests, which saves us a bit of overhead. 

However, we could potentially mount it as a route filter under `v1` hierarchy, not sure what's better..

Bug: https://phabricator.wikimedia.org/T149444
cc @wikimedia/services

### DIFF
--- a/lib/content_location_filter.js
+++ b/lib/content_location_filter.js
@@ -1,0 +1,42 @@
+"use strict";
+
+const HyperSwitch = require('hyperswitch');
+const Template = HyperSwitch.Template;
+const URI = HyperSwitch.URI;
+const mwUtil = require('./mwUtil');
+
+let basePathTemplate;
+
+module.exports = (hyper, req, next, options) => {
+    basePathTemplate = basePathTemplate || new Template({
+        uri: options.templates.base_uri_template
+    });
+    if (req.method !== 'get') {
+        return next(hyper, req);
+    } else {
+        const baseUri = basePathTemplate.expand({
+            request: {
+                params: {
+                    // This is a bit of a hack, but we know the domain
+                    // will be the first in the path, so use that info
+                    domain: req.uri.path[0]
+                }
+            }
+        }).uri;
+        const attachLocation = (res) => {
+            if (res.status !== 301 && res.status !== 302) {
+                res.headers = res.headers || {};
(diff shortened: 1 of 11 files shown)


lib/content_location_filter.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+    if (req.method !== 'get') {
+        return next(hyper, req);
+    } else {
+        const baseUri = basePathTemplate.expand({
+            request: {
+                params: {
+                    // This is a bit of a hack, but we know the domain
+                    // will be the first in the path, so use that info
+                    domain: req.uri.path[0]
```
gwicke: This makes me wonder if the /v1/ route filter would be better overall. In that case, we should be able to use `req.params.domain`. Another benefit of a route filter on /v1/ would be that it would apply to internal sub-requests to public end points as well.

gwicke:
> However, we could potentially mount it as a route filter under v1 hierarchy, not sure what's better..

To me that sounds slightly cleaner overall.

Pchelolo:
@gwicke Updated it all. I don't understan what's up with travis today, it's just not working...

projects/wikimedia.org.yaml
```diff
@@ -42,6 +42,10 @@ paths:
             - path: ./lib/normalize_title_filter.js
               options:
                 redirect_cache_control: '{{options.purged_cache_control}}'
+            - path: lib/content_location_filter.js
+              options:
+                templates:
+                  base_uri_template: 'https://{{request.params.domain}}/api/rest_v1'
```
gwicke: In other URLs we support direct access to parameters without a prefix (just `{{domain}}`). Should we do the same here? This would require passing req.params as the global object to the template, instead of all of req.

Pchelolo: It will work even if we just remove `request.params`, but I wanted to leave it for better self-documentation

gwicke: I see, mostly wondered about why this was different. Sounds good to me.

PR #702, opened 2016-11-02
Added support for providing redirected content for CORS request

Pchelolo:
Basically this implements the strategy explained in https://phabricator.wikimedia.org/T149295#2764406 

cc @wikimedia/services

### DIFF
--- a/lib/access_check_filter.js
+++ b/lib/access_check_filter.js
@@ -4,6 +4,35 @@ const mwUtil = require('./mwUtil');
 const P = require('bluebird');
 const URI = require('hyperswitch').URI;
 
+function redirect(content, location, options) {
+    const vary = content.headers.vary ? `${content.headers.vary}, origin` : 'origin';
+    return {
+        status: 302,
+        headers: Object.assign(content.headers, {
+            location,
+            'cache-control': options.redirect_cache_control || 'no-cache',
+            vary
+        }),
+        body: content.body
+    };
+}
+
+function response(hyper, req, path, titleParamName, location) {
+    const pathBeforeTitle = path.substring(0, path.indexOf(`{${titleParamName}}`));
+    const contentLocation = new URI(pathBeforeTitle, req.params, true).toString() + location;
+    return hyper.request({
+        method: req.method,
+        uri: new URI(contentLocation),
+        headers: req.headers,
+        body: req.body
+    })
+    .tap((res) => {
+        res.headers = res.headers || {};
+        res.headers['cache-control'] = 'no-cache';
+        res.headers.vary = res.headers.vary ? `${res.headers.vary}, origin` : 'origin';
+    });
+}
+
 module.exports = (hyper, req, next, options, specInfo) => {
     const rp = req.params;
     const titleParamName = options.title || 'title';
@@ -32,19 +61,14 @@ module.exports = (hyper, req, next, options, specInfo) => {
                     && !mwUtil.isNoCacheRequest(req)) {
                 const newParams = Object.assign({}, rp);
                 newParams[titleParamName] = redirectTarget;
-                let location = mwUtil.createRelativeTitleRedirect(specInfo.path,
-                    req, {
-                        newReqParams: newParams,
+                const location = mwUtil.createRelativeTitleRedirect(specInfo.path,
+                    req, redirectTarget, {
                         titleParamName,
                         dropPathAfterTitle: true,
                     });
 
-                if (mwUtil.isSelfRedirect(req, location)) {
-                    location = mwUtil.addQueryString(location, { redirect: false });
-                }
-
                 let contentPromise;
-                if (options.attach_body_to_redirect) {
+                if (options.attach_body_to_redirect || mwUtil.isCrossOrigin(req)) {
                     contentPromise = P.resolve(content);
                 } else {
                     contentPromise = P.resolve({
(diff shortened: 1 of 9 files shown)


lib/access_check_filter.js
```diff
@@ -4,6 +4,35 @@ const mwUtil = require('./mwUtil');
...
+        headers: Object.assign(content.headers, {
+            location,
+            'cache-control': options.redirect_cache_control || 'no-cache',
+            vary
+        }),
+        body: content.body
+    };
+}
+
+function response(hyper, req, path, titleParamName, location) {
```
gwicke: Minor: `response` is very general -- would `redirectResponse` be more accurate?

lib/access_check_filter.js
```diff
@@ -4,6 +4,35 @@ const mwUtil = require('./mwUtil');
...
+    const pathBeforeTitle = path.substring(0, path.indexOf(`{${titleParamName}}`));
+    const contentLocation = new URI(pathBeforeTitle, req.params, true).toString() + location;
+    return hyper.request({
+        method: req.method,
+        uri: new URI(contentLocation),
```
gwicke: If I understand this right, we are constructing an internal path, and then perform an internal request. I'm wondering if we could instead use `location` (converted to absolute URL). The potential upsides might be simpler code here, and the ability to leverage Varnish caching. The underlying `preq` library would also take care of following multiple redirects.

lib/access_check_filter.js
gwicke: Or `resolvedRedirectResponse`.

lib/access_check_filter.js
Pchelolo: @gwicke Oh, sorry, I should've written a comment here. Basically the plan is to do exactly that, but right now for providing the correct `content-location` header we rely on the content-location that will be set on the response to this request. The content-location code is not deployed yet, so we can't rely on it. 

More importantly, we would need to wait until all entries in Varnish expire before switching to it here to be able to provide correct content-location. After we merge this I will create a task for that.

lib/content_location_filter.js
```diff
@@ -18,19 +18,26 @@ module.exports = (hyper, req, next, options) => {
...
         const attachLocation = (res) => {
-            if (res.status !== 301 && res.status !== 302) {
-                res.headers = res.headers || {};
-                Object.assign(res.headers, {
-                    'content-location': baseUri
-                        + new URI(req.uri.path.slice(2))
-                        + mwUtil.getQueryString(req)
-                });
+            res.headers = res.headers || {};
+            if (res.status === 301 || res.status === 302
```
gwicke: I might be missing something.. what would the pros / cons be of handling CORS redirects here?

Pchelolo: Em??? We don't need to set the `content-location` header for a redirect response, it already has the `location` header

lib/access_check_filter.js
gwicke: Varnish max TTL is currently capped at two days, so content-location wouldn't be everywhere only for a short time. Since nobody is relying on content-location yet, I wouldn't worry about it yet.

Pchelolo: @gwicke Our unit tests rely on it :)

gwicke: I see, good point. Since this is a more general testing issue, perhaps it would be worth adding a http filter rule that redirects requests to the REST API directly to RB itself when testing?

lib/content_location_filter.js
Pchelolo: Ah, got it. Actually that's a good idea since we have some similar code in a couple of places. Let me try to explore this option.

lib/access_check_filter.js
Pchelolo: To implement going to Varnish I need #703 because we would need the ability to resolve relative redirects to absolute

Pchelolo:
@gwicke Updated:
- Used Varnish to fetch the redirected content
- Put all the CORS-related code in one filter.

Overall I like the new solution way more the the old one.

lib/content_location_filter.js
```diff
@@ -3,15 +3,51 @@
...
+function resolvedRedirectResponse(hyper, req, res) {
+    let getContentURI;
+    if (isAbsoluteRedirect(res.headers.location)) {
+        getContentURI = P.resolve(res.headers.location);
+    } else {
+        getContentURI = resolveRelativeRedirect(hyper, req, res);
+    }
+    return getContentURI.then((contentURI) => hyper.request({
+        method: req.method,
+        uri: new URI(contentURI)
```
gwicke: I *think* this implicitly avoids cyclical redirects by not forwarding the `origin` header, and thus keeping those redirects HTTP-only & iterative. Since this is quite subtle & important, lets add a comment to make sure that we don't break this in the future.

Pchelolo: Added a comment and a test

lib/mwUtil.js
```diff
@@ -301,14 +301,31 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
             .replace(pathBeforeTitle, '');
     const pathSuffixCount = (pathSuffix.match(/\//g) || []).length;
     const backString = Array.apply(null, { length: pathSuffixCount }).map(() => '../').join('');
+    let location;
     if (!options.dropPathAfterTitle) {
         const pathPatternAfterTitle = path.substring(path.indexOf('{title}') - 1);
-        return backString
+        location = backString
```
Pchelolo: @gwicke The semantics is different the `prefixPath`. `backString` contains the `../../` elements exactly the number of time we need to go back from the current URI. `prefixPath` feels kinda different to me.

gwicke: I see, that wasn't clear to me. Perhaps something like `relativePathPrefix` ?

lib/mwUtil.js
```diff
@@ -301,14 +301,31 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
...
+};
+
+/**
+  * Checks if the request is a CORS request
+  *
+  * @param {Object} req The request to check
+  * @return {boolean} Whether the request is CORS or not
+  */
+mwUtil.isCrossOrigin = (req) => {
+    return req && req.headers && req.headers.origin && req.headers.origin !== req.params.domain;
```
gwicke: Minor: In some situations it might be more desirable to fail if no `req` or `req.headers` is passed in. Returning `false` here might instead cover up a bug. Perhaps we could just remove those extra checks?

gwicke:
Looks great overall! The way all the redirect / cors stuff is in one place is really nice.

PR #705, opened 2016-11-03
Switch to the restriction table based access check filter

Pchelolo:
I don't propose to deploy this yet, but it's handy to have it as a PR for labs testing.

cc @wikimedia/services


Pchelolo:
Found 2 blockers for this during testing: 
- https://phabricator.wikimedia.org/T149966
- https://phabricator.wikimedia.org/T149965

Closing the PR since it's gonna take some time.

PR #706, opened 2016-11-04
Bump summary thump size to 640

Pchelolo:
There's been a broad discussion on how to improve the image quality for the article of the day, and per https://phabricator.wikimedia.org/T149450#2769784 the agreement was to generically increase the thumb size to 640 and downsize on the client. 

Also made a minor bump to the spec version so that stored content could be rerendered automatically.

Bug: https://phabricator.wikimedia.org/T149450

### DIFF
--- a/v1/summary.yaml
+++ b/v1/summary.yaml
@@ -29,7 +29,7 @@ paths:
 
         Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
       produces:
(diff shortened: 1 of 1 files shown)


gwicke:
LGTM. How will this work for existing clients, during the transition phase?

Pchelolo:
@gwicke see https://phabricator.wikimedia.org/T150062 for discussion of that question

gwicke:
So this is basically on hold, as we are updating the featured article response only for now.

Pchelolo:
@gwicke I guess we can just close it since it seems like the general direction has shifted to using the new thumb API

PR #708, opened 2016-11-05
metrics: Replace project parameter descriptions with text from analytics/aqs.git

Krenair:
all-projects only works for aggregate.

### DIFF
--- a/v1/metrics.yaml
+++ b/v1/metrics.yaml
@@ -49,7 +49,7 @@ paths:
       parameters:
         - name: project
           in: path
-          description: If you want to filter by project, use the domain of any Wikimedia project, for example 'en.wikipedia.org', 'www.mediawiki.org' or 'commons.wikimedia.org'. If you are interested in all pageviews regardless of project, use all-projects.
+          description: The name of any Wikimedia project formatted like {language code}.{project name}, for example en.wikipedia. You may pass en.wikipedia.org and the .org will be stripped off.  For projects like commons without language codes, use commons.wikimedia.  For projects like www.mediawiki.org, you can use that full string, or just use mediawiki or mediawiki.org.
           type: string
           required: true
         - name: access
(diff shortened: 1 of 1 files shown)


Krenair:
test failure looks unrelated

Krenair:
@wikimedia/analytics

v1/metrics.yaml
```diff
@@ -49,7 +49,7 @@ paths:
       parameters:
         - name: project
           in: path
-          description: If you want to filter by project, use the domain of any Wikimedia project, for example 'en.wikipedia.org', 'www.mediawiki.org' or 'commons.wikimedia.org'. If you are interested in all pageviews regardless of project, use all-projects.
+          description: The name of any Wikimedia project formatted like {language code}.{project name}, for example en.wikipedia. You may pass en.wikipedia.org and the .org will be stripped off.  For projects like commons without language codes, use commons.wikimedia.  For projects like www.mediawiki.org, you can use that full string, or just use mediawiki or mediawiki.org.
```
gwicke: We actually suggested using the domains over the custom project names on purpose, as we feel that this is more obvious and consistent.

gwicke:
Perhaps we should consider updating the aqs docs instead, for consistency?

v1/metrics.yaml
milimetric: The domains work, and that's why this documentation explained it the way it was.  This explanation is confusing and I think we should just use the previous version everywhere, especially towards the public.

We store data the way we do for good reason, and we can't hard-code to only allow the domains because of that and special projects like all-projects.  We're close to adding wiki-projects as pseudo-sub-projects here and that will complicate things even more.  So I see no reason to expose the internals of how this works to consumers.

Krenair:
I don't really mind which way you guys decide to do it as long as the all-projects issue is resolved.

milimetric:
Thanks for spotting that, indeed we should take off the all-projects part,
but I like the previous explanation from the end user's point of view.  Ok
with everyone if we close this and I just submit a new PR with
"all-projects" stripped from the other two endpoints?

Krenair:
:thumbsup:

milimetric:
https://github.com/wikimedia/restbase/pull/710

PR #712, opened 2016-11-14
Restrict format for feed endpoint

Pchelolo:
To avoid cache fragmentation restrict dates to be 0-padded all the time.

Bug: https://phabricator.wikimedia.org/T149148


gwicke:
:+1:, happy to merge once travis has done its part.

PR #713, opened 2016-11-15
Set content revision for summary

Pchelolo:
Need to include the revision on the `ETag` because we'd need to use it for access control.

Depends on https://github.com/wikimedia/swagger-router/pull/50

cc @wikimedia/services

### DIFF
--- a/sys/key_value.js
+++ b/sys/key_value.js
@@ -19,7 +19,7 @@ function returnRevision(req) {
         if (dbResult.body && dbResult.body.items && dbResult.body.items.length) {
             const row = dbResult.body.items[0];
             let headers = {
-                etag: mwUtil.makeETag('0', row.tid),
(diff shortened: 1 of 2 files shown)


Pchelolo:
Depends on https://github.com/wikimedia/swagger-router/pull/50

gwicke:
Since this depends on the new hyperswitch version, perhaps it's worth explicitly bumping the dependency version in package.json?

Pchelolo:
@gwicke On new `swagger-router` version which's a transitive dependency. So do you think it worths releasing a new `hyperswitch` version just to bump it?

gwicke:
@Pchelolo, main advantage would be to communicate that people need to update their dependencies. Although in practice the number of people actually noticing this might be small. No strong feelings either way from my side. Happy to merge as-is.

Pchelolo:
@gwicke I don't think it worths the bother to bump versions explicitly..

PR #714, opened 2016-11-15
Added public announcements endpoint

Pchelolo:
Pending deployment of https://gerrit.wikimedia.org/r/#/c/320712/
Bug: https://phabricator.wikimedia.org/T150181
cc @wikimedia/services

### DIFF
--- a/v1/announcements.yaml
+++ b/v1/announcements.yaml
@@ -1,0 +1,137 @@
+# Announcements endpoint powered by MCS
+paths:
+  /announcements:
+    get:
+      tags:
+        - Feed
+      summary: Current announcements to display
+      description: |
+        Gets announcements for display in the official Wikipedia iOS and Android apps.
+
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      produces:
+        - application/json
+        - application/problem+json
+      responses:
+        '200':
+          description: Announcements for the given Wiki
+          schema:
+            $ref: '#/definitions/announcement'
+        '501':
+          description: Unsupported Wiki
+          schema:
+            $ref: '#/definitions/problem'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_announcements_from_mcs:
+            request:
+              method: get
+              uri: '{{options.host}}/{domain}/v1/feed/announcements'
+              gzip: false
+            response:
+              status: '{{get_announcements_from_mcs.status}}'
+              headers: '{{merge(get_announcements_from_mcs.headers, { "cache-control": options.announcement_cache_control})}}'
+              body: '{{get_announcements_from_mcs.body}}'
+      x-amples:
+        - title: Retrieve announcements
+          request:
+            params:
+                domain: en.wikipedia.org
+          response:
+            status: 200
+            headers:
+              content-type: application/json
+            body:
+              announce:
+                - id: /.+/
+                  type: /.+/
+                  start_time: /.+/
+                  end_time: /.+/
+                  platforms: [ /.+/ ]
+                  text: /.+/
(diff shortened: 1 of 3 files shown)


Pchelolo:
One Q is whether we want the endpoint to show up in the docs in all wikipedias or not. Habing it everywhere is easy and it's what this PR is doing. Limiting it to just enwiki is possible, but would require setting up a new project for it, a config deploy etc, so it's much more involved. Thoughts? Opinions?

Pchelolo:
Tests are currently failing due to parsoid issue in beta cluster.

gwicke:
> One Q is whether we want the endpoint to show up in the docs in all wikipedias or not. Habing it everywhere is easy and it's what this PR is doing. Limiting it to just enwiki is possible, but would require setting up a new project for it, a config deploy etc, so it's much more involved. Thoughts? Opinions?

If we returned empty lists of announcements for all languages, then exposing the end point as experimental everywhere could work. The API would work consistently everywhere, but wouldn't show any announcements for non-English languages. Of course, this assumes that the intention is to add support for non-English languages soon(ish).

@mdholloway @berndsi: Does this sound like a sensible direction to you? Are you planning to add support for projects other than enwiki in the foreseeable future?

berndsi:
Good idea. Yes, the idea is to open this up to other wikis in the future.

berndsi:
Or should we do an empty response (204), without payload?

berndsi:
Proposal at https://gerrit.wikimedia.org/r/#/c/322016/

status: 200
`{ announce: [] }`

berndsi:
> One Q is whether we want the endpoint to show up in the docs in all wikipedias or not.

Having it hidden would be slightly preferable right now. It's only meant to be consumed by apps, and of course experimental, at this time. But it's not a big deal if it shows up, too. But maybe you could add a note that this is not intended for third-party use.

v1/announcements.yaml
```diff
@@ -0,0 +1,149 @@
...
+      produces:
+        - application/json
+        - application/problem+json
+      responses:
+        '200':
+          description: Announcements for the given Wiki
+          schema:
+            $ref: '#/definitions/announcementsResponse'
+        '404':
+          description: Unsupported Wiki
```
berndsi: Should be removed since I removed in in MCS, too.

v1/announcements.yaml
```diff
@@ -0,0 +1,149 @@
...
+        - title: Retrieve announcements
+          request:
+            params:
+                domain: en.wikipedia.org
+          response:
+            status: 200
+            headers:
+              content-type: application/json
+            body:
+              announce:
```
gwicke: Lets remove the assertion in here that expects an announcement to be returned.

Pchelolo:
This needs some updates

Pchelolo:
@d00rman @berndsi Updated

v1/announcements.yaml
```diff
@@ -35,28 +27,7 @@ paths:
...
-                  start_time: /.+/
-                  end_time: /.+/
-                  platforms: [ /.+/ ]
-                  text: /.+/
-                  action:
-                    title: /.+/
-                    url: /.+/
-                  caption_HTML: /.+/
-                  countries: [ /.+/ ]
+      x-monitor: false
```
gwicke: Hmm, should we still have some basic monitoring? Just status 200 & the presence of an announce property would already ensure that we didn't break this end point in MCS or RB.

Pchelolo: The checker script can't check just presence of the property unfortunately. But I can add a test for 200 status code

gwicke:
LGTM overall.

Pchelolo:
@gwicke Updated to fix your latest comments

gwicke:
Waiting for travis..

PR #715, opened 2016-11-17
Emit 640 TFA thumbnail

Pchelolo:
Hacky fix for the TFA thumb size we've been talking about during the meeting.

Bug: https://phabricator.wikimedia.org/T149450


berndsi:
:)

PR #718, opened 2016-11-18
Revert the order of page_restrictions rev key

Pchelolo:
Final polish needed to enable the restrictions table. Still need to reload the data to the new table and backfill the redirects, but overall we're good to go. All unit tests pass (sqlite depends on https://github.com/wikimedia/restbase-mod-table-sqlite/pull/37)

Done a quite extensive testing pass in vagrant, all looks good. Vagrant depends on https://gerrit.wikimedia.org/r/#/c/322358/

Bug: https://phabricator.wikimedia.org/T149965

### DIFF
--- a/lib/access_check_filter.js
+++ b/lib/access_check_filter.js
@@ -32,13 +32,15 @@ module.exports = (hyper, req, next, options, specInfo) => {
     )
     .spread((content, restriction) => {
         if (restriction || content.headers.location) {
+            const revInfo = mwUtil.parseETag(content.headers.etag);
+
             if (restriction && restriction.body) {
-                const revInfo = mwUtil.parseETag(content.headers.etag);
                 mwUtil.applyAccessChecks(restriction.body, revInfo.rev);
             }
 
             // Use redirect target from restrictions table or content response.
             const redirectTarget = restriction && restriction.body
+                && parseInt(revInfo.rev, 10) === parseInt(restriction.body.rev, 10)
                 && restriction.body.redirect || content.headers.location;
             if (redirectTarget
                     && req.query.redirect !== false
@@ -64,6 +66,9 @@ module.exports = (hyper, req, next, options, specInfo) => {
                 }
 
                 return contentPromise.then(content => redirect(content, location, options));
+            } else if (content.status === 302) {
+                content.status = 200;
+                delete content.headers.location;
             }
         }
         return content;
--- a/lib/access_check_filter.js
+++ b/lib/access_check_filter.js
@@ -41,7 +41,8 @@ module.exports = (hyper, req, next, options, specInfo) => {
             // Use redirect target from restrictions table or content response.
             const redirectTarget = restriction && restriction.body
                 && parseInt(revInfo.rev, 10) === parseInt(restriction.body.rev, 10)
-                && restriction.body.redirect || content.headers.location;
+                && restriction.body.redirect
+                || content.headers.location && decodeURIComponent(content.headers.location);
             if (redirectTarget
                     && req.query.redirect !== false
                     && !mwUtil.isNoCacheRequest(req)) {
@@ -67,6 +68,9 @@ module.exports = (hyper, req, next, options, specInfo) => {
 
                 return contentPromise.then(content => redirect(content, location, options));
             } else if (content.status === 302) {
+                // Parsoid.js emits a redirect when the new content is parsed because
+                // there could be a race condition with updating a restriction table.
+                // If the client doesn't want a redirect - remove it.
                 content.status = 200;
                 delete content.headers.location;
             }
--- a/sys/page_revisions.js
+++ b/sys/page_revisions.js
@@ -254,6 +254,8 @@ class PRS {
         return this.storeRestrictions(hyper, req, {
             title: req.params.title,
             // We're storing page_deletions with magic 0 rev just to update the static column
+            // 0 is used to allow "LE" query always find at least something if no restrictions
+            // are present for a particular revision.
             rev: 0,
             page_deleted: revision.page_deleted
         });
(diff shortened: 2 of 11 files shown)


sys/page_revisions.js
```diff
@@ -32,7 +32,7 @@ const tableName = 'title_revisions';
  * @type {string}
  * @const
  */
-const restrictionsTableName = 'restrictions';
+const restrictionsTableName = 'page_restrictions';
```
gwicke: A better name in any case.

Pchelolo: @gwicke Ye, we've needed to rename it anyways..

sys/parsoid.js
```diff
@@ -318,7 +318,8 @@ class ParsoidService {
                             // This revision is actually a redirect. Pass redirect target
                             // to caller, and let it rewrite the location header.
                             resp.status = 302;
-                            resp.headers.location = encodeURIComponent(redirectTarget);
+                            // Filter will take care of uri-encoding
+                            resp.headers.location = redirectTarget;
```
gwicke: This strikes me as a bit brittle. Is this mainly meant as a performance optimization, or is there another reason to defer encoding?

Pchelolo: @gwicke To be able to share the same codepath within the filter

sys/page_revisions.js
```diff
@@ -261,7 +262,8 @@ class PRS {
     storePageDeletion(hyper, req, revision) {
         return this.storeRestrictions(hyper, req, {
             title: req.params.title,
-            rev: req.params.revision || revision.rev,
+            // We're storing page_deletions with magic 0 rev just to update the static column
```
gwicke: Perhaps also worth mentioning that 0 is chosen so that it will always match for any query for less-or-equal some revision greater than 0.

lib/access_check_filter.js
```diff
@@ -64,6 +66,9 @@ module.exports = (hyper, req, next, options, specInfo) => {
                 }
 
                 return contentPromise.then((content) => redirect(content, location, options));
+            } else if (content.status === 302) {
```
gwicke: Is the intention here to override stored redirects?

sys/parsoid.js
gwicke: Could the decoding be done once in the filter & the result be stored in a local variable? The performance impact should be pretty small, and in exchange we'd avoid introducing an exception to the "regular HTTP headers" principle & assumptions about which filters are active.

gwicke:
LGTM overall, just a couple of questions in line.

lib/access_check_filter.js
Pchelolo: We return 302 when content's not in storage && it's a redirect. For `?redirect=false` and some other conditions we don't what 302 but want 200 instead

sys/parsoid.js
Pchelolo: A that's possible. But that would mean doing encode-decode-encode cycle instead of just one `encode`

lib/access_check_filter.js
gwicke: To me it looks as if this condition would fire when content *was* in storage, but the restriction table did not have the redirect. Is that intended?

Pchelolo: @gwicke nono. We're only setting 302 in parsoid.js when the content was *not* in storage because in that case restriction table update would be concurrent with the content request because we've just parsed out the redirect.

Pchelolo: When we're taking the content from the storage we don't set 302 and rely on the filter to get it from the restrictions table

gwicke: I see -- that's pretty subtle. Do you think an explaining comment here could be helpful to remind ourselves about how this interacts with parsoid.js?

Pchelolo:
@gwicke Updated to address your comments

Pchelolo:
I've rebased the PR on top of the latest master. The tests are currently failing, see the Phablet ticket for an explanation why are they failing and a discussion on what to do with that.

Pchelolo:
I've adapted the patch to the new storage (we don't need to supply a tacky artificial fixed `_tid` any more!!) and made the tests pass by making them sequential. The tests need more attention, see discussion on the ticket.

Pchelolo:
I did manually test this locally and in vagrant for various scenarios of page creation/renaming/deleting/redirecting etc and it seems to work correct for all the cases I'm managed to imagine. So the next step would be to create the proxy to run both filters in parallel, although I'm not sure that in this case it's gonna be a valid test - we don't want to request the content twice and also calling one filter might affect the behavior of another one.

lib/access_check_filter.js
```diff
@@ -64,6 +67,12 @@ module.exports = (hyper, req, next, options, specInfo) => {
                 }
 
                 return contentPromise.then(content => redirect(content, location, options));
+            } else if (content.status === 302) {
+                // Parsoid.js emits a redirect when the new content is parsed because
+                // there could be a race condition with updating a restriction table.
```
gwicke: Is this about the time window between a new restriction being created (ex: revision suppressed), and the update for this being processed in ChangeProp?

sys/page_revisions.js
gwicke: {{done}}

gwicke:
:+1:

lib/access_check_filter.js
Pchelolo: Yup.

Pchelolo:
DO NOT MERGE BEFORE I REBASE AND FIX IT UP A BIT!

Pchelolo:
Finally ready!

lib/access_check_filter.js
```diff
@@ -64,6 +67,12 @@ module.exports = (hyper, req, next, options, specInfo) => {
...
+                // If the client doesn't want a redirect - remove it.
```
d00rman: How can we know for sure that the client doesn't want a redirect here simply based on the status code (which implies otherwise) ?

Pchelolo: This is inside an else-if clause, if client wanted a redirect or we wanted to provide it - we wouldn've going into the if clause and never even evaluated this condition comparing the status to 302

PR #719, opened 2016-11-22
Wikidata: Resurrect the /media/math endpoint

d00rman:
Wikidata pages that are editable and use VE, use Mathoid as well, so
resurrect the `/media/math/` endpoint.

Bug: [T150737](https://phabricator.wikimedia.org/T150737)


d00rman:

physikerwelt:
👍

PR #720, opened 2016-11-25
Parallel tests

Pchelolo:
Having lots of tests is great, but now it's takes ages to complete all of them - on my laptop with sqlite it takes 130 seconds. Parallelise some test execution where possible - time down to 90 seconds - almost 30% win in speed. Also, it's nice to examine possible race condition by running some of the tests concurrently. I didn't have a goal to parallelise all parallelazable tests, just added where it seemed to fit.

cc @wikimedia/services

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,22 +46,23 @@
     "entities": "^1.1.1"
   },
   "devDependencies": {
-    "js-yaml": "^3.6.1",
-    "preq": "^0.4.10",
+    "ajv": "^4.7.0",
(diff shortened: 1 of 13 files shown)


Pchelolo:
Win in travis is even more dramatic - from 7.5 minutes to 3 minutes.

d00rman:
Wow, awesome work @Pchelolo :clap:

Pchelolo:
@d00rman I'm trying to do the same for change-prop, but it's a bit trickier

PR #721, opened 2016-11-29
Restore graphoid end point for wikidata

gwicke:
Bug: https://phabricator.wikimedia.org/T151895


gwicke:
/cc @wikimedia/services 

Tests are expected to fail initially because of an unrelated Parsoid beta labs issue that is being fixed right now.

PR #723, opened 2016-12-02
Minor: s/responce/response/

d00rman:


d00rman:
cc @wikimedia/services

PR #724, opened 2016-12-02
Citoid: Fix the monitoring example in the spec; release v0.16.1

d00rman:
The example provided in Citoid's spec was wrongly indented. Also, the
automating checking script run in prod does not understand bibtex, so
use the `mediawiki` format, which returns a JSON.


d00rman:
cc @wikimedia/services

PR #726, opened 2016-12-07
Parsoid version bump and various test fixes

Pchelolo:
The spec test is not related, just it was failing too so I've fixed it.

### DIFF
--- a/test/features/schema_tests.js
+++ b/test/features/schema_tests.js
@@ -13,8 +13,14 @@ parallel('Responses should conform to the provided JSON schema of the response',
     var ajv = new Ajv({});
 
     function getToday() {
+        function zeroPad(num) {
+            if (num < 10) {
+                return `0${num}`;
+            }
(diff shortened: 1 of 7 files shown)


test/features/schema_tests.js
```diff
@@ -13,8 +13,14 @@ parallel('Responses should conform to the provided JSON schema of the response',
...
     function getToday() {
+        function zeroPad(num) {
+            if (num < 10) {
+                return `0${num}`;
+            }
+            return `${num}`;
+        }
         const now = new Date();
-        return `${now.getUTCFullYear()}/${now.getUTCMonth() + 1}/${now.getUTCDate()}`;
+        return `${now.getUTCFullYear()}/${zeroPad(now.getUTCMonth() + 1)}/${zeroPad(now.getUTCDate())}`;
```
gwicke: Alternative: `now.toISOString().replace(/T.*$/,'')`

PR #727, opened 2016-12-07
Implemented a trending endpoint

Pchelolo:
Ready for review. The big open question is whether we actually need to store this data.

cc @wikimedia/services @jdlrobson

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -42,6 +42,7 @@ default_project: &default_project
           secret: secret
         trending:
           host: <THE HOST GOES HERE>
+          cache-control: s-maxage=600, max-age=600
         skip_updates: false
 
 # A different project template, sharing configuration options.
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -40,10 +40,17 @@ default_project: &default_project
           cache_control: s-maxage=600, max-age=600
           uri: https://pdf-electron.wmflabs.org
           secret: secret
+        skip_updates: false
+
+# A separate project for en.wikipedia because it is more feature-rich
+en.wikipedia.org: &en.wikipedia.org
+  x-modules:
+    - path: projects/wmf_enwiki.yaml
+      options:
+        <<: *default_options
         trending:
           host: https://trending-beta.wmflabs.org
-          cache-control: s-maxage=600, max-age=600
-        skip_updates: false
+          cache_control: s-maxage=1800, max-age=1800
 
 # A different project template, sharing configuration options.
 wikimedia.org: &wikimedia.org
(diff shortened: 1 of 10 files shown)


d00rman:
I highly doubt that we need to keep historic data for this, honestly. Trending article change during one day as it is, so just proxying and caching ought to be fine.

@jdlrobson @coreyfloyd @berndsi thoughts on historical needs?

projects/wmf_default.yaml
```diff
@@ -63,6 +63,8 @@ paths:
                                 "ttl": 7200}, options.mobileapps)}}'
                 - path: v1/announcements.yaml
                   options: '{{merge({"announcement_cache_control": "s-maxage=86400, max-age=86400"}, options.mobileapps)}}'
+                - path: v1/trending.js
```
d00rman: Hm hm. I wonder if we should now split enwiki off, seeing that trending data is available only for enwiki (for now).

lib/mwUtil.js
```diff
@@ -414,4 +414,86 @@ mwUtil.hydrateResponse = (response, fetch) => {
     .thenReturn(response);
 };
 
+/**
+ * Checks whether the date is today or in the past in UTC-0 timezone
+ *
+ * @param {Date} date a date to check
+ * @return {boolean} true if the date is in the past
+ */
+mwUtil.isHistoric = (date) => {
```
d00rman: Regardless of whether we will store historical data for the trending edits or not, :+1: for separating this from the feed context.

v1/trending.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    getTrending(hyper, req) {
+        mwUtil.verifyDateParams(req);
+        const rp = req.params;
+        const date = mwUtil.getDateSafe(rp);
+        const dateKey = mwUtil.dateToKey(date);
+        const dateArr = dateKey.split('-');
+        const populateSummaries = (res) => {
+            function fetchSummary(uri) {
```
d00rman: Could this also perhaps be put in `mwUtil` ?

v1/trending.js
```diff
@@ -0,0 +1,104 @@
...
+    constructor(options) {
+        this.options = options;
+    }
+
+    _assembleResult(result, dateArr) {
+        // assemble the final response to be returned
+        return {
+            status: 200,
+            headers: {
+                'cache-control': this.options.feed_cache_control,
```
d00rman: `feed_cache_control`? Why when we have `this.options.cache_control` ?

Pchelolo:
@d00rman Addressed your comments

d00rman:
:+1: Thnx @Pchelolo. I still think it would be prudent to split enwiki's project off of the others, just as we've done with enwiktionary.

config.example.wikimedia.yaml
```diff
@@ -40,6 +40,9 @@ default_project: &default_project
           cache_control: s-maxage=600, max-age=600
           uri: https://pdf-electron.wmflabs.org
           secret: secret
+        trending:
+          host: https://trending-beta.wmflabs.org
+          cache-control: s-maxage=600, max-age=600
```
d00rman: Let's start conservatively with 30 or 45 mins and then decrease it once we see how the service behaves in production.

Pchelolo:
@d00rman Done - increased caching to 30 minutes and separated out en.wiki

d00rman:
Puppet config change is [here](https://gerrit.wikimedia.org/r/326529).

PR #729, opened 2016-12-13
Support revisions access to mobile-sections

Pchelolo:
Ready for review.

For now we'll have both revisions and not revisions bucket. Once the revision bucket gets filled up well enough we'll delete the non-revisioned bucket and switch completely to a revision one.

Bug: https://phabricator.wikimedia.org/T153133

cc @wikimedia/services


berndsi:
Does the reloading of data you mentioned have the same effect as forcing a rerender of mobile-sections?

Pchelolo:
@berndsi No, that's what I'm trying to avoid. I want to write a script that will go over all the content that's currently stored in RESTBase and push it to the new storage bucket to avoid generating lots of load on MCS

berndsi:
@Pchelolo Thanks. I think eventually we'd like to bump the version in the content-type response header and force a rerender of page that have not been edited in a while. But I think this can wait until mid January. I don't want to cause any instability during the holiday break.

Pchelolo:
Rebased and removed a leftover console.log in tests. Ready to be deployed (it's a no-op anyway right now)

Pchelolo:
Any objections against merging and deploying this? @d00rman @gwicke

Pchelolo:
I've completely reimplemented this going forward with the idea that we wouldn't run a reload script, just wait for the new table to fill up naturally.

Pchelolo:
@d00rman Updated, we don't purge or store not-oldest revisions any more.

Pchelolo:
@d00rman And added support for the `no-cache` header. Now it's finally ready

sys/mobileapps.js
```diff
@@ -0,0 +1,281 @@
...
+            .catch({ status: 404 }, () => {
+                // We have no revisions for this title, so it's certainly latest.
+                return -1;
+            });
+        } else {
+            requests.latestRev = P.resolve(-1);
+        }
+
+        const shouldStoreNewRev = (latestRev) =>
+            mwUtils.isNoCacheRequest(req) || !rp.revision || rp.revision >= latestRev;
```
d00rman: What if we have `no-cache` **and** a revision which is `rp.revision <= latestRev` ?

Pchelolo: Removed the check for `no-cache` - it doesn't really make sense to serenader older revisions since we don't store them because of the `latest_hash` policy.

PR #730, opened 2016-12-18
Blacklist User:Cyberpower678/RfX_Report on enwiki

d00rman:
This page is edited by Cyberbot every 15 minutes on average and it is
transcluded in a lot of other user pages, so it has the potential of
creating very wide rows in storage on a large number of pages, so
disabling it.


eevans:
:+1:

PR #736, opened 2017-01-03
Indicate RB needs at least Node 4 to run

d00rman:


d00rman:
cc @wikimedia/services

Pchelolo:
Would be nice to reflect that in package.json too

d00rman:
> Would be nice to reflect that in package.json too

{{done}}

Pchelolo:
👍

PR #741, opened 2017-01-18
Return non-free page images too

Pchelolo:
The default result of the PageImages extension is going to change to show only free images. For us it's OK to show non-free image too, so add the parameter.

Right now this change is a no-op, but let's do it before we forget.

Bug: https://phabricator.wikimedia.org/T147317
cc @wikimedia/services


berndsi:
+1 Thank you for thinking about this.

phuedx:
👌 ❤️

PR #743, opened 2017-01-20
Added logging when some feed pieces are missing

Pchelolo:
Quite frequently we're missing some pieces of the feed from the response for unknown reasons. Added some login to help us investigate the problem.

Bug: https://phabricator.wikimedia.org/T150703

### DIFF
--- a/v1/feed.js
+++ b/v1/feed.js
@@ -48,6 +48,13 @@ class Feed {
             props[part] = hyper.get({
                 uri: uriArray.join('/'),
                 query: { aggregated: true }
+            })
+            .tap((res) => {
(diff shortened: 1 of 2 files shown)


v1/feed.js
```diff
@@ -48,6 +48,13 @@ class Feed {
             props[part] = hyper.get({
                 uri: uriArray.join('/'),
                 query: { aggregated: true }
+            })
+            .tap((res) => {
+                if (res.status !== 200 || !res.body || !Object.keys(res.body)) {
```
d00rman: `!Object.keys(res.body).length`

PR #746, opened 2017-01-23
Added option for 'monthly' granularity in metrics endpoint

Pchelolo:


nuria:
Looks good

Pchelolo:
The tests are failing because the apache mirror we use to fetch correct Cassandra version is down. Unrelated to the PR, merging.

PR #747, opened 2017-01-23
Log feed only for en.wiki

Pchelolo:
Doesn't make sense to log for others, en.wiki is enough of testing what's happening.


Pchelolo:
Self-merging, it's trivial.

PR #750, opened 2017-01-26
Abstracted the common feed functionality

Pchelolo:
In preparation for introducing the anniversaries endpoint.

cc @wikimedia/services

### DIFF
--- a/lib/base_feed.js
+++ b/lib/base_feed.js
@@ -1,0 +1,192 @@
+'use strict';
+
+
+const P = require('bluebird');
+const HyperSwitch = require('hyperswitch');
+const uuid = require('cassandra-uuid').TimeUuid;
+const mwUtil = require('./mwUtil');
+const URI = HyperSwitch.URI;
+
+const DEFAULT_TTL = 3600;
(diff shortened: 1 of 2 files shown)


lib/base_feed.js
```diff
@@ -0,0 +1,192 @@
...
+        const props = {};
+        parts.forEach((part) => {
+            const def = this.options.part_uris[part];
+            const uriArray = [this.options.host, rp.domain].concat(def.uri);
+            if (def.date) {
+                Array.prototype.push.apply(uriArray, dateArr);
+            }
+            props[part] = hyper.get({
+                uri: uriArray.join('/'),
+                query: { aggregated: true }
```
d00rman: Nit: unlikely that this will be supported in anniversaries.

d00rman:
@Pchelolo this needs a rebase now :/

Pchelolo:
@d00rman And the rebase is done.

lib/base_feed.js
Pchelolo: @d00rman Updated. I've decided not to abstract the actual request to the MCS, this could be better done with inheritance then by setting complex parameter

PR #753, opened 2017-01-30
Removed old outdated mobile-sections tables

Pchelolo:
We should let it bake for a while to make sure the new tables are filled with content and then drop the old tables.

Bug: https://phabricator.wikimedia.org/T156662

### DIFF
--- a/sys/mobileapps.js
+++ b/sys/mobileapps.js
@@ -18,21 +18,15 @@ class MobileApps {
         }
 
         const rp = req.params;
-        let fetchPaths;
+        const fetchPaths = {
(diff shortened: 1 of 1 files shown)


d00rman:
:+1:

Pchelolo:
Ok, we've got ~2.2 million titles already, I think we can switch to new tables next week

Pchelolo:
Per discussion on teams meeting let's move this! @d00rman @gwicke @eevans ?

gwicke:
This is scheduled for deployment Wednesday (tomorrow) morning.

PR #758, opened 2017-02-03
Added 'coordinates' property to the summary

Pchelolo:
Bug: https://phabricator.wikimedia.org/T157068

cc @wikimedia/services @coreyfloyd

### DIFF
--- a/v1/summary.js
+++ b/v1/summary.js
@@ -34,6 +34,10 @@ module.exports = (options) => ({
             const coord = coords[0];
             delete coord.primary;
             delete coord.globe;
+            if (coord.lat === undefined || coord.lon === undefined) {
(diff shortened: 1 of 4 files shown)


v1/summary.js
```diff
@@ -25,6 +25,16 @@ module.exports = (options) => ({
...
+        },
+        processCoords(coords) {
+            if (!coords || !coords.length) {
+                return undefined;
+            }
+
+            const coord = coords[0];
+            delete coord.primary;
+            delete coord.globe;
+            return coord;
```
d00rman: Since `lat` and `lon` are `requiredProperties` should we double-check here we have them and return `undefined` if we don't? It only seems prudent.

PR #760, opened 2017-02-09
Use printable view for PDF printing

gwicke:
The WM Germany team hooked up specific CSS fixes with the printable
view. These fixes are only delivered when the user agent matches
electron, and when ?printable=yes is specified.

So far, we used the cached, regular page view. This is better for
performance, as it leverages Varnish cache results. However, it does not
receive the specific fixes until those have become part of the normal
print styles.

So, this patch (temporarily) adds the ?printable=yes parameter to the
source URL.

### DIFF
--- a/v1/pdf.yaml
+++ b/v1/pdf.yaml
@@ -60,7 +60,7 @@ paths:
         - get_pdf_from_backend:
             request:
               method: get
-              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url=https://{{domain}}/wiki/{title}'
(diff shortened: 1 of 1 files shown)


addshore:
Hmmm, AFAIK this should not be needed?
The module will be included on ALL requests made to any mw page that are made with the electron user agent.
https://github.com/wikimedia/mediawiki-extensions-ElectronPdfService/blob/master/src/ElectronPdfServiceHooks.php#L66
There is not a check for printable=yes there

gwicke:
@addshore: Clearly that will only work when the HTML is not cached in Varnish. We don't vary on user agent, so Varnish doesn't care about electron. The printable view is not cached, so the temporary CSS injection will actually work there.

addshore:
@gwicke Ahh, yes, then this change makes sense!

PR #762, opened 2017-02-10
Extract common response schemas to a separate spec for reusability

Pchelolo:
Bug: https://phabricator.wikimedia.org/T157095


d00rman:
Very nice :+1:

Pchelolo:
@d00rman With swagger spec 3.0 there will be a better way to achieve the same, but its not clear when it's coming...

d00rman:
Yup yup. Let's be patient :)

PR #764, opened 2017-02-14
A pass to stabilise endpoints

Pchelolo:
1. Made a pass over all endpoints increasing stability per 1.0 release
2. Still unclear what to do with the metadata endpoints https://phabricator.wikimedia.org/T158100
3. @milimetric @nuria I've marked the page view API endpoints as [stable](https://www.mediawiki.org/wiki/API_versioning#Stable) - do you think that's appropriate? Do you plan any major changes for these APIs?
4. Finally deleted the `mobile-text` endpoint - it's not used and it seems like it will never be since it was an experiment and stuff went to a different path since. @berndsi Is that correct?

cc @wikimedia/services

### DIFF
--- a/v1/announcements.yaml
+++ b/v1/announcements.yaml
@@ -8,7 +8,7 @@ paths:
       description: |
         Gets announcements for display in the official Wikipedia iOS and Android apps.
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
       produces:
         - application/json
         - application/problem+json
--- a/v1/announcements.yaml
+++ b/v1/announcements.yaml
@@ -8,7 +8,7 @@ paths:
       description: |
         Gets announcements for display in the official Wikipedia iOS and Android apps.
 
-        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
       produces:
         - application/json
         - application/problem+json
--- a/v1/summary.yaml
+++ b/v1/summary.yaml
@@ -27,7 +27,7 @@ paths:
         sentences, as well as information about a thumbnail that represents
         the page.
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
       produces:
         - application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/Summary/1.1.0"
       parameters:
--- a/v1/citoid.yaml
+++ b/v1/citoid.yaml
@@ -30,7 +30,7 @@ paths:
           - `bibtex`: format used in conjunction with LaTeX documents.
             See [bibtex.org](http://www.bibtex.org/).
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
       produces:
         - application/json; charset=utf-8;
         - application/x-bibtex; charset=utf-8
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -21,7 +21,7 @@ paths:
         - Page content
       summary: List page-related API entry points.
       description: |
-        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Stable)
+        Stability: [stable](https://www.mediawiki.org/wiki/API_versioning#Stable)
       produces:
         - application/json
       responses:
@@ -212,7 +212,7 @@ paths:
         - Page content
       summary: Get latest HTML for a title.
       description: |
-        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+        Stability: [stable](https://www.mediawiki.org/wiki/API_versioning#Stable)
       parameters:
         - name: title
           in: path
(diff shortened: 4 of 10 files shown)


v1/mobileapps.yaml
```diff
@@ -270,54 +270,3 @@ paths:
                             from_backend.headers) }}'
               body: '{{from_backend.body}}'
       x-monitor: false
-
-  /mobile-text/{title}:
```
d00rman: We already tried to do that, and discovered there are clients using it, cf [T139029](https://phabricator.wikimedia.org/T139029), so we should be careful this time around.

v1/onthisday.yaml
```diff
@@ -16,7 +16,7 @@ paths:
          - Selected: Featured, most important events
          - All: Everything that happened on the provided day of the year
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
```
d00rman: We have just released it. I think this is premature.

v1/pdf.yaml
```diff
@@ -24,7 +24,7 @@ paths:
       description: |
         Renders the page content as PDF.
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
```
d00rman: Idem. Still not used anywhere.

v1/summary.yaml
```diff
@@ -27,7 +27,7 @@ paths:
         sentences, as well as information about a thumbnail that represents
         the page.
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [stable](https://www.mediawiki.org/wiki/API_versioning#Stable)
```
d00rman: Hm, with all of the tickets filed by the reading team, I think this deserves to be `unstable` ?

v1/trending.yaml
```diff
@@ -8,7 +8,7 @@ paths:
       description: |
         Provides a list of articles that are trending right now (receiving the most edits)
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
```
d00rman: Same as pdf and on this day, premature IMHO.

v1/summary.yaml
Pchelolo: The 'stable' doesn't mean the endpoint result is fixed forever, we're still gonna tweak it, just in a backwards-compatible way. Summary is used in so many places now that we have absolutely no possibility to make backwards incompatible changes or remove it..

d00rman:
I think we should follow the experimental -> unstable -> stable path, IMHO.

Also, we should discuss when do we want these changes to take effect and give **at least** 2 to 3 months to people to update/change/etc before we proceed (especially for deleted endpoints).

Pchelolo:
> Also, we should discuss when do we want these changes to take effect and give at least 2 to 3 months to people to update/change/etc before we proceed (especially for deleted endpoints).

So, I guess I'm gonna split this into 3 pieces:
1. Stability bumps - no need to deprecate or wait for anything.
2. Removal of `mobile-text` endpoint - I'll write a deprecation email, mark is as deprecated and I guess we can remove it pretty soon, grafana shows no usage over the last several months.
3. Removal of all the rest we're planning to remove/hide

The goal of this quarter is to release REST API 1.0, and that's setting the upper time limit on removing stuff that's not gonna be in 1.0.

v1/onthisday.yaml
berndsi: Agreed. The iOS app hasn't even started using it. It's likely that there would be change requests once it's actually used by the iOS app.

v1/mobileapps.yaml
berndsi: +1 to what @d00rman said. I made a section about [route usage](https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/RESTBase_services_for_apps#Route_usage) on the MCS page, with some links to some Grafana reports. I remember seeing mobile-text there but cannot find it at the moment. Is it supposed to be at https://grafana.wikimedia.org/dashboard/db/restbase?panelId=15&fullscreen&from=1487062171322&to=1487071346871?

Pchelolo: @berndsi Yes, but those graphs only shows top-20 endpoints by request rate. To see all of them you can go to grafana-admin and change the query.

nuria:
Not sure what "stable" means on this context. Let me know and I can confirm.

nuria:
Ah, sorry, just saw link. Correct in our case, changes will always be backwards compatible.

gwicke:
@nuria, as linked in the API description, end point stability and the associated versioning policy is described at https://www.mediawiki.org/wiki/API_versioning.

v1/announcements.yaml
```diff
@@ -8,7 +8,7 @@ paths:
       description: |
         Gets announcements for display in the official Wikipedia iOS and Android apps.
 
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
```
berndsi: I'd probably leave this one at experimental for now. It's unclear if or when the apps would move to using centralnotice instead. This endpoint was meant for the time when FR-tech was busy at the end of last year.

Pchelolo:
Updated. I think now I've incorporated all the concerns into the PR

Pchelolo:
So, it's been hanging for a while. Time to merge-deploy @gwicke @eevans @d00rman ?

PR #767, opened 2017-02-16
Onthisday documentation improvements

berndsi:
This is mainly to get rid of the typo 'ovents'. Then I thought we could reuse some of the wording from the description earlier in the file.


Pchelolo:
+1

PR #768, opened 2017-02-22
Support new MCS response format

Pchelolo:
See https://gerrit.wikimedia.org/r/#/c/338288/1 for more context

### DIFF
--- a/v1/feed.js
+++ b/v1/feed.js
@@ -125,7 +125,12 @@ class Feed extends BaseFeed {
         const body = {};
         Object.keys(result).forEach((key) => {
             if (result[key].body && Object.keys(result[key].body).length) {
-                body[key] = result[key].body;
+                // TODO: temp code to support transition to new MCS response format
+                if (key === 'onthisday' && result.onthisday.body.selected) {
+                    body[key] = result.onthisday.body.selected;
+                } else {
+                    body[key] = result[key].body;
+                }
             }
         });
         return body;
--- a/v1/feed.js
+++ b/v1/feed.js
@@ -47,7 +47,7 @@ const PARTS_URIS = {
     },
     onthisday: {
         reqTemplate: new Template({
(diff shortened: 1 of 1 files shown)


berndsi:
-1 With this and the new MCS patch you only get `selected` filled out in  http://localhost:7231/en.wikipedia.org/v1/feed/onthisday/all/01/01 . The other four are empty arrays.

Pchelolo:
The 'onthisday' in RESTBase doesn't use the individual endpoints, but /all endpoint which shouldn't have been changed?

d00rman:
@berndsi perhaps you are testing this with the upcoming version of MCS instead of the currently-deployed one?

berndsi:
@Pchelolo Then I don't understand why we need a code change in RESTBase.
@d00rman Yes, I test this with both master and the new MCS patch applied. The goal of this RB patch is that it works with both.

Pchelolo:
> @Pchelolo Then I don't understand why we need a code change in RESTBase.

The feed/featured has selected anniversaries in it and it's using the individual endpoint to fetch it.

berndsi:
Right, I forgot about that since I haven't seen the selected anniversaries appear in practice. Example: https://en.wikipedia.org/api/rest_v1/feed/featured/2017/02/22 . I do remember seeing code that should do this in theory.

Pchelolo:
@berndsi Ok, updated the PR. Apparently it was broken because I forgot to add a `/feed/` portion in one of the URI templates.

berndsi:
Weird. Not sure what happened. I can't reproduce the issue I reported earlier. Maybe a DB issue since I blew away my old sqlite.db file, and now it works.

Pchelolo:
@berndsi To avoid it going to the storage for the content, try doing requests with `-H 'Cache-Control: no-cache`

PR #769, opened 2017-02-23
Deprecated endpoints that would be deleted

Pchelolo:
Bug: https://phabricator.wikimedia.org/T158100
Bug: https://phabricator.wikimedia.org/T158128
cc @wikimedia/services

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -41,7 +41,7 @@ paths:
         - Page content
       summary: List all pages.
       description: |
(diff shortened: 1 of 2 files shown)


v1/content.yaml
```diff
@@ -41,7 +41,7 @@ paths:
         - Page content
       summary: List all pages.
       description: |
-        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental). Don't rely on this.
+        Stability: Deprecated. To be removed on May, 1st 2017
```
d00rman: Let's bold the word *Deprecated*. Idem for the others

PR #770, opened 2017-03-08
Allow specifying the trending period

Pchelolo:
Bug: https://phabricator.wikimedia.org/T156680
cc @wikimedia/services @jdlrobson

### DIFF
--- a/v1/trending.yaml
+++ b/v1/trending.yaml
@@ -1,6 +1,6 @@
 # Trending-edits endpoints powered by the the trending service
 paths:
-  /trending/edits:
+  /trending/edits{/period}:
     get:
       tags:
         - Feed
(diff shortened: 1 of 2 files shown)


v1/trending.yaml
```diff
@@ -9,6 +9,16 @@ paths:
         Provides a list of articles that are trending right now (receiving the most edits)
 
         Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      parameters:
+        - name: period
+          description: |
+            Optional number in hours to only see pages that became trending in
```
d00rman: Slightly clearer, perhaps: *The period, expressed in hours, for which to obtain the list of trending articles. Default: 1 hour*

Pchelolo: I new there will be a comment on this :)

PR #776, opened 2017-03-21
Minor feed bug for no-cache request for non-existent historic content

Pchelolo:

### DIFF
--- a/lib/base_feed.js
+++ b/lib/base_feed.js
@@ -98,6 +98,7 @@ class BaseFeed {
                 return P.join(
                     getContent(this._getHistoricBucketName())
                     .catch({ status: 404 }, () => ({
+                        status: 204,
                         headers: {},
                         body: {}
                     })),
--- a/lib/base_feed.js
+++ b/lib/base_feed.js
@@ -97,14 +97,15 @@ class BaseFeed {
                 // we're able to regenerate and reuse others
                 return P.join(
                     getContent(this._getHistoricBucketName())
-                    .catch({ status: 404 }, () => ({
-                        status: 204,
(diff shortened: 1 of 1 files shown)


gwicke:
What is the reason for returning 204 instead of 404 here? Are there specific expectations in the app about this?

Pchelolo:
@gwicke Basically we want to completely ignore the `404` error and return a normal result so that we end up in the `.then`. We cannot `return 404` because hyper switch disallows plain returning erroneous codes and automagically wraps them into an HTTPError and throws.

gwicke:
@Pchelolo, could we instead .catch the 404 where we need to handle it?

Pchelolo:
@gwicke This is where we need to handle it that's why we 'catch' it here. 
The logic is: for the historic content with a `no-cache` request, we need to check the stored content, because not everything can be regenerated historically and regenerate all pieces of the feed that can be regenerated and re-save those which can't. If there's no historic content (404) - that's fine, just return an 'empty' response. You cannot 'catch' one part of the `P.join` request and not the other part anyhow differently from what we're doing here.

d00rman:
Right, this makes sense to me since we need to return all of the components that we can when regenerating the historic feed :+1:

gwicke:
My concern is more about using 204 when semantically a 404 seems like what we really would want here. If the concern is just the .join, can't we just return a 404 code that is not an error from the .catch? That wouldn't trigger any HTTPError wrapping, and would thus not interfere with promise chaining.

Pchelolo:
@gwicke Done.

gwicke:
:+1:

PR #777, opened 2017-03-22
Spec: Add link to the deprecated stability class

d00rman:


d00rman:
cc @wikimedia/services

d00rman:
Travis fails because it can't fetch the Cassandra tgz (again)...

PR #778, opened 2017-03-23
Blacklist: two User pages on commons

d00rman:


d00rman:
Trivial, so self-merging

Pchelolo:
Ha, I was just now looking why the error rate is elevated and found these as well.

PR #779, opened 2017-03-23
Properly name commons in the blacklist

Pchelolo:
It's 'commons.wikimedia.org', not 'commons.wikipedia.org'


Pchelolo:
Self merging, trivial.

PR #780, opened 2017-03-23
Support regexes in blacklists

Pchelolo:
Some of the pages we blacklist are subpages of some use/bot page. Often they are partitioned by month/date/etc, so we need to support regexes to block all the subpages right away. 

Also, added more blacklists based on the report of the most problematic pages in Change-Prop: https://gist.github.com/Pchelolo/4cbfd3c38623234395cb3c6ee1a85d9a That's generated by running through the `error` topic in kafka, filtering for 500/504 errors and getting the most problematic ones.

### DIFF
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -120,70 +120,64 @@ paths:
                     # rarely viewed in any case.
                     rerenderBlacklist:
                       ca.wikipedia.org:
-                        Usuari:TronaBot/log:Activitat_reversors_per_hores: true
+                        - 'Usuari:TronaBot/log:Activitat_reversors_per_hores'
                       ceb.wikipedia.org:
-                        Gumagamit:Lsjbot/Anomalier-PRIVAT: true
-                        Gumagamit:Lsjbot/Kartrutor2: true
+                        - 'Gumagamit:Lsjbot/Anomalier-PRIVAT'
+                        - 'Gumagamit:Lsjbot/Kartrutor2'
                       commons.wikimedia.org:
-                        Commons:Quality_images_candidates/candidate_list: true
-                        Commons:Quality_images/Subject/Places/Natural_structures: true
-                        Commons:WikiProject_Aviation/recent_uploads/2014_February_17: true
-                        User:Darkweasel94/Vienna/2016_December_14-16: true
-                        User:J_budissin/Uploads/BiH/2016_December_11-20: true
-                        User:Jarekt/d: true
-                        User:JJMC89_bot/report/BSicons/changes/2017-03: true
-                        User:Marcus_Cyron/OgreBotCeramics/2013_May_21-31: true
-                        User:Nicolas_Rück_(WMDE)/Supported_by_Wikimedia_Deutschland_2015/1: true
-                        User:Nicolas_Rück_(WMDE)/Supported_by_Wikimedia_Deutschland_2015/2: true
-                        User:OgreBot/Uploads_by_new_users: true
-                        User:Oxyman/Buildings_in_London/2014_October_11-20: true
-                        User:Oxyman/London/2014_October_1-10: true
-                        User:Stunteltje/gallery: true
-                        User:PIERRE_ANDRE_LECLERCQ/gallery: true
+                        - 'Commons:Quality_images_candidates/candidate_list'
+                        - 'Commons:Quality_images/Subject/Places/Natural_structures'
+                        - '/Commons:WikiProject_Aviation\/recent_uploads\//'
+                        - '/^User:Darkweasel94\/Vienna\//'
+                        - '/^User:J_budissin\/Uploads\/BiH\//'
+                        - 'User:Jarekt/d'
+                        - '/^User:JJMC89_bot\/report\/BSicons\/changes\//'
+                        - '/^User:Marcus_Cyron\/OgreBotCeramics\//'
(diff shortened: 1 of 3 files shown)


projects/wmf_default.yaml
```diff
@@ -126,9 +126,11 @@ paths:
                         Gumagamit:Lsjbot/Kartrutor2: true
                       commons.wikimedia.org:
                         Commons:Quality_images_candidates/candidate_list: true
+                        Commons:Quality_images/Subject/Places/Natural_structures: true
                         User:Darkweasel94/Vienna/2016_December_14-16: true
                         User:J_budissin/Uploads/BiH/2016_December_11-20: true
                         User:Jarekt/d: true
+                        User:Nicolas_Rück_(WMDE)/Supported_by_Wikimedia_Deutschland_2015/1: true
```
subbuss: I pasted User:Nicolas_R%C3%BCck_(WMDE)/Supported_by_Wikimedia_Deutschland_2015/2 in IRC .. but maybe both /1 and /2 are a problem.

subbuss:
s/parsed/Parsoid/ in the commit message

projects/wmf_default.yaml
Pchelolo: Oh, indeed, sorry. I've made a little script that analyzes kafka error topic over the last week and gives out the most problematic pages (with  the highest number of 500/504 errors) and these 2 are definitely problematic. Will update to include /2 as well.

sys/parsoid.js
```diff
@@ -159,11 +159,27 @@ function _dependenciesUpdate(hyper, req) {
     });
 }
 
+function compileReRenderBlacklist(blacklist) {
+    const result = {};
+    blacklist = blacklist || {};
+    Object.keys(blacklist).forEach((domain) => {
+        result[domain] = blacklist[domain].map((entry) => {
```
gwicke: Minor optimization opportunity: You could build up a single regexp per domain with `|`.

gwicke:
LGTM overall. It might be worth compiling a single regexp per domain, to avoid the need to sequentially test all regexps and strings in the common case of no match.

sys/parsoid.js
Pchelolo: Nice idea. Will modify. I think we have the code that does exactly that somewhere but I've completely forgotten where that code is...

Pchelolo:
@gwicke Update the PR with a big regex compilation.

sys/parsoid.js
```diff
@@ -159,11 +159,42 @@ function _dependenciesUpdate(hyper, req) {
...
+
+    /**
+     * From a list of regexes and strings, constructs a regex that
+     * matches any of list items
+     */
+    const constructRegex = (variants) => {
+        let regex = (variants || []).map((regexString) => {
+            regexString = regexString.trim();
+            if (/^\/.+\/$/.test(regexString)) {
+                return `(:?${regexString.substring(1, regexString.length - 1)})`;
```
d00rman: Did you mean `(?:` ... ? :)

sys/parsoid.js
```diff
@@ -159,11 +159,42 @@ function _dependenciesUpdate(hyper, req) {
...
+            }
+            // Instead of comparing strings
+            return `(:?^${decodeURIComponent(regexString)
```
d00rman: Idem

sys/parsoid.js
```diff
@@ -159,11 +159,42 @@ function _dependenciesUpdate(hyper, req) {
...
+     * From a list of regexes and strings, constructs a regex that
+     * matches any of list items
+     */
+    const constructRegex = (variants) => {
+        let regex = (variants || []).map((regexString) => {
+            regexString = regexString.trim();
+            if (/^\/.+\/$/.test(regexString)) {
+                return `(:?${regexString.substring(1, regexString.length - 1)})`;
+            }
+            // Instead of comparing strings
```
d00rman: *Compare strings, instead*

sys/parsoid.js
gwicke: This reads like a google search suggestion.

sys/parsoid.js
```diff
@@ -159,11 +159,42 @@ function _dependenciesUpdate(hyper, req) {
...
 }
 
+function compileReRenderBlacklist(blacklist) {
+    const result = {};
+
+    /**
+     * From a list of regexes and strings, constructs a regex that
+     * matches any of list items
+     */
+    const constructRegex = (variants) => {
```
d00rman: `variants` makes me think of language variants. `pages` perhaps?

sys/parsoid.js
Pchelolo: Oh when will I remember this...?

sys/parsoid.js
```diff
@@ -159,11 +159,42 @@ function _dependenciesUpdate(hyper, req) {
...
+        let regex = (variants || []).map((regexString) => {
+            regexString = regexString.trim();
+            if (/^\/.+\/$/.test(regexString)) {
+                return `(?:${regexString.substring(1, regexString.length - 1)})`;
+            }
+            // Compare strings, instead
+            return `(?:^${decodeURIComponent(regexString)
+                .replace(/[-[\]/{}()*+?.\\^$|]/g, "\\$&")})`;
```
d00rman: We want to match the whole string, not just its beginning, so you are missing a `$` at the end

eevans:
:heart:

PR #786, opened 2017-03-31
Update to the new eslint config and fix errors

d00rman:


d00rman:
cc @wikimedia/services

PR #787, opened 2017-04-03
Test config: Switch to the deployment-prep Electron instance

d00rman:


d00rman:
cc @wikimedia/services

Pchelolo:
Example config should be updated as well

d00rman:
@Pchelolo good catch! {{done}}

Pchelolo:
+1 Waiting for travis

PR #790, opened 2017-04-06
Action Module: Fix parameters check

d00rman:
The action module requires either apiUriTemplate or baseUriTemplate to work properly, but it doesn't need both to be specified.


d00rman:
cc @wikimedia/services

PR #793, opened 2017-04-11
Update all API descriptions; more legalese

gwicke:
- Actually update the descriptions for all project specs.
- Add some more legalese by legal's request. While licensing is also
touched on in the ToUs, they feel that it won't hurt to make it really
explicit.

### DIFF
--- a/projects/wikimedia.org.yaml
+++ b/projects/wikimedia.org.yaml
@@ -4,20 +4,30 @@ paths:
         # swagger options, overriding the shared ones from the merged specs (?)
       - spec:
           info:
-            version: 1.0.0-beta
+            version: 1.0.0
             title: Wikimedia REST API
             description: >
-                This API aims to provide coherent and low-latency access to
-                Wikimedia content and services. It is currently in beta testing, so
-                things aren't completely locked down yet. Each entry point has
-                explicit stability markers to inform you about development status
-                and change policy, according to [our API version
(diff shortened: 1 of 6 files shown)


projects/wikimedia.org.yaml
```diff
@@ -7,17 +7,27 @@ paths:
             version: 1.0.0-beta
```
Pchelolo: Forgot `beta` here

projects/wmf_wiktionary.yaml
```diff
@@ -8,17 +8,27 @@ paths:
             version: 1.0.0-beta
```
Pchelolo: Forgot `beta` here

Pchelolo:
We have the `common.yaml` spec that contains the reusable definitions. Can we maybe put this text there and merge it into the resulting spec?

gwicke:
> We have the common.yaml spec that contains the reusable definitions. Can we maybe put this text there and merge it into the resulting spec?

Do you mean content.yaml? If so, that's at a lower (route) level. The description always comes from the top-level spec, which necessarily needs to differ between our projects.

Pchelolo:
> Do you mean content.yaml? If so, that's at a lower (route) level. The description always comes from the top-level spec, which necessarily needs to differ between our projects.

Right, right, bas suggestion... 

LGTM +1

PR #795, opened 2017-04-13
Allow client-side cache-control for summaries

Pchelolo:
Allow 5-minute client-side caching for the summary content. 

Depends-On: https://gerrit.wikimedia.org/r/348124
Bug: https://phabricator.wikimedia.org/T161284

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -35,6 +35,8 @@ default_project: &default_project
           host: http://citoid-beta.wmflabs.org
         events: {}
         purged_cache_control: test_purged_cache_control
+        # Cache control for purged endpoints allowing short-term client caching
+        purged_cache_control_client_cache: test_purged_cache_control_with_client_caching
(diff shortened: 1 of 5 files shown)


config.test.yaml
```diff
@@ -35,6 +35,8 @@ default_project: &default_project
           host: http://citoid-beta.wmflabs.org
         events: {}
         purged_cache_control: test_purged_cache_control
+        # Cache control for purged endpoints allowing short-term client caching
+        purged_cache_control_client_cache: test_purged_cache_control_with_client_caching
```
gwicke: This is an odd value to use, but I guess it eliminates the change of satisfying the test by a static cache-control response.

Pchelolo: @gwicke That's exactly the idea behind this - make sure we take the value exactly where we want it from, so in test config we have unique self-describing strings

PR #796, opened 2017-04-13
Add the translation public API specification

d00rman:
This PR adds the public API specification for the functionality
exposed in CXServer. As a first step, it is going to be exposed only on
the global wikimedia.org domain, as it doesn't really depend on a
specific wiki/project (it is Wikipedia-centric, though).

Bug: [T107914](https://phabricator.wikimedia.org/T107914)

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -890,7 +890,7 @@ paths:
               body: '{{request.body}}'
       x-monitor: false
 
-  /segmented/{title}{/revision}:
+  /segments/{title}{/revision}:
     get:
       tags:
         - Page content
@@ -941,7 +941,7 @@ paths:
           schema:
             $ref: '#/definitions/problem'
         '404':
-          description: Unknown page, revision or tid
+          description: Unknown page or revision
           schema:
             $ref: '#/definitions/problem'
         default:
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -890,6 +890,70 @@ paths:
               body: '{{request.body}}'
       x-monitor: false
 
+  /segmented/{title}{/revision}:
+    get:
+      tags:
+        - Page content
+      summary: Fetches a segmented page to be used in machine translation
+      description: |
+        Use this end point to fetch the segmented content of a page. Clients should
+        use the returned content in conjunction with the [language transform
+        API](https://wikimedia.org/api/rest_v1/#!/Transform).
+
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      produces:
+        - application/json
+      parameters:
+        - name: title
+          in: path
+          description: "Page title. Use underscores instead of spaces. Example: `Main_Page`."
+          type: string
+          required: true
+        - name: revision
+          in: path
+          description: The revision id
+          type: integer
+          required: false
+      responses:
+        '200':
+          description: The segmented page for the given title and revision
+          schema:
+            type: object
+            properties:
+              sourceLanguage:
+                type: string
+                description: The source language of the page
+              title:
+                type: string
+                description: The title of the segmented page returned
+              revision:
+                type: integer
+                description: The revision ID of the segmented page
+              segmentedContent:
+                type: string
+                description: The segmented HTML body of the contents of the page
+        '400':
+          description: Invalid revision
+          schema:
+            $ref: '#/definitions/problem'
+        '403':
+          description: Access to the specific revision is restricted
+          schema:
+            $ref: '#/definitions/problem'
+        '404':
+          description: Unknown page, revision or tid
+          schema:
+            $ref: '#/definitions/problem'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/page/{domain}/{title}/{revision}'
+      x-monitor: false
+
 definitions:
   revisions:
     description: The result format for revision listing
--- a/v1/transform-global.yaml
+++ b/v1/transform-global.yaml
@@ -1,0 +1,254 @@
+info:
+  version: '1.0.0'
+  title: Global Transform API
+  description: Global Transform API
+  termsofservice: https://github.com/wikimedia/restbase
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /html/{from}/to/{to}{/provider}:
+    post:
+      tags: ['Transform']
+      summary: Machine-translate content
+      description: |
+        Fetches the machine translation for the posted content from the source
+        to the destination language.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: from
+          in: path
+          description: The source language code
+          type: string
+          required: true
+        - name: to
+          in: path
+          description: The target language code
+          type: string
+          required: true
+        - name: provider
+          in: path
+          description: The machine translation provider id
+          type: string
+          required: false
+          enum:
+           - Apertium
+           - Yandex
+           - Youdao
+        - name: html
+          in: formData
+          description: The HTML content to translate
+          type: string
+          required: true
+          x-textarea: true
+      responses:
+        '200':
+          description: The translated content
+          schema:
+            $ref: '#/definitions/cx_mt'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/mt/{from}/{to}/{provider}'
+  /word/{from}/to/{to}/{word}{/provider}:
+    get:
+      tags: ['Transform']
+      summary: Fetch the dictionary meaning of a word
+      description: |
+        Fetches the dictionary meaning of a word from a language and displays
+        it in the target language.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      produces:
+        - application/json
+      parameters:
+        - name: from
+          in: path
+          description: The source language code
+          type: string
+          required: true
+        - name: to
+          in: path
+          description: The target language code
+          type: string
+          required: true
+        - name: word
+          in: path
+          description: The word to lookup
+          type: string
+          required: true
+        - name: provider
+          in: path
+          description: The dictionary provider id
+          type: string
+          required: false
+          enum:
+           - JsonDict
+           - Dictd
+      responses:
+        '200':
+          description: the dictionary translation for the given word
+          schema:
+            $ref: '#/definitions/cx_dict'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/dictionary/{word}/{from}/{to}/{provider}'
+  /list/pair/{from}/{to}/:
+    get:
+      tags: ['Transform']
+      summary: Lists the tools available for a language pair
+      description: |
+        Fetches the list of tools that are available for the given pair of languages.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      produces:
+        - application/json
+      parameters:
+        - name: from
+          in: path
+          description: The source language code
+          type: string
+          required: true
+        - name: to
+          in: path
+          description: The target language code
+          type: string
+          required: true
+      responses:
+        '200':
+          description: the list of tools available for the language pair
+          schema:
+            $ref: '#/definitions/cx_list_tools'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/list/pair/{from}/{to}'
+  /list/tool/{tool}/:
+    get:
+      tags: ['Transform']
+      summary: Lists the tools and language pairs available for the given tool category
+      description: |
+        Fetches the list of tools and all of the language pairs it can translate
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      produces:
+        - application/json
+      parameters:
+        - name: tool
+          in: path
+          description: The tool category to list tools and language pairs for
+          type: string
+          required: true
+          enum:
+            - mt
+            - dictionary
+      responses:
+        '200':
+          description: the list of language pairs available for a given translation tool
+          schema:
+            $ref: '#/definitions/cx_list_pairs_for_tool'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/list/tool/{tool}'
+  /list/languagepairs/:
+    get:
+      tags: ['Transform']
+      summary: Lists the tools and language pairs available for the given tool category
+      description: |
+        Fetches the list of tools and all of the language pairs it can translate
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      produces:
+        - application/json
+      responses:
+        '200':
+          description: the list of source and target languages supported by the API
+          schema:
+            $ref: '#/definitions/cx_languagepairs'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/list/languagepairs'
+
+definitions:
+  cx_mt:
+    type: object
+    properties:
+      contents:
+        type: string
+        description: the translated content
+  cx_dict:
+    type: object
+    properties:
+      source:
+        type: string
+        description: the original word to look up
+      translations:
+        type: array
+        description: the translations found
+        items:
+          type: object
+          properties:
+            phrase:
+              type: string
+              description: the translated phrase
+            info:
+              type: string
+              description: extra information about the phrase
+            sources:
+              type: string
+              description: the source dictionary used for the translation
+  cx_list_tools:
+    type: object
+    properties:
+      tools:
+        type: array
+        description: the list of tools available for the given language pair
+        items:
+          type: string
+          description: the tool available
+  cx_list_pairs_for_tool:
+    type: object
+  cx_languagepairs:
+    type: object
+    properties:
+      source:
+        type: array
+        description: the list of available source languages
+        items:
+          type: string
+          description: one source language
+      target:
+        type: array
+        description: the list of available destination languages
+        items:
+          type: string
+          description: one destination language
--- a/v1/transform-global.yaml
+++ b/v1/transform-global.yaml
@@ -7,7 +7,7 @@ info:
     name: Apache licence, v2
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /html/{from}/to/{to}{/provider}:
+  /html/from/{from_lang}/to/{to_lang}{/provider}:
     post:
       tags: ['Transform']
       summary: Machine-translate content
@@ -21,12 +21,12 @@ paths:
       produces:
         - application/json
       parameters:
-        - name: from
+        - name: from_lang
           in: path
           description: The source language code
           type: string
           required: true
-        - name: to
+        - name: to_lang
           in: path
           description: The target language code
           type: string
@@ -41,7 +41,7 @@ paths:
            - Yandex
            - Youdao
         - name: html
-          in: formData
+          in: body
           description: The HTML content to translate
           type: string
           required: true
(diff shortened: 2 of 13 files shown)


d00rman:
cc @wikimedia/services @kartikm

v1/translate.yaml
```diff
@@ -0,0 +1,254 @@
...
+  title: Content Translation API
+  description: Content Translation API
+  termsofservice: https://github.com/wikimedia/restbase
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /machine/{from}/{to}{/provider}:
+    post:
+      tags: ['Translate']
```
Pchelolo: Could add the tag description as well

v1/translate.yaml
```diff
@@ -0,0 +1,254 @@
...
+      summary: Machine-translate content
+      description: |
+        Fetches the machine translation for the posted content from the source
+        to the destination language.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      consumes:
+        - application/x-www-form-urlencoded
```
Pchelolo: Hm.. I thought we wanted to mostly switch POST endpoints to consuming JSON, didn't we?

v1/translate.yaml
```diff
@@ -0,0 +1,254 @@
...
+          description: The target language code
+          type: string
+          required: true
+        - name: provider
+          in: path
+          description: The machine translation provider id
+          type: string
+          required: false
+          enum:
+           - Apertium
```
Pchelolo: Would they indeed be capitalized? Feels weird to have capitalized params in the URI

v1/translate.yaml
```diff
@@ -0,0 +1,254 @@
...
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.host}}/v1/dictionary/{word}/{from}/{to}/{provider}'
+  /list/pair/{from}/{to}:
```
Pchelolo: We do listings with the trailing `/` and it's kinda a listing. Maybe we could reuse the same concept here?

Pchelolo:
Perhaps it worths choosing the cheapest endpoint and add `x-amples` to it?

d00rman:
> Perhaps it worths choosing the cheapest endpoint and add x-amples to it?

Yup, yup, this PR is not ready to go. When CXServer implements the needed changes (some endpoints don't even exist yet), we need to revise this and monitor at least one route.

Pchelolo:
Then can you indicate that it's dependent on smith in the description? Or I might accidentally merge it :)

d00rman:
Heh, I can't review and -1 my own PR :-1:

d00rman:
Updated the desc @Pchelolo

Pchelolo:
I've -1 👎 your PR, you're welcome :)

v1/translate.yaml
kartikm: Apertium is correct project name.

v1/translate.yaml
```diff
@@ -0,0 +1,254 @@
+info:
+  version: '1.0.0'
+  title: Content Translation API
+  description: Content Translation API
+  termsofservice: https://github.com/wikimedia/restbase
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /machine/{from}/{to}{/provider}:
```
Nikerabbit: `machine` doesn't seem to be a very good name. Even `translate` is better IMHO.

v1/translate.yaml
Nikerabbit: Yes they are currently capitalized and I would keep it that way.

v1/translate.yaml
Nikerabbit: What is this tag intended to encapsulate? I worry there might be confusion with the Translate extension.

v1/translate.yaml
Pchelolo: @Nikerabbit The whole spec is mounted under `/translate` prefix. So the full path will be: `/api/rest_v1/translate/machine/{from}/{to}{/provider}`. I'm not defending the `machine` part though, I don't really like it either, it suggests that not all of the translation is machine-driven

v1/translate.yaml
Pchelolo: @Nikerabbit The tags are used to group the endpoints in the documentation page. See https://en.wikipedia.org/api/rest_v1/ - the endpoints are grouped by tag. So introducing a `translate` tag we will add a new top-level section and group all of these endpoints together int he doc.

v1/translate.yaml
d00rman: It is machine-translated, after all. Dictionary look-up does not constitute translations at all (technically), so I don't see a problem with the name. Do you have a better suggestion @Nikerabbit ?

v1/translate.yaml
Nikerabbit: Okay. It seems the other ones use nouns, so maybe `Translation` here? Can't think of anything more generic that would reflect the scope better, unless we go all the way to things like `Language`. That could then include language identification, spell-checking etc. related services in the future.

v1/translate.yaml
Nikerabbit: `translate/translate` is bit repetitive (except if we change the prefix, see my earlier comment on the tag) but I would still consider it sensible. Other options could be `mt` or `machine-translation`

d00rman: Even if we changed the top-level route to `translation`, `translation/translate` is still redundant. `mt` is too CXServer-specific (and cryptic from the user's perspective), while `machine-translation` is too long IMHO. I don't understand the problem you have with `machine` - `translate/machine/` describes correctly what is about to happen: a text will be machine-translated.

Nikerabbit: It is not the kind of lingo people use in this field. To me it appears as if it is about some machinery related to translation (compare with _translation engine_), especially if it appears out of context like it does here. Machine translation is a thing, but if you swap the order of words it no longer makes sense. And like said above, there are no other forms of translations this would contrast with. Hence I think some repetition is the smallest evil.

d00rman: How about `translate/engine/...` ?

gwicke: Another option could be to extend [the pre-existing `transform` hierarchy](https://en.wikipedia.org/api/rest_v1/#/Transforms), for example via `/transform/language/{from_lang}/to/{to_lang}`.

v1/translate.yaml
gwicke: I agree with Pchelolo that it does stand out quite a bit, as we are using lowercase for all other fixed identifiers in the API. I can definitely see this to be a source of errors, as many developers expect such constants to be lowercase.

d00rman: I'm in the same boat. Having these capitalised is not the best option. I'd vote for CXServer to be able to accept case-insensitive input and we could then put the lowercase variant in the docs, while allowing clients to use either. @kartikm @Nikerabbit would that work for you guys?

v1/translate.yaml
d00rman: I was thinking about `transform` too. The advantage of putting it below it is that **it is** a type of transform. But then we need to think about the source and destination languages. If I call `https://en.wikipedia.org/api/rest_v1/transform/html/language/...`, then is `en` the source or the destination language? Obviously, because we already have the domain, we don't need two languages, but which of these do we not include?

It's not an easy answer. This is why for this round I opted to have a separate `/translate` hierarchy under the global domain as a first step. I do think we (eventually) need to integrate content translation into the proper domains, but it needs careful consideration.

gwicke: > But then we need to think about the source and destination languages. If I call https://en.wikipedia.org/api/rest_v1/transform/html/language/..., then is en the source or the destination language? Obviously, because we already have the domain, we don't need two languages, but which of these do we not include?

Yeah, this is the "source" vs. "destination" question. For requesting existing content from a wiki in another language, I personally find the "source" variant more intuitive (request to enwiki to get translated enwiki content), but for a translator creating content for a specific target wiki, the other way around makes sense too. For pure transforms (where both input & output can presumably be arbitrary languages), I don't have any intuition at all. So agreed,

> It's not an easy answer. 

What do you think about using "html" as the source type, vs. "machine" or "engine"?

santhoshtr: `/translate/html/{from_lang}/to/{to_lang}` looks good to me. /translate/machine looks bit odd to me too.

Nikerabbit: It doesn't really matter whether it is html or plain-text, we can translate both. In fact the fact that we always wrap the result unconditionally in a div is an anti-feature: https://phabricator.wikimedia.org/T87656

But calling it `html` would allow adding complemeting `text` entry point that does not do the wrapping, thus solving the reported issue, so I too like this suggestion.

v1/translate.yaml
Nikerabbit: -1 but not -2 from me. I don't think there will be many errors, the users should figure out the list of available engines via the listing APIs, not typing them manually. Hence they are not fixed. In fact the enum is wrong and does not list all possible values (which again depend on the language pair). Apertium was used as an example in the old API because it is the only service that does not require an authorization header (which is missing from these docs btw).

See also https://gerrit.wikimedia.org/r/#/c/346991/8/modules/mw.cx.MachineTranslationManager.js for an example where this pattern is taken as the convention.

d00rman: @Nikerabbit I don't understand why you wouldn't allow clients to send the engine name in either case? As I see it there is no difference. And if we can avoid even the small amount of errors, we should.

v1/translate.yaml
d00rman: Sounds good to me. Will move to `/transform`. @Nikerabbit @kartikm @santhoshtr any thoughts on having this under the domains' hierarchies? That is, instead of `https://wikimedia.org/api/rest_v1/transform/html/{src_lang}/to/{dest_lang}` to have `https://en.wikipedia.org/api/rest_v1/transform/html/to/{dest_lang}` ? Concretely, the question is: would it be acceptable to clients to assume that the domain against which the request is made is the source of the translation?

v1/translate.yaml
Nikerabbit: I am not opposed to allow case insensitive matching, and I don't think it causes caching issues here. I'm just saying that my expected behavior is that clients should either fetch the languages from our API and return the same keys byte for byte, or the users defining them in the configuration should also check from our API and copy byte for byte.

kartikm:
@Pchelolo @d00rman What's pending in this PR?

Pchelolo:
So apparently all the dependencies for this were resolved and we can merge it now. @d00rman objections?

d00rman:
According to [this comment](https://phabricator.wikimedia.org/T107914#3192851) we need to expose the segmented HTML as well. I propose to go with `https://{domain}/api/rest_v1/page/segmented/{title}`. Opinions?

Also, given the need for the segmented HTML, I would prefer to perhaps move the API from the global domain into each project's `transform` hierarchy. The logic I'm proposing is to have the project domain be interpreted as the source language, so that `https://en.wp.org/api/rest_v1/transform/html/to/{lang}{/provider}` would mean *translate the supplied text from English to `{lang}`*.

d00rman:
Ping @kartikm @santhoshtr @Nikerabbit @wikimedia/services about ^

Nikerabbit:
1) The api for segmenting looks okay, except it's missing the revision parameter. I think it makes sense to be per-domain.

2) I don't support moving the translation API to per domain, because:
- it would limit the supported languages to the existing Wikipedias we have
- it doesn't make sense to provide the sami API in multiple locations (Wikipedia.org, Wiktionary.org)
- because of history, the domain names in use don't map cleanly to language codes. This inconsistency should not be spread further.

I mean, it would make sense if the only accepted input was a page on that domain, but it is not. It takes any text as an input.

kartikm:
@d00rman Ping for @Nikerabbit 's comment.

d00rman:
> * it would limit the supported languages to the existing Wikipedias we have
> * because of history, the domain names in use don't map cleanly to language codes. This inconsistency should not be spread further.

RESTBase uses [the `siteinfo` API](https://www.mediawiki.org/wiki/API:Siteinfo) to retrieve information about each domain it manages. Amongst other things, it provides the project's language code which can be used as the source language.

> * it doesn't make sense to provide the sami API in multiple locations (Wikipedia.org, Wiktionary.org)

From the perspective of CXServer, maybe (since the language pairs don't change between these projects), but from the overall REST API perspective, being able to translate something on that domain/project is the property of that domain/project only.

> I mean, it would make sense if the only accepted input was a page on that domain, but it is not. It takes any text as an input.

Ok, I think we are talking about two different things here :) You are referring to CXServer's functionality and capabilities and I am looking from the perspective of the API. It is my understanding that CXServer's main purpose is to accompany the ContentTranslation extension, which can function only in the context of a WM/MW project. As such, it can be used to translate articles from one project and import them into another one. Is that assessment correct? (I assume so, otherwise the need for having the segmented end point makes little to no sense, at least to me and will need further explanation.) If so, having project-level APIs is what we really need and should drop the one on the global domain.

If that, however, is only a subset of the ways CXServer **is used** (as opposed to _can be used_), then I think we should have both: project-level ones and on the global domain. That way, if inter-project translation is needed, the extension can easily use the project-level API (through the VRS, which eliminates all need for any special configuration of the extension itself w.r.t. the API). If an external client wishes to simply translate text using a source language that is not available in any project (but is in CXServer), then they can use the global domain.

Note, however, that the focus of the public REST API is on the content and functionality available **for the WM domains and projects**, not external uses. This means that supporting every possible use-case for back-end services is not our primary goal, since for such cases one can always set up an instance somewhere outside the WMF production environment (including Labs).

Nikerabbit:
(Lost my draft reply when I restarted my browser)

I am not sure how much this would simplify CX configuration. With CX the translation always happens in the **target** language wiki. This means we, in your proposed scheme, would need to contact the restbase in the **source** language wiki for segmentation and translation, which still requires a knowledge from CX to map language codes to the project domains (we have this logic already).

There is also Translate extension running on Meta, MediaWiki.org and so on which calls the translation api for the page translation & banner translation features (and where the source language is not always the wiki content language!). It doesn't and shouldn't care about the project domains at all, but if you are okay for keeping the global domain, then it's fine. I'm just dubious about the benefits of having the translation api in the project domain, given it requires all users (except the simplest, limited ones) to implement the logic of mapping language codes to project domains, which would then be immediately mapped back to language codes by the Restbase api when it calls CXServer.

And like I said above, I have no concerns about the segmentation api and support moving it to the project domains.

d00rman:
I have added the `https://{domain}/api/rest_v1/page/segmented/{title}{/revision}` end point. This depends on [Gerrit 360692](https://gerrit.wikimedia.org/r/#/c/360692/).

@Nikerabbit given your comment above, would then make more sense to set it up in the target language? Something like `https://en.wp.org/api/rest_v1/transform/html/from/{lang}{/provider}` ? The reason why I want to have it under the project domains as well is that that makes the API discoverable (as one can see the translation options/end points aggregated under the same documentation page for the project without the need to go to https://wikimedia.org/ap/rest_v1/ nor know about its existence).

d00rman:
Ping @kartikm @santhoshtr @Nikerabbit @wikimedia/services about ^

It has been almost 3 months that we have been bike-shedding about this (rather slowly), so I propose we set a time limit on this discussion and take action. I propose to settle on the open question by the end of this week and proceed with public deployment/announcement next week.

Nikerabbit:
Having the API in the target language wiki would work for CX.

kartikm:
@d00rman Can you check @Nikerabbit 's comment? Let's set deadline by next week to finish this and deploy!

v1/content.yaml
```diff
@@ -890,6 +890,70 @@ paths:
...
+        '400':
+          description: Invalid revision
+          schema:
+            $ref: '#/definitions/problem'
+        '403':
+          description: Access to the specific revision is restricted
+          schema:
+            $ref: '#/definitions/problem'
+        '404':
+          description: Unknown page, revision or tid
```
Pchelolo: Nit: There's no tie parameter here

v1/transform-global.yaml
```diff
@@ -0,0 +1,254 @@
...
+          in: path
+          description: The machine translation provider id
+          type: string
+          required: false
+          enum:
+           - Apertium
+           - Yandex
+           - Youdao
+        - name: html
+          in: formData
```
Pchelolo: If it's `formData` it doesn't really consume `application/json`, it actually consumes `multipart/form-data`. It it's consuming JSON, the `in` should be `body`

v1/transform-global.yaml
```diff
@@ -0,0 +1,254 @@
...
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/list/tool/{tool}'
+  /list/languagepairs/:
+    get:
+      tags: ['Transform']
+      summary: Lists the tools and language pairs available for the given tool category
```
Pchelolo: Copy-paste from previous

v1/transform-lang.js
```diff
@@ -0,0 +1,47 @@
...
+        return mwUtil.getSiteInfo(hyper, req).get('general').get('lang').then(lang => hyper.post({
+            uri: `${this._options.cx_host}/v1/mt/${rp.from}/${lang}/${rp.provider}`,
+            body: req.body
+        }));
+    }
+
+    doDict(hyper, req) {
+        const rp = req.params;
+        return mwUtil.getSiteInfo(hyper, req).get('general').get('lang').then(lang => hyper.get({
+            uri: `${this._options.cx_host}/v1/dictionary/${rp.word}/${rp.from}/` +
```
Pchelolo: Let's URI-encode the word, it might contain some weird characters

v1/transform-lang.yaml
```diff
@@ -0,0 +1,124 @@
...
+          in: path
+          description: The machine translation provider id
+          type: string
+          required: false
+          enum:
+           - Apertium
+           - Yandex
+           - Youdao
+        - name: html
+          in: formData
```
Pchelolo: Same thing about formData and JSON

v1/transform-lang.yaml
```diff
@@ -0,0 +1,124 @@
...
+            $ref: '#/definitions/cx_dict'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      operationId: doDict
+
+definitions:
+  cx_mt:
```
Pchelolo: You can put these to the common definitions spec to avoid repeating it here

d00rman:
This is now ready for review. I added two project-scoped end points, bringing the total number of project routes to three:
- `GET https://{domain}/api/rest_v1/page/segmented/{title}{/revision}`
- `POST https://{domain}/api/rest_v1/transform/html/from/{from}{/provider}`
- `GET https://{domain}/api/rest_v1/transform/word/from/{from}/{word}{/provider}`

If this looks good to you all, we can move on it next week.

d00rman:
@Pchelolo addressed the comments

Pchelolo:
+1

v1/content.yaml
```diff
@@ -890,6 +890,70 @@ paths:
               body: '{{request.body}}'
       x-monitor: false
 
+  /segmented/{title}{/revision}:
```
gwicke: Minor: Perhaps "segments" sounds slightly more consistent with other more noun-oriented names?

v1/transform-global.yaml
```diff
@@ -0,0 +1,227 @@
+info:
+  version: '1.0.0'
+  title: Global Transform API
+  description: Global Transform API
+  termsofservice: https://github.com/wikimedia/restbase
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /html/{from}/to/{to}{/provider}:
```
gwicke: This might be easy to confuse with the similar-sounding `POST /transform/html/to/wikitext{/title}{/revision}` entrypoint. Maybe it would make it a bit clearer to use more specific parameter names like `{from_language}` and `{to_language}`, or name the entire entrypoint something like `/html-lang/{from}/to/{to}{/provider}`.

d00rman: Good point @gwicke . To be consistent with the other transform end points, how about `transform/html/from/{from_lang}/to/{to_lang}{/provider}` ?

gwicke: >  To be consistent with the other transform end points, how about `transform/html/from/{from_lang}/to/{to_lang}{/provider}` ?

The structure is slightly different, as other end points tend to be organized around `/{from_format}/to/{to_format}/`. The new endpoint is not converting the format, but the language within a given format (HTML).

I think the renamed parameters could work. I slightly prefer the look and feel of `/transform/html-lang/{from}/to/{to}{/provider}`, but don't feel too strongly about it. It seems to more clearly communicate what is being transformed (`html-lang`).

d00rman: I agree that putting it under `transform` strongly suggests a format transform (due to historical reasons), but it's still a transformation. If we want to communicate strongly that it's about language, then it would probably need to be moved under `translate` or a similar new hierarchy, but I don't think that is the way to go, as language translation for the same content is a form of transformation.

kartikm:
@d00rman @Pchelolo See: https://phabricator.wikimedia.org/T107914#3455255 - can we get stack trace to debug this?

d00rman:
A bug got through, PR #840 fixes it.

PR #801, opened 2017-04-18
Blacklist: Add pl.wikisource, hy.wp and blacklist a group in frwiki

d00rman:


d00rman:
cc @wikimedia/services

projects/wmf_default.yaml
```diff
@@ -171,9 +150,12 @@ paths:
                       fr.wikipedia.org:
                         - 'Utilisateur:ZéroBot/Log/Erreurs'
                         - 'Utilisateur:SyntaxTerror/Ajouts_du_modèle_Autorité'
+                        - '/^Utilisateur:[\s\S]+[Bb]rouillon'
```
d00rman: This encompasses all user pages that are drafts. But more generally, I think we could blacklist all user pages that have subpages, i.e. `/^Utilisteur:.+\/.+/`. Thoughts?

PR #802, opened 2017-04-19
Report siteinfo fetch failures with 500 status code

gwicke:
We should distinguish invalid titles (returning 400) from siteinfo fetch
failures. The latter are not triggered by a property of the request, so
400 is inappropriate.

This patch catches fetch failures, and turns them into 500 responses.
Invalid titles still return 400.

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -33,15 +33,18 @@ mwUtil.makeETag = (rev, tid, suffix) => {
  * Normalizes the request.params.title and returns it back
  */
 mwUtil.normalizeTitle = (hyper, req, title) =>  mwUtil.getSiteInfo(hyper, req)
-.then(siteInfo => Title.newFromText(title, siteInfo))
-.catch((e) => {
-    throw new HTTPError({
-        status: 400,
-        body: {
-            type: 'bad_request',
-            detail: e.message
-        }
-    });
+.then((siteInfo) => {
+    try {
+        return Title.newFromText(title, siteInfo);
+    } catch (e) {
+        throw new HTTPError({
+            status: 400,
+            body: {
+                type: 'bad_request',
+                detail: e.message
+            }
+        });
+    }
 });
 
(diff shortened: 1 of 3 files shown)


lib/mwUtil.js
```diff
@@ -33,6 +33,15 @@ mwUtil.makeETag = (rev, tid, suffix) => {
  * Normalizes the request.params.title and returns it back
  */
 mwUtil.normalizeTitle = (hyper, req, title) =>  mwUtil.getSiteInfo(hyper, req)
+.catch((e) => {
```
Pchelolo: The SiteInfo is used in various other places, so a better spot to do that is in the action module, here: https://github.com/wikimedia/restbase/blob/master/sys/action.js#L266

gwicke: My concern was that not all uses of that module might use a domain that is expected to exist. If the domain is user-controlled (not configured in RB), then the 500 might not be helpful. Should we just make & document that assumption?

Pchelolo: @gwicke Em... It's a MW specific RB module, how would the domain be not controlled by RB?

gwicke: > @gwicke Em... It's a MW specific RB module, how would the domain be not controlled by RB?

For example, if we fetched the siteinfo for a domain passed in as a path parameter, on some wikimedia.org end point. To my knowledge this is not the case right now & does not seem very likely, so I'm happy to make the assumption that the domain is validated.

Pchelolo: @gwicke I think it's better to think that the domain should exist, since using this for a non-existent (user-provided) domain is an error. Let's document that the expectation is that the domain exist?

gwicke: Yes, that's what I meant with validated / RB controlled. Will add a note that this is the expectation.

Pchelolo:
👏 🎉 👍

PR #803, opened 2017-04-19
Start enforcing Parsoid content versions

gwicke:
Start re-rendering Parsoid content (/html/{title},
/html/{title}/{revision} and data-parsoid) if the stored revision does
not match the API spec.

Note that we do not need to set vary on this response, as we are only
comparing the spec version & the returned version in this filter. We
ignore client-supplied `accept` headers for now.

### DIFF
--- a/lib/ensure_content_type.js
+++ b/lib/ensure_content_type.js
@@ -72,8 +72,9 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
                     expected: expectedContentType,
                     actual: res.headers['content-type']
                 });
-                // Limit response caching to a few seconds
-                res.headers['cache-control'] = 'max-age=10, s-maxage=10';
+                // Disable response caching, as we aren't setting a vary
+                // either.
+                res.headers['cache-control'] = 'max-age=0, s-maxage=0';
             }
         }
         // Default: Just return.
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -211,6 +211,8 @@ paths:
         options:
           allowInlineStyles: true
     get:
+      x-route-filters:
+        - path: lib/ensure_content_type.js
       tags:
         - Page content
       summary: Get latest HTML for a title.
@@ -481,6 +483,8 @@ paths:
         options:
           allowInlineStyles: true
     get:
+      x-route-filters:
+        - path: lib/ensure_content_type.js
       tags:
         - Page content
       summary: Get HTML for a specific title/revision & optionally timeuuid.
@@ -601,6 +605,8 @@ paths:
           redirect_cache_control: '{{options.purged_cache_control}}'
           attach_body_to_redirect: true
     get:
+      x-route-filters:
+        - path: lib/ensure_content_type.js
       tags:
         - Page content
       summary: Get data-parsoid metadata for a specific title/revision/tid.


gwicke:
For discussion @wikimedia/services

Pchelolo:
+1 but have to talk to @subbuss about elevated Parsoid load.

berndsi:
Would be great to have for mobile-sections, too.

Pchelolo:
kk @berndsi I'll make a separate PR for mobile tomorrow morning. In the meantime, could you double-check which exact content type MCS emits for the mobile-sections content?

gwicke:
I looked at the mobile section content types, and they currently all return the same content-type with a profile pointing to https://www.mediawiki.org/wiki/Specs/mobile-sections/0.9.3. The actual JSON formats differ between entry points, so to avoid claiming the same content type for different JSON responses I would propose to differentiate the content types first, with one profile per actual JSON structure. Following the Parsoid lead, the version number can be shared (0.9.3 in this case), but you might want to vary the `mobile-sections` part for each structure. Some documentation (ex: a link) would be great at the profile URL.

berndsi:
@gwicke The latest version is actually `0.10.0`. 
(content-type with a profile pointing to https://www.mediawiki.org/wiki/Specs/mobile-sections/0.10.0), as defined in 
https://phabricator.wikimedia.org/diffusion/GMOA/browse/master/lib/mobile-util.js;ba3ad5b83b357bc0ad6181db7f4b1da74b7b1f23$10.
That version was deployed on April 10.

So, I should create https://www.mediawiki.org/wiki/Specs/mobile-sections/0.10.0, https://www.mediawiki.org/wiki/Specs/mobile-sections-lead/0.10.0, and https://www.mediawiki.org/wiki/Specs/mobile-sections-remaining/0.10.0?
I plan to bump version numbers more often in the future to follow semver (whenever anything changes I'd bump the patch number, a bit more the the subminor version).

gwicke:
@berndsi, I guess me looking at fairly old stored content illustrates why this matters ;)

If all three formats are different (only looked at the first two closely), then yes it would make sense to assign different types. If maintaining seperate pages is too much work, you could perhaps also link to a section anchor within a shared doc, but I would recommend to keep the version, and document changes from the previous version on that page.

Regarding version update frequency: I think this depends on what you really need to distinguish / expect in consumers. https://www.mediawiki.org/wiki/API_versioning#Background:_How_we_interpret_semver_in_format_negotiations describes how we plan to interpret semver minor versions. For client breaking changes, you should increment the major version. To avoid too many re-renders on minor versions, we intend to ignore those. However, this is not yet implemented. Currently, we re-render whenever the stored version is less than the expected version.

lib/ensure_content_type.js
```diff
@@ -72,8 +72,9 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
                     expected: expectedContentType,
                     actual: res.headers['content-type']
                 });
-                // Limit response caching to a few seconds
-                res.headers['cache-control'] = 'max-age=10, s-maxage=10';
+                // Disable response caching, as we aren't setting a vary
+                // either.
+                res.headers['cache-control'] = 'max-age=0, s-maxage=0';
```
d00rman: I don't think we need this. Version bumps do not happen that often to warrant this.

gwicke: This is not related to the frequency of version bumps. It's just about making sure that failed upgrades don't get stuck in caches.

d00rman:
I think this is a good move for latest revisions, but I'm not able to convince myself whether we should do it for all revisions. The moment that we re-render a revision because of the version content, we will likely not be serving the HTML that was valid at the time of the update.

> +1 but have to talk to @subbuss about elevated Parsoid load.

The load should not be a problem. Version bumps happen rarely, pertain only to pages that haven't been updated since the version bump and such requests are processed in the cluster serving live traffic.

lib/ensure_content_type.js
d00rman: They can get stuck only for a finite amount of time, and the frequency (and possibility) of that happening is rather low, so why disable caching altogether?

gwicke: Overriding caching in a filter is always a bit tricky, since we might actually end up unintentionally *extending* the cache period in some cases. Avoiding that would require parsing existing cache-control headers, which adds complexity. Given that such upgrade failures are error conditions, should be very rare, and we can't set a very long TTL anyway, it does not seem to be worth the complexity or risk to me.

gwicke:
> I think this is a good move for latest revisions, but I'm not able to convince myself whether we should do it for all revisions. The moment that we re-render a revision because of the version content, we will likely not be serving the HTML that was valid at the time of the update.

The goal here is to give clients predictable content to work with. You are right that a few prominent use cases only use the "latest" end point, but there are also other important ones (like VisualEditor, HTML dumps, external RC followers) that request content by revision. Typically, it is actually the latest revision they are explicitly requesting.

I think your concern about historical templates being re-rendered is not that critical as long as we are only offering fairly recent revisions. I agree that it will be an important consideration once we actually archive, but by then we should have the Parsoid html2html end point for version updates, and won't be using the simple re-render strategy implemented in this filter.

gwicke:
Wrong button..

PR #805, opened 2017-04-26
Store new version if only the content-type changed

Pchelolo:
After starting to rerender on the content-type mismatch #803 we need to store the new version if only the content-type has changed but HTML is still the same to avoid infinite rerendering on every request for pages where the content-type bump doesn't affect the HTML. 

cc @wikimedia/services


Pchelolo:
Need to merge this before deploying #804 and #805

d00rman:
:100:

PR #806, opened 2017-04-26
Only update and purge the summary if it actually changed

Pchelolo:
Bug: https://phabricator.wikimedia.org/T163924
cc @wikimedia/services


sys/key_value.js
```diff
@@ -194,6 +195,32 @@ class KVBucket {
...
             return { status: 400 };
         });
+
+        if (req.query.ignore_duplicates) {
+            return hyper.get({
+                uri: new URI([rp.domain, 'sys', 'key_value', rp.bucket, rp.key])
+            })
+            .then((oldContent) => {
+                if (stringify(req.body) === stringify(oldContent.body) &&
+                    (!req.headers['content-type']
```
d00rman: Nit: let's indent this one more level so as not to confuse it with the statements inside the `if` block

sys/key_value.js
```diff
@@ -194,6 +195,32 @@ class KVBucket {
...
+                        }
+                    };
+                } else {
+                    throw new HTTPError({
+                        status: 404
+                    });
+                }
+            })
+            .catch({ status: 404 }, doPut);
+        } else {
```
d00rman: Nit: no need for `else` here, as there is a `return` inside the `if`

d00rman:
Some nits, but otherwise good work! :+1:

Pchelolo:
@d00rman Updated!

sys/key_value.js
```diff
@@ -194,6 +195,29 @@ class KVBucket {
...
+            return hyper.get({
+                uri: new URI([rp.domain, 'sys', 'key_value', rp.bucket, rp.key])
+            })
+            .then((oldContent) => {
+                if (stringify(req.body) === stringify(oldContent.body) &&
+                        (!req.headers['content-type']
+                        || req.headers['content-type'] === oldContent.headers['content-type'])) {
+                    hyper.metrics.increment(`sys_kv_${req.params.bucket}.unchanged_rev_render`);
+                    return {
+                        status: 200,
```
gwicke: According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6, 204 (no content) might be more appropriate here.

Pchelolo: @gwicke Done!

v1/definition.yaml
```diff
@@ -123,7 +123,18 @@ paths:
             request:
               method: put
               uri: /{domain}/sys/key_value/term.definition/{request.params.term}
-              headers: '{{extract.headers}}'
+              headers: '{{merge({"if-none-hash-match": "*"}, extract.headers)}}'
+              body: '{{extract.body}}'
+            # With the ignore_duplicates parameter the storage will return 200 if the content is not changed,
```
d00rman: *With the if-none-hash-match header the storage will return 204*...

v1/summary.yaml
```diff
@@ -139,21 +139,30 @@ paths:
...
                 description: '{{extractDescription(extract.body.items[0].terms)}}'
                 coordinates: '{{processCoords(extract.body.items[0].coordinates)}}'
-
+        - store_and_return:
+            request:
+              method: put
+              uri: /{domain}/sys/key_value/summary/{request.params.title}
+              headers: '{{merge({"if-none-hash-match": "*"}, extract.headers)}}'
+              body: '{{extract.body}}'
+            # With the ignore_duplicates parameter the storage will return 200 if the content is not changed,
```
d00rman: Idem as above

v1/definition.yaml
Pchelolo: Errr.... Fixed.

PR #808, opened 2017-04-27
Return 412 when the if-none-hash-match condition fails

gwicke:


gwicke:
@wikimedia/services

gwicke:
ping @wikimedia/services

PR #818, opened 2017-05-23
Implemented a naive content transfomation filter

Pchelolo:
A pretty naive filter to work as a foundation for later potential generalization.

Bug: https://phabricator.wikimedia.org/T166163
cc @wikimedia/services

### DIFF
--- a/v1/summary_transform_filter.js
+++ b/v1/summary_transform_filter.js
@@ -4,6 +4,7 @@ const cType = require('content-type');
 const semver = require('semver');
 
 const TAGS_MATCH = /<\/?[a-zA-Z][\w-]*(?:\s+[a-zA-Z_-]+(?:=(?:"[^"]*"|'[^']*'))?)*\s*\/?>/g;
+const LATEST_1_VERSION = '1.1.2';
 
 function splitProfile(profile) {
     const match = /^(.*)\/([0-9.]+)$/.exec(profile);
@@ -29,18 +30,18 @@ module.exports = (hyper, req, next) => {
         // for compatibility with older clients.
         // TODO: remove eventually and return the latest content-type by default
         // Normally we would set the latest version here
-        acceptVersion = '1.1.2';
+        acceptVersion = LATEST_1_VERSION;
     }
 
     return next(hyper, req)
     .then((res) => {
         if (semver.lt(acceptVersion, '2.0.0')) {
-            res.body.extract = res.body.extract.replace(/<[^>]+\/?>/g, '');
+            res.body.extract = res.body.extract.replace(TAGS_MATCH, '');
             res.headers['content-type'] = res.headers['content-type']
-                .replace(/[0-9.]+"$/, `${acceptVersion}"`);
+                .replace(/[0-9.]+"$/, `${LATEST_1_VERSION}"`);
         }
 
-        res.headers.vary = 'Accept';
+        res.headers.vary = 'accept';
         return res;
     });
 };
--- a/v1/summary_transform_filter.js
+++ b/v1/summary_transform_filter.js
@@ -3,6 +3,8 @@
 const cType = require('content-type');
 const semver = require('semver');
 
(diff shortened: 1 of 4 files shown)


v1/summary_transform_filter.js
```diff
@@ -0,0 +1,45 @@
...
+        // for compatibility with older clients.
+        // TODO: remove eventually and return the latest content-type by default
+        // Normally we would set the latest version here
+        acceptVersion = '1.1.2';
+    }
+
+    return next(hyper, req)
+    .then((res) => {
+        if (semver.lt(acceptVersion, '2.0.0')) {
+            res.body.extract = res.body.extract.replace(/<[^>]+\/?>/g, '');
```
gwicke: HTML attribute values can contain literal `>` characters, so we'll have to do a bit more work to remove HTML tags. The following pattern borrowed from [web-html-stream](https://github.com/wikimedia/web-html-stream/blob/master/lib/index.js#L10-L11) should work for HTML5 serialization: `/<\/?[a-zA-Z][a-zA-Z0-9_-]*(?:\s+[a-zA-Z_-]+(?:=(?:"[^"]*"|'[^']*'))?)*\s*\/?>/g`

v1/summary_transform_filter.js
```diff
@@ -0,0 +1,45 @@
...
+            res.headers['content-type'] = res.headers['content-type']
+                .replace(/[0-9.]+"$/, `${acceptVersion}"`);
```
gwicke: Would it make sense to return the highest 1.x version instead?

Pchelolo: Em, we should return what the client requested shouldn't we?

v1/summary_transform_filter.js
Pchelolo: Updated.

v1/summary_transform_filter.js
gwicke: Normally, the policy is to return the latest content matching the requested major version. This means definitely returning the latest minor version, and typically (but not always) the latest patch version. So far the policy only guaranteed returning the latest minor version compatible with the requested major version, and left the patch version handling unspecified. I just went ahead & made this more explicit [in the API versioning policy](https://www.mediawiki.org/wiki/API_versioning#Background:_How_we_interpret_semver_in_format_negotiations).

Pchelolo:
Updated yet again to address the comments

gwicke:
LGTM overall, but lets wait until next Thursday's meeting to figure out

- whether we will return 1.x by default in response to requests without an explicit `accept` header, and
- whether we should also include the title-related changes discussed in https://phabricator.wikimedia.org/T164291.

v1/summary_transform_filter.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+
+const cType = require('content-type');
+const semver = require('semver');
+
+const TAGS_MATCH = /<\/?[a-zA-Z][\w-]*(?:\s+[a-zA-Z_-]+(?:=(?:"[^"]*"|'[^']*'))?)*\s*\/?>/g;
+const LATEST_1_VERSION = '1.1.2';
+
+function splitProfile(profile) {
+    const match = /^(.*)\/([0-9.]+)$/.exec(profile);
```
d00rman: Why `.*` ? `.+` seems more prudent since /` is required by the RegExp before the version

v1/summary_transform_filter.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+
+const cType = require('content-type');
+const semver = require('semver');
+
+const TAGS_MATCH = /<\/?[a-zA-Z][\w-]*(?:\s+[a-zA-Z_-]+(?:=(?:"[^"]*"|'[^']*'))?)*\s*\/?>/g;
```
d00rman: A comment documenting the RegExp would be highly appreciated ;)

Pchelolo:
It's been decided to proceed on this by adding a separate `extract_html` field. Closing this PR.

PR #820, opened 2017-06-01
Added html_extract property to the summary endpoint

Pchelolo:
Per discussion on the meeting apps are not interested in migrating to the HTML summaries, so I've added a new `html_extract` property to the summary response.

Size computation for feed response:

Feed plain text -  58K
Feed with html - 94K
Feed plain text gzipped - 18K
Feed with html gzipped - 21K

So for traffic sizes the difference is negligible. 

This also bumps the content type version starting an automatic storage update, but you might still see cached versions from Varnish for 14 more days.

Bug: https://phabricator.wikimedia.org/T165017

### DIFF
--- a/v1/common_schemas.yaml
+++ b/v1/common_schemas.yaml
@@ -64,7 +64,7 @@ definitions:
       extract:
         type: string
         description: First several sentences of an article in plain text
-      html_extract:
+      extract_html:
         type: string
         description: First several sentences of an article in simple HTML format
       thumbnail:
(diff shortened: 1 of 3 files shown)


gwicke:
:+1:

berndsi:
Thanks for the quick turnaround. Just wanted to mention that the new field is called `extract_html` (not `html_extract` as the the first message indicates).

Pchelolo:
@berndsi Ye, I've written the PR title first and then renamed the field..

PR #821, opened 2017-06-02
Summary: Ensure automatic monitoring checks for the HTML extract

d00rman:
This is a follow-up on PR #820

Bug: [T165017](https://phabricator.wikimedia.org/T165017)


d00rman:
cc @wikimedia/services

PR #825, opened 2017-06-12
Add Citoid and MobileApps paths; release v0.16.7

d00rman:


d00rman:
cc @wikimedia/services

config.example.yaml
```diff
@@ -60,6 +60,16 @@ services:
...
+                  citoid:
+                    host: http://citoid-beta.wmflabs.org
+                  mathoid:
+                    host: http://mathoid-tester.wmflabs.org
+                  mobileapps:
+                    host: http://appservice.wmflabs.org
+                  # 10 days Varnish caching, one day client-side
+                  purged_cache_control: s-maxage=864000, max-age=86400
+                  # Cache control for purged endpoints allowing short-term client caching
+                  purged_cache_control_client_cache: s-maxage=1209600, max-age=300
```
Pchelolo: Hm... We don't have Varnish in MW-containers and we unlikely will, so I don't think having purged cache-control makes any sense here

d00rman: Right, but we have it in the projects, so better to keep it for compatibility.

gwicke: Unless we know that no external (and un-purged) proxies see the s-maxage header, it might be better to set the value to 0. This way, we can be sure that there is no surprising caching in transparent proxies when the wiki is accessed via HTTP.

d00rman: Good idea. Set both `s-maxage` params to `0`.

d00rman:
Published as [v0.16.7](https://github.com/wikimedia/restbase/tree/v0.16.7)

PR #826, opened 2017-06-12
Update dependencies

Pchelolo:
Solves the peer dependency issue that didn't allow installing dependencies on node 4.

cc @wikimedia/services


package.json
```diff
@@ -37,35 +37,35 @@
...
-    "mocha-lcov-reporter": "^1.2.0",
-    "mocha.parallel": "^0.14.0",
-    "nock": "^9.0.2",
+    "mocha-lcov-reporter": "^1.3.0",
+    "mocha.parallel": "^0.15.2",
+    "nock": "^9.0.13",
     "preq": "^0.5.2",
     "restbase-mod-table-sqlite": "^0.1.20",
-    "swagger-test": "0.5.1"
+    "swagger-test": "0.7.2"
```
d00rman: This should be reverted to 0.5.1 as that version is known to work for us.

PR #828, opened 2017-06-20
Sample version mis-match warning 1:1000

gwicke:
After a backend service is upgraded, we tend to get a burst of "need to
update the spec" warnings. We don't really need to see each of those
warnings, so this patch introduces significant sampling at 1:1000.

I also changed the log message slightly to make it clear which part
needs updating.

### DIFF
--- a/lib/ensure_content_type.js
+++ b/lib/ensure_content_type.js
@@ -33,12 +33,16 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
 
         function updateSpecWarning() {
             // Returned content type version is higher than what the
-            // spec promises. Log a warning about the need to update the spec.
-            hyper.log('warn/content-type/spec_outdated', {
-                msg: 'Spec needs update to reflect latest content type.',
-                expected: expectedContentType,
-                actual: res.headers['content-type']
-            });
+            // spec promises.
+            if (Math.random() < 0.001) {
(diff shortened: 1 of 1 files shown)


lib/ensure_content_type.js
```diff
@@ -33,12 +33,16 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
...
-            // spec promises. Log a warning about the need to update the spec.
-            hyper.log('warn/content-type/spec_outdated', {
-                msg: 'Spec needs update to reflect latest content type.',
-                expected: expectedContentType,
-                actual: res.headers['content-type']
-            });
+            // spec promises.
+            if (Math.random() < 0.001) {
+                // Log a warning about the need to update the spec. Sampled 1:1000.
+                hyper.log('warn/content-type/spec_outdated', {
```
Pchelolo: Maybe better to decrease the level to `INFO` and make sampling use our standard sampled_levels through the config?

gwicke: An outdated spec is not normal & requires attention, so I think a warning is the best way to ensure that users interested in warnings see this. 

Sampled_levels on the other hand is an opt-in mechanism that lets us log specific items from lower-priority levels, which would not otherwise be logged. We shouldn't rely on users setting that up for specific messages just to see warnings.

Pchelolo: ok, makes sense..

PR #829, opened 2017-06-21
Unblacklist one page from commons to check how automatic black listing work

Pchelolo:
To be deployed after https://github.com/wikimedia/change-propagation/pull/189

cc @wikimedia/services


gwicke:
LGTM.

eevans:
👍

PR #832, opened 2017-06-29
Protect special characters in PDF render backend requests

gwicke:
Since we are passing a URL in a query string, we need to double-escape
special characters to ensure the preservation of special characters like
question marks.

This patch uses the existing, but internal _encodeURIComponent function.
This is primarily for expediency. We could also consider promoting
encodeURIComponent to a public function in swagger-router, and then
using that version.

Bug: T169223

### DIFF
--- a/v1/pdf.yaml
+++ b/v1/pdf.yaml
@@ -60,7 +60,8 @@ paths:
         - get_pdf_from_backend:
             request:
               method: get
-              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url=https://{{domain}}/wiki/{title}%3Fprintable=yes'
+              # Note: The title needs to be encoded twice.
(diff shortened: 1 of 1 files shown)


Pchelolo:
+1

v1/pdf.yaml
```diff
@@ -60,7 +60,8 @@ paths:
         - get_pdf_from_backend:
             request:
               method: get
-              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url=https://{{domain}}/wiki/{title}%3Fprintable=yes'
+              # Note: The title needs to be encoded twice.
+              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url=https://{{domain}}/wiki/{_encodeURIComponent(title)}%3Fprintable=yes'
```
d00rman: @gwicke why `_` in front of `encodeURIComponent()` ?

gwicke: That is how it is exposed in swagger-router. The commit message mentions this.

d00rman: Oh right, sorry about the noise.

PR #834, opened 2017-07-06
Add cassandra 3 to travis to check it out

Pchelolo:
Let's check if our current stuff works with Cassandra 3

NOT WORKING YET! Probably will have to fetch travis docker images and try to debug locally first.

cc @wikimedia/services


eevans:
LGTM (but you probably want to squash some of those changesets :))

Pchelolo:
Ye,  we definitely need to squash :)

PR #835, opened 2017-07-14
Add the recommendation public api specification

schana:
@d00rman From what I can tell, this seems to be the next step in making the service publicly accessible, but I wasn't able to find much documentation on adding something to RESTBase. Could you review and/or point me in the right direction? Thanks

https://phabricator.wikimedia.org/T170877


d00rman:
Closing in favour of PR #849

PR #836, opened 2017-07-17
Add multi-segments bucket

Pchelolo:
Multi-segment bucket created, parsed and mobile converted.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -28,7 +28,7 @@ default_project: &default_project
               domains: /./
         parsoid:
           host: http://parsoid-beta.wmflabs.org
-          time_to_live: 84600
+          time_to_live: 1
         action:
           apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
           baseUriTemplate: "{{'https://{domain}/api/rest_v1'}}"
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -33,6 +33,8 @@ default_project: &default_project
           cache_control: s-maxage=86400, max-age=86400
         citoid:
           host: http://citoid-beta.wmflabs.org
+        recommendation:
+          host: https://recommendation-api-beta.wmflabs.org
         # 10 days Varnish caching, one day client-side
         purged_cache_control: s-maxage=864000, max-age=86400
         # Cache control for purged endpoints allowing short-term client caching
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -17,7 +17,7 @@ default_project: &default_project
             - name: default.group.local
               domains: /./
         table_ng:
-          backend: cassandra
+          backend: cassandra-ng
           hosts: [localhost]
           keyspace: system
           username: cassandra
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -16,6 +16,16 @@ default_project: &default_project
           storage_groups:
             - name: default.group.local
               domains: /./
+        table_ng:
+          backend: cassandra
+          hosts: [localhost]
+          keyspace: system
+          username: cassandra
+          password: cassandra
+          defaultConsistency: one # or 'localQuorum' for production
+          storage_groups:
+            - name: test.group.local
+              domains: /./
         parsoid:
           host: http://parsoid-beta.wmflabs.org
         action:
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -28,6 +28,7 @@ default_project: &default_project
               domains: /./
         parsoid:
           host: http://parsoid-beta.wmflabs.org
+          time_to_live: 84600
         action:
           apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
           baseUriTemplate: "{{'https://{domain}/api/rest_v1'}}"
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "bluebird": "^3.5.0",
     "cassandra-uuid": "^0.0.2",
     "restbase-mod-table-cassandra": "^0.11.2",
-    "restbase-mod-table-cassandra-ng": "git@github.com:wikimedia/restbase-mod-table-cassandra.git",
+    "restbase-mod-table-cassandra-ng": "git@github.com:wikimedia/restbase-mod-table-cassandra.git#redesign",
     "service-runner": "^2.3.0",
     "json-stable-stringify": "git+https://github.com/wikimedia/json-stable-stringify#master",
     "content-type": "git+https://github.com/wikimedia/content-type#master",
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "bluebird": "^3.5.0",
     "cassandra-uuid": "^0.0.2",
     "restbase-mod-table-cassandra": "^0.11.2",
+    "restbase-mod-table-cassandra-ng": "git@github.com:wikimedia/restbase-mod-table-cassandra.git",
     "service-runner": "^2.3.0",
     "json-stable-stringify": "git+https://github.com/wikimedia/json-stable-stringify#master",
     "content-type": "git+https://github.com/wikimedia/content-type#master",
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "mocha.parallel": "^0.15.2",
     "nock": "^9.0.13",
     "preq": "^0.5.2",
+    "restbase-mod-table-sqlite-ng": "^0.1.20",
     "restbase-mod-table-sqlite": "^0.1.20",
     "swagger-test": "0.5.1"
   },
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "bluebird": "^3.5.0",
     "cassandra-uuid": "^0.0.2",
     "restbase-mod-table-cassandra": "^0.11.2",
-    "restbase-mod-table-cassandra-ng": "git@github.com:wikimedia/restbase-mod-table-cassandra.git#redesign",
+    "restbase-mod-table-cassandra-ng": "git+https://github.com/wikimedia/restbase-mod-table-cassandra#redesign",
     "service-runner": "^2.3.0",
     "json-stable-stringify": "git+https://github.com/wikimedia/json-stable-stringify#master",
     "content-type": "git+https://github.com/wikimedia/content-type#master",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "0.16.7",
+  "version": "0.16.8",
   "description": "REST storage and service dispatcher",
   "main": "index.js",
   "scripts": {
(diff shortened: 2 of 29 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "bluebird": "^3.5.0",
     "cassandra-uuid": "^0.0.2",
     "restbase-mod-table-cassandra": "^0.11.1",
+    "restbase-mod-table-cassandra-ng": "git@github.com:wikimedia/restbase-mod-table-cassandra.git",
```
eevans: I think this git url needs the branch name, too (ala #redesign)

projects/wmf_default.yaml
```diff
@@ -96,6 +96,14 @@ paths:
                 - path: sys/table.js
                   options:
                     conf: '{{options.table}}'
+            /table-ng: &sys_table
```
d00rman: `s/sys_table/sys_table_ng/`

Pchelolo:
I've created this WIP PR to discuss the general approach. The idea is that the new bucket (currently called `parsed_bucket`, but we will generalize it to `multi_content_bucket` or something. So it's got the API that allows to fetch individual types of content separately, but for storage you can only save all types of content simultaneously. Also the version of the previously stored content will be provided as a header. 

Listings are supported currently to make the tests pass, but they actually don't make a lot of sense since normally the listing should contain only a single revision/render. So we can either continue supporting `/html/{title}/` endpoint for debugging purposes or just drop it completely. 

One more thing - for non-latest-revision support I propose to use existing stashing tables.

sys/parsoid.js
```diff
@@ -225,6 +225,17 @@ class ParsoidService {
         return new URI(path);
     }
 
+    getNGBucketURI(rp, format, tid) {
+        const path = [rp.domain, 'sys', 'parsoid_bucket', `${format}`, rp.title];
```
d00rman: `s/'${format}'/format/` ?

sys/parsoid_bucket.js
```diff
@@ -0,0 +1,241 @@
...
+        }
+    };
+}
+
+class ParsoidBucket {
+    createBucket(hyper, req) {
+        const rp = req.params;
+        return P.all([
+            hyper.put({
+                uri: new URI([rp.domain, 'sys', 'table-ng', 'html-ng']),
```
d00rman: I am not sure we should suffix the names of the tables with `-ng`. Ideally we want something that does not clash with the current naming scheme (which allows easier upgrades for other users), but still something that makes sense. `parsoid-html` perhaps?

d00rman: Oh, and `table-ng` should probably be made a constant and used as such, since once we switch all of the use cases to the new storage design `/sys/table/` will no longer be used and we will be able to safely rename this one.

sys/parsoid_bucket.js
```diff
@@ -0,0 +1,241 @@
...
+                        { attribute: 'key', type: 'hash' },
+                        { attribute: 'tid', type: 'range', order: 'desc' },
+                    ]
+                }
+            })
+        ]).thenReturn({ status: 201 });
+    }
+
+    makeSchema(opts) {
+        const schemaVersionMajor = 2;
```
d00rman: This can be set to `1` as these are new tables.

gwicke:
One small remark: It might be less ambiguous longer term to use version numbers (table2 or table_2) instead of -ng.

Pchelolo:
@gwicke There's separate underlying tables

gwicke:
> There's separate underlying tables

In that case, how does the backend know about which of the elements is primary for consistency purposes?

Pchelolo:
@gwicke In this version is just knows, but in the generalized version of the bucket (which comes later) it will be configurable.

sys/parsoid_bucket.js
```diff
@@ -0,0 +1,418 @@
...
+                },
+                limit: 1,
+                proj: [ 'rev', 'tid']
+            }
+        }).catchReturn({ status: 404 }, undefined)
+        .then((res) => {
+            if (!res || !res.body.items.length) {
+                // Noting was ever there - put the first render and no need to update the index
+                return createContentStoreRequests(hyper, req,rev, tid);
+            } else if (res && res.body.items.length && res.body.items[0].rev < rev) {
```
d00rman: Simpler `else if (res.body.items[0].rev < rev)`

config.example.wikimedia.yaml
```diff
@@ -16,6 +16,16 @@ default_project: &default_project
           storage_groups:
             - name: default.group.local
               domains: /./
+        table_ng:
```
gwicke: Perhaps use table_2?

sys/parsoid.js
```diff
@@ -344,7 +392,9 @@ class ParsoidService {
                         body: res.body[format].body
                     };
                     resp.headers.etag = mwUtil.makeETag(rp.revision, tid);
-                    return this.saveParsoidResult(hyper, req, format, tid, res)
+                    return this.saveParsoidResultToLatest(hyper, req, tid, res)
+                    .catch({ status: 412 }, () =>
```
gwicke: Where is this 412 coming from, and what is it based on? The request in saveParsoidResultToLatest does not look conditional.

Pchelolo: @gwicke So the 412 was the closes one to what we need to indicate here. It's coming from here: https://github.com/wikimedia/restbase/pull/836/files#diff-2505eb75397a4cf81edd753431f60036R369

This is the case where we're trying to store something old to the `latest_and_recent` table. So it's kinda a "Precondition failed". If you have an idea of how to better communicate that info - that would be great.

gwicke: Thanks for the pointer. I was somehow assuming that this was hitting the table storage backend, and didn't find the 412 there.

Regarding the interface, I agree that 412 kind of makes sense here. It just seems surprising without making a conditional request. The condition we are after is "if rev > previous rev". None of the existing conditional request headers fit that squarely, but perhaps it would still be clearer to  communicate the condition in the request. The best I can come up with using existing conditions would be `if-none-match: W/"12345"`: Only store if existing content does not *weakly* match rev 12345. I guess another option would be to create a new header, for example `if-version-lt: 12345`.

For some of these more specialized buckets we could also consider calling them directly as modules, rather than making HTTP calls to them through the router. This might be a bit easier to understand, and would avoid some routing overhead. Main loss would be less detailed metrics about internal bucket requests.

Pchelolo: > The best I can come up with using existing conditions would be if-none-match: W/"12345": Only store if existing content does not weakly match rev 12345. I guess another option would be to create a new header, for example if-version-lt: 12345.

The problem with this idea is that on this level we do not know the value of "12345" It's only know inside the `parsoid_bucket`...

> For some of these more specialized buckets we could also consider calling them directly as modules, rather than making HTTP calls to them through the router. 

The `parsed_bucket` is indeed kinda a submodule of `parsed` module, so library-style call could make a lot of sense here. But we are going to generalize the `parsed_bucket` to reuse it elsewhere, in that sense the library-style call is worse... 

Maybe replace 412 with 400? Like "You're trying to put the latest stuff but the revId is not actually latest so it's illegal"?

gwicke:
From a correctness perspective, the main thing that stands out to me is that requests for old, superseded revisions can be served from the current table storage, without guarantees about remaining available after the initial request. Even without probabilistic deletions, previous revisions can remain in the current table indefinitely if edits are consistently spaced out more than 10 days apart (revision-timeline TTL). Even with more frequent edits and without probabilistic deletions, previous revisions are always retained until the next edit happens, at which point they are likely to be deleted. From a user perspective, this means that edits based off old revisions will now have a decent chance of failing.

sys/parsoid.js
gwicke: > The problem with this idea is that on this level we do not know the value of "12345" It's only know inside the parsoid_bucket...

The number we need to know here is the revision we are about to store, which I believe we know.

Pchelolo:
> From a correctness perspective, the main thing that stands out to me is that requests for old, superseded revisions can be served from the current table storage, without guarantees about remaining available after the initial request.

Yes, this is issue number 2 that I was talking about on the sync up. We have 2 ways of dealing with that issue: static column or an additional read from the index. Both variants are really easily implementable and they are very self-contained in the getRevision method of the `parsoid_bucket` module. I can implement one of them right now and we can test that, but we wanted to first make a run without any of them to check for baseline performance. We're not extensively testing the read pipeline in the dev cluster right now anyway. If you think it's better to go with one of the options right away - I can update the PR

sys/parsoid.js
Pchelolo: Hm... So the standard `If-Match/If-None-Match` doesn't fit because of the non-strict matching... Custom header is a possible solution indeed, but I'm not sure what do we gain from it?

Other response status candidate is:

> **409 Conflict**
Indicates that the request could not be processed because of conflict in the request, such as an edit conflict between multiple simultaneous updates.

gwicke:
@Pchelolo, I think the main advantage of finishing the design & implementation as far as possible will be to avoid the need for a lot of repeat testing & review. That said, I know that there are a lot of competing priorities right now as well. Perhaps we can at least finalize the design & make sure that we have covered all bases in theory?

d00rman:
I believe the portion of requests that might fail due to this unaddressed issue is pretty low, since there is only one scenario (that I can think of) in which that happens:

1. user A starts an edit of a non-latest revision
1. `24h - delta` time units after that, user B starts an edit of the same revision
1. user B attempts a save after more than `delta` time units

To my knowledge, users editing the same revision usually happens during edit wars, and those usually refer to the latest revision and happen in a time span much lower than 24h. I do agree with the sentiment that

> [...] the main advantage of finishing the design & implementation as far as possible will be to avoid the need for a lot of repeat testing & review

However, I question whether either of the current ideas we are floating should be employed given the extreme edge case in which this issue presents itself. Concretely, @eevans expressed some concerns wrt using a static column, while introducing an additional read to the index table hurts performance for all of the common cases.

gwicke:
> I believe the portion of requests that might fail due to this unaddressed issue is pretty low, since there is only one scenario (that I can think of) in which that happens

Old revisions read from the "current" table can be deleted at any time as part of background updates. This means that the TTL protections are essentially gone for a large portion (likely most) edits based off old revisions.

Pchelolo:
> Old revisions read from the "current" table can be deleted at any time as part of background updates. This means that the TTL protections are essentially gone for a large portion (likely most) edits based off old revisions.

The only case when that happens is if someone edits an old revision, and then 24ish hours later someone edits the same old revision again. It doesn't seem very likely. But as I've wrote - we can achieve 100% correctness with a tiny little function in the storage module. We just don't need to do that right now as we can't really test it in dev cluster.

Pchelolo:
Ok, if we wanna start switching mobile apps early next week we need to merge this first before I'm able to generalize this and write a PR for mobile apps

projects/wmf_default.yaml
```diff
@@ -114,17 +114,26 @@ paths:
...
                     conf: '{{options.table_ng}}'
             /parsoid_bucket:
               x-modules:
-                - path: sys/parsoid_bucket.js
+                - path: sys/multi_content_bucket.js
                   options:
                     time_to_live: '{{default(options.parsoid.time_to_live, 84600)}}'
                     delete_probability: '{{default(options.parsoid.delete_probability, 1)}}'
+                    table_name_prefix: parsoid
+                    main_content_type:
```
eevans: Complete bike-shedding, feel free to ignore, but how about s/main/primary/?

eevans: > Complete bike-shedding, feel free to ignore, but how about s/main/primary/?

Or not...given how pervasive the term is throughout the code; Perhaps not worth the effort.

d00rman:
Overall, LGTM :+1: Very nice separation and generalisation of the bucket. I think we can start testing with this now.

PR #837, opened 2017-07-18
Disable sending previous content to Parsoid

Pchelolo:
During the transition period to the new storage we want the pipeline to be as simple as possible. As the previous content is not used in Parsoid yet, let's just not send it for now. 

We can resurrect that later with great caution, as I think that with our storage model the previous versions should still be available, but for now it takes away square from my storage diagram and makes it safer to reason about the pipeline. 

Bug: https://phabricator.wikimedia.org/T170997

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -264,7 +264,7 @@ class ParsoidService {
         const reqRevision = rp.revision;
         // Helper for retrieving original content from storage & posting it to
         // the Parsoid pagebundle end point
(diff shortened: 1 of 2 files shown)


d00rman:
LGTM :+1:

eevans:
LGTM too; @Pchelolo To confirm: It is your intention to merge this to master?

arlolra:
LGTM

Pchelolo:
@eevans Ya, why not

eevans:
> @eevans Ya, why not

K

PR #838, opened 2017-07-19
Log a warning when revision-mismatch code actually plays out

Pchelolo:
To investigate what's actually going on here.

cc @wikimedia/services


eevans:
LGTM 👍

PR #843, opened 2017-07-25
Fix basic `example` config

gwicke:
- Restore basic `example` project config, without advanced features like
cxserver or citoid.
- Move advanced "dev" functionality into a more advanced "dev.yaml"
project.
- Move cxserver "section" content into its own module, so that it can be
enabled selectively.

Bug: https://phabricator.wikimedia.org/T171592

### DIFF
--- a/projects/dev.yaml
+++ b/projects/dev.yaml
@@ -1,4 +1,4 @@
-# RESTBase project config for small wiki installs
+# RESTBase project config for development environments
 paths:
   /{api:v1}:
     x-modules:
@@ -16,10 +16,27 @@ paths:
               x-modules:
                 - path: v1/content.yaml
                   options: '{{options}}'
+                - path: v1/content_segments.yaml
+                  options:
+                      purged_cache_control: '{{options.purged_cache_control}}'
+                      cx_host: '{{options.transform.cx_host}}'
                 - path: v1/common_schemas.yaml # Doesn't really matter where to mount it.
+                - path: v1/mobileapps.yaml
+                  options: '{{merge({"response_cache_control": options.purged_cache_control},
+                                options.mobileapps)}}'
             /transform:
               x-modules:
                 - path: v1/transform.yaml
+                - path: v1/transform-lang.js
+                  options: '{{options.transform}}'
+            /media:
+              x-modules:
+                - path: v1/mathoid.yaml
+                  options: '{{options.mathoid}}'
+            /data:
+              x-modules:
+                - path: v1/citoid.js
+                  options: '{{options.citoid}}'
         options: '{{options}}'
 
   /{api:sys}:
@@ -59,5 +76,10 @@ paths:
                   options:
                     parsoidHost: '{{options.parsoid.host}}'
                     response_cache_control: '{{options.purged_cache_control}}'
+            /mobileapps:
+              x-modules:
+                - path: sys/mobileapps.js
+                  options: '{{merge({"response_cache_control": options.purged_cache_control},
+                                options.mobileapps)}}'
         options: '{{options}}'
 
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -48,6 +48,8 @@ paths:
             /page:
               x-modules:
                 - path: v1/content.yaml
+                  options: '{{options}}'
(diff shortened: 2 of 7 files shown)


projects/dev.yaml
```diff
@@ -0,0 +1,104 @@
...
+# - cp config.example.yaml config.yaml
+# - double-check and possibly modify lines marked with XXX, then start restbase with
+#
+#   node server
+#
+# - If all went well, http://localhost:7231/localhost/v1/page/html/Main_Page
+# should show your wiki's [[Main Page]].
+
+# First, we define some project templates. These are referenced / shared
+# between domains in the root_spec further down.
```
Pchelolo: This whole intro looks like a copy-paste from a wrong location.
- It's not SQLite or Cassandra - whatever you pass in the options will be used
- It's not for small wiki installs
- It's not for config.example.yaml, but for config.dev.yaml or smth

projects/example.yaml
```diff
@@ -34,27 +34,11 @@ paths:
...
-                  options: '{{options.transform}}'
-            /media:
-              x-modules:
-                - path: v1/mathoid.yaml
-                  options: '{{options.mathoid}}'
-            /data:
-              x-modules:
-                - path: v1/citoid.js
-                  options: '{{options.citoid}}'
-        options: '{{options}}'
```
Pchelolo: That should still be needed I think.

Pchelolo:
How does the swagger-ui look with this change? Does `segments` property fall to the right tag category and get sorted into a good position?

gwicke:
> How does the swagger-ui look with this change? Does segments property fall to the right tag category and get sorted into a good position?

The position is the same as before, so yeah it still works as expected. Making the merge order predictable was one of the reasons for using arrays for module definitions.

projects/example.yaml
gwicke: Yup, fixed in amend.

projects/dev.yaml
gwicke: Yeah, it was the same header as in example.yaml. Removed it from both.

projects/dev.yaml
```diff
@@ -0,0 +1,85 @@
...
+                      cx_host: '{{options.transform.cx_host}}'
+                - path: v1/common_schemas.yaml # Doesn't really matter where to mount it.
+                - path: v1/mobileapps.yaml
+                  options: '{{merge({"response_cache_control": options.purged_cache_control},
+                                options.mobileapps)}}'
+            /transform:
+              x-modules:
+                - path: v1/transform.yaml
+                - path: v1/transform-lang.js
+                  options: '{{ merge({"cx_host": "http://localhost:12345/"}, options.transform) }}'
```
Pchelolo: Ulalala... Why are we merging it with some random port? If it is supported in dev env, the proper config should be provided. If it's not - the whole section regarding the cxserver must not be here.

projects/dev.yaml
```diff
@@ -0,0 +1,85 @@
...
+                redirect_cache_control: '{{options.purged_cache_control}}'
+          paths:
+            /page:
+              x-modules:
+                - path: v1/content.yaml
+                  options: '{{options}}'
+                - path: v1/content_segments.yaml
+                  options:
+                      purged_cache_control: '{{options.purged_cache_control}}'
+                      cx_host: '{{options.transform.cx_host}}'
```
Pchelolo: So we do the `merge` later but not here. We need to settle on one option - we either support cxserver here and require the option or we don't and don't add the endpoints at all.

gwicke: Yeah, lets just require it here.

Pchelolo: Are you sure we need it at all? Do we want CXServer in our container-based dev env?

gwicke: It won't work initially, but this is very similar to mobileapps and citoid. If we can make cxserver work in the dev environment's memory limits, then that would be nice.

d00rman: CXServer is not resource-hungry, but Apertium does sometimes consume some extra memory. However, given the scale for which the dev environment is intended that should not be a problem in practice. +1 on getting it into the dev environment.

PR #844, opened 2017-07-26
Improve filename encoding for PDF responses

gwicke:
- Use quoted string syntax for the plain filename parameter. Protecting
raw `"` remains a TODO that will require new functionality in
hyperswitch. In the meantime, adding quotes already protects many other
chars.
- Correctly pass the mandatory encoding in extended filename syntax.
This avoids Firefox picking up apostrophes in the filename as a
delimiter.

Bug: https://phabricator.wikimedia.org/T171747


gwicke:
@Pchelolo, the correct escaping would be a backslash. We could add a special-case "quoted_string" function that handles the escaping & quote adding.

PR #848, opened 2017-07-31
Update summary and licensing information; add pageview license

gwicke:
Bug: [T170602](https://phabricator.wikimedia.org/T170602)

### DIFF
--- a/projects/wikimedia.org.yaml
+++ b/projects/wikimedia.org.yaml
@@ -8,33 +8,44 @@ paths:
             title: Wikimedia REST API
             description: >
                 This API provides cacheable and straightforward access to
-                Wikimedia content and data, in machine-readable formats. Each
-                entry point has explicit stability markers to inform you about
-                development status and change policy, according to [our API
-                version policy](https://www.mediawiki.org/wiki/API_versioning).
+                Wikimedia content and data, in machine-readable formats.
 
-                See https://www.mediawiki.org/wiki/REST_API for background and details.
+                ### Global Rules
 
-                ### High-volume access
-                  - Limit your clients to no more than 200 requests/s to this API.
-                  - Set a unique `User-Agent` or `Api-User-Agent` header that
-                    allows us to contact you quickly. Email addresses or URLs
-                    of contact pages work well.
+                - Limit your clients to no more than 200 requests/s to this API.
+                  Each API endpoint's documentation may detail more specific usage limits.
+                - Set a unique `User-Agent` or `Api-User-Agent` header that
+                  allows us to contact you quickly. Email addresses or URLs
+                  of contact pages work well.
 
                 By using this API, you agree to Wikimedia's 
                 [Terms of Use](https://wikimediafoundation.org/wiki/Terms_of_Use) and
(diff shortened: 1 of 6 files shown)


Pchelolo:
Having no idea about the specific licenses, LGTM from the code perspective

gwicke:
LGTM in general to me as well. Lets wait a moment to see if we can generalize this for the other projects as well (see mail thread).

d00rman:
Why is CC0 being added to the mathoid end points as well @marcelrf ?

gwicke:
@d00rman, math is CC-BY-SA 3.0. This is explicit because we just dropped the global license disclaimer. There is a follow-up discussion about maybe re-adding the global disclaimer with something like "Unless otherwise specified in the endpoint documentation, content is licensed under CC-BY-SA 3.0.". We could then consider dropping the explicit mention for the majority of endpoints using that license, and use the same text for other projects.

d00rman:
Oh right right, thnx @gwicke !

projects/wikimedia.org.yaml
```diff
@@ -8,33 +8,44 @@ paths:
             title: Wikimedia REST API
             description: >
                 This API provides cacheable and straightforward access to
-                Wikimedia content and data, in machine-readable formats. Each
-                entry point has explicit stability markers to inform you about
-                development status and change policy, according to [our API
-                version policy](https://www.mediawiki.org/wiki/API_versioning).
+                Wikimedia content and data, in machine-readable formats.
```
nuria: Don't we need to add " in machine-readable formats, for the project specified in the URL." according to zhou's comments?

projects/wikimedia.org.yaml
```diff
@@ -8,33 +8,44 @@ paths:
...
-                [Privacy Policy](https://wikimediafoundation.org/wiki/Privacy_policy), and
-                you irrevocably agree to release modifications or additions
-                made through this API under the 
+                [Privacy Policy](https://wikimediafoundation.org/wiki/Privacy_policy).
+                Unless otherwise specified in the endpoint documentation
+                below, content accessed via this API is licensed under the
                 [CC-BY-SA 3.0](https://creativecommons.org/licenses/by-sa/3.0/) 
-                License and the [GFDL](https://www.gnu.org/copyleft/fdl.html).
+                and [GFDL](https://www.gnu.org/copyleft/fdl.html) licenses,
+                and you irrevocably agree to release modifications or
```
nuria: I think this needs to be "and you irrevocably agree to release modifications or additions made through this API under the CC-BY-SA 3.0 license and the GFDL."

gwicke: I dropped the repetition. I don't think it is needed for clarity.

projects/wikimedia.org.yaml
gwicke: Some of the data is cross-project as well. For licensing purposes, it shouldn't matter anyway, as differing licenses are called out explicitly.

projects/wikimedia.org.yaml
nuria: Then, +1 I think you have to merge as i do not have permits correct?

PR #849, opened 2017-08-01
Recommendation API Public End Point

d00rman:
This PR exposes the recommendation API service's end point publicly at `https://{domain}/api/rest_v1/data/recommendation/translation/{from_lang}{/seed_article}`

This PR is a continuation of PR #835 .

Bug: [T170877](https://phabricator.wikimedia.org/T170877)

### DIFF
--- a/v1/recommend.yaml
+++ b/v1/recommend.yaml
@@ -51,6 +51,7 @@ paths:
         - get_from_backend:
             request:
               uri: '{{options.host}}/{domain}/v1/translation/articles/{from_lang}/{seed_article}'
+      x-monitor: false
 definitions:
   translation_result:
     type: object
--- a/v1/recommend.yaml
+++ b/v1/recommend.yaml
@@ -1,0 +1,77 @@
+info:
+  version: '1.0.0'
+  title: Recommendation API
+  description: Recommendation API
+  termsofservice: https://github.com/wikimedia/restbase
+  license:
+    name: Apache license, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /translation/articles/{source}{/seed}:
+    get:
+      tags:
+        - Recommend
+      summary: Recommend articles for translation.
+      description: |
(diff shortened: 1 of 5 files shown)


d00rman:
cc @wikimedia/services @schana

v1/recommend.yaml
```diff
@@ -0,0 +1,76 @@
...
+          in: query
+          description: The max number of articles to return
+          type: integer
+          required: false
+          default: 24
+      responses:
+        '200':
+          description: the list of articles recommended for translation
+          schema:
+            $ref: '#/definitions/translation_result'
```
Pchelolo: Translation?

v1/recommend.yaml
```diff
@@ -0,0 +1,76 @@
...
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_backend:
+            request:
+              uri: '{{options.host}}/{domain}/v1/translation/articles/{from_lang}/{seed_article}'
```
Pchelolo: That wouldn't work. `seed_article` is optional so it will be `/en/undefined` if it's missing.

v1/recommend.yaml
```diff
@@ -0,0 +1,76 @@
...
+      x-monitor: false
+definitions:
+  translation_result:
```
Pchelolo: Recommendation_result better perhaps?

v1/recommend.yaml
```diff
@@ -0,0 +1,76 @@
...
+    type: object
+    properties:
+      count:
+        type: integer
+        description: the number of recommendations returned
+      articles:
```
Pchelolo: Normally for lists of articles we use `items`? Maybe better to stick with a pattern?

d00rman: Very good catch! We will need to implement that in the service first though, so leaving this comment open until we do so.

d00rman:
@Pchelolo comments addressed.

PR #850, opened 2017-08-01
Temporary fallback to the new storage buckets in order to truncate th…

Pchelolo:
DO NOT MERGE!!!!!!!!!!!!

This is needed to temporary switch to another set of buckets in order to truncate the storage while having VE working. We need to wait for @eevans to come back for the truncation part, so do not merge for now. This is temporary code. This needs to be extra careful deploying this, because unit tests do not examine these routes.

cc @wikimedia/services

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -213,8 +213,11 @@ class ParsoidService {
         };
     }
 
-    getBucketURI(rp, format, tid, useKeyRevValue) {
-        const bucket = useKeyRevValue ? 'key_rev_value' : this.options.bucket_type;
+    getBucketURI(rp, format, tid, forceKeyRevValue, useNewBucket) {
+        const bucket = forceKeyRevValue ? 'key_rev_value' : this.options.bucket_type;
+        if (useNewBucket) {
+            format = `${format}-2`;
+        }
         const path = [rp.domain, 'sys', bucket, `parsoid.${format}`, rp.title];
         if (rp.revision) {
             path.push(rp.revision);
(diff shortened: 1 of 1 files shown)


d00rman:
LGTM :+1: but let's merge/deploy once @eevans is ready to assist with the truncation/deletion.

sys/parsoid.js
```diff
@@ -213,8 +213,11 @@ class ParsoidService {
         };
     }
 
-    getBucketURI(rp, format, tid, useKeyRevValue) {
-        const bucket = useKeyRevValue ? 'key_rev_value' : this.options.bucket_type;
+    getBucketURI(rp, format, tid, forceKeyRevValue, useNewBucket) {
```
gwicke: Minor niggle: It might be more readable & require less changes across the module to introduce a second "getOldBucketURI" method, and then call that only in the few places we need to access it.

gwicke:
LGTM as well, only one minor & very optional niggle inline. Feel free to ignore, especially considering that we'll likely revert this soon anyway.

Pchelolo:
> Feel free to ignore, especially considering that we'll likely revert this soon anyway.

I will ignore the niggle this time, because as the new codepath is not covered by unit tests (and can't be covered by them because we clean the DB in the beginning), I've already spent a significant amount of time manually testing this. Introducing any changes will require repeating manual testing and wasting time essentially

gwicke:
> I will ignore the niggle this time
:+1:

Pchelolo:
Ok. Then we will move on this once @eevans is back and when we're actually ready to truncate.

PR #853, opened 2017-08-14
Remove references to old parsoid tables

Pchelolo:
To prepare for the truncation we need to delete references to old tables and delete the fallback.

To be deployed not earlier then Tuesday afternoon Aug 15

cc @wikimedia/services

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -904,12 +904,14 @@ module.exports = (options) => {
                     version: 1
                 }
             },
+            /* This keyspace is not used yet - no need to create it
             {
                 uri: `/{domain}/sys/key_rev_value/parsoid.data-mw`,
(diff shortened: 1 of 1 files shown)


sys/parsoid.js
```diff
@@ -888,18 +865,6 @@ module.exports = (options) => {
...
-                        type: 'latest',
-                        count: 1,
-                        grace_ttl: 86400
-                    },
-                    valueType: 'blob',
-                    version: 1
-                }
-            },
-            {
                 uri: `/{domain}/sys/${options.bucket_type}/parsoid.html-2`,
```
d00rman: Since we don't use `options.bucket_type` any more in the code, we should probably fix this to `key_rev_value`

PR #854, opened 2017-08-17
Update security headers

gwicke:
- Allow access to response headers of HEAD requests.
- Allow access to `content-length` response header. This is useful for
clients wishing to avoid downloading large pages or resources, by first
checking for the response size with a HEAD request.
- Restrict cross-origin referrer header forwarding to relay the origin
(domain) only. This avoids revealing the precise page / API result a
client navigated from.

Bug: https://phabricator.wikimedia.org/T173509

### DIFF
--- a/test/features/pagecontent/pagecontent.js
+++ b/test/features/pagecontent/pagecontent.js
@@ -21,10 +21,11 @@ describe('item requests', function() {
         .then(function(res) {
             assert.deepEqual(res.status, 200);
             assert.deepEqual(res.headers['access-control-allow-origin'], '*');
-            assert.deepEqual(res.headers['access-control-allow-methods'], 'GET');
-            assert.deepEqual(res.headers['access-control-allow-headers'], 'accept, content-type, cache-control, ' +
+            assert.deepEqual(res.headers['access-control-allow-methods'], 'GET,HEAD');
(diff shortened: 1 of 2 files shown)


test/features/pagecontent/pagecontent.js
```diff
@@ -21,8 +21,8 @@ describe('item requests', function() {
         .then(function(res) {
             assert.deepEqual(res.status, 200);
             assert.deepEqual(res.headers['access-control-allow-origin'], '*');
-            assert.deepEqual(res.headers['access-control-allow-methods'], 'GET');
-            assert.deepEqual(res.headers['access-control-allow-headers'], 'accept, content-type, cache-control, ' +
+            assert.deepEqual(res.headers['access-control-allow-methods'], 'GET,HEAD');
+            assert.deepEqual(res.headers['access-control-allow-headers'], 'accept, content-type, content-length, cache-control, ' +
```
Pchelolo: Let's test for `referrer-policy` as well since we're testing for all the CORS headers here.

gwicke: Good point. Added.

Pchelolo:
+1 Let's wait for travis and merge

PR #855, opened 2017-08-22
Correct the spec for cxserver machine translation

santhoshtr:
1. The consume type is application/x-www-form-urlencoded like
   parsoid.
2. The html for mt should be set in formData with key as html like
   parsoid.

Fix for https://phabricator.wikimedia.org/T173801

### DIFF
--- a/v1/transform-global.yaml
+++ b/v1/transform-global.yaml
@@ -17,7 +17,7 @@ paths:
 
         Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
       consumes:
-        - application/json
+        - application/x-www-form-urlencoded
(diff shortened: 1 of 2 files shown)


Pchelolo:
grrrrr.... travis failures are unrelated but I still want it to be green.. restarting the build one more time and merging afterwards

Pchelolo:
Looks good. Merged. do you need it deployed soon?

kartikm:
@Pchelolo Yes, this need to deploy as soon as possible, so we can go ahead with cxserver deployment after this. Thanks for the merge!

Pchelolo:
@kartikm oke, will deploy today.

PR #856, opened 2017-09-01
Use MCS for text extracts in summary

Pchelolo:
UNTESTED! DO NOT MERGE!
Since MCS is not deployed yet.

cc @wikimedia/services @berndsi @jdlrobson

### DIFF
--- a/v1/summary.yaml
+++ b/v1/summary.yaml
@@ -109,14 +109,15 @@ paths:
               body: '{{storage.body}}'
 
         # Storage miss. Call the Action API to get the textextract.
-        - extract:
+        - get_extracts:
+            request: get
+            uri: '{{options.host}}/{domain}/v1/page/preview-html/{title}'
+          get_properties:
             request:
               method: post
               uri: /{domain}/sys/action/query
               body:
-                prop: 'info|extracts|pageimages|revisions|pageterms|coordinates'
-                exsentences: 5
-                exintro: true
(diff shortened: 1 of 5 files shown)


v1/summary.yaml
```diff
@@ -109,39 +109,43 @@ paths:
...
-                lang: '{{extract.body.items[0].pagelanguagehtmlcode}}'
-                dir: '{{extract.body.items[0].pagelanguagedir}}'
-                timestamp: '{{getRevision(extract.body.items[0].revisions).timestamp}}'
-                description: '{{extractDescription(extract.body.items[0].terms)}}'
-                coordinates: '{{processCoords(extract.body.items[0].coordinates)}}'
+                title: '{{get_properties.body.items[0].title}}'
+                displaytitle: '{{get_properties.body.items[0].displaytitle}}'
+                pageid: '{{get_properties.body.items[0].pageid}}'
+                extract: '{{get_extracts.body.extract}}'
+                extract_html: '{{get_extracts.body.extract_html}}'
```
berndsi: The `get_extract` response also has a `type` property. I think we'd want to expose this one as well. At least for the new version.

berndsi: @jdlrobson also added a new `revision` property. Not sure about that one since it could also be retrieved from the etag but probably doesn't hurt.

Pchelolo: Done Done

PR #858, opened 2017-09-06
Separate mobileapps switch to the new bucket

Pchelolo:
cc @wikimedia/services

### DIFF
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -119,22 +119,6 @@ paths:
                 - path: sys/table.js
                   options:
                     conf: '{{options.table_ng}}'
-            /parsoid_bucket:
-              x-modules:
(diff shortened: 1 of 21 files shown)


projects/wmf_default.yaml
```diff
@@ -114,6 +114,27 @@ paths:
...
+            /table3: &sys_table_ng
+              x-modules:
+                - path: sys/table.js
+                  options:
+                    conf: '{{options.table_ng}}'
+            /parsoid_bucket:
+              x-modules:
+                - path: sys/multi_content_bucket.js
+                  options:
+                    time_to_live: '{{default(options.parsoid.time_to_live, 84600)}}'
```
gwicke: [minor, also below]: One day is 86400 seconds.

gwicke:
LGTM, but I did not review the bucket code in detail.

PR #862, opened 2017-09-12
Respect the table prefix for timelines

Pchelolo:
We shouldn't share the timelines between use-cases, so need to respect the table prefix there.

cc @wikimedia/services

### DIFF
--- a/sys/multi_content_bucket.js
+++ b/sys/multi_content_bucket.js
@@ -168,9 +168,9 @@ class MultiContentBucket {
         }))
         .concat([
             {
-                uri: new URI([rp.domain, 'sys', 'table3', 'revision-timeline']),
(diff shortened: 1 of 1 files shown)


eevans:
If Travis is good, so am I; LGTM

Pchelolo:
We shouldn't merge/deploy it yet until we finish testing in dev cluster as this will rename the timeline tables and potentially break everything.

gwicke:
Travis is now happy as well.

d00rman:
> We shouldn't merge/deploy it yet until we finish testing in dev cluster as this will rename the timeline tables and potentially break everything.

Well, not really. If the data is absent from the timeline table (which would be the case after deploying this, as they would be empty), older revisions/renders won't get deleted until a new one comes along.

d00rman:
Merging as we need this before putting it out into prod

PR #863, opened 2017-09-13
Flagged revision handling

d00rman:
This PR adds support for handling flagged revisions in RESTBase. The logic is as follows. First, we check whether the *FlaggedRevs* extension is enabled on the target project. If that is not the case **or the client provided a revision**, the regular code path is followed. If the extension is installed, then we launch two requests in parallel: the regular one and a call to the MW API to check for flagged revisions for the given title. If there is a flagged revision for the title and its stable revision ID matches that of the returned content, return the content, otherwise issue another request for the stable revision.

Bug: [T169116](https://phabricator.wikimedia.org/T169116)

### DIFF
--- a/lib/flagged_revs_filter.js
+++ b/lib/flagged_revs_filter.js
@@ -1,0 +1,53 @@
+'use strict';
+
+
+const P = require('bluebird');
+const URI = require('hyperswitch').URI;
+const mwUtil = require('./mwUtil');
+
+
+module.exports = (hyper, req, next, options, specInfo) => {
+
+    const rp = req.params;
+
+    if (mwUtil.isNoCacheRequest(req) || rp.revision || rp.revision === 0) {
+        // this is an update request or the client
+        // provided a revision, so skip the check entirely
+        return next(hyper, req);
+    }
+
+    return mwUtil.getSiteInfo(hyper, req).then((siteinfo) => {
+        if (!siteinfo.extensions['Flagged Revisions']) {
+            // the project does not have the FlaggedRevs extension
+            // enabled so proceed as normal
+            return next(hyper, req);
+        }
+        const queryReq = {
+            uri: new URI([rp.domain, 'sys', 'action', 'query']),
+            method: 'post',
+            body: {
+                prop: 'flagged',
+                titles: rp.title,
+                formatversion: 2
+            }
+        };
+        return P.props({
+            content: next(hyper, req),
+            revinfo: hyper.post(queryReq).catch((e) => { return null; })
+        }).then((res) => {
+            const content = res.content;
+            const retrev = mwUtil.parseETag(content.headers.etag).rev;
+            const revinfo = res.revinfo && res.revinfo.body.items[0];
+            if (!revinfo || !revinfo.flagged || revinfo.flagged.stable_revid === retrev) {
+                // the page has no flagged revisions, or the client is
+                // fetching a stable revision, so it's safe to return it
+                return content;
+            }
+            // the client is not looking for a specific revision
+            // so redo the request with the stable revision
+            req.params.revision = revinfo.flagged.stable_revid;
+            return next(hyper, req);
+        });
+    });
+
+};
--- a/lib/flagged_revs_filter.js
+++ b/lib/flagged_revs_filter.js
@@ -46,7 +46,8 @@ module.exports = (hyper, req, next, options, specInfo) => {
             // the client is not looking for a specific revision
             // so redo the request with the stable revision
             req.params.revision = revinfo.flagged.stable_revid;
-            return next(hyper, req);
+            req.uri = new URI(`${req.uri.toString()}/${revinfo.flagged.stable_revid}`);
+            return hyper.request(req);
         });
     });
 
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -198,7 +198,7 @@ class ParsoidService {
             getPageBundle: this.pagebundle.bind(this),
             // Revision retrieval per format
             getWikitext: this.getFormat.bind(this, 'wikitext'),
-            getHtml: this.getHtml.bind(this),
+            getHtml: this.getFormat.bind(this, 'html'),
             getDataParsoid: this.getFormat.bind(this, 'data-parsoid'),
             // Listings
             listWikitextRevisions: this.listRevisions.bind(this, 'wikitext'),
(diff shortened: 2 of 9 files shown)


sys/parsoid.js
```diff
@@ -455,6 +455,67 @@ class ParsoidService {
...
+                if (!rp.revision) {
+                    // the client is looking for the latest revision, so
+                    // redo the request with the stable revision
+                    req.params.revision = revinfo.flagged.stable_revid;
+                    return this.getFormat('html', hyper, req);
+                }
+                // the client supplied a revision, but that one is
+                // flagged, so bail out
+                // XXX: this is a problematic case, we need to find
+                // a solution for this
```
berndsi: This sounds like a flagged revision is a bad thing. While I'm sure my understanding of flagged revisions is not complete, I have a different understanding of flagged revisions. If a revision is flagged it's marked as good/reviewed, even when just DeferredChanges is enabled, AFAICS.

Example: http://deferred-changes.wmflabs.org/wiki/Special:ApiSandbox#action=query&format=json&prop=flagged&meta=&titles=Test_passive_deferred
returns `"stable_revid": 19`

http://deferred-changes.wmflabs.org/w/index.php?title=Test_passive_deferred&action=history
has two changes:
* rev 19: reviewed as good/flagged
* rev 20: unreviewed/not flagged

gwicke: Yeah, expected behavior for the explicit revision request should be to just return the content, as before. This ensures that these revisions can still be edited, among other things.

Pchelolo:
I see 2 issues with this patch:
- It doesn't really address all the other types of the content - you can still get the unapproved mobile content, summary etc. I think we should make this a filter.
- The VE issue is a BIG one, this practically makes VE completely unusable on wikis with flagged revisions enabled. I think if we have to sacrifice something - let's sacrifice crawlers, not VE.

gwicke:
> The VE issue is a BIG one, this practically makes VE completely unusable on wikis with flagged revisions enabled. I think if we have to sacrifice something - let's sacrifice crawlers, not VE.

Agreed. Lets implement the behavior discussed in https://phabricator.wikimedia.org/T169116 ?

sys/parsoid.js
```diff
@@ -455,6 +455,67 @@ class ParsoidService {
...
+            }
+            return P.props({
+                html: this.getFormat('html', hyper, req),
+                revinfo: hyper.post(queryReq)
+            }).then((res) => {
+                const html = res.html;
+                const retrev = mwUtil.parseETag(html.headers.etag).rev;
+                const revinfo = res.revinfo.body.items[0];
+                if (!revinfo.flagged || revinfo.flagged.stable_revid === retrev) {
+                    // the revision is not flagged, or the client is
```
tgr: // the page has no flagged revision...

would be a more accurate way to put it.

sys/parsoid.js
```diff
@@ -455,6 +455,67 @@ class ParsoidService {
...
+                    // fetching a stable revision, so it's safe to return it
+                    return html;
+                }
+                if (!rp.revision) {
+                    // the client is looking for the latest revision, so
```
tgr: // the client is not looking for a specific revision of the article...

might be clearer; when the client is actually looking for the latest revision, they will end up in the HTTPError branch.

sys/parsoid.js
```diff
@@ -455,6 +455,67 @@ class ParsoidService {
...
+                    // redo the request with the stable revision
+                    req.params.revision = revinfo.flagged.stable_revid;
+                    return this.getFormat('html', hyper, req);
+                }
+                // the client supplied a revision, but that one is
```
tgr: // the client supplied a revision, but a different revision is flagged...

sys/parsoid.js
tgr: Sounds like the real problem is that something else links to the latest revision instead of the non-revision-specific URL. That cannot be fixed here, and just returning the last revision when asked to do so will not make anything worse (crawlers getting the last revision instead of the stable one via RESTBase is already the status quo, right?).

tgr:
Note that patrolling and flagging are two different things; a wiki can have both (e.g. enwiki). Patrolling should be ignored completely when selecting which revision is shown.

Depending on stability configuration, the flagged version might not be the one that gets displayed by MediaWiki. This configuration can be changed per-page by administrators. Here is an example: [history](https://de.wikipedia.beta.wmflabs.org/w/index.php?title=FRtest&action=history), [API](https://de.wikipedia.beta.wmflabs.org/wiki/Spezial:ApiSandbox#action=query&format=json&prop=revisions%7Cflagged&titles=FRtest&rvprop=ids%7Cflags%7Cuser%7Cflagged&rvlimit=5), [config page](https://de.wikipedia.beta.wmflabs.org/w/index.php?title=Special:Stabilization&page=FRtest). The flagged revision is 20657 but anons will see the current revision 20661.
This is an edge case so it might make sense to not deal with it but if that's the case it should be stated explicitly.

Note also that this configuration is handled internally as a page protection so it can have an expiration time (again see the example). If RESTBase does take the stability configuration into account, the cache expiration time should probably be shortened accordingly.

sys/parsoid.js
berndsi: I was mostly looking at the comments. Once they get updated with something like @tgr's proposals then most parts start making more sense to me.

Pchelolo:
Needs a rebase to make travis pass

d00rman:
> Needs a rebase to make travis pass

A lot of other refactoring too :)

d00rman:
> Depending on stability configuration, the flagged version might not be the one that gets displayed by MediaWiki. This configuration can be changed per-page by administrators. Here is an example: history, API, config page. The flagged revision is 20657 but anons will see the current revision 20661.
> This is an edge case so it might make sense to not deal with it but if that's the case it should be stated explicitly.

Ok, you have really confused me with this example. I was under the impression that `stable_rev` tells me which revision should be returned/displayed but this does not seem to be the case. Furthermore, the API response makes no other hint at that fact. Since we are here just working on an intermediary solution, would returning the revision `stable_rev` points at work?

> 
> Note also that this configuration is handled internally as a page protection so it can have an expiration time (again see the example). If RESTBase does take the stability configuration into account, the cache expiration time should probably be shortened accordingly.

This is a good tip that we will definitely keep in mind when building the real solution to this. This is not a problem for now, because we will consult the MW API on each request.

On that note, we will move forward with the plan of honouring flagged revs only for revision-less requests as a first iteration, so as not to break VE at this time.

tgr:
> Ok, you have really confused me with this example. I was under the impression that stable_rev tells me which revision should be returned/displayed but this does not seem to be the case.

Yeah, FlaggedRevs is probably the most confusing piece of software in Wikmedia production (although, is that's any consolation, it used to be far worse - a lot of configuration options were removed because no one used them). stable_rev actually tells you which is the last revision that was marked as meeting for the community's standards for review; whether the stable revision or the latest one is displayed to anons is configurable (both per-wiki and per-page).

> Furthermore, the API response makes no other hint at that fact.

You can call `prop=flagged` which will return with something like `"protection_level": ""`. I don't think there is a way to check through the API what the site-wide default is, but I doubt there is any wiki which does not default to showing the stable revision to anons.

> Since we are here just working on an intermediary solution, would returning the revision stable_rev points at work?

Yeah, typically there are very few pages with non-default display settings (you can check via `Special:ConfiguredPages`; right now there are [two on dewiki](https://de.wikipedia.org/wiki/Spezial:Konfigurierte_Seiten) and [twentyish on fiwki](https://fi.wikipedia.org/wiki/Toiminnot:ConfiguredPages) for example, so I doubt anyone cares when those settings are ignored. But it would be nice to have a pointer in the source code that that's happening.

d00rman:
I reformatted the code into a filter that is applied to all of the relevant pieces of content (`html`, `mobile-sections*`, `pdf`). This version of the filter is applied only to requests for the latest content of a title.

lib/flagged_revs_filter.js
```diff
@@ -0,0 +1,53 @@
...
+const P = require('bluebird');
+const URI = require('hyperswitch').URI;
+const mwUtil = require('./mwUtil');
+
+
+module.exports = (hyper, req, next, options, specInfo) => {
+
+    const rp = req.params;
+
+    if (mwUtil.isNoCacheRequest(req) || rp.revision || rp.revision === 0) {
```
Pchelolo: What's up with a zero? Is it possible to have a 0 revision number?

Pchelolo:
Also it would be neat to add some tests, but that might be too hard to create stable tests for this. In case it is, never mind

lib/flagged_revs_filter.js
d00rman: `0` is used for stashing, and since it's false-y, it needs special treatment here.

Pchelolo: In stashing code we `'0'` is added afterwards inside the parsed module. https://github.com/wikimedia/restbase/blob/master/sys/parsoid.js#L769

So at this level 0 shouldn't be tolerated.

d00rman: This check will not be exercised as we don't apply the filter on POSTs nor on requests containing a revision inside the spec, but it's future-proof. AFAIK, once we add the revision `0`, that is picked up by VE and used on subsequent requests, so I'd prefer to keep it there. Invalid or bogus revision handling should be left to the Parsoid module either way, we should just ensure that the filter does the right thing.

Pchelolo: I was nit picking :) Agreed. It doesn't hurt to have the check.

Pchelolo:
+1 from me, but not merging since we need to establish a release plan. Given that we're experimenting with the new storage on mobile introducing even more change simultaneously sounds like a bad idea. I propose to have this change sit around until the next Monday - if all goes good we completely switch mobile to the new storage on Wednesday and that gives us a nice window to verify the new storage before introducing yet another change. Thoughts?

d00rman:
Fixed a bug and added some tests. I agree with not moving forward on this until Monday.

gwicke:
I really like how this turned out. Great work, @d00rman!

test/features/pagecontent/pagecontent.js
```diff
@@ -342,6 +343,28 @@ describe('page content access', function() {
...
+        });
+    });
+
+    it('should return the stable revision for flagged titles', function() {
+        return preq.get({
+            uri: server.config.deLabsBucketURL + '/html/FRtest'
+        }).then(function(res) {
+            const retRev = mwUtil.parseETag(res.headers.etag).rev;
+            assert.deepEqual(res.status, 200);
+            assert.deepEqual(retRev, 20657);
```
Pchelolo: Em why is the stable the same as the latest? Are we actually testing anything here? If they're the same the tests will pass even without the flagged rev filter.

d00rman: This tests asserts `deepEqual()`, the other does `notDeepEqual()` ;)

Pchelolo: Oh indeed. My bad, sorry.

PR #865, opened 2017-09-20
Summary: Fix: Switch to HTTPS for the `original` field as well

d00rman:
Bug: [T176263](https://phabricator.wikimedia.org/T176263)


Pchelolo:
Hm it's really weird - for some reason travis tests with Cassandra 3 hangs because Cassandra doesn't start. I remember we've had a same problem as we've just introduced testing against Cassandra 3, but I thought I've fixed it with #834 by decreasing the heap sizes. Perhaps travis changed something?

PR #869, opened 2017-09-20
Make sure schema changes are applied sequentially

gwicke:
Cassandra does not support concurrent schema changes. For this reason,
we have been very careful to apply all schema changes sequentially:

- Start one RB instance at a time in a deploy.
- Start one worker at a time in service-runner.
- Create one cf/keyspace at a time in the storage backend, and wait for
schema agreement.

However, multi_content_bucket introduced a regression by using .map() to
apply several schema changes in parallel. This did not work very well
with Cassandra.

This patch replaces .map() with .each(), returning to sequential
execution.


gwicke:
Travis sadly has trouble starting Cassandra.

PR #871, opened 2017-09-21
Do not cache feeds for the current day in cassandra

Pchelolo:
We were caching the result of the feed request for the current day in Cassandra, but it doesn't really make any sense - regenerating it is quite fast. 

Bug: https://phabricator.wikimedia.org/T176233
cc @wikimedia/services

### DIFF
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -83,11 +83,11 @@ paths:
             /feed:
               x-modules:
                 - path: v1/feed.js
-                  options: '{{merge({"feed_cache_control": "s-maxage=300, max-age=60", "ttl": 7200}, options.mobileapps)}}'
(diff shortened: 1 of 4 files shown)


Pchelolo:
Rebased, travis is passing now.

projects/wmf_default.yaml
```diff
@@ -83,11 +83,11 @@ paths:
             /feed:
               x-modules:
                 - path: v1/feed.js
-                  options: '{{merge({"feed_cache_control": "s-maxage=300, max-age=60", "ttl": 7200}, options.mobileapps)}}'
+                  options: '{{merge({"feed_cache_control": "s-maxage=300, max-age=60", "ttl": 300}, options.mobileapps)}}'
```
d00rman: The `ttl` is not used any more, so it can be safely removed (here and elsewhere).

Pchelolo: Done

PR #872, opened 2017-09-21
Mobile Proxy: Produce the diff if there is a body mismatch

d00rman:
Temporary code to see what differs between the two buckets in the logs.

### DIFF
--- a/sys/mobile_proxy.js
+++ b/sys/mobile_proxy.js
@@ -2,9 +2,10 @@
 
 const P = require('bluebird');
 const HyperSwitch = require('hyperswitch');
-const URI = HyperSwitch.URI;
+const stringify = require('json-stable-stringify');
 const mwUtils = require('../lib/mwUtil');
 
+const URI = HyperSwitch.URI;
 const spec = HyperSwitch.utils.loadSpec(`${__dirname}/mobileapps.yaml`);
 
 class MobileApps {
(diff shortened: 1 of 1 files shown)


sys/mobile_proxy.js
```diff
@@ -45,12 +136,18 @@ class MobileApps {
...
+                const oldBody = oldBucket.body;
+                const newBody = newBucket.body;
+                if (stringify(oldBody) !== stringify(newBody)) {
+                    const diff = this._getDiff(oldBody, newBody);
+                    if (diff.length) {
+                        hyper.log('error/mobileapps/mismatch', {
+                            message: 'Content mismatch between old and new bucket',
+                            old_etag: oldBucket.headers.etag,
+                            new_etag: newBucket.headers.etag,
+                            diff
```
Pchelolo: Let's stringily the diff so that log stash doesn't try to analyze all of it's properties - we might hit into the problem with field types.

sys/mobile_proxy.js
```diff
@@ -45,12 +136,18 @@ class MobileApps {
...
-                if (JSON.stringify(oldBucket.body) !== JSON.stringify(newBucket.body)) {
-                    hyper.log('error/mobileapps', {
-                        message: 'Content mismatch between old and new bucket',
-                        old_etag: oldBucket.headers.etag,
-                        new_etag: newBucket.headers.etag
-                    });
+                const oldBody = oldBucket.body;
+                const newBody = newBucket.body;
+                if (stringify(oldBody) !== stringify(newBody)) {
+                    const diff = this._getDiff(oldBody, newBody);
```
Pchelolo: Let's surround with try/catch to avoid failing the request instead there's a bug in the `diff`?

PR #873, opened 2017-09-21
Remove old mobileapps module

Pchelolo:
Be brave! Be bold!

Right now we're seeing no issues with new storage stuff, so I've prepared the complete removal of old mobile buckets in advance. Once we're confident in the new storage we can go ahead with this.

cc @wikimedia/services


d00rman:
:+1: 

I'd propose the following time-line: start using only the new storage on Monday (by setting `storage: new` in the config) and if all is well merge and deploy this on Wednesday.

d00rman:
Actually @Pchelolo I think we can move on this today. Everything looks good with the new storage.

Pchelolo:
Rebased on master. Ready to deploy.

PR #874, opened 2017-09-22
Travis: Drop Node v4, add Node v8

d00rman:


Pchelolo:
I suppose we can drop 4 already

d00rman:
> I suppose we can drop 4 already

Good idea. {{done}}

PR #876, opened 2017-09-26
Temp: Temporarily log the request body if something goes wrong

Pchelolo:
Will deploy for 10 minutes to collect some data and then revert.

Bug: https://phabricator.wikimedia.org/T176728


gwicke:
@Pchelolo, are you waiting for a review for this?

Pchelolo:
@gwicke I wanna deploy this for 10-15 mins on Monday so that we could get to the bottom of this bug. At this point it seems testing in prod like this is the only option.

d00rman:
We made this change live on `restbase1007` just to collect logs, so closing this.

PR #877, opened 2017-09-27
Introduce the `/page/title/{title}/{revision}` route

d00rman:
It is meant to replace `/page/revision/{revision}` route. This is a first step towards that effort.

Bug: [T158100](https://phabricator.wikimedia.org/T158100)

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -200,6 +200,78 @@ paths:
                 page: '{{page}}'
       x-monitor: false
 
+  /title/{title}/{revision}:
+    get:
+      tags:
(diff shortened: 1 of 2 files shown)


v1/content.yaml
```diff
@@ -200,6 +200,78 @@ paths:
                 page: '{{page}}'
       x-monitor: false
 
+  /title/{title}/{revision}:
```
Pchelolo: Why are we adding a completely separate endpoint instead of just adding and optional `{/revision}` parameter to the existing one?

PR #878, opened 2017-09-28
Use new storage for Parsoid and make a proxy

Pchelolo:
A rebase of #857 on top of the current master to be deployed in dev cluster for testing.

Also added a `parsoid_proxy` analogues with the `mobile_proxy` we've used for mobile transition with a difference that we can switch on the new parsed storage domain-by-domain. The kill switch is simple - just make the default backend to `old`.

We should deploy this to the dev cluster, run change-prop for a while and then we can deploy this to labs and try poking VisualEditor. The to prod and enable only for test wikis and poke VE there. We can't really do a full-scale deploy of this because we don't have capacity in the new Cassandra cluster.

cc @wikimedia/services


projects/dev.yaml
```diff
@@ -86,5 +86,19 @@ paths:
                 - path: sys/mobileapps.js
                   options: '{{merge({"response_cache_control": options.purged_cache_control},
                                 options.mobileapps)}}'
+            /mobile_bucket:
```
d00rman: Did you mean(TM) `/parsoid_bucket`?

projects/dev.yaml
```diff
@@ -86,5 +86,19 @@ paths:
...
+              x-modules:
+                - path: sys/multi_content_bucket.js
+                  options:
+                    # TODO: Decide on the retention period
```
d00rman: We know the TTL here so this comment is redundant :)

sys/parsoid.js
```diff
@@ -246,23 +289,41 @@ class ParsoidService {
...
+        return hyper.put({
+            uri: this.getNGBucketURI(rp, 'all', tid),
+            body: {
+                html: parsoidResp.body.html,
+                'data-parsoid': parsoidResp.body['data-parsoid'],
+                'section-offsets': {
+                    headers: {
+                        'content-type': 'application/json'
+                    },
+                    body: parsoidResp.body['data-parsoid'].body.sectionOffsets
```
d00rman: If we need to set the content for data-parsoid, then we need to set it here as well. If we don't, then we don't need it anywhere.

test/features/pagecontent/pagecontent.js
```diff
@@ -1,4 +1,4 @@
-'use strict';
+    'use strict';
```
d00rman: Extra WS not needed.

sys/parsoid.js
d00rman: Uh, I placed the wrong comment on the wrong line, sorry about that. My point is that here in the request we explicitly set the `content-type` header for `section-offsets`, but not for `data-parsoid`, so the way I see it we should either set the header for both components or for neither of them, shouldn't we?

Pchelolo: @d00rman no, not really. For `data-parsoid` we set `parsoidResp.body['data-parsoid']` and for the `pagebundle` response from parsoid that property contains both `headers` and `body`. For `section-offsets` we store a subject of the `data-parsoid` body, so we need to specify the content type to something. We could either reuse the same headers as `data-parsoid` has or just state it's application/json. I'm indifferent between both approaches though.

d00rman: Ah indeed @Pchelolo !

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,161 @@
...
+                    });
+                })
+            )
+            .then((results) => {
+                const oldContent = results[0];
+                const newContent = results[1];
+
+                if (oldContent && newContent && mwUtil.isNoCacheRequest(req)) {
+                    if (oldContent.body !== newContent.body) {
+                        hyper.log('error/mobileapps', {
```
d00rman: `s/mobileapps/parsoid/`

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,161 @@
...
+        return {
+            method: req.method,
+            uri: new URI(path),
+            query: req.query,
+            headers: req.headers,
+            body: req.body
+        };
+    }
+
+    _createDoubleProcessingRequest(hyper, req, oldPath, newPath, listing) {
```
d00rman: Since `parsoid_old` and `parsoid_new` never change, perhaps it would be enough to have just one `path` param and prefix it with the appropriate component directly here below while making the requests?

Pchelolo:
@d00rman Addressed all your comments and also found (and fixed) a little bug. Should be GTG now

d00rman:
LGTM, but now the *should render & re-render independent revisions, if-unmodified-since support* test is failing, returning 500 instead of 412.

Pchelolo:
@d00rman right, sorry, fixed that, accidentally put the `throw e` inside the if block.. Tests pass now

Pchelolo:
Although we can deploy this to dev/labs for testing or even deploy to prod without using the new storage, I've realized that we need a bit more that this to be able to actually switch to the new content - we need to respect the `grace_ttl` period and fallback to the old storage for 24 hours after we switch an actual production wiki. I'm inclined to add support for that in a follow up PR, this one is pretty complex as it is.

d00rman:
Oh I realised that this patch puts the new parsoid module in `/sys/parsoid.js`, which means that we cannot actually deploy it as-is without causing changes to the way things are currently set out in production. Perhaps rather than having `parsoid` and `parsoid_old` we should have `parsoid_new` and `parsoid`? We can then rename the module later on, once we have moved everything over?

Also, needs the ability to specify domains in the config as a regex to ease things during the transition.

d00rman:
Hm, no, disregard my last comment about `parsoid` and `parsoid_old`, all seems good there. So should we switch this to go against master with the config options defaulting to `old` if not set at all to achieve full compatibility with the current status quo?

Pchelolo:
Closing as replaced by #894

PR #879, opened 2017-09-28
Switch feeds to the new storage. Step 1

Pchelolo:
Creates new buckets in the new storage. We will need to transfer the content before completely switching, so this is only the first step.

cc @wikimedia/services


d00rman:
:+1:

PR #880, opened 2017-10-03
Switch the feeds to the new storage completely

Pchelolo:
Use the new storage for all the feeds.

cc @wikimedia/services


Pchelolo:
We didn't think that we have to actually modify the key_value bucket, not just move that to the new storage, so I've made #881 to address that, making this PR obsolete.

PR #881, opened 2017-10-03
Make a proper key_value for the new storage

Pchelolo:
1. Rename `key_value` to `key_value_old` and use it everywhere
2. Adapt the `key_value` to the new storage - remove unused columns and make a `tid` a normal column, not a part of a primary key
3. Create new buckets for the new feeds.

cc @wikimedia/services

### DIFF
--- a/sys/key_value.js
+++ b/sys/key_value.js
@@ -11,7 +11,7 @@ const stringify = require('json-stable-stringify');
 const HTTPError = HyperSwitch.HTTPError;
 const URI = HyperSwitch.URI;
 
-const spec = HyperSwitch.utils.loadSpec(`${__dirname}/key_value_old.yaml`);
+const spec = HyperSwitch.utils.loadSpec(`${__dirname}/key_value.yaml`);
 
 // Format a revision response. Shared between different ways to retrieve a
 // revision (latest & with explicit revision).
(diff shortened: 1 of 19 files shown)


d00rman:
This seems a bit too invasive, IMHO. Wouldn't it be better to use the proxy approach we've been using? Which `key_value` to use could then be specified in the resources section during bucket creation, something along the lines of:

```yaml
version: 1
valueType: 'json'
backend: 'new'
```

with the default being `backend: 'old'`

Pchelolo:
Then we'd need to specify new/old on very request because we don't really have a space to store whether a particular bucket was created in new/old storage. We can cache that in memory though..

Why do you think it's invasive? it's mostly a mass rename of `key_value` to `key_value_old`. We can not do that and name the new bucket `key_value_new`, but then we'd need to do a mass rename anyways after we switch all to the new bucket.

Pchelolo:
I've tried switching all the use-cases to the new `key_value` implementation and the new storage and run the tests and everything looks nice.

sys/key_value.js
```diff
@@ -157,26 +134,23 @@ class KVBucket {
 
 
     putRevision(hyper, req) {
-        // TODO: support other formats! See cassandra backend getRevision impl.
         const rp = req.params;
-        let tid = rp.tid && mwUtil.coerceTid(rp.tid, 'key_value');
+        let tid = rp.tid && mwUtil.coerceTid(rp.tid, 'key_value_old');
```
d00rman: Nit: `s/key_value_old/key_value/`

PR #884, opened 2017-10-05
Remove /revision/{revision} endpoint

Pchelolo:
We've decided to remove the `/revision/{revision}` endpoint a while ago, so let's finally get rid of it.

WIP-ish we can't really release it like this, because change-prop depends on this endpoint. So once the review is done I will split it into separate stages.

One additional thing that should be done is to test it together with #718 

cc @wikimedia/services


Pchelolo:
Run all the tests with the restriction_table access checks and all the tests pass!

gwicke:
:+1: 

Any objections to merging this right away?

sys/page_revisions.js
```diff
@@ -83,15 +83,7 @@ class PRS {
...
-            ],
-            secondaryIndexes: {
-                by_rev: [
-                    { attribute: 'rev', type: 'hash' },
-                    { attribute: 'tid', type: 'range', order: 'desc' },
-                    { attribute: 'title', type: 'range', order: 'asc' },
-                    { attribute: 'restrictions', type: 'proj' }
-                ]
-            }
+            ]
```
d00rman: Should we also bump the version since it has changed?

sys/page_revisions.js
```diff
@@ -269,110 +261,117 @@ class PRS {
...
-        const getLatestTitleRevision = () => {
-            return hyper.get({
+        const titleRevisionRequest = () => {
+            const revReqObject = {
                 uri: this.tableURI(rp.domain),
                 body: {
                     table: tableName,
                     attributes: {
-                        title: rp.title
+                        title: rp.title,
```
d00rman: No comma needed at the end

d00rman:
LGTM overall, a couple of nits in-lined. Also, needs a rebase.

> WIP-ish we can't really release it like this, because change-prop depends on this endpoint.

We are good on that front. As of [Gerrit 380932](https://gerrit.wikimedia.org/r/#/c/380932/), CP uses `/page/title/{title}{/revision}` for this.

Pchelolo:
Rebased and bumped the schema version.

sys/page_revisions.js
Pchelolo: Did bump the version. We should not forget that the storage module doesn't actually delete the tables after deleting the index, so we will need to drop key spaces manually.

d00rman:
LGTM :+1: Let's merge this right before deployment.

PR #885, opened 2017-10-06
metrics: Update doc for timestamp format

kulla:
The parameters "start" and "end" also accept the format YYMMDDHH.
See https://phabricator.wikimedia.org/T177671

### DIFF
--- a/v1/metrics.yaml
+++ b/v1/metrics.yaml
@@ -81,12 +81,12 @@ paths:
           required: true
         - name: start
           in: path
-          description: The date of the first day to include, in YYYYMMDD format
+          description: The date of the first day to include, in YYYYMMDD or YYYYMMDDHH format
           type: string
           required: true
         - name: end
           in: path
-          description: The date of the last day to include, in YYYYMMDD format
+          description: The date of the last day to include, in YYYYMMDD or YYYYMMDDHH format
           type: string
           required: true
       responses:


v1/metrics.yaml
```diff
@@ -81,12 +81,12 @@ paths:
           required: true
         - name: start
           in: path
-          description: The date of the first day to include, in YYYYMMDD format
+          description: The date of the first day to include, in YYYYMMDD or YYMMDDHH format
```
berndsi: Not sure where YYMMDDHH is coming from. I think it should be YYYYMMDDHH instead. Otherwise how would the API know if you are using YYYYMMDD or YYMMDDHH? It's the same amount of numbers. In other words, YYYYMMDD[HH]. The HH part is optional. The response seems to always include it,
 though, BTW.

v1/metrics.yaml
```diff
@@ -81,12 +81,12 @@ paths:
...
           type: string
           required: true
         - name: end
           in: path
-          description: The date of the last day to include, in YYYYMMDD format
+          description: The date of the last day to include, in YYYYMMDD or YYMMDDHH format
```
berndsi: Same as above

v1/metrics.yaml
kulla: Yeah, you are right. It was late yesterday, when I made the commit :-)

kulla:
@berndsi I updated the merge request accordingly to your comment...

PR #887, opened 2017-10-10
Add Reading List Service

tgr:
Adds a bunch of endpoints under `/data/lists/` which act as a proxy towards MediaWiki's `action=readinglists`, `meta=readinglists`, `list=readinglistentries` and `meta=readinglistorder` API modules, and towards the summary endpoint on an arbitrary wiki. Task: [T168986](https://phabricator.wikimedia.org/T168986)

### DIFF
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -101,6 +101,8 @@ paths:
               x-modules:
                 - path: v1/citoid.js
                   options: '{{options.citoid}}'
+                - path: v1/lists.js
+                  options: '{{options.lists}}'
                 - path: v1/recommend.yaml
                   options: '{{options.recommendation}}'
         options: '{{options}}'
--- a/sys/action.js
+++ b/sys/action.js
@@ -101,14 +101,20 @@ function apiError(apiErr) {
     return new HTTPError(ret);
 }
 
-function buildQueryResponse(apiReq, res) {
+function checkQueryResponse(apiReq, res) {
     if (res.status !== 200) {
         throw apiError({
             info: `Unexpected response status (${res.status}) from the PHP action API.`
         });
     } else if (!res.body || res.body.error) {
         throw apiError((res.body || {}).error);
-    } else if (!res.body.query || (!res.body.query.pages && !res.body.query.userinfo)) {
+    }
+    return res;
+}
+
+function buildQueryResponse(apiReq, res) {
+    checkQueryResponse(apiReq, res);
+    if (!res.body.query || (!res.body.query.pages && !res.body.query.userinfo)) {
         throw new HTTPError({
             status: 404,
             body: {
@@ -219,6 +225,13 @@ class ActionService {
         }, buildQueryResponse);
     }
 
+    rawQuery(hyper, req) {
+        return this._doRequest(hyper, req, {
+            format: 'json',
+            formatversion: 2
+        }, checkQueryResponse);
+    }
+
     edit(hyper, req) {
         return this._doRequest(hyper, req, {
             action: 'edit',
@@ -296,6 +309,11 @@ module.exports = (options) => {
                         operationId: 'mwApiQuery'
                     }
                 },
+                '/rawquery': {
+                    all: {
+                        operationId: 'mwRawApiQuery'
+                    }
+                },
                 '/siteinfo': {
                     all: {
                         operationId: 'mwApiSiteInfo'
@@ -310,6 +328,7 @@ module.exports = (options) => {
         },
         operations: {
             mwApiQuery: actionService.query.bind(actionService),
+            mwRawApiQuery: actionService.rawQuery.bind(actionService),
             mwApiEdit: actionService.edit.bind(actionService),
             mwApiSiteInfo: actionService.siteinfo.bind(actionService)
         }
--- a/v1/lists.js
+++ b/v1/lists.js
@@ -1,57 +1,128 @@
 'use strict';
 
+const P = require('bluebird');
+const mwUtil = require('../lib/mwUtil');
 const HyperSwitch = require('hyperswitch');
+const URI = HyperSwitch.URI;
 const spec = HyperSwitch.utils.loadSpec(`${__dirname}/lists.yaml`);
 
-module.exports = options => ({
-    spec,
-    globals: {
-        options,
-        /**
-         * Transform the continuation data into a string so it is easier for clients to deal with.
-         * @param {!Object|undefined} continuation Continuation object returned by the
-         *   MediaWiki API.
-         * @return {!String|undefined} Continuation string.
-         */
-        flattenContinuation(continuation) {
-            return JSON.stringify(continuation);
-        },
-        /**
-         * Inverse of flattenContinuation.
-         * @param {!String|undefined} continuation Continuation string
-         *   returned by flattenContinuation().
-         * @return {!Object} Continuation object.
-         */
-        unflattenContinuation(continuation) {
-            const sanitizedContinuation = {};
-            if ( typeof continuation === 'string' ) {
-                try {
-                    continuation = JSON.parse(continuation);
-                } catch (e) {
-                    this.options.log('error/unflatten', {
-                        msg: e.stack,
-                        json: continuation,
-                    });
-                    throw e;
-                }
-                // Make sure nothing malicious can be done by splicing the continuation data
-                // into the API parameters.
-                const allowedKeys = ['continue', 'rlcontinue', 'rlecontinue'];
-                for (let key of allowedKeys) {
-                    if (typeof continuation[key] !== 'object') {
-                        sanitizedContinuation[key] = continuation[key];
-                    }
+class ReadingLists {
+    /**
+     * Transform the continuation data into a string so it is easier for clients to deal with.
+     * @param {!Object|undefined} continuation Continuation object returned by the MediaWiki API.
+     * @return {!String|undefined} Continuation string.
+     */
+    flattenContinuation(continuation) {
+        return JSON.stringify(continuation);
+    }
+
+    /**
+     * Inverse of flattenContinuation.
+     * @param {!String|undefined} continuation Continuation string returned by flattenContinuation()
+     * @return {!Object} Continuation object.
+     */
+    unflattenContinuation(continuation) {
+        const sanitizedContinuation = {};
+        if (typeof continuation === 'string') {
+            try {
+                continuation = JSON.parse(continuation);
+            } catch (e) {
+                this.options.log('error/unflatten', {
+                    msg: e.message,
+                    json: continuation,
+                });
+                throw e;
+            }
+            // Make sure nothing malicious can be done by splicing the continuation data
+            // into the API parameters.
+            const allowedKeys = ['continue', 'rlcontinue', 'rlecontinue'];
+            for (const key of allowedKeys) {
+                if (typeof continuation[key] !== 'object') {
+                    sanitizedContinuation[key] = continuation[key];
                 }
             }
-            return sanitizedContinuation;
+        }
+        return sanitizedContinuation;
+    }
+
+    /**
+     * Convert an array of values into the format expected by the MediaWiki API.
+     * @param {!Array} list A list containing strings and numbers.
+     * @return {!String}
+     */
+    flattenMultivalue(list) {
+        return list.join('|');
+    }
+
+    /**
+     * Handle the /list/{id}/entries endpoint (get entries of a list).
+     * @param {!HyperSwitch} hyper
+     * @param {!Object} req The request object as provided by HyperSwitch.
+     * @return {!Promise<Object>} A response promise.
+     */
+    getListEntries(hyper, req) {
+        return hyper.post({
+            uri: new URI([req.params.domain, 'sys', 'action', 'rawquery']),
+            body: {
+                action: 'query',
+                list: 'readinglistentries',
+                rlelists: req.params.id,
+                rlelimit: 'max',
+                continue: this.unflattenContinuation(req.query.next).continue,
+                rlecontinue: this.unflattenContinuation(req.query.next).rlecontinue,
+            },
+        })
+        .then((res) => {
+            const entries = res.body.query.readinglistentries;
+            const next = this.flattenContinuation(res.body.continue);
+
+            return this.hydrateSummaries({
+                status: 200,
+                headers: {
+                    'content-type': 'application/json; charset=utf-8;'
+                        + 'profile="https://www.mediawiki.org/wiki/Specs/Lists/0.1"',
+                    'cache-control': 'no-cache',
+                },
+                body: {
+                    entries,
+                    next,
+                }
+            }, hyper, req);
+        });
+    }
+
+    /**
+     * Add data from the summary endpoint to an array of list entries.
+     * @param {!Object} res Response object.
+     * @param {!HyperSwitch} hyper
+     * @param {!Object} req The request object as provided by HyperSwitch.
+     * @return {!Promise<Array>} The objects, enriched with summaries.
+     */
+    hydrateSummaries(res, hyper, req) {
+        return P.map(res.body.entries, (entry) => {
+            return mwUtil.getSiteInfo(hyper, req, entry.project).then((siteinfo) => {
+                entry.$merge = [
+                    `${siteinfo.baseUri}/page/summary/${encodeURIComponent(entry.title)}`,
+                ];
+            }).catch(() => {});
+        })
+        .then(() => mwUtil.hydrateResponse(res, uri => mwUtil.fetchSummary(hyper, uri)));
+    }
+}
+
+module.exports = (options) => {
+    const rl = new ReadingLists(options);
+
+    return {
+        spec,
+        globals: {
+            options,
+            flattenContinuation: rl.flattenContinuation.bind(rl),
+            unflattenContinuation: rl.unflattenContinuation.bind(rl),
+            flattenMultivalue: rl.flattenMultivalue.bind(rl),
         },
-        /**
-         * Convert an array of values into the format expected by the MediaWiki API.
-         * @param {!Array} list A list containing strings and numbers.
-         * @return {!String}
-         */
-        flattenMultivalue(list) {
-            return list.join('|');
+        operations: {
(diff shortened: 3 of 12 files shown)


v1/lists.js
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Handle the /list/{id}/entries endpoint (get entries of a list).
+     * @param {!HyperSwitch} hyper
+     * @param {!Object} req The request object as provided by HyperSwitch.
+     * @return {!Promise<Object>} A response promise.
+     */
+    getListEntries(hyper, req) {
+        return hyper.post({
+            uri: new URI([req.params.domain, 'sys', 'action', 'rawquery']),
```
d00rman: There is actually no need for `rawquery` here, you can use `query` since the response contains the `query` field, and hence will not fail.

v1/lists.js
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Add data from the summary endpoint to an array of list entries.
+     * @param {!Array} entries An array of objects with two properties:
+     *   - project: the domain of the wiki
+     *   - title: the title of the page
+     * @return {!Promise<Array>} The objects, enriched with summaries.
+     */
+    hydrateSummaries(entries) {
```
d00rman: For this you can use `hydrateResponse()` and `fetchSummary()` from `lib/mwUtil.js`. For a sample usage c.f. [how the feed end point does it](https://github.com/wikimedia/restbase/blob/564e23da394ae480ec03e286f9ccf3fcbacf37f9/v1/feed.js#L58).

v1/lists.yaml
```diff
@@ -0,0 +1,1011 @@
...
+              uri: /{domain}/sys/action/rawquery
+              body:
+                action: readinglists
+                command: teardown
+                token: '{{request.query.csrf_token}}'
+            return:
+              status: 200
+              headers:
+                content-type: application/json; charset=utf-8
+                cache-control: "no-cache"
```
d00rman: Did you mean(TM): `cache-control: max-age=0, s-maxage=0` ?

v1/lists.yaml
```diff
@@ -0,0 +1,1011 @@
...
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-route-filters:
+        - path: ./lib/mediawiki_auth_filter.js
+      x-request-handler:
+        - forward_to_mw:
+            request:
+              uri: /{domain}/sys/action/rawquery
```
d00rman: Since this is an `action=query`, `/sys/action/query` should be used to enforce the response having a `query` field.

v1/lists.yaml
```diff
@@ -0,0 +1,1011 @@
...
+              status: 200
+              headers:
+                content-type: application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/Lists/0.1"
+                cache-control: "no-cache"
+              body:
+                lists: '{{forward_to_mw.body.query.readinglists}}'
+                next: '{{flattenContinuation(forward_to_mw.body.continue)}}'
+      x-monitor: true
+      x-amples:
+        - title: Get all my lists
```
d00rman: How will this get all the lists for a user if there is no token to identify them?

v1/lists.yaml
```diff
@@ -0,0 +1,1011 @@
...
+          request:
```
d00rman: Superfluous, remove if there is nothing to be said about the request.

projects/wmf_default.yaml
```diff
@@ -101,6 +101,8 @@ paths:
               x-modules:
                 - path: v1/citoid.js
                   options: '{{options.citoid}}'
+                - path: v1/lists.js
```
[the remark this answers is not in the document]
tgr: Done. Didn't do wikimedia.org as there are no cookies to forward there; also I think there is no MW API behind it?).

v1/lists.js
tgr: `query` throws an error if there is no `query.pages` field in the response. Reading lists don't deal with pages (as understood by the MW API which has no concept of remote pages) so that wouldn't work here.

v1/lists.js
tgr: `fetchSummary()` uses HyperSwitch so AIUI it assumes a local endpoint; I need to use the summary endpoint belonging to `entry.project` here. `hydrateResponse()` seems to be intended for responses which already include the URL to be hydrated; if I add the `$merge` fields manually before calling it, I don't think I can simplify the code much. Also it doesn't have concurrency limits, and in theory we could be processing a thosand entries here (although the actual limits will be much lower until [T177462](https://phabricator.wikimedia.org/T177462) gets resolved).

I think the only extra feature `hydrateResponse()` has is deduplication, and reading lists cannot contain the same page twice.

v1/lists.yaml
tgr: AIUI `cache-control: max-age=0, s-maxage=0` means "should actually send the request", `cache-control: no-cache` means "must actually send the request". This is a POST endpoint that changes data and returns a success flag, so the latter seems more appropriate.

v1/lists.yaml
tgr: As above, the response does not have any pages.

v1/lists.yaml
tgr: Good question. I was thinking of adding the session cookie to the Swagger spec as a security scheme object with a vendor extension, and then making service-checker handle it. Do you think that makes sense? I am writing some mocha tests that have a flag for using nock (much like `tests/pagecontent/save_api.js`) so that could also be used to do smoke testing (including write endpoints) - I guess that would make `x-monitor` pointless?

In any case I'll remove it from this commit. It can be readded later of there is a sensible way of using it.

v1/lists.js
```diff
@@ -0,0 +1,190 @@
...
+     * @param {!String} continuation Continuation string returned by flattenContinuation().
+     * @return {!Object|undefined} Continuation object.
+     */
+    unflattenContinuation(continuation) {
+        const sanitizedContinuation = {};
+        if (typeof continuation === 'string') {
+            try {
+                continuation = JSON.parse(continuation);
+            } catch (e) {
+                this.options.log('error/unflatten', {
```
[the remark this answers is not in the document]
tgr: It doesn't log any details though. I guess the stack trace is pretty pointless but the raw json and the error message from the parser are helpful for debugging.

Eventually I'd like to use `mwUtil.encodePagingToken` or something similar so that users cannot trigger any error-level logging here, but I considered that less important than getting the service set up on beta.

tgr:
I can move things to .js if you prefer; using the yaml format seemed more readable to me. For `/lists/{id}/entries` I couldn't use it as there is no way to make subrequests for every item in a list via yaml. (I think it could be added via `x-route-filters` but that would be a bit hacky.)

Rebased on top of master; reading_lists_beta needs a pull from master.

I'm not sure what to do with the schema error. I have a definition like this:
<pre>
"definitions": {
      "list_common": {
       "name": {
            "type": "string",
            "example": "Planets"
        },
        "description": {
            "type": "string",
            "example": "Planets of the Solar System"
        }
    }
}
</pre>
and that triggers a `data.description should be string` error. That seems to be a bug in the schema checker. Is that something under our control or does it come with Swagger?

tgr:
The remaining test errors seem unrelated.

v1/lists.js
```diff
@@ -0,0 +1,192 @@
...
+     * @return {!Object} Continuation object.
+     */
+    unflattenContinuation(continuation) {
+        const sanitizedContinuation = {};
+        if (typeof continuation === 'string') {
+            try {
+                continuation = JSON.parse(continuation);
+            } catch (e) {
+                this.options.log('error/unflatten', {
+                    msg: e.stack,
```
Pchelolo: The stack is not a great message for a log entry, something descriptive would be much better.

v1/lists.js
Pchelolo: Actually concurrency limits seem like a good idea in general, I will add it l8r.

`fetchSummary` doesn't assume local endpoint, it will fetch whatever URI you provide it.

v1/lists.js
```diff
@@ -0,0 +1,192 @@
...
+            }
+        }, { concurrency: this.options.summaryConcurrency });
+    }
+
+    /**
+     * Verify that a domain is owned by Wikimedia and has a RESTBase API.
+     * @param {!String} domain
+     * @return {bool}
+     */
+    isValidDomain(domain) {
```
Pchelolo: There's mwUtils.getSiteInfo call - we could modify this check to verify that the site exists via fetching the SiteInfo - this will avoid hardcoding the projects. It will not add any latency because once the SiteInfo is fetched it's cached in memory.

tgr: Allowing users to make RESTBase send requests to third party websites is a bad idea, even if it's just a single request per domain per RESTBase deploy. It could be used to DDoS third-party websites, it could be used to DoS RESTBase by filling up its memory with data from arbitrary domains, depending on the implementation (how it interacts with mediawiki_auth_filter) it might even be used to steal session cookies. And it's not a reliable check, a third-party site could still run MediaWiki or pretend to do so.

Yes hardcoding MediaWiki domains sucks but the alternatives are more scary as far as I can see. (Apart from leaving hydration to the client which IMO makes more sense but that's not my decision to make.) I guess I should move the domain list to the configuration at least.

tgr: Actually, we can probably use `action=sitematrix` to get the data. Which is still Wikimedia-specific but now it's someone else's problem :)

Pchelolo: @tgr internet is not accessible from within the WMF RESTBase cluster so there's no way to call a third party website. If you're thinking about third-party installs where that might not be true, hard-coding WMF domains will not help the either.

tgr: Haven't thought of that. Anyway, using sitematrix seems like a nice approach that might be useful for other things (and in theory could work for third parties, although you'd have to mirror WMF's MediaWiki configuration very closely for that API to work). It does make Extension:SiteMatrix a soft dependency (there will be no summary hidration without it) - that doesn't seem to bad. Maybe there should be a way to disable domain verification for dev installs, though. Or a way to import WMF sites into the local SiteMatrix data.

d00rman: `siteInfo` would be more useful in this context, since in practice all RB processes end up requesting the site info for all of the domains, so there is no overhead for it.

Pchelolo: @tgr here's the [request we currently make to the API](https://github.com/wikimedia/restbase/blob/master/sys/action.js#L240-L248) in order to fetch the necessary information about a wiki. What are the advantages of the `sitematrix` you see over the current approach?

v1/lists.js
```diff
@@ -0,0 +1,151 @@
...
+     * @param {!HyperSwitch} hyper
+     * @param {!Object} req The request object as provided by HyperSwitch.
+     * @return {!Promise<Array>} The objects, enriched with summaries.
+     */
+    hydrateSummaries(entries, hyper, req) {
+        return mwUtil.getSiteMatrix(hyper, req).then((sitematrix) => {
+            return P.map(entries, (entry) => {
+                const site = sitematrix[entry.project];
+                if (site && !site.private) {
+                    const uri = `https://${entry.project}/api/rest_v1/page/summary/`
```
Pchelolo: So this is an interesting one here: if you go for the `/rest_v1` URI the request will leave RESTBase and hit Varnish. In case it's a cache miss in varnish it will come back to RESTBase and go to Cassandra for the summary. When doing these hydrations before for other things, going directly to Cassandra here (using internal `/sys` summary URI) was slightly faster. However that was before the PagePreviews feature was introduced, so most likely now the hit ratio in Varnish will be better and going to Varnish will be slightly faster.

Having said that, I still think we need to go to the internal URI for summary here. The difference in latency will be negligible but going to Varnish will mess up all the request rate statistics here. 

Here's an example how to go directly to RESTBase: https://github.com/wikimedia/restbase/blob/master/v1/related.js#L38-L39

v1/lists.js
```diff
@@ -0,0 +1,151 @@
...
+                    return mwUtil.fetchSummary(hyper, uri)
+                        .then((res) => {
+                            if (res) {
+                                entry.summary = res;
+                            }
+                            return entry;
+                        });
+                }
+                return entry;
+            }, { concurrency: this.options.summaryConcurrency });
```
Pchelolo: I still think duplicating code that mostly exists in the `hydrateSummaries` is not great and I would personally prefer if we mangle the result here to use the `$merge` and filter out private sites and then reuse the `hydrate` thingy. Just like [here](https://github.com/wikimedia/restbase/blob/master/v1/related.js#L31-L48)

Adding support for concurrency to the generic `hydrateResponse` is a great idea anyway and I will do that regardless of whether we decide to add it now or not although 10 is a bit too low I think. We can easily do 100.

Anyways feel free to disregard this comment, I don't have a very strong opinion here.

v1/lists.yaml
Pchelolo: According to http://www.faqs.org/rfcs/rfc2616.html section 9.5 POST responses are only cached if appropriate `cache-control` headers are present, which means that if the header is absent the response shouldn't be cached meaning that this is not needed.

v1/lists.yaml
```diff
@@ -0,0 +1,939 @@
...
+              headers:
+                content-type: application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/Lists/0.1"
+                cache-control: "no-cache"
+              body:
+                lists: '{{forward_to_mw.body.query.readinglists}}'
+                next: '{{flattenContinuation(forward_to_mw.body.continue)}}'
+    post:
+      tags:
+        - Reading lists
+      summary: Create a new lists for the current user.
```
Pchelolo: in `a new lists` "list" should not be plural

v1/lists.yaml
```diff
@@ -0,0 +1,939 @@
...
+                description: '{{request.body.description}}'
+                color: '{{request.body.color}}'
+                image: '{{request.body.image}}'
+                icon: '{{request.body.icon}}'
+                token: '{{request.query.csrf_token}}'
+            return:
+              status: 200
+              headers:
+                content-type: application/json; charset=utf-8
+                cache-control: "no-cache"
```
Pchelolo: Again not sure it's needed for the POST request

v1/lists.yaml
```diff
@@ -0,0 +1,939 @@
...
+      summary: Update a list.
+      description: |
+        List must belong to current user and request must be authenticated with
+        a MediaWiki session cookie.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      parameters:
+        - name: id
+          in: path
+          schema:
```
Pchelolo: You don't really need a schema for an int, can just do `type: integer`

Pchelolo:
T176689 is fixed by https://github.com/wikimedia/hyperswitch/pull/84 so you can remove the TODO when that's merged

v1/lists.js
berndsi: @Pchelolo Couldn't we still go to varnish but give it a special X-Analytics header value to indicate that this is an internal request?

v1/lists.js
tgr: I guess I just feel bad about allowing RESTBase to send requests to domains it knows nothing about. I can't really think of any way it could be exploited in a DMZ network, so I'll discard it and use siteinfo instead if you feel strongly about it. Defense in depth is always nice though, and the overhead is minimal (one fetch per thread lifetime).

Alternatively, I could move this code to the MediaWiki extension (which already tries to deal with domain parsing to some extent, to figure out which domains can be represented as interwikis) and just flag domains as unsafe, or even reject them. That's probably a good idea even if `sitematrix` or `siteinfo` does get used here.

As for duplication, I don't see much room for improvement there. `siteinfo` is per-domain, `sitematrix` is global so they have to be cached separately. `sitematrix` is not a query module (it should be, but it was written before that concept existed, I think) so the API requests cannot be merged. It's not necessarily present on a non-WMF installation / dev machine / etc so error handling needs to be different. And the rest of the code is just response formatting (`sitematrix` is old and weird; it has some fairly unfortunate choices such as using a hash as some kind of mix between a normal and an associative array; it doesn't honor `formatversion=2`) which is specific to that endpoint. I really don't see what could be merged there.

v1/lists.js
Pchelolo: @berndsi why bother? there's no real benefit in doing that..

v1/lists.js
Pchelolo: @tgr we can change that behaviour. Just need to ask @berndsi if that will affect mobile clients that's using feeds. Or we can put that under a parameter flag like`removeFailed = true|false`

v1/lists.js
Pchelolo: hm @tgr interesting question and a very valid concern.. Ok, I think I'd be fine with going to Varnish here. 

One thing is that I'd avoid hard-coding `/api/rest_v1` - in the `mwUtils.getSiteInfo` we have a property `baseURITemplate` that's configurable - better to use that to construct the URIs so that if a wiki has no `api/rest_v1` rewrite RESTBase would still be usable.

sys/action.js
```diff
@@ -296,11 +399,21 @@ module.exports = (options) => {
...
+                    all: {
+                        operationId: 'mwRawApiQuery'
+                    }
+                },
                 '/siteinfo': {
                     all: {
                         operationId: 'mwApiSiteInfo'
                     }
                 },
+                '/sitematrix': {
```
Pchelolo: I wanna reiterate a question that was lost somewhere along the review process - what's the advantage of using the sitematrix instead of already implemented siteinfo?

v1/lists.js
berndsi: @Pchelolo My comment was just prompted by your earlier comment:

> so most likely now the hit ratio in Varnish will be better and going to Varnish will be slightly faster.

Not saying that we have to do it, obviously, just wanted to note that we could if the benefits outweigh the cons.

sys/action.js
tgr: @Pchelolo I answered that [here](https://github.com/wikimedia/restbase/pull/887#discussion_r144390020), don't have any arguments beyond that. I'l replace it with siteinfo.

Pchelolo:
Looking so much better! There's a lint error, but other than that I think we're ready

PR #888, opened 2017-10-12
Handle null values in mwUtil.hydrateResponse

tgr:
Avoid errors caused by `typeof null === 'object'`.


Pchelolo:
+1, lemme wait for travis before merging

PR #892, opened 2017-10-18
Summary new storage and a proxy

Pchelolo:
cc @wikimedia/services


v1/summary_proxy.js
```diff
@@ -0,0 +1,49 @@
...
+                            uri: oldURI,
+                            headers: req.headers,
+                            query: req.query
+                        }),
+                        newContent: hyper.get({
+                            uri: newURI,
+                            headers: req.headers,
+                            query: req.query
+                        })
+                        .catch(e => undefined),
```
d00rman: Would be good to at least log the failure so that we know something's up.

v1/summary_proxy.js
```diff
@@ -0,0 +1,49 @@
...
+                    })
+                    .get('oldContent');
```
d00rman: If this is a no-cache request, i think we can afford to check the output of both and compare them and log if there are any differences, just like we did for mobileapps.

v1/summary_proxy.js
```diff
@@ -0,0 +1,49 @@
...
+const mwUtil = require('../lib/mwUtil');
+const URI = require('hyperswitch').URI;
+const P = require('bluebird');
+
+const spec = require('hyperswitch').utils.loadSpec(`${__dirname}/summary.yaml`);
+
+module.exports = (options) => {
+    const isNewStorageEnabled = (() => {
+        const checkEnableRegex = options.new_storage_enabled
+            && mwUtil.constructRegex(options.new_storage_enabled);
```
d00rman: Since the options are static, it would be more efficient to construct the regex once on start-up and then use it in the compares, instead of constructing it every time.

Pchelolo: it's not doing it every time, it's memoized. Just wanted to use a fancy technic :)

PR #894, opened 2017-10-18
Use new storage for Parsoid and make a proxy

Pchelolo:
Time to move on! 

Right now without enabling anything in the config it's a no-op.

Prior discussion is in #878 

cc @wikimedia/services

### DIFF
--- a/sys/parsoid_proxy.js
+++ b/sys/parsoid_proxy.js
@@ -70,7 +70,8 @@ class ParsoidProxy {
             return hyper.request(oldRequest);
         }
         if (backend === 'new') {
-            return hyper.request(newRequest);
+            return hyper.request(newRequest)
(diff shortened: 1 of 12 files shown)


sys/parsoid_proxy.js
```diff
@@ -0,0 +1,174 @@
...
+const URI = HyperSwitch.URI;
+
+const mwUtil = require('../lib/mwUtil');
+
+const spec = HyperSwitch.utils.loadSpec(`${__dirname}/parsoid.yaml`);
+
+class ParsoidProxy {
+    constructor(options) {
+        this.options = options || {};
+        if (this.options.backends) {
```
d00rman: `options.parsoid.backends` is passed in here, so this will never be true. You need `this.options`

Pchelolo: Fixed.

PR #896, opened 2017-10-18
Switch definitions to new storage

Pchelolo:
I think after we switch summary we don't need to make a proxy for the definitions.

cc @wikimedia/services

### DIFF
--- a/v1/definition.yaml
+++ b/v1/definition.yaml
@@ -85,7 +85,7 @@ paths:
         # Set up a simple key-value bucket.
         - init:
             method: 'put'
-            uri: /{domain}/sys/key_value/term.definition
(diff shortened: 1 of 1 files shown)


v1/definition.yaml
```diff
@@ -85,7 +85,7 @@ paths:
         # Set up a simple key-value bucket.
         - init:
             method: 'put'
-            uri: /{domain}/sys/key_value_old/term.definition
+            uri: /{domain}/sys/key_value/term.definition
```
d00rman: For the schema-creation process of definitions, we [added the `-ng` prefix](https://phabricator.wikimedia.org/T179420#3738999) which I think is good for compatibility. We should add it here as well.

PR #899, opened 2017-10-30
Switch parsed stashing to new storage for new parsed module implementation

Pchelolo:
cc @wikimedia/service

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -193,17 +193,6 @@ class ParsoidService {
         };
     }
 
-    getBucketURI(rp, format, tid) {
-        const path = [rp.domain, 'sys', 'key_rev_value', `parsoid.${format}-2`, rp.title];
-        if (rp.revision) {
-            path.push(rp.revision);
-            if (tid) {
-                path.push(tid);
-            }
-        }
-        return new URI(path);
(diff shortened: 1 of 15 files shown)


d00rman:
LGTM, but let's not merge it until we have the Cassandra 3 tables in place

d00rman:
Hm, Travis does not concur with my +1, lots of `Error: Header \"%s\" value must not be undefined` errors in there.

sys/parsoid.js
```diff
@@ -227,15 +216,7 @@ class ParsoidService {
...
-        // TODO: We're adding the -ng only to avoid table name clashing in dev cluster. Remove!
-        const path = [rp.domain, 'sys', 'key_rev_value', `parsoid.stash.${format}-ng`, rp.title];
-        if (rp.revision) {
-            path.push(rp.revision);
-            if (tid) {
-                path.push(tid);
-            }
-        }
-        return new URI(path);
+        return this.getStashBucketURI(rp, format, tid);
```
d00rman: Hm, the new tables have been created with the `-ng` suffix, so we need to use it here.

Pchelolo:
Fixed node 8 test failure

PR #900, opened 2017-10-30
Use only new summary storage

Pchelolo:
Revert everything we did to create a summary proxy and only use the new storage. 

Also rebased on the most recent developments.

cc @wikimedia/services


d00rman:
LGTM, but I think we might want to run a dump of at least enwki and dewiki to ensure the smallest impacts on users before merging and deploying. Alternatively, we may want to think about migrating the content, but IMO that's an overkill.

Pchelolo:
@d00rman we can switch to the new storage only without deploying this by using the new_storage_enabled config. It's for the future.

Pchelolo:
rebased, dumps are done, we're ready to go with this.

PR #907, opened 2017-11-03
Reading lists: add tests

tgr:
Work in progress; I'm hoping to get guidance on how to finish it. The problems are:
* nock and hyperswitch seem interact poorly. When there is a request to an URL that has not been mocked, nock throws an error, but instead if that bubbling up to the test runner, the whole thing breaks (there is no output, subsequent tests do not even get run). I have to step through in a debugger / litter console.log statements around (half of which mysteriously fail to work) to get some idea of what's happening, making testing a very frustrating process. Am I causing that somehow?
* nock error messages seem fairly useless (e.g. I miss the expected URL/payload by just a tiny bit and I just get something nonspecific message about an expected request not having happened. Is that just how nock is or am I using it wrong?
* How should I handle local test running in general? The existing test server config wasn't written with vagrant support in mind; I'd need some puppet-generated config file to inject the port number into. What's the best way to integrate that with the existing config? Maybe extract the whole `config` object from `server.js` into a file that can be specified via env vars? Also right now the tests seem to hardcode where they run. If I wanted tests which can be run on production, beta, local/vagrant, what would be the best way to add that capability to `server.js`? (Right now it's a theoretical question since I'm using nock so the tests have to be run locally. I was thinking of eventually adding two alternative setup phases to each test - mocking with nock vs. setting up a fixture via real requests to the API, so I can use some kind of config flag to switch the tests between unit(ish) and smoke test mode.)

### DIFF
--- a/test/features/lists.js
+++ b/test/features/lists.js
@@ -1,0 +1,1182 @@
+'use strict';
+
+const nock = require('nock');
+const assert = require('../utils/assert.js');
+const server = require('../utils/server.js');
+const querystring = require('querystring');
+const preq = require('preq');
+
+
+describe('reading lists', function() {
+    const csrfToken = '<mock>';
+    const sessionCookies = '<mock>';
+    const config = {
+        domain: 'dev.wiki.local.wmftest.net',
+        baseURL: 'http://localhost:7231/dev.wiki.local.wmftest.net/v1',
+        apiDomain: 'http://localhost:10123',
+        apiPath: '/w/api.php',
+    };
+
+    function getApi() {
+        return nock(config.apiDomain, {
+            reqheaders: {
+                'Cookie': sessionCookies,
+                'Host': config.domain,
+            },
+        })
+        .defaultReplyHeaders({
+            'Content-Type': 'application/json; charset=utf-8',
+        });
+    }
+
+    // workaround for https://github.com/node-nock/nock/issues/852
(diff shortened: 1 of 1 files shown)


Pchelolo:
Oh yes, `nock` is pretty terrible and really hard to use.

A couple of very quick tips on it that will solve some of the issues you've described. I will dive into the actual code a bit later in the day:

1. You should run `nock.cleanAll()` in a finally block after each test - if you don't do that failure of one test will bring down all the tests using nock as the unhandled interceptors are global, also .done should be called in `then` not `finally` as in case of error this function does something weird to the nock state. We use `nock` in `change-propagation` tests pretty successfully by sticking to the following pattern for an individual test: https://github.com/wikimedia/change-propagation/blob/master/test/feature/update_rules.js#L90-L120

2. There's a feature in nock to allow [unmocked](https://github.com/node-nock/nock#allow-unmocked-requests-on-a-mocked-hostname) requests to pass through - that might save you some time on mocking SiteInfo etc

3. For debugging there's a `logging` option in `nock` - that's extremely useful the debugging.

And regarding vagrant - I didn't quite understand your question. What do you want to achieve?

mateusbs17:
This code/task continued and has some updates at https://github.com/wikimedia/restbase/pull/1055

Pchelolo:
Replaced by #1055

PR #908, opened 2017-11-07
Don't store the feed results

Pchelolo:
We've reached the consensus that we will not store the results for the feed in Cassandra. This will eventually allow to simplify the code a lot, but right now I wanted to make the first experiment with `storeHistory === false` - it's equivalent to having no buckets at all - to access the latency increase and see if we eventually really want that.

Bug: https://phabricator.wikimedia.org/T179412

cc @wikimedia/services @coreyfloyd @berndsi

### DIFF
--- a/lib/base_feed.js
+++ b/lib/base_feed.js
@@ -69,7 +69,7 @@ class BaseFeed {
                 body: res.body
             });
         };
-        const requesCurrentContentFromMCS = () => this._makeFeedRequests(hyper, req, false)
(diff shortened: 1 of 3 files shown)


d00rman:
What's the status of this @Pchelolo ? Can we move forward with it?

Pchelolo:
We have to rebase it and deploy?

d00rman:
Yeah

Pchelolo:
Rebased.

PR #910, opened 2017-11-16
Remove order-related information

tgr:
Bug: T180402


Pchelolo:
I don't really understand the reasoning behind removing that, but as a code review it's a +1 or course.

tgr:
AIUI apps decided they don't really need it, and the backend implementation was a performance concern with the current constraints (the discussion of which is a bit fractured but see [T177462](https://phabricator.wikimedia.org/T177462) and [T168984#3659835](https://phabricator.wikimedia.org/T168984#3659835)) so removing it makes meeting performance expectations simpler.

Pchelolo:
Ya thank you' asked out of curiosity

PR #913, opened 2017-11-24
Stabilize example config

Pchelolo:
Make the example configs actually work so that third party or dev installations can use it.

cc @wikimedia/services


config.example.yaml
```diff
@@ -60,10 +60,6 @@ services:
                     retry_delay: 250
                     retry_limit: 10
                     show_sql: false
-                  citoid:
```
d00rman: Why is this being removed?

CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+## 0.18.0 (2017-11-27)
+
+**Breaking changes**
+
+- `RESTBase` now depends on Cassandra 3.11. In case SQLite backen is used no changes us required.
```
d00rman: `s/backen/back-end/`

CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
...
+- After updating `RESTBase` new keyspaces will be created for everything, so it's effectively starting the storage from scratch. All the content stored in `RESTBase` can be rerendered, so no need to transfer the data, but old keyspaces can be deleted afterwards. New keyspace names have `_ng` suffix.
```
d00rman: `s/keyspace names have _ng/keyspace names have the _ng/`

Also, let's limit the line lengths to 80 chars.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "0.17.0",
+  "version": "0.18.0",
```
d00rman: It's not time to release the new version yet. We need to switch everything over before that can happen.

config.example.yaml
Pchelolo: Because the example.yaml config don't specify citoid or mathoid modules, so the setting is useless.

package.json
Pchelolo: I that case maybe let's not add the changelog now but just fix the example config and postpone adding the changelog until we actually make the release?

d00rman: Sure, +1

PR #914, opened 2017-11-29
Reading lists: add sorting

tgr:
Add `sort` parameter to `/lists/` and `/lists/{id}/entries/`, set sort/dir for MW API accordingly.

Also add sort parameters to `/lists/changes/since/{date}` (has no effect since those are the default settings, but good for self-documentation).

(Also get rid of a bunch of old FIXME comments. I wanted to see if it's possible to just return an object with multiple keys from a helper method and merge that with the rest of the query parameters, but I'm getting used to the current syntax.)

Bug: [T180402](https://phabricator.wikimedia.org/T180402)

### DIFF
--- a/v1/lists.js
+++ b/v1/lists.js
@@ -70,18 +70,34 @@ class ReadingLists {
     }
 
     /**
+     * Get the sort parameters for the action API.
+     * @param {!String} sort Sort mode ('name' or 'updated').
+     * @return {!Object} { sort: <rlsort/rlesort parameter>, dir: <rldir/rledir parameter> }
+     */
+    getSortParameters(sort) {
+        sort = sort || 'updated';
+        return {
+            sort,
+            dir: (sort === 'updated') ? 'descending' : 'ascending',
+        };
+    }
+
+    /**
      * Handle the /list/{id}/entries endpoint (get entries of a list).
      * @param {!HyperSwitch} hyper
      * @param {!Object} req The request object as provided by HyperSwitch.
      * @return {!Promise<Object>} A response promise.
      */
     getListEntries(hyper, req) {
+        const sortParameters = this.getSortParameters(req.query.sort);
         return hyper.post({
             uri: new URI([req.params.domain, 'sys', 'action', 'rawquery']),
             body: {
                 action: 'query',
                 list: 'readinglistentries',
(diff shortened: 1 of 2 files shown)


v1/lists.js
```diff
@@ -82,6 +94,8 @@ class ReadingLists {
                 action: 'query',
                 list: 'readinglistentries',
                 rlelists: req.params.id,
+                rlesort: this.getSortParameters(req.query.sort).sort,
```
d00rman: Nit: this is the same as `rlesort: req.query.sort`. If you do want to call the function nevertheless and use its return value (because the logic inside it might change), better to do it before returning the request and store the result in a variable so that it's not called twice with the same param as the return value can be reused on the line below. That not only saves time, but also reduces the memory footprint.

v1/lists.yaml
```diff
@@ -346,6 +357,15 @@ paths:
           description: Continuation parameter from previous request
           type: string
           required: false
+        - name: sort
+          in: query
+          description: |
+            Sort order
+            - `name` (default): by page title, ascending;
+            - `updated`: by last modification date, descending.
```
berndsi: Maybe more a client question. I'm leaning towards `updated` should be the default. Where is the default behavior implemented? Do we have tests?

tgr: Defaults are on implemented in the MediaWiki API. I should probably implement it in the proxy though, it's not nice to depend on the defaults of another API.

I have no opinion on default behavior, I'll change it to `updated`.

Tests are still in WIP: #907 
On the MW side, there are DB tests but no API tests. I'll get there eventually; the API layer has very little logic in it though.

tgr:
RESTBase does validate parameters which have an enum constraint (although does not set the default; I'll do that manually).

v1/lists.yaml
berndsi: Sounds good. My opinion on the default behavior is not very strong. Before you change it it might be good to check with @dbrant and whoever works on the iOS portion of this, what they actually prefer.
The main reason I brought it up is because I think that `updated` probably makes for a faster sync experience but I haven't fully thought through if the reverse chronological order makes issues more likely or not.

tgr:
The remaining test errors seem unrelated.

PR #915, opened 2017-11-29
Add ability to select domain HTTP scheme for PDF exports

Silex:
My wiki server runs on HTTP, not HTTPS.

There is probably a better way to do this, but since issues are closed on this repository I'm forced to create a pull request.

### DIFF
--- a/v1/pdf.yaml
+++ b/v1/pdf.yaml
@@ -61,7 +61,7 @@ paths:
             request:
               method: get
               # Note: The title needs to be encoded twice.
-              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url=https://{{domain}}/wiki/{_encodeURIComponent(title)}%3Fprintable=yes'
+              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url={{default(options.scheme, "https")}}://{{domain}}/wiki/{_encodeURIComponent(title)}%3Fprintable=yes'
             return:
               status: 200
               headers:
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -55,6 +55,7 @@ default_project: &default_project
           cache_control: s-maxage=600, max-age=600
           uri: https://pdfrender-beta.wmflabs.org
           secret: secret
+          scheme: https
         transform:
           cx_host: https://cxserver-beta.wmflabs.org
         skip_updates: false


Pchelolo:
Issues can be filed on http://phabricator.wikimedia.org 

What're you trying to achieve here? Are you running the electron renderer service in your environment?    If so, please beware that WMF is going to abandon the electron service  very soon and replace it with a new one, that most likely will have different API

Silex:
Yes, we are running the electron renderer service in our environment.

What, yet another PDF solution replacement? Is that planned for 1.30?

Pchelolo:
@Silex the electron renderer has proven to be unreliable at the WMF scale, see https://phabricator.wikimedia.org/T174916 so we're developing a yet another one. It's not gonna affect the MW side of things, so it's not bound to any particular release. If you're happy with how electron works you can keep it.

v1/pdf.yaml
```diff
@@ -61,7 +61,7 @@ paths:
             request:
               method: get
               # Note: The title needs to be encoded twice.
-              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url=https://{{domain}}/wiki/{_encodeURIComponent(title)}%3Fprintable=yes'
+              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url={{options.scheme}}://{{domain}}/wiki/{_encodeURIComponent(title)}%3Fprintable=yes'
```
Pchelolo: Could you update it to default to https? something like `{{default(options.scheme, "https")}}`?

Silex:
Alright thanks for the comment, I did your suggestion.

I wondered if there's a way to reuse `baseUriTemplate` or something instead of specifying the scheme explicitely?

Pchelolo:
If you specify a template here, it will be used inside of another template and I don't think that will work without a bigger restructuring of the code. So the solution you've proposed seems like the easiest one.

Silex:
Alright, we are good to go then :wink: 

Thanks for the review.

Pchelolo:
👍 Will merge closer to the deployment time

d00rman:
@Silex I agree that the constant changes are hard to keep up with currently. We are in the middle of some big changes, so hopefully things will stabilise soon. As for the change in PDF rendering back-end, the Wikimedia production environment will soon switch to a [puppeteer-based solution](https://github.com/wikimedia/mediawiki-services-chromium-render) but that shouldn't affect the RESTBase side of things at all, so you could continue using the current Electron solution indefinitely or (better) switch at a later time.

Silex:
A'ight :+1:

PR #916, opened 2017-12-04
Fix cxserver MT for transform-global

santhoshtr:
The HTML content was not passed to backend server

Bug: T173801


d00rman:
Duh, so obvious! Thank you for looking into it, @santhoshtr!

PR #918, opened 2017-12-05
Add configurable new summary endpoint and fix eslint

Pchelolo:
Allow using MCS for summary endpoint. It's a temporary state - we will configure beta cluster to use MCS and production to use old implementation - this will allow us to unblock reading to test it in beta and not block us from doing production deployments.

Also fixed eslint.

Some test are failing related to different expectation for sections - I guess Parsoid was deployed in beta. Will update the tests separately.

cc @wikimedia/services @coreyfloyd @berndsi @mdholloway


d00rman:
For future reference, we should not put multiple fixes/unrelated things in one PR, as it makes it harder to both review it and revert it later in case something goes wrong.

PR #919, opened 2017-12-05
Update section tests to match new parsoid output

Pchelolo:
1. Update section tests expectations
2. In case we're using the content-type upgrade filter with sections it didn't really play well - we were expecting HTML content-type but instead got application/json and tried to upgrade. Fixed that.

### DIFF
--- a/test/features/parsoid/transform.js
+++ b/test/features/parsoid/transform.js
@@ -111,7 +111,7 @@ parallel('transform api', function() {
         .then(function (res) {
             assert.deepEqual(res.status, 200);
             assert.contentType(res, contentTypes.html);
-            var pattern = /^<h2.*> Heading <\/h2>$/;
+            var pattern = /^<section data-mw-section-id="[^"]+" id="[^"]+"><h2 id="Heading"> Heading <\/h2><\/section>$/;
             if (!pattern.test(res.body)) {
                 throw new Error('Expected pattern in response: ' + pattern
                         + '\nSaw: ' + res.body);
@@ -203,8 +203,8 @@ parallel('transform api', function() {
             body: {
                 changes: {
                     mwAQ: [],
-                    First_Section: [{ html: "<h2>First Section replaced</h2>" }],
-                    Second_Section: [{ html: "<h2>Second Section replaced</h2>" }]
+                    mwAw: [{ html: "<h2>First Section replaced</h2>" }],
+                    mwBA: [{ html: "<h2>Second Section replaced</h2>" }]
                 }
             }
         })
@@ -227,8 +227,8 @@ parallel('transform api', function() {
             body: {
                 changes: JSON.stringify({
                     mwAQ: [],
-                    First_Section: [{ html: "<h2>First Section replaced</h2>" }],
-                    Second_Section: [{ html: "<h2>Second Section replaced</h2>" }]
+                    mwAw: [{ html: "<h2>First Section replaced</h2>" }],
+                    mwBA: [{ html: "<h2>Second Section replaced</h2>" }]
                 }),
                 scrub_wikitext: 'true'
             }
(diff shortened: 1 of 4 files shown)


test/features/parsoid/transform.js
```diff
@@ -80,214 +77,214 @@ parallel('transform api', function() {
...
-                throw new Error('Expected pattern in response: ' + pattern
-                        + '\nSaw: ' + res.body);
+                throw new Error(`Expected pattern in response: ${pattern
+                }\nSaw: ${res.body}`);
             }
         });
     });
 
-    it('wt2html with body_only', function () {
+    it('wt2html with body_only', () => {
```
subbuss: With body only, section tags should be suppressed. we have a mocha test for it.

v1/content.yaml
```diff
@@ -152,6 +152,7 @@ paths:
           required: false
       produces:
         - text/html; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/HTML/1.5.0"
```
subbuss: Move to 1.6.0

subbuss:
Your comment at the top should be part of the commit message.

Also, ideally, you should split the ES6 changes from the section test changes so it is easier to spot section related changes. But, maybe for future PRs like this if you don't want to do for this round.

v1/content.yaml
Pchelolo: @subbuss We're not ready to bump the expected version yet - this is just to fix the failing tests. We still want to be able to deploy RESTBase one time before we will be ready to for you to bump the version. So we will bump it in a separate one.

Pchelolo:
@subbuss Ye, good call, after I've stashed it all in one place I've realized It was better not to.. Will separate them

test/features/parsoid/transform.js
```diff
@@ -111,7 +111,7 @@ parallel('transform api', function() {
         .then(function (res) {
             assert.deepEqual(res.status, 200);
             assert.contentType(res, contentTypes.html);
-            var pattern = /^<h2.*> Heading <\/h2>$/;
+            var pattern = /^<section data-mw-section-id="[^"]+" id="[^"]+"><h2 id="Heading"> Heading <\/h2><\/section>$/;
```
Pchelolo: @subbuss I've seen your comment regarding this that but it disappeared for some reason (maybe me force-pushing stuff)

> With body only, section tags should be suppressed. we have a mocha test for it.

Can I have a little more background on that? Parsoid should be stripping them right? That means we don't properly pass the `body_only` parameter from RESTBase as I understand...

v1/content.yaml
subbuss: ok

test/features/parsoid/transform.js
subbuss: I meant: Parsoid suppresses section tagging in body_only mode because VE requested that. https://github.com/wikimedia/parsoid/commit/e6a6f26141506a93ab5f47c09af0f9ecb49f67d6

Pchelolo: Oh, ok. This suggests that we have some other bug and `bodyOnly` mode is not passed properly to Parsoid. Looking into that.

Pchelolo: Hm... Perhaps that fix's not deployed in beta yet? 

`curl -XPOST http://parsoid-beta.wmflabs.org/en.wikipedia.org/v3/transform/wikitext/to/html --data "wikitext=''foo''&body_only=1" ` gives me `<section data-mw-section-id="0" data-parsoid="{}"><p data-parsoid='{"dsr":[0,7,0,0]}'><i data-parsoid='{"dsr":[0,7,2,2]}'>foo</i></p></section>` - so it does have the section wrapping..

Pchelolo: Ya, just verified - it's not deployed in beta. I will rollback this particular change making a mental note that it's fine that this test fails.

subbuss: ok, let me deploy to beta now.

subbuss: oh, actually, it will require a patch, etc. do you need this now?

Pchelolo: @subbuss no rush here at all, we can live with a failing test for a few days

PR #921, opened 2017-12-06
Add CORS headers to error responses as well

Pchelolo:
Bug: https://phabricator.wikimedia.org/T182103

cc @wikimedia/services @milimetric @niedzielski

### DIFF
--- a/lib/security_response_header_filter.js
+++ b/lib/security_response_header_filter.js
@@ -50,7 +50,7 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
         resPromise = next(hyper, req);
     }
 
-    return resPromise.then((res) => {
+    const addHeaders = (res) => {
         if (!res.headers) {
             res.headers = {};
         }
@@ -109,5 +109,7 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
         // Finally, assign the csp header variants
         cspHeaderVariants.forEach((name) => { rh[name] = csp; });
         return res;
-    });
+    };
+
+    return resPromise.then(addHeaders, addHeaders);
 };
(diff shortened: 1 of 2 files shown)


lib/security_response_header_filter.js
```diff
@@ -109,5 +109,10 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
         // Finally, assign the csp header variants
         cspHeaderVariants.forEach((name) => { rh[name] = csp; });
         return res;
+    };
+
+    return resPromise.then(addHeaders, (e) => {
+        e = addHeaders(e);
```
d00rman: I think we want to make sure `e` is a response here, and not an Error.

Pchelolo: `e` should be `HTTPError` here, shouldn't it?

d00rman: Yes, but we allow things like `throw new Error(blah)` too in the code, so it might not be what we expect it to be afaik.

lib/security_response_header_filter.js
```diff
@@ -109,5 +110,18 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
...
         cspHeaderVariants.forEach((name) => { rh[name] = csp; });
         return res;
+    };
+
+    return resPromise.then(addHeaders, (e) => {
+        if (e.constructor.name !== 'HTTPError') {
+            e = new HTTPError({
+                status: 500,
+                body: {
+                    error: e
```
d00rman: This should resemble more the way [hyperswitch does it](https://github.com/wikimedia/hyperswitch/blob/fa0a97c35d62d96f2c6094bcd1503457b0b931ac/lib/server.js#L369-L387)

Pchelolo: Indeed. I was trying to find that code... 

I will make a patch for HyperSwitch to expose that as a factory method in HTTPError to avoid copy-paste

d00rman: Even better! :+1:

Pchelolo:
Stalled until https://github.com/wikimedia/hyperswitch/pull/87

PR #922, opened 2017-12-07
Remove color/image/icon

tgr:
Follows up [Icc712cb8abc025b992ed25afbf4f67d4e67386f9](https://gerrit.wikimedia.org/r/#/c/395666/).

Bug: [T180092](https://phabricator.wikimedia.org/T180092)

### DIFF
--- a/v1/lists.yaml
+++ b/v1/lists.yaml
@@ -29,18 +29,6 @@ x-yaml-anchors:
     description:
       type: string
       example: "Planets of the Solar System"
-    color:
-      type: string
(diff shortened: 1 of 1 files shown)


d00rman:
LGTM. Just a note: it's better to first remove things from the public API response and then remove it from the back-end, so as to minimise inconsistencies and problems (stating this for future changes).

tgr:
In this case the motivation was that the backend is going to be deployed to production sooner (and that would make schema changes more complicated).

PR #924, opened 2017-12-08
Merge reading lists feature branch

tgr:
Bug: [T181107](https://phabricator.wikimedia.org/T181107)


d00rman:
@tgr the PR needs a rebase against current master.

Pchelolo:
@tgr merge conflicts...

projects/dev.yaml
```diff
@@ -37,6 +37,8 @@ paths:
               x-modules:
                 - path: v1/citoid.js
                   options: '{{options.citoid}}'
+                - path: v1/lists.js
+                  options: '{{options.lists}}'
```
Pchelolo: Just to verify I'm correct - the options are not actually ever used, right? I'm asking cause I don't see them being specified in the `config.wikimedia.yaml` so I guess this was added for the future uses and we don't need to accompany this PR with a prod config change.

Pchelolo:
I've made a final pass through the code and everything looks good. After a rebase this is good to go 🎉

projects/dev.yaml
tgr: Yeah, it's not used currently. I will probably use it for [T180416](https://phabricator.wikimedia.org/T180416) but nothing substantial.

tgr:
Rebased. I don't think the test error is related.

I added one more trivial patch (#926).

PR #925, opened 2017-12-08
Bump Parsoid html version to 1.6.0

Pchelolo:
In preparation for deploy.

The tests will fail right now cause Parsoid is not deployed. To make the tests pass we need to
1. Deploy version bump to beta
2. Merge #919 
3. Rebase this one

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -107,8 +107,8 @@ wiktionary_project: &wiktionary_project
 # Hacky way to parametrize RESTBase tests. TODO: Move to config?
 test:
   content_types:
-    html: text/html; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/HTML/1.5.0"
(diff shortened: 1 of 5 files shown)


Pchelolo:
I've tested that the storage transition together with a version bump play well with VE in my vagrant, but I would still like to do several tests in beta before we proceed with this into production as there's so many things changing simultaneously that it's super easy to overlook something.

Here are the tests I have in mind:
1. Start editing a page in VE before deploying Parsoid/RESTBase, save after all done (plus with a page that uses a template that's edited after VE opening)
2. Start an edit after Parsoid is deployed but RB is not, save after RB is deployed (plus a page that uses a template that is edited after VE opening)

Pchelolo:
@d00rman btw, I gave you push access to my RESTBase fork, so if you work on this while I'm sleeping you can rebase my PR yourself.

PR #929, opened 2017-12-11
Fix failure if there's no content-type returned

Pchelolo:
https://phabricator.wikimedia.org/T182619

### DIFF
--- a/lib/ensure_content_type.js
+++ b/lib/ensure_content_type.js
@@ -47,7 +47,13 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
             return res;
         }
 
-        if (res.headers && res.headers['content-type'] !== expectedContentType) {
+        if (res.headers && res.headers === undefined) {
+            delete res.headers['content-type'];
+        }
+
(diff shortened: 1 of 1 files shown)


Pchelolo:
Self-merge to deploy to beta cluster to allow testing of Page Previews T182639

lib/ensure_content_type.js
```diff
@@ -47,7 +47,13 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
             return res;
         }
 
-        if (res.headers && res.headers['content-type'] !== expectedContentType) {
+        if (res.headers && res.headers === undefined) {
```
d00rman: I fail to see how this will ever evaluate to `true`. If `res.headers === undefined` then `res.headers` will always yield a false outcome, which means that the outcome will always be `false`. I also fail to see the benefit of deleting the `content-type` key in this circumstance.

Pchelolo: Oh this is obviously a bug, it should read `res.headers['content-type'] === undefined`

PR #931, opened 2017-12-12
Try old bucket in case original content TID was not matching with the received TID

Pchelolo:
Seems like we don't need this after all.

Bug: https://phabricator.wikimedia.org/T182728


Pchelolo:
WE will not proceed with this path.

PR #935, opened 2017-12-18
Switch Mathoid to new storage and remove references to Cassandra 2

d00rman:
Perform the last migration of Mathoid storage and clean up the code and configs by removing references to Cassandra 2 storage.

Bug: [T179419](https://phabricator.wikimedia.org/T179419)
Bug: [T179416](https://phabricator.wikimedia.org/T179416)


Pchelolo:
Seems that we can go even further and get rid of the `-ng` suffix everywhere, but we can do that as a separate PR.

d00rman:
> Seems that we can go even further and get rid of the -ng suffix everywhere, but we can do that as a separate PR.

I wonder if that would be backwards-compatible. If not, I'd be against it.

Pchelolo:
> I wonder if that would be backwards-compatible. If not, I'd be against it.

In most places it should be compatible. The only thing I think we wouldn't be able to change is the actual key space names in Cassandra.

d00rman:
> In most places it should be compatible. The only thing I think we wouldn't be able to change is the actual key space names in Cassandra.

Indeed. But yeah, let's concern ourselves with that in a follow-up PR. If we do it, we should remember to do it before we release a new version of RESTBase ;)

PR #936, opened 2017-12-18
Parsoid: Add the time_to_live config option to projects

d00rman:
The Parsoid module uses the time_to_live configuration option to determine whether content has half-expired and thus needs to be re-rendered. Alas, this config option if not passed to it by the projects. This PR fixes that.

### DIFF
--- a/projects/dev.yaml
+++ b/projects/dev.yaml
@@ -83,6 +83,8 @@ paths:
                   options:
                     parsoidHost: '{{options.parsoid.host}}'
                     response_cache_control: '{{options.purged_cache_control}}'
+                    grace_ttl: '{{default(options.parsoid.grace_ttl, 86400)}}'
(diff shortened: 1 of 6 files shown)


Pchelolo:
Should we better reuse `options.grace_ttl` for that?

d00rman:
> Should we better reuse options.grace_ttl for that?

If we reuse grace_ttl, then we shouldn't have two different configuration options to begin with.

Pchelolo:
Ye, that's exactly what I mean. I think that `time_to_live` is just a type and it should've been `grace_ttl`

d00rman:
> Ye, that's exactly what I mean. I think that time_to_live is just a type and it should've been grace_ttl

Agreed. Let's consolidate them in a separate PR and resolve the issue at hand here?

Pchelolo:
ok, I'm fine with a separate PR

PR #937, opened 2018-01-05
Fix duplicate pages in on this day results.

montehurd:
The `events` for the following days are examples having dupe pages:

`onthisday/events/09/14`
"Telecommunications companies MCI Communications and WorldCom complete their $37 billion merger to form MCI WorldCom."

`onthisday/events/09/01`
"St. Petersburg, Russia, changes its name to Petrograd."

Bug: T175974

### DIFF
--- a/v1/onthisday.js
+++ b/v1/onthisday.js
@@ -26,18 +26,18 @@ class Feed extends BaseFeed {
                 [req.params.type]: res.body[req.params.type]
             };
         }
-        let hydratedResponse = super._hydrateResponse(hyper, req, res);
+        const hydratedResponse = super._hydrateResponse(hyper, req, res);
 
         // Hydration resolves re-directs so we need to de-dupe titles here *after* hydration.
         const removeDuplicateTitlesFromHydratedResponsePages = (response) => {
-          Object.keys(response.body).forEach(key => {
-            response.body[key].forEach(event => {
-              event.pages = event.pages.filter(
-                (item1, index, self) =>
-                  self.findIndex(item2 => item2.title === item1.title) === index
-              )
+            Object.keys(response.body).forEach((key) => {
+                response.body[key].forEach((event) => {
+                    event.pages = event.pages.filter(
+                        (item1, index, self) =>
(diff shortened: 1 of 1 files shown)


v1/onthisday.js
```diff
@@ -26,7 +26,23 @@ class Feed extends BaseFeed {
                 [req.params.type]: res.body[req.params.type]
             };
         }
-        return super._hydrateResponse(hyper, req, res);
+        const hydratedResponse = super._hydrateResponse(hyper, req, res);
```
d00rman: It's actually faster to use Objects to determine whether a value should be filtered out:

```js
return super._hydrateResponse(hyper, req, res).then((response) => {
  Object.keys(response.body).forEach((key) => {
    if (!Array.isArray(response.body[key])) {
      return;
    }
    response.body[key].forEach((elem) => {
      if (!elem || !Array.isArray(elem.pages)) {
        return;
      }
      const titles = {};
      elem.pages = elem.pages.filter((item) => {
        if (titles[item.title]) {
          return false;
        }
        titles[item.title] = true;
        return true;
      });
    });
  });
  return response;
});
```

montehurd: @d00rman 
Updated! I kept the declarative naming, but I don't feel super strongly about it if you prefer it more concise.

montehurd:
@d00rman 

Quick question - the `REQUEST_TEMPLATE` is set to...
`{{options.host}}/{{domain}}/v1/feed/onthisday/all/{{mm}}/{{dd}}`
...but shouldn't it be something like...
`{{options.host}}/{{domain}}/v1/feed/onthisday/{type}/{{mm}}/{{dd}}`?

(**type** being either `births`, `deaths`, `events`, `holidays`, `selected` or `all`)

d00rman:
@montehurd Nope, because we need `all` for storage purposes. When another type is requested, it's filtered out [here](https://github.com/wikimedia/restbase/blob/c530aa6d18b513aaa29b867643e2d6622d295317/v1/onthisday.js#L24) so the client gets only what they asked for.

PR #938, opened 2018-01-05
Recursively change the URI protocol to the summary response

d00rman:
Instead of hard-coding the elements for which the protocol needs to be changed, recursively traverse the whole response body and apply the transformation to anything that looks like a URI.

Also introduce a new `protocol` configuration option that allows the user to specify which protocol to apply:
- `http`: use HTTP
- `https`: use HTTPS
- default: use the protocol-relative `//` notation

Replaces #934
Bug: [T179875](https://phabricator.wikimedia.org/T179875)


d00rman:
cc @Pchelolo @berndsi @mdholloway @phuedx

@berndsi @mdholloway we are still lacking the confirmation that Android can handle protocol-relative `//` notation.

phuedx:
👍

berndsi:
Adding @dbrant + @cooltey.
The Android app doesn't handle protocol-relative thumbnail URLs. So, I guess this is a no go. Here are some visuals. I can provide more but I guess that's plenty enough to see that it would be a bad experience. 

(For the Android devs: I ran this locally with a patched RbPageSummary.java file using `thumbnail.getUrl().substring(6);` in line 54.)
device-2018-01-05-173557
device-2018-01-05-173620

d00rman:
Thanks for the tests @berndsi . It's a pitty protocol-relative notation can't be used, but that's OK, as this patch allows one to configure the protocol style and type, so that we can have different protocols based on the environment (`https` for prod and Beta, `http` for local and Vagrant installs).

PR #946, opened 2018-01-18
Update dependencies and release v0.18.0

d00rman:
Bug: [T184199](https://phabricator.wikimedia.org/T184199)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "0.17.0",
+  "version": "0.18.0",
   "description": "REST storage and service dispatcher",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


d00rman:
cc @wikimedia/services .

A question here is: could this potentially be v1.0.0 instead of v0.18.0 ?

Pchelolo:
> A question here is: could this potentially be v1.0.0 instead of v0.18.0 ?

I think we should make 1.0 somewhat more remarkable than just a version bump. With an announcement etc

d00rman:
> I think we should make 1.0 somewhat more remarkable than just a version bump. With an announcement etc

I feel the same, that's why I put v0.18.0 here.

d00rman:
Published as [v0.18.0](https://github.com/wikimedia/restbase/tree/v0.18.0)

PR #952, opened 2018-02-11
 Adjust max list size and expose in Swagger response

tgr:
There does not seem to be any way for clients to get root-level fields of the Swagger spec so copy it into each relevant path block.

Also adjust max item number to reflect that the configuration will change soon ([Ia3c1e631](https://gerrit.wikimedia.org/r/#/c/409712/)).

Bug: [T186296](https://phabricator.wikimedia.org/T186296)

This is on top of #944.

### DIFF
--- a/v1/lists.yaml
+++ b/v1/lists.yaml
@@ -12,7 +12,7 @@ info:
     url: https://www.apache.org/licenses/LICENSE-2.0
   x-configuration:
     maxListsPerUser: &maxListsPerUser 100
-    maxEntriesPerList: &maxEntriesPerList 1000
+    maxEntriesPerList: &maxEntriesPerList 5000
     deletedRetentionDays: &deletedRetentionDays 30
 x-yaml-anchors:
   csrf_token: &csrf_token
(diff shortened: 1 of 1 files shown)


Pchelolo:
So why does this PR contain stuff from the batching PR #944 ? Since I've merged #944 could you please rebase this one so that it's easier to review?

tgr:
Rebased. Wasn't sure how to do stacked changes otherwise.

Pchelolo:
Thank you! Wouldn't it be better to include the  `maxItems` in the schema definition for requests/responses? that way it would be a bit harder to get the max size on the client side, but it's a way more standard way of doing this and not introducing a custom key... Although I'm not sure if that worths the trouble...

tgr:
I'm not sure if it would be accurate, `maxItems` tells how many items the request can include (so technically that would be something like 5000 - <currrent list size> for the batch endpoint, and there isn't really any good place to put it on the non-batch endpoint). Also the batch endpoint does not have a limit for items sent in a single request currently, but it probably should and it should probably be smaller than the list size, and `maxItems` seems more accurate for that.

I could put it on the `GET` endpoints, but those are paged so the maximum number of items you can get in a single request is actually 100, and again `maxItems` seems more accurate for that.

Pchelolo:
Ok, makes sense.

tgr:
Filed [T187103](https://phabricator.wikimedia.org/T187103) about adding `maxItems` for paging / limiting writes in a single request.

PR #953, opened 2018-02-12
Provide timestamp to continue sync from

tgr:
The sync (`GET /lists/changes/since/{date}`) API takes a timestamp to list events from. Clients need a way to determine what timestamp to use on their next sync call; that decision can be nontrivial due to issues with multiple changes in the same second, transactioned changes that are committed significantly later than inserted etc. To avoid every client having to reimplement that logic, the API
will provide these timestamps.

The logic will eventually be hosted in the MediaWiki API (see [I8247148b](https://gerrit.wikimedia.org/r/409731)); to ease deployment coordination, it will be temporarily backfilled in RESTBase.

This is on top of #952.

Bug: [T182706](https://phabricator.wikimedia.org/T182706)

### DIFF
--- a/v1/lists.js
+++ b/v1/lists.js
@@ -102,6 +102,35 @@ class ReadingLists {
     }
 
     /**
+     * Get a timestamp that's safe to use in GET /lists/changes/since/{timestamp} assuming
+     * the client's state is based on the current response. This deals with things database rows
+     * items being committed in a different order than their 'created' fields would suggest.
+     * See T182706 for details.
+     *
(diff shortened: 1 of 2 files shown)


Pchelolo:
Same thing, could you please rebase?

Pchelolo:
Also there's some linting errors in tests

Pchelolo:
Looks good! When do you need all these changes to be deployed?

tgr:
#944 is the urgent one but it would be nice if the others could also be deployed today.

PR #960, opened 2018-02-23
Reading lists: use full objects from MW API in response

tgr:
Includes a couple documentation updates that should have been in
previous patches.

Relies on [I097d6ba](https://gerrit.wikimedia.org/r/#/c/413650/) (won't break without it, just lie about the
API response format).

Bug: [T186041](https://phabricator.wikimedia.org/T186041)

### DIFF
--- a/v1/lists.yaml
+++ b/v1/lists.yaml
@@ -192,7 +192,8 @@ paths:
         - Reading lists
       summary: Create a new list for the current user.
       description: |
-        Creates a new empty list.
+        Creates a new empty list. On name conflict, does nothing and returns the data of an
+        existing list.
 
         Request must be authenticated with a MediaWiki session cookie.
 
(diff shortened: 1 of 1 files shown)


v1/lists.yaml
```diff
@@ -269,7 +277,21 @@ paths:
...
+          description: The updated data for the list.
+          schema:
+            type: object
+            properties:
+              id:
+                type: integer
+                description: |
+                  List ID.
+
+                  Deprecated, will be removed. Use the full list object.
```
Pchelolo: Idem

v1/lists.yaml
```diff
@@ -251,7 +259,7 @@ paths:
       summary: Update a list.
       description: |
         List must belong to current user and request must be authenticated with
-        a MediaWiki session cookie.
+        a MediaWiki session cookie. If the name is changed, the new name must not be in use.
```
Pchelolo: Can we also specify what happens when there's a naming conflict? I guess 409 could be a nice response code

Pchelolo:
A very optional and small comment inlined, otherwise LGTM

v1/lists.yaml
tgr: It results in a `readinglists-db-error-duplicate-list` error. Is 409 really applicable? The conflict isn't really with the current status of the resource, it's with another resource. 409 seems more appropriate for things like an edit conflict.

PR #961, opened 2018-02-26
Remove the old summary implementation

Pchelolo:
After we complete the migration to the new MCS-based summary we'd need to clean up the code and remove the old implementation.


d00rman:
Hm, this makes a hard dependency on MCS for summary. Maybe we could keep the old version around for other users who might be happy with the TextExtracts variant?

Pchelolo:
> Hm, this makes a hard dependency on MCS for summary. Maybe we could keep the old version around for other users who might be happy with the TextExtracts variant?

Do you think such users even exist?

d00rman:
@Pchelolo AFAIK, if people install RESTBase, it's most likely for VE. If we keep this code in, then they get page previews via RB for free which is a nice usage scale up

Pchelolo:
Ok, makes sence.

PR #962, opened 2018-02-26
Use container-based travis for tests

Pchelolo:
Attempt to check if the travis-side bug was fixed.

cc @wikimedia/services


Pchelolo:
OK, seems like the travis is still not working.. Closing.

PR #963, opened 2018-02-27
Use regexes to generalize content-type expectations in tests

Pchelolo:
From time to time Parsoid bumps the content-type version which breaks RESTBase tests that expect a specific content-type version to be served. I don't think that expecting the exact version number in tests provides any value and is kinda painful cause we need to update RESTBase every time the version in PArsoid is bumped. Let's use less restrictive regexes to match the version.

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -89,9 +89,9 @@ wiktionary_project: &wiktionary_project
 # Hacky way to parametrize RESTBase tests. TODO: Move to config?
 test:
   content_types:
-    html: '/text\/html; charset=utf-8; profile="https:\/\/www\.mediawiki\.org\/wiki\/Specs\/HTML\/[\d.]+"$/'
-    data-parsoid: '/application\/json; charset=utf-8; profile="https:\/\/www\.mediawiki\.org\/wiki\/Specs\/data-parsoid/[\d.]+"/'
-    wikitext: '/text\/plain; charset=utf-8; profile="https:\/\/www.mediawiki.org\/wiki\/Specs\/wikitext\/[\d.]+"/'
+    html: '/^text\/html; charset=utf-8; profile="https:\/\/www\.mediawiki\.org\/wiki\/Specs\/HTML\/[\d.]+"$/'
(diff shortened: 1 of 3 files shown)


config.test.yaml
```diff
@@ -89,9 +89,9 @@ wiktionary_project: &wiktionary_project
 # Hacky way to parametrize RESTBase tests. TODO: Move to config?
 test:
   content_types:
-    html: text/html; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/HTML/1.6.0"
-    data-parsoid: application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/data-parsoid/1.6.0"
-    wikitext: text/plain; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/wikitext/1.0.0"
+    html: '/text\/html; charset=utf-8; profile="https:\/\/www\.mediawiki\.org\/wiki\/Specs\/HTML\/[\d.]+"$/'
```
berndsi: Have you considered adding a `^` at the beginning of the regexs here and below?

test/utils/assert.js
```diff
@@ -1,15 +1,19 @@
...
 
 /**
  * Asserts whether content type was as expected
  */
 function contentType(res, expected) {
-    var actual = res.headers['content-type'];
-    deepEqual(actual, expected,
-        'Expected content-type to be ' + expected + ', but was ' + actual);
+    const actual = res.headers['content-type'];
+    if (/^\/.+\/$/.test(expected)) {
```
berndsi: BTW you could also use `expected.source` to see if it's a regex. If it's not then it would return undefined.

> 'foo'.source
undefined

Pchelolo: Here it's never an instance of a `RegEx` cause it's a string read from the yaml, so testing for `/` in the beginning and the end is the only way

config.test.yaml
Pchelolo: Done, good idea.

PR #964, opened 2018-02-28
Fix namespaces issues when generating PDFs

Silex:
Hello,

Don't merge this as it is not working.

I need your guidance:

- why do some parameters use one `{` and others two `{{` ?
- why doesn't my fix work?

The goal is to url encode the whole `url` parameter, otherwise it fails for wikis that use namespaces (it url-encodes the slash that is part of the title, but not the slash that is part of `http://` and the pdf electron service is confused). Also maybe it encodes the title twices?

At the moment the only fix that works is:

 ```
uri: '{{options.uri}}/pdf?accessKey={options.secret}&url={{default(options.scheme, "https")}}://{{domain}}/wiki/{title}%3Fprintable=yes'
```

### DIFF
--- a/v1/pdf.yaml
+++ b/v1/pdf.yaml
@@ -60,8 +60,7 @@ paths:
         - get_pdf_from_backend:
             request:
               method: get
-              # Note: The title needs to be encoded twice.
-              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url={{default(options.scheme, "https")}}://{{domain}}/wiki/{_encodeURIComponent(title)}%3Fprintable=yes'
+              uri: '{{options.uri}}/pdf?accessKey={options.secret}&url={_encodeUriComponent(default(options.scheme, "https") + "://" + domain + "/wiki/" + title + "/printable=yes")}'
             return:
               status: 200
               headers:

### CODE REVIEW

**Comment by Pchelolo** [comment]
> why do some parameters use one { and others two {{ ?

A single one uses `encodeURIComponent` under the hood the double one doesn't.

> Also maybe it encodes the title twices?

It does due to #832

I'm not really sure which problem are you trying to solve. Could you provide an example of the. request that fails?

**Comment by Silex** [comment]
The URL I click `Export to PDF` on is:

```
http://10.10.10.239/wiki/PMS/Bellefontaine
```

The request that the pdf electron service receives is:

```
[2018-03-01T08:50:41.059Z] global@172.23.0.2 - GET 200 /pdf?accessKey=secret&url=http%3A%2F%2Fweb%2Fwiki%2FPMS%252FBellefontaine%3Fprintable%3Dyes 63717 "-" 111.413 ms
```

We see that title is URL-encoded twice, more specifically the `/` (`%252F` should be `%2F`)

**Comment by Silex** [comment]
@Pchelolo: ping

**Comment by Pchelolo** [comment]
So the `+` is not supported inside the templates, it's more like string interpolation.

As for the double-encoding - that was added by #832 and I think it's correct - when the PDF service receives the double-encoded query parameter it first unencodes it before passing inside the service and then it uses it to pass to the chromium renderer - so it's like you put the uri-encoded string into the browser. 

So I still think the current code is correct, so perhaps your problem lies somewhere else?

**Comment by Silex** [comment]
The README at https://github.com/msokk/electron-render-service never mention we should URL-encode twice. Also it would not really make sense for such a service to require that given its API.

Anway, when I use the original code the PDF that is generated for namespaced URLs contains:

```
Not Found

The requested URL /wiki/PMS/Bellefontaine was not found on this server.

Apache/2.4.18 (Ubuntu) Server at web Port 80
```

When I "Download as PDF" on an article that is not namespaced then everything works as expected.

When I don't url-encode the title at all everything works all the time, but I suspect this is vulnerable to some kind of attack.

**Comment by Silex** [comment]
> So the + is not supported inside the templates, it's more like string interpolation.

What is the name of the template engine? This info is very hard to find. If it's a custom one, does it support "recursion"? I mean would something like this work:

``` yaml
uri: '{{options.uri}}/pdf?accessKey={options.secret}&url={"{{default(options.scheme, "https")}:}//{{domain}}/wiki/{{title}}/printable=yes"}'
```

Silex:
@Pchelolo: ping :wink:

Silex:
@Pchelolo: ping

Silex:
Closing without an answer?!

PR #965, opened 2018-03-12
Lower down rate limit for all the Analytics Edit metrics to 25 rps

elukey:
We have recently been hit by a bot making 100rps (already rate limited)
scanning the whole mw-history and causing a bit of pressure to the Druid
backend. Druid is not as fast as Cassandra for some type of queries
(especially the ones for big time windows), so it makes sense in my
opinion to lower down the maximumg rps allowed for the moment.

### DIFF
--- a/v1/metrics.yaml
+++ b/v1/metrics.yaml
@@ -1430,7 +1430,7 @@ paths:
           options:
             limits:
               internal: 25
-              external: 100
+              external: 25
       x-request-handler:
         - get_from_backend:
             request:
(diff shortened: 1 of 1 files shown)


v1/metrics.yaml
```diff
@@ -1429,7 +1429,7 @@ paths:
           name: ratelimit_route
           options:
             limits:
-              internal: 100
+              internal: 25
               external: 100
```
marcelrf: Shouldn't the external limit be updated here as well?

elukey: yes!

elukey:
@Pchelolo, @d00rman any thoughts/recommendations?

Pchelolo:
LGTM

d00rman:
FYI @elukey this is now live in production.

PR #967, opened 2018-03-13
Expose lint errors for a page/revision via a GET request.

Pchelolo:
Bug: T164006

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -624,7 +624,6 @@ paths:
       summary: Get the linter errors for a specific title/revision.
       description: |
         Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
-      operationId: getFormatRevision
       produces:
         - application/json
       parameters:
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -612,6 +612,80 @@ paths:
                 x-restbase-parentrevision: '{{x-restbase-parentrevision}}'
       x-monitor: false
 
+  /lint/{title}{/revision}:
+    x-route-filters:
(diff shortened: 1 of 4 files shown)


subbuss:
LGTM.

v1/content.yaml
```diff
@@ -612,6 +612,80 @@ paths:
...
+        options:
+          redirect_cache_control: '{{options.response_cache_control}}'
+      - path: lib/security_response_header_filter.js
+    get:
+      tags:
+        - Page content
+      summary: Get the linter errors for a specific title/revision.
+      description: |
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      operationId: getFormatRevision
```
d00rman: c/p error? Also, no need for an operation id here.

v1/content.yaml
```diff
@@ -612,6 +612,80 @@ paths:
                 x-restbase-parentrevision: '{{x-restbase-parentrevision}}'
       x-monitor: false
 
+  /lint/{title}{/revision}:
```
d00rman: `lint` doesn't seem too informative to me. Perhaps `lintstatus` or `linterrors` ? Any suggestions @subbuss / @Pchelolo ?

Pchelolo: `lint` was chosen for consistency with the transform endpoint `transform/wikitext/to/lint`...

d00rman: Oh, excellent point @Pchelolo

v1/content.yaml
Pchelolo: Done

PR #973, opened 2018-03-20
Expose /page/metadata and /page/references endpoints via RESTBase

berndsi:
This is pretty much the same as was done for /page/media. Going to do references later.

Bug: T190198
Bug: T189750

### DIFF
--- a/projects/dev.yaml
+++ b/projects/dev.yaml
@@ -33,6 +33,10 @@ paths:
                   options:
                     response_cache_control: '{{options.purged_cache_control}}'
                     host: '{{options.mobileapps.host}}'
+                - path: v1/references.yaml
+                  options:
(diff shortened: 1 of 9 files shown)


Pchelolo:
LGTM. We need a rule to purge it in change-prop, I'll make necessary changes. 

When do you want this deployed?

berndsi:
I think this can wait until I got the pr for references as well. Tomorrow
would be fine.

Pchelolo:
The test failures are unrelated. Needs to be rebased on #974 after that one is merged.

berndsi:
Since I'll have to also change some of the same files for the new references endpoint, should I just add to this PR?

Pchelolo:
yup @berndsi go for it

PR #975, opened 2018-03-20
Lists API: Minor: Fix param type in getContinueFrom()

d00rman:


d00rman:
Trivial, so self-merging.

PR #977, opened 2018-03-26
Add tag descriptions for citations and recommendations

d00rman:


d00rman:
Trivial, so self-merging.

PR #983, opened 2018-04-10
API: Include anchors in redirect targets

d00rman:
If the redirect is explicit and it contains a URL fragment, we should honour it.

Bug: [T148645](https://phabricator.wikimedia.org/T148645)

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -350,6 +350,12 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
         location = backString
             + new URI(pathPatternAfterTitle, newReqParams, true).toString().substr(1)
             + mwUtil.getQueryString(req);
+    } else if (/#/.test(newReqParams.title)) {
+        const titleFragments = encodeURIComponent(newReqParams.title).split(/%23/);
+        /* eslint-disable prefer-template */
+        location = backString + titleFragments.shift() + mwUtil.getQueryString(req)
+            + '#' + titleFragments.join('%23');
+        /* eslint-enable prefer-template */
     } else {
         location = backString + encodeURIComponent(newReqParams.title) + mwUtil.getQueryString(req);
     }
@@ -370,7 +376,7 @@ mwUtil.isCrossOrigin = (req) => {
     return req && req.headers && req.headers.origin && req.headers.origin !== req.params.domain;
 };
 
-const redirectRegEx = /<link rel="mw:PageProp\/redirect" href="\.\/([^"#]+)(?:#[^"]*)?"/;
+const redirectRegEx = /<link rel="mw:PageProp\/redirect" href="\.\/([^"]+)"/;
 mwUtil.extractRedirect = (html) => {
     const redirectMatch = redirectRegEx.exec(html);
     if (redirectMatch) {
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -383,7 +383,8 @@ class ParsoidService {
                             // This revision is actually a redirect. Pass redirect target
                             // to caller, and let it rewrite the location header.
                             resp.status = 302;
-                            resp.headers.location = encodeURIComponent(redirectTarget);
+                            resp.headers.location = encodeURIComponent(redirectTarget)
+                                .replace(/%23/, '#');
                             return hyper.post({
                                 uri: new URI([rp.domain, 'sys', 'page_revisions',
                                     'restrictions', rp.title, rp.revision]),
--- a/test/features/pagecontent/redirects.js
+++ b/test/features/pagecontent/redirects.js
@@ -208,7 +208,7 @@ describe('Redirects', function() {
         })
         .then(function(res) {
             assert.deepEqual(res.status, 302);
-            assert.deepEqual(res.headers.location, 'Main_Page');
+            assert.deepEqual(res.headers.location, 'Main_Page#Test%23123');
             assert.deepEqual(res.body.length > 0, true);
             assert.deepEqual(res.headers['cache-control'], 'test_purged_cache_control');
         });


d00rman:

Pchelolo:
Adding a test would be real nice

berndsi:
@d00rman Looks like the # is getting encoded:
```
curl -sI http://localhost:7231/en.wikipedia.org/v1/page/mobile-sections/Anion
HTTP/1.1 302 
[...]
location: Ion%23Anions_and_cations
```

d00rman:
> Adding a test would be real nice

@Pchelolo, it turns out there is already a test for this, so I just modified it :P

> Looks like the # is getting encoded:

@berndsi should be fixed now.

Pchelolo:
+1

berndsi:
LGTM 
```
curl -sI http://localhost:7231/en.wikipedia.org/v1/page/mobile-sections/Anion | grep \^location
location: Ion#Anions_and_cations
```

d00rman:
@berndsi does MCS correctly pick up the anchor?

berndsi:
@d00rman Not sure why it would. Redirects get handled in the RESTBase layer, not MCS.

d00rman:
@berndsi `mobile-sections` are JSONs, so my question is if the app asks for `Ion#Anions_and_cations`, will it display the correct thing? (Basically, I wonder what does the anchor really mean in a non-HTML response and how is it interpreted)

berndsi:
@d00rman The Android app doesn't not handle them yet since there was nothing it could do about it but now that it could the ball would be in the Android app teams court to implement it. See discussion starting at https://phabricator.wikimedia.org/#3499123.

d00rman:
> @d00rman The Android app doesn't not handle them yet since there was nothing it could do about it but now that it could the ball would be in the Android app teams court to implement it. 

This is exactly the reason I was asking. This effectively means that until we are sure how would the App respond to getting anchors in the response, we cannot merge and deploy this PR. From my POW, we need to answer two questions:

1) How would the current version of the Android App behave if it started receiving URLs with anchors in them? Would it break? Would it start showing the wrong content?
2) How/when would the implementation come to the Android App?

Obviously, (2) is still a matter of discussion and implementation, but (1) needs to be known before we can get this PR out.

> See discussion starting at https://phabricator.wikimedia.org/#3499123.

Oops, this link points to nowhere, even after `s/#/T/` :P

d00rman:
@berndsi ping re ^

d00rman:
OK, great, that's good news! Thnx @berndsi.

I think you can expect this to be deployed soon. After the deployment, we will need to run dumps to re-populate the redirect fields.

berndsi:
@d00rman It's not ideal from an Android app perspective to have to do this. Having the redirected field in the JSON output would be much more convenient. If I can't convince you to inject the redirected field into the JSON response then this is the best compromise I can come up with for mobile-sections.

We also should note that this only deals with mobile-sections, not the summary. I'm not sure at this time how the app should handle redirects with fragments for the summary.

d00rman:
@Pchelolo Do you think it's time we merged this?

lib/mwUtil.js
```diff
@@ -311,7 +311,8 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
             + new URI(pathPatternAfterTitle, newReqParams, true).toString().substr(1)
             + mwUtil.getQueryString(req);
     } else {
-        location = backString + encodeURIComponent(newReqParams.title) + mwUtil.getQueryString(req);
+        location = backString + encodeURIComponent(newReqParams.title).replace(/%23/g, '#')
```
Pchelolo: According to RFC the fragment can not contain the `#` and it must be percent-encoded. So we only need to replace the first occurrence of `%23`, cause if we replace all something like `https://bla.com/lala%23bla%23bla` will become `https://bla.com/lala#bla#bla` which is icorrect, it should be https://bla.com/lala#bla%23bla

lib/mwUtil.js
```diff
@@ -311,7 +311,8 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
...
+            + mwUtil.getQueryString(req);
```
Pchelolo: I believe the order is incorrect - everything after `#` is client-side, so the query string must be before the hashbang

sys/parsoid.js
```diff
@@ -398,7 +398,8 @@ class ParsoidService {
                             // This revision is actually a redirect. Pass redirect target
                             // to caller, and let it rewrite the location header.
                             resp.status = 302;
-                            resp.headers.location = encodeURIComponent(redirectTarget);
+                            resp.headers.location = encodeURIComponent(redirectTarget)
+                                .replace(/%23/g, '#');
```
Pchelolo: Again, not `/g`

test/features/pagecontent/redirects.js
```diff
@@ -208,7 +208,7 @@ describe('Redirects', function() {
         })
         .then(function(res) {
             assert.deepEqual(res.status, 302);
-            assert.deepEqual(res.headers.location, 'Main_Page');
+            assert.deepEqual(res.headers.location, 'Main_Page#Test');
```
Pchelolo: Would be good to add a test for a hashbang containing %23 and also for query order.

lib/mwUtil.js
d00rman: Really good catch! {{done}}

sys/parsoid.js
d00rman: Done

test/features/pagecontent/redirects.js
d00rman: Done

lib/mwUtil.js
d00rman: As it turns out, MW does a bad job at it - https://en.wikipedia.beta.wmflabs.org/w/index.php?title=User:Pchelolo/Redirect_Test_Hash&redirect=no :)

PR #984, opened 2018-04-16
Pageviews: Use `keep-alive` for AQS connections

d00rman:
When there is a surge in traffic for the Pageviews API, a lot of time is spent setting up connections to AQS. However, both RESTBase and AQS are able to withstand high traffic pressure, so keep the connections to it open.

Bug: [T190213](https://phabricator.wikimedia.org/T190213)


d00rman:
cc @elukey

Pchelolo:
+1, restarted travis cause it failed for unrelated reason

elukey:
Looks good to me!

PR #986, opened 2018-04-23
Log when the timeline index is empty

Pchelolo:
cc @wikimedia/services

### DIFF
--- a/sys/multi_content_bucket.js
+++ b/sys/multi_content_bucket.js
@@ -333,7 +333,7 @@ class MultiContentBucket {
         }).catchReturn({ status: 404 }, undefined)
         .then((res) => {
             if (!res || !res.body.items.length) {
-                // Noting was ever there - put the first render and no need to update the index
+                // Nothing was ever there - put the first render and no need to update the index
                 return this._createContentStoreRequests(hyper, req,rev, tid);
             } else if (res && res.body.items.length && res.body.items[0].rev < rev) {
                 // New revision is being written - update revision index and do the revision deletes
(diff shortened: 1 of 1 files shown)


sys/multi_content_bucket.js
```diff
@@ -378,7 +378,11 @@ class MultiContentBucket {
                         });
                     })
                     .catch({ status: 404 }, () => {
-                        // Ignore the 404 if we don't have the timeline.
+                        // Log the 404 if we don't have the timeline.
+                        hyper.log('error/noindex', {
+                            msg: 'Empty revision timeline',
+                            page_title: rp.key
```
eevans: Do we need the domain logged as well, or will that be available elsewhere?

sys/multi_content_bucket.js
```diff
@@ -426,7 +430,12 @@ class MultiContentBucket {
                         });
                     })
                     .catch({ status: 404 }, () => {
-                        // Ignore the 404 if we don't have the timeline.
+                        // Log the 404 if we don't have the timeline.
+                        hyper.log('error/noindex', {
+                            msg: 'Empty render timeline',
+                            page_title: rp.key,
+                            page_revision: rev
```
eevans: Idem.

sys/multi_content_bucket.js
Pchelolo: The domain should be added by default.

eevans:
LGTM :+1: ; Are these test failures new?  They seem wholly unrelated.

d00rman:
Perhaps `warn` would have been a more appropriate logging level given that we've been ignoring these so far and that, supposedly, a 404 here does not signal an actual failure

PR #980, opened 2018-04-05
Expose CSS endpoints

berndsi:
Another Page Content Service endpoint. This patch provides two
page-independent CSS responses under /data/css/mobile/:
* /data/css/mobile/base
* /data/css/mobile/site

Bug: T190846

PR #988, opened 2018-04-24
Minor: Remove unneeded CSS test file

d00rman:
The monitoring examples in the CSS YAML specification already perform these tests, so no need to have them.

This is a follow-up on PR #980


d00rman:

PR #993, opened 2018-05-08
Transform: Update /list/{tool}/{/from}/{/to} end point

kartikm:
CX is using this end-point, so adding it in RB API.

### DIFF
--- a/v1/transform-global.yaml
+++ b/v1/transform-global.yaml
@@ -147,6 +147,49 @@ paths:
         - get_from_cx:
             request:
               uri: '{{options.cx_host}}/v1/list/pair/{from}/{to}'
+  /list/{tool}/{from}/{to}:
+    get:
+      tags: ['Transform']
+      summary: Lists all language pairs that tool supports
+      description: |
+        Fetches the list of all of the language pairs for given tool
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      produces:
+        - application/json
+      parameters:
+        - name: tool
+          in: path
+          description: The tool name
+          type: string
+          required: true
+          enum:
+            - mt
+            - dictionary
+        - name: from
+          in: path
+          description: The source language code
+          type: string
+          required: true
+        - name: to
+          in: path
+          description: The target language code
+          type: string
+          required: true
+      responses:
+        '200':
+          description: the list of language pairs available for a given tool
+          schema:
+            $ref: '#/definitions/cx_list_pairs_for_tool'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-monitor: false
+      x-request-handler:
+        - get_from_cx:
+            request:
+              uri: '{{options.cx_host}}/v1/list/{tool}/{from}/{to}'
   /list/tool/{tool}/:
     get:
       tags: ['Transform']
--- a/v1/transform-global.yaml
+++ b/v1/transform-global.yaml
@@ -147,7 +147,7 @@ paths:
         - get_from_cx:
             request:
               uri: '{{options.cx_host}}/v1/list/pair/{from}/{to}'
-  /list/tool/{tool}/{/from}/{/to}:
+  /list/tool/{tool}{/from}{/to}:
     get:
       tags: ['Transform']
       summary: Lists the tools and language pairs available for the given tool category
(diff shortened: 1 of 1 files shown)


Pchelolo:
LGTM, but let's wait for @d00rman

v1/transform-global.yaml
```diff
@@ -147,6 +147,49 @@ paths:
         - get_from_cx:
             request:
               uri: '{{options.cx_host}}/v1/list/pair/{from}/{to}'
+  /list/{tool}/{from}/{to}:
```
d00rman: Based on the end point's description, this is supposed to do the same as `/list/tool/{tool}`, but `from` and `to` and required here, so how can this end point list "all the language pairs for a given tool" ? What is this end point supposed to do exactly that is different from the one below?

Even if there is a legitimate need for this end point, we need to find a different path for it, because it conflicts with `/list/pair/` and `/list/tool/`

kartikm: Updating /list/tool/{tool} with from and to should be fine. Currently, /list/tool/{tool} gives list of tools available for all pairs, while /list/tool/{tool}/{from}/{to} will be useful to query particular tool. This is used in MT card in CX. @santhoshtr @Nikerabbit Please see this.

d00rman: The existing `/list/tool/{tool}` end point could then be expanded to allow for optional `from` and `to` parametres, so that if they are not given, then list all pairs, if only `from` is given list the pairs for the given source language and if both are given, then basically return just them if applicable. Would such a `/list/tool/{tool}{/from}{/to}` end point cover your use cases?

v1/transform-global.yaml
```diff
@@ -147,7 +147,7 @@ paths:
         - get_from_cx:
             request:
               uri: '{{options.cx_host}}/v1/list/pair/{from}/{to}'
-  /list/tool/{tool}/:
+  /list/tool/{tool}/{from}/{to}:
```
d00rman: `from` and `to` are optional, so `{/from}{/to}`

v1/transform-global.yaml
```diff
@@ -179,7 +189,7 @@ paths:
       x-request-handler:
         - get_from_cx:
             request:
-              uri: '{{options.cx_host}}/v1/list/tool/{tool}'
+              uri: '{{options.cx_host}}/v1/list/{tool}/{from}/{to}'
```
d00rman: Idem: `{/from}{/to}`

v1/transform-global.yaml
```diff
@@ -147,7 +147,7 @@ paths:
         - get_from_cx:
             request:
               uri: '{{options.cx_host}}/v1/list/pair/{from}/{to}'
-  /list/tool/{tool}/:
+  /list/tool/{tool}/{/from}/{/to}:
```
d00rman: These go without the slash: `/list/tool/{tool}{/from}{/to}`

v1/transform-global.yaml
```diff
@@ -179,7 +189,7 @@ paths:
       x-request-handler:
         - get_from_cx:
             request:
-              uri: '{{options.cx_host}}/v1/list/tool/{tool}'
+              uri: '{{options.cx_host}}/v1/list/{tool}/{/from}/{/to}'
```
d00rman: Idem.

v1/transform-global.yaml
d00rman: We are back to the old way... This needs to be: `/list/tool/{tool}{/from}{/to}`

v1/transform-global.yaml
d00rman: Same as above.

kartikm:
@d00rman Asking @santhoshtr about collapsing two endpoints in cxerver. I'll fix it there once we finalize it.

kartikm:
@d00rman Santhosh has https://gerrit.wikimedia.org/r/#/c/432340/ to deprecate /list/tool which is unused. I'll do testing of it and update here.

kartikm:
> Ok, we keep going back and forth on this. So, you keep pushing changes that either do /list/tool/{tool}/{from}/{to} or /list/tool/{tool}/{/from}/{/too}. However, the form we need is /list/tool/{tool}{/from}{/to} -- note the lack of / between {tool} and {/from} and {/to} :) Let's get this right!

Thanks. Fixed.

> Perfect, thank you! Once that is merged and deployed and the above fixed, we can proceed.

I plan to deploy this in Production on Monday with testing today and tomorrow. Will update here.

d00rman:
looking good now! Let's wait until the CXServer side is deployed to proceed.

kartikm:
@d00rman CXServer change is merged and deployed.

d00rman:
@kartikm, the end point has been deployed. Please see https://phabricator.wikimedia.org/T194660 for a needed follow-up.

kartikm:
Thanks @d00rman

PR #998, opened 2018-06-04
Expose feed content availability endpoint

mdholloway:


Pchelolo:
Hm... So we only return language codes? How about project names? Maybe we will have feed for projects other then wikipedia in the future?

v1/availability.yaml
```diff
@@ -0,0 +1,109 @@
...
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_feed_content_availability:
+            request:
+              method: get
+              uri: '{{options.host}}/{domain}/v1/feed/availability'
+              gzip: false
```
d00rman: Why no gzip here?

d00rman:
> Maybe we will have feed for projects other then wikipedia in the future?

+1 to that. Also, looking at the output format, it could be interesting / easy for clients to also have domain-dependent responses with booleans,

v1/availability.yaml
mdholloway: Just an oversight, will update.

mdholloway:
> Hm... So we only return language codes? How about project names? Maybe we will have feed for projects other then wikipedia in the future?

That's a good point.  Maybe it could be future-proofed by adding projects at the top level of the hierarchy (just "wikipedia" for now), like so:

```
{ 
  "wikipedia": {
    "todays_featured_article": [...],
    "most_read": [...],
    ...
  }
}
```

> Also, looking at the output format, it could be interesting / easy for clients to also have domain-dependent responses with booleans,

I'm not quite sure what you're suggesting here.  Did the comment get cut off early?

d00rman:
> Maybe it could be future-proofed by adding projects at the top level of the hierarchy (just "wikipedia" for now), like so: [...]

+1.

> I'm not quite sure what you're suggesting here.

Ah, sorry. I meant that the same format could be used to get domain-specific features, e.g. for enwiki the response of the analogous end point could be:

```yaml
todays_featured_article: true
most_read: true
...
```

(whereas for a wiktionary these would all be `false`). In other words, clients could discover if a specific project supports these or not, instead of asking for the full list and then parsing and filtering. But this is just an idea for the future, I don't think we need to go with it now. It just seems helpful to have that info for external clients.

Pchelolo:
> That's a good point. Maybe it could be future-proofed by adding projects at the top level of the hierarchy (just "wikipedia" for now), like so:

Maybe just use dbname from mediawiki eg `enwiki`, `wikidatawikiw` etc?

d00rman:
> Maybe just use dbname from mediawiki eg enwiki, wikidatawikiw etc?

I'd be -1 on that. The db names are already too magical and shouldn't be relied upon, really. Domain parts are a much better approach IMHO.

v1/availability.yaml
```diff
@@ -0,0 +1,109 @@
...
+              on_this_day: [ /.+/ ]
+
+# copied from MCS spec.yaml
+definitions:
+  availability:
+    type: object
+    properties:
+      todays_featured_article:
+        type: array
+        description: language codes for wikis with this feature enabled, or [ '*' ] for universally enabled
```
berndsi: This and the ones below need updates to match the MCS change.

v1/availability.yaml
mdholloway: done

Pchelolo:
LGTM. I can deploy it whenever needed

v1/availability.yaml
mdholloway: done

mdholloway:
@Pchelolo Great!  Let's deploy right after the mobileapps update is deployed during today's services deploy window.

mdholloway:
Rebased.

d00rman:
Hm, https://gerrit.wikimedia.org/r/#/c/437469/ seems to go against what we have been discussing here. Instead of encompassing `wikipedia: { ... }`, the aforementioned patch adds all of the FQDNs. It's effectively the same, but it's more transmitted data.

My personal like is `org.wikipedia: { ... }` or `org: { wikipedia: { ... } }` (but the latter might be too verbose for our needs)

mdholloway:
I'd be fine with `org.wikipedia: { ... }`.  @Pchelolo @berndsi Thoughts?

Pchelolo:
> It's effectively the same, but it's more transmitted data.

It looks and feels simpler and with gzip all the `wikipedia.org` will be like 2 extra bytes or something around that. Given that this endpoint will we accessed extremely rarely I think it's better to go for simplicity than trying to save a couple of bytes.

mdholloway:
I would tend to prefer the existing output for the reasons @Pchelolo mentioned.

berndsi:
I slightly prefer `"org.wikipedia": { ... }` but I could live with either version.

d00rman:
Alright, let's find something else to bike-shed over, in that case :) Merging.

d00rman:
It's already late back here in Europe, so expect this to be deployed first thing tomorrow morning (hint, hint: would be nice to get #1000 in the same round as well :P)

PR #999, opened 2018-06-05
Tests: Fix the transform end point expected outcome

d00rman:


d00rman:
Fixed now. The remaining failure seems unrelated to this PR. Merging.

PR #1000, opened 2018-06-05
Update required media item fields

mdholloway:
As of https://gerrit.wikimedia.org/r/#/c/436360/ (adding support for
Mathoid images) media items may not have the usual set of Commons
metadata.  This updates the media spec to fix the monitoring check.

### DIFF
--- a/v1/media.yaml
+++ b/v1/media.yaml
@@ -111,17 +111,8 @@ paths:
               revision: /.+/
               tid: /.+/
               items:
-                - titles:
-                    canonical: /.+/
-                    normalized: /.+/
-                    display: /.+/
-                  thumbnail:
-                    source: /.+/
-                    width: /.+/
-                    height: /.+/
-                    mime: /.+/
-                  license:
-                    type: /.+/
+                - section_id: /.+/
+                  type: /.+/
 
 # copied from MCS spec.yaml
 definitions:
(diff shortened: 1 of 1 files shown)


Pchelolo:
Congrats @mdholloway - you've made PR #1000 :)

mdholloway:
\o/

Pchelolo:
1 test is still failing for the media endpoint.

d00rman:
@mdholloway It would also be good for RB's and MCS' specs to have the same `x-maples` for the same end points, so that when things start failing we can identify the source more easily. Right now RB is alerting because of the change in MCS, but MCS is not.

Pchelolo:
@mdholloway test is still failing with `body item original not found in the response`

mdholloway:
> @mdholloway It would also be good for RB's and MCS' specs to have the same x-maples for the same end points, so that when things start failing we can identify the source more easily. Right now RB is alerting because of the change in MCS, but MCS is not.

See https://gerrit.wikimedia.org/r/#/c/437636/.  Small change to follow in restbase as well.

v1/media.yaml
```diff
@@ -254,11 +245,9 @@ definitions:
         type: string
         description: description of the image from Wikimedia Commons
     required:
-      - titles
-      - file_page
+      - section_id
       - type
-      - thumbnail
-      - license
+      - original
```
d00rman: There is something wrong here @mdholloway : the spec test requires `section_id` and `type` to be present, but the definition additionally requires the `original` field. These need to match.

PR #1004, opened 2018-06-12
Include `page_language` into the title_revision table.

Pchelolo:
For language variants support it's not enough to look
whether the domain language supports multiple variants,
since on every wiki each page can actually have it's own
lang, and since LanguageConverter is enabled universally,
the individual pages could be transformed even on wikis
where we generically don't support variants.

Thus, we need to include `pagelanguage` property into
the title_revision table in order to be able to check,
whether to try the transformtion, or just to skip it.
Also will be useful to do Varnish cache splitting more
intelegently.

Bug: https://phabricator.wikimedia.org/T190689

### DIFF
--- a/sys/page_revisions.js
+++ b/sys/page_revisions.js
@@ -43,7 +43,7 @@ class PRS {
     getTableSchema() {
         return {
             table: tableName,
-            version: 1,
+            version: 2,
             attributes: {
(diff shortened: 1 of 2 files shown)


Pchelolo:
This requires a schema migration!

PR #1005, opened 2018-06-12
Add support for language variant filter

Pchelolo:
WIP: Pending actual parsoid transformations support.

cc @wikimedia/services

### DIFF
--- a/lib/language_variants_filter.js
+++ b/lib/language_variants_filter.js
@@ -9,7 +9,18 @@ module.exports = (hyper, req, next, options, specInfo) => {
     const acceptLanguage = req.headers && req.headers['accept-language'];
     if (!acceptLanguage || mwUtil.isNoCacheRequest(req)) {
         delete req.headers['accept-language'];
-        return next(hyper, req);
+        return next(hyper, req)
+        // TODO: Store the Vary header so that if the page is possible
+        // to be converted we return a proper Vary even when accept-language
+        // is not specified.
+        // For now we hacky check by domain.
+        .then((res) => {
+            const langCode = req.params.domain.substring(0, req.params.domain.indexOf('.'));
+            if (mwUtil.canConvertLangVariant(hyper, req, langCode)) {
+                res.headers.vary = 'accept-language';
+            }
+            return res;
+        });
     }
 
     const revTableURI = [rp.domain, 'sys', 'page_revisions', 'page', rp.title];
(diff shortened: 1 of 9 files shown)


lib/mwUtil.js
```diff
@@ -66,6 +66,31 @@ mwUtil.parseETag = (etag) => {
...
+    }
+    return mwUtil.getSiteInfo(hyper, req)
+    .then((siteInfo) => {
+        // First, whether specialized variants exist for the page lang on a wiki
+        const pageLangVariants = siteInfo.languagevariants[pageLanguage];
+        if (!pageLangVariants) {
+            return false;
+        }
+        // Second, if the requested language variant exist for page language
+        return pageLangVariants.indexOf(acceptLanguage) !== -1;
```
d00rman: It would be good if the check were case-insensitive.

lib/language_variants_filter.js
```diff
@@ -0,0 +1,60 @@
...
+        .then((shouldConvert) => {
+            if (!shouldConvert) {
+                delete req.headers['accept-language'];
+                return next(hyper, req);
+            }
+
+            if (/\/page\/html\//.test(specInfo.path)) {
+                // It's HTML, hit Parsoid for conversion
+                return next(hyper, req)
+                .then(html => hyper.post({
```
d00rman: All this block looks like it belongs in the Parsoid module rather than here since we keep all Parsoid-related interfaces and logic there.

lib/language_variants_filter.js
```diff
@@ -0,0 +1,60 @@
...
+                                source: null,
+                                target: acceptLanguage
+                            }
+                        }
+                    }
+                }));
+                // We can skip setting Vary: accept-language as Parsoid sets it.
+            } else {
+                // TODO: It's something else, so just forward to MCS
+                return next(hyper, req);
```
d00rman: Right now this will also store the result while forwarding the A-L header, won't it?

sys/action.js
```diff
@@ -292,13 +292,31 @@ class ActionService {
...
+                //   'lang' => ['variant1', 'variant2' ... ]
+                // }
+                // for ease of use.
+                const origVariants = res.body.query.languagevariants;
+                const variants = {};
+                if (origVariants) {
+                    Object.keys(origVariants).forEach((lang) => {
+                        variants[lang] = Object.keys(origVariants[lang])
+                        // Filter out non-specific variants like `en`, `zh` etc.
+                        .filter(variant => /-/.test(variant));
```
d00rman: What about multi-lang wikis? Wouldn't `en` be a valid "variant" for commons, e.g.?

sys/parsoid.js
```diff
@@ -735,6 +735,7 @@ class ParsoidService {
             headers: {
                 'content-type': 'application/json',
                 'user-agent': req['user-agent'],
+                'content-language': req.headers['content-language']
```
d00rman: Did you mean `accept-language` ?

lib/language_variants_filter.js
Pchelolo: It's basically just making a request to the parsoid module here. We can probably move a bit more logic in there, but not too much..

sys/action.js
Pchelolo: In case of commons/wikidata it's all pretty different - this is just about conversion, commons actually would return a completely different response. I don't think parsoid even supports that case.

sys/parsoid.js
Pchelolo: Nope, I meant `content-language` - parsoid requires that header to be set for the conversion request.

PR #1007, opened 2018-06-19
Store the Vary header per page that we get from Parsoid

Pchelolo:
Now Parsoid sets proper per-page Vary header if language variant transformation is available for that page. We need to store it. 

This requires a Cassandra schema change!

cc @wikimedia/services


d00rman:
We'll also need to strip `accept` from Parsoid's `vary` header, as they will be setting it as per [T197702](https://phabricator.wikimedia.org/T197702)

Pchelolo:
> We'll also need to strip accept from Parsoid's vary header, as they will be setting it as per T197702

The new version of the PR does so

sys/parsoid.js
```diff
@@ -166,6 +166,16 @@ function compileReRenderBlacklist(blacklist) {
     return result;
 }
 
+function stripAcceptFromVary(res) {
+    if (res.headers.vary) {
```
d00rman: Let's make this `if (res && res.headers && res.headers.vary)` to be on the safe side.

Pchelolo: Done.

d00rman:
LGTM. Let's wait with the merge until after the fields have been created in dev and production.

Pchelolo:
So, let's apply the schema transforms? https://phabricator.wikimedia.org/T197789

PR #1010, opened 2018-06-21
Support language variants for summary and mobile content

Pchelolo:

### DIFF
--- a/lib/normalize_headers_filter.js
+++ b/lib/normalize_headers_filter.js
@@ -1,0 +1,12 @@
+"use strict";
+
+const mwUtil = require('./mwUtil');
+
+module.exports = function normalizeHeaders(hyper, req, next) {
+    if (mwUtil.isNoCacheRequest(req)) {
+        req.headers['cache-control'] = 'no-cache';
+    } else {
+        delete req.headers['cache-control'];
+    }
+    return next(hyper, req);
+};
--- a/lib/normalize_headers_filter.js
+++ b/lib/normalize_headers_filter.js
@@ -5,7 +5,7 @@ const mwUtil = require('./mwUtil');
 module.exports = function normalizeHeaders(hyper, req, next) {
     if (mwUtil.isNoCacheRequest(req)) {
         req.headers['cache-control'] = 'no-cache';
-    } else {
+    } else if (req.headers && req.headers['cache-control']) {
         delete req.headers['cache-control'];
     }
     return next(hyper, req);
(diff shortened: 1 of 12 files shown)


v1/summary.yaml
```diff
@@ -145,15 +145,16 @@ paths:
             request:
               method: put
               uri: /{domain}/sys/key_value/page_summary/{request.params.title}
-              headers: '{{merge({"if-none-hash-match": "*"}, extract.headers)}}'
+              headers: '{{merge({"if-none-hash-match": "*"}, extract.headers, "cache-control": request.headers.cache-control)}}'
```
d00rman: Why is the request's `cache-control` being added here and below?

Pchelolo: @d00rman to pass through the `no-store` header.

Pchelolo:
Doesn't Varnish strip those headers?

Pchelolo:
Rebased

Pchelolo:
All the dependencies were deployed, so we're ready to go fowrward with this.

d00rman:
> Doesn't Varnish strip those headers?

Yes, Varnish does, but what about internal clients? Better safe than sorry

v1/summary.yaml
```diff
@@ -145,15 +145,16 @@ paths:
...
               body: '{{extract.body}}'
             # With the if-none-hash-match header the storage will return 412
             # if the content is not changed. In that case, return from the
             # handler completely, and avoid issuing purges.
+            # Also if we didn't really store anything, return as well.
             catch:
-              status: 412
+              status: [ 412, 202 ]
```
d00rman: Nit: let's sort these numbers.

v1/summary.yaml
```diff
@@ -145,15 +145,16 @@ paths:
...
             return_if:
-              status: 412
+              status: [ 412, 202 ]
```
d00rman: Idem.

v1/summary.yaml
Pchelolo: Done

lib/normalize_headers_filter.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+const mwUtil = require('./mwUtil');
+
+module.exports = function normalizeHeaders(hyper, req, next) {
+    if (mwUtil.isNoCacheRequest(req)) {
+        req.headers['cache-control'] = 'no-cache';
+    } else {
```
d00rman: Nit: `else if (req.headers && req.headers['cache-control'])` (I wonder if this is needed, i.e. is there somewhere where we ensure the existence of the headers hash?)

PR #1011, opened 2018-06-25
Public documentation for Accept-Language header

Pchelolo:
WIP until we enable it for everything.

cc @wikimedia/services

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -152,6 +152,13 @@ paths:
             To get a 200 response instead, supply `false` to the `redirect` parameter.
           type: boolean
           required: false
+        - name: Accept-Language
+          in: header
+          description: >
+            The desired language variant code for wikis where LanguageConverter is enabled.
+            Example: `sr-el` for latin trascription of the Serbian language.
+          required: false
+          type: string
       produces:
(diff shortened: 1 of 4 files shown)


v1/content.yaml
```diff
@@ -152,6 +152,13 @@ paths:
             To get a 200 response instead, supply `false` to the `redirect` parameter.
           type: boolean
           required: false
+        - name: Accept-Language
+          in: header
+          description: >
+            The desired language variant code for wikis where LanguageConverter is enabled.
+            Example: `sr-el` for latin trascription of the Serbian language.
```
jdforrester: * Latin, * transcription

v1/content.yaml
```diff
@@ -438,6 +445,13 @@ paths:
...
```
jdforrester: Same here and below.

v1/content.yaml
Pchelolo: Sorry, didn't quite understand what you meant here

berndsi: "Latin" should be upper case. "trascription" is a typo, missing an n. Should be "transcription".

Pchelolo: oh.. didn't notice the typo, so the `*` made me confused. Sorry.

jdforrester: My fault, I should have been clearer, sorry. :-)

PR #1013, opened 2018-07-09
Update the related endpoint spec

Pchelolo:
https://phabricator.wikimedia.org/T199099


mdholloway:
LGTM.  There are a couple of pdfrender tests failing in Travis but they pass on my machine and aren't related to this change anyway.

PR #1017, opened 2018-07-10
Deduplicate most-read results

mdholloway:
Builds on some deduplication functionality added earlier for on-this-day by generalizing it and adding an optional updater function to update original items based on their duplicates.  This fixes a situation in which pages can appear more than once in the most-read/trending list if they are moved at the height of their popularity. 

Alternate solution to the mobileapps-only patch in https://gerrit.wikimedia.org/r/#/c/mediawiki/services/mobileapps/+/434806/. 

Bug: https://phabricator.wikimedia.org/T195390

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -566,12 +566,19 @@ mwUtil.constructRegex = (list) => {
 /**
  * Remove entries with duplicate 'title' values from an array of objects
  * @param {?Array} arr Array of page/article objects
+ * @param {?Function} upd Optional function to update the original item based on the duplicate.
+ *   Takes two parameters (orig, dupe): orig - the original object, dupe - the duplicate
  * @return {?Array} arr deduplicated by 'title'
  */
-mwUtil.removeDuplicateTitles = (arr) => {
+mwUtil.removeDuplicateTitles = (arr, upd) => {
     const titles = {};
-    return arr.filter((item) => {
+    return arr.filter((item, idx, init) => {
         if (titles[item.title]) {
+            if (upd) {
+                // eslint-disable-next-line no-unused-vars
+                let orig = init[init.findIndex(el => el.title === item.title)];
+                orig = upd(orig, item);
+            }
             return false;
         }
         titles[item.title] = true;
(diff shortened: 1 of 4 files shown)


d00rman:
I would agree with your comment @mdholloway on the gerrit patch: de-duplication is a business-logic decision and should probably be better handled in MCS rather than here. But, that seems to be a chicken and egg problem since duplicate results arise after hydrating the summaries.

I would be for self-containment and vote for moving duplicates resolution (both for most read and onthisday) to MCS. @Pchelolo what do you think?

Pchelolo:
So, do I understand correctly that duplicates appear because of redirects? Does that mean, that analytics count several 'names' of the same article as separate articles? Does that make sence? Shouldn't that be a pageviews service task to merge redirects (if we want to merge the redirects). Personally, I could see value in the pageviews service providing both merged and non-merged statisticts, with redirect-merged stats being a bit more important, but let's cc @milimetric @ottomata @nuria and see what actually is happening in the pageviews service.

nuria:
FYI that mediawiki does not store historical redirect information, so, yes, several 'names' of an article are counted as different articles. 

Example (note missspelling, is really "Barack") "Barak Obama" 
https://wikimedia.org/api/rest_v1/metrics/pageviews/per-article/en.wikipedia/all-access/all-agents/Barak_Obama/daily/2015100100/2015103100

mdholloway:
> I would be for self-containment and vote for moving duplicates resolution (both for most read and onthisday) to MCS.
This is my preference as well.  @berndsi, any objection?

Pchelolo:
>> I would be for self-containment and vote for moving duplicates resolution (both for most read and onthisday) to MCS.

> This is my preference as well. @berndsi, any objection?

None here if it could be implemented efficiently. I believe that redirects in the individual articles are resolved while hydrating the summaries, so you will need to do the redirect resolution work again. How about moving the hydration into MCS as well? :)

Pchelolo:
Updated previous comment if someone didn't notice

berndsi:
If we move hydration to MCS as well then I don't have an objection to moving duplicate resolution to MCS as well.

mdholloway:
It's a plan! https://phabricator.wikimedia.org/T199455

PR #1018, opened 2018-07-10
Hygiene: Always use https for wikimedia/wmflabs URLs in example configs

mdholloway:


config.example.wikimedia.yaml
```diff
@@ -17,27 +17,27 @@ default_project: &default_project
...
         parsoid:
-          host: http://parsoid-beta.wmflabs.org
+          host: https://parsoid-beta.wmflabs.org
           grace_ttl: 1
         action:
           apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
           baseUriTemplate: "{{'https://{domain}/api/rest_v1'}}"
         graphoid:
-          host: http://graphoid.wikimedia.org
+          host: https://graphoid.wikimedia.org
```
jdforrester: Not routable?

d00rman: Indeed, this doesn't exist any more. Let's change this to `graphoid-beta.wmflabs.org`

Pchelolo:
If we're touching this, we can also explicitly prohibit http in x-request-filters around here https://github.com/wikimedia/restbase/pull/1018/files#diff-541b6e195e9da580fc0bd42761ee761fR88

(remove `?` from the `https?` regex)

Pchelolo:
Let's see what travis says, but I don't anticipate and tests to fail since they use test config. Did you try manually using the example config and making some random requests to RB? Does it still work?

mdholloway:
Yes, I tried out several requests to various endpoints and all seems fine.

Pchelolo:
Cool. Thank you @mdholloway

PR #1021, opened 2018-07-12
Media: Remove revision parameter

mdholloway:
Updates this to reflect that the MCS media endpoint no longer supports
specific revision requests as of the updates in the most recent
deployment.

### DIFF
--- a/v1/media.yaml
+++ b/v1/media.yaml
@@ -11,7 +11,7 @@ info:
     name: Apache licence, v2
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /media/{title}{/revision}:
+  /media/{title}:
     x-route-filters:
       - path: ./lib/revision_table_access_check_filter.js
         options:
@@ -34,12 +34,6 @@ paths:
           description: "Page title. Use underscores instead of spaces. Example: `Main_Page`."
           type: string
           required: true
-        - name: revision
-          in: path
-          description: >
-            Optional page revision. Note that older revisions are not stored, so
-            request latency with the revision would be higher.
-          required: false
         - name: redirect
           in: query
           description: >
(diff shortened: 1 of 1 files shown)


Pchelolo:
Why doesn't it? It would be fairly weird for the user if they read one revision of a page, and while they're reading a new revision with completely different media pops up

mdholloway:
It can't because the MediaWiki API query now used to get the file page titles only considers the latest revision, and there's apparently no way to specify a previous revision.

If that's a problem for clients then that's a strong argument in favor of reverting the change (which didn't really help on performance as intended).

Pchelolo:
> If that's a problem for clients then that's a strong argument in favor of reverting the change (which didn't really help on performance as intended).

For that we better ask the apps teams or product managers, I was just wondering what happened

mdholloway:
AFAIK the apps only request pages by title and not by rev ID, but let's ping @dbrant and @joewalsh for comment.

jdforrester:
> It can't because the MediaWiki API query now used to get the file page titles only considers the latest revision, and there's apparently no way to specify a previous revision.

The reason is that the information we're requesting is stored in the `page_props` table, which only has data for the latest revision by design, and so for performance reasons the API doesn't let you try to ask it to calculate live the information requested each time.

berndsi:
I would prefer keeping the ability to specify an older version. The media information is something which could reasonably be requested quite some time after the corresponding page was loaded. It's important that the media info matches the loaded page.

mdholloway:
Yeah, especially since the patch didn't actually move the needle on performance, it's probably not worth giving up the ability to specify a revision.  I'll revert.

mdholloway:
Reversion patch: https://gerrit.wikimedia.org/r/#/c/mediawiki/services/mobileapps/+/445551/

PR #1023, opened 2018-07-12
Expose Javascript endpoint

berndsi:
Another Page Content Service endpoint. This patch provides one
page-independent JavaScript response under /data/javascript/mobile/:
* /data/javascript/mobile/pagelib

This is somewhat similar to the CSS endpoints.

### DIFF
--- a/projects/dev.yaml
+++ b/projects/dev.yaml
@@ -55,6 +55,9 @@ paths:
                 - path: v1/css.yaml
                   options:
                     host: '{{options.mobileapps.host}}'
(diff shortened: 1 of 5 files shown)


berndsi:
We plan to deploy this during our regular Services deploy window today.

d00rman:
K thank you @berndsi . Please ping us on this PR once that happens so we can proceed.

berndsi:
@d00rman The MCS part is deployed now.

PR #1024, opened 2018-07-12
Page Content Service: Expose mobile-html endpoint

berndsi:
This is somewhat similar to the other page specific endpoints: summary, metadata, media, references.

Bug: https://phabricator.wikimedia.org/T199491


berndsi:
Renamed to mobile-html + rebased.

mdholloway:
The MCS change is deployed, and CI is now passing.

test/features/mobile-html.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
```
d00rman: This is not needed because of the `x-amples` below in `v1/mobile-html.yaml`

PR #1027, opened 2018-07-24
Transform end points: Raise limit to 25 rps

d00rman:
After splitting the Parsoid traffic for external and internal updates
between the two DCs, the one serving live traffic can handle more load,
so allow clients to benefit from that.

### DIFF
--- a/v1/transform.yaml
+++ b/v1/transform.yaml
@@ -31,7 +31,7 @@ paths:
         readable.
 
         - Stability: [stable](https://www.mediawiki.org/wiki/API_versioning#Stable)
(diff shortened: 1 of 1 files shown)


d00rman:
cc @Pchelolo @subbuss @arlolra

arlolra:
eqiad does look like it can handle more load, but what's the impetus for doing this?  Are some clients being limited?  And I guess we gotta keep in mind the total load in case of DC failover.

arlolra:
Nevermind, @subbuss explained this to me.

PR #1028, opened 2018-07-26
mobile-html: Do not anchor the ETag field in x-amples

d00rman:
The ETag header might have the optional `W/` portion in front of the
revision number, so do not anchor the regex testing it.

Bug: [T199491](https://phabricator.wikimedia.org/T199491)

### DIFF
--- a/v1/mobile-html.yaml
+++ b/v1/mobile-html.yaml
@@ -132,6 +132,6 @@ paths:
             status: 200
             headers:
               content-type: /^text\/html/
-              etag: /^"830543386\/[0-9a-f-]+"$/
+              etag: /830543386\/[0-9a-f-]+/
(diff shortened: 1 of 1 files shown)


d00rman:
cc @Pchelolo @berndsi @mdholloway 

Will self-merge and deploy since RESTBase checks in production are now failing because of this.

v1/mobile-html.yaml
```diff
@@ -132,6 +132,6 @@ paths:
             status: 200
             headers:
               content-type: /^text\/html/
-              etag: /^"830543386\/[0-9a-f-]+"$/
+              etag: /830543386\/[0-9a-f-]+/
```
berndsi: How about `/^W?"830543386\/[0-9a-f-]+"$/` instead?

PR #1032, opened 2018-07-31
Update references spec

berndsi:
Some minor updates, from https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/449326/3..4
which were necessary to make the spec validation work with a live example, which has a null id and no section_heading.

### DIFF
--- a/v1/references.yaml
+++ b/v1/references.yaml
@@ -155,7 +155,8 @@ definitions:
     description: A single reference list
     properties:
       id:
-        type: string
+        type: ['string', 'null']
+        nullable: true
         description: Identifier for the whole reference list. May be null.
       section_heading:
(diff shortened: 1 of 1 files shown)


v1/references.yaml
```diff
@@ -155,7 +155,8 @@ definitions:
     description: A single reference list
     properties:
       id:
-        type: string
+        type: ['string', 'null']
+        nullable: true
```
d00rman: Is `nullable` needed if `null` is one of the possible types/values?

berndsi: nullable is not needed yet. I just kept it so we don't forget to add it when we move to Swagger 3.0 or higher.

PR #1034, opened 2018-07-31
Add application/problem+json to produces stanza for remaining endpoints

berndsi:
This covers the remaining endpoints exposed through RESTBase,
whenever the response had something like:
        default:
          description: Error
          schema:
            $ref: '#/definitions/problem'

In one case, for content.yaml, I replaced an application/json with
application/problem+json since I didn't see a reason for it to
respond with application/json.

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -24,6 +24,7 @@ paths:
         Stability: [stable](https://www.mediawiki.org/wiki/API_versioning#Stable)
       produces:
         - application/json
+        - application/problem+json
       responses:
         '200':
           description: A list of page-related API end points.
@@ -47,6 +48,7 @@ paths:
         Stability: [stable](https://www.mediawiki.org/wiki/API_versioning#Stable).
       produces:
         - application/json
+        - application/problem+json
       parameters:
         - name: title
           in: path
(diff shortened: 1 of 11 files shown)


v1/content.yaml
```diff
@@ -161,7 +163,7 @@ paths:
           type: string
       produces:
         - text/html; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/HTML/1.6.0"
-        - application/json
+        - application/problem+json
```
berndsi: Note this is a replace. I don't think Parsoid would produce regular `application/json` here.

v1/content.yaml
```diff
@@ -416,7 +420,7 @@ paths:
       operationId: getFormatRevision
       produces:
         - text/html; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/HTML/1.6.0"
-        - application/json
+        - application/problem+json
```
berndsi: Same as above.

v1/content.yaml
Pchelolo: This is incorrect. this endpoint can return actual json in case section ids are provided

berndsi: Ok, brought both of these back. Thanks!

PR #1035, opened 2018-07-31
Fix simple Eslint issues in some tests

berndsi:
Also removed a few occurrences of unused variables when I noticed them.

### DIFF
--- a/test/features/buckets/key_value_bucket.js
+++ b/test/features/buckets/key_value_bucket.js
@@ -1,7 +1,5 @@
 'use strict';
 
-/* global describe, it, before */
-
 const preq   = require('preq');
 const assert = require('../../utils/assert.js');
 const server = require('../../utils/server.js');
--- a/test/features/buckets/key_value_bucket.js
+++ b/test/features/buckets/key_value_bucket.js
@@ -1,4 +1,6 @@
-"use strict";
+'use strict';
+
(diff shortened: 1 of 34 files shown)


Pchelolo:
omg that's a lot :) I never had the time to eslint the test code, so it's just excluded from the linter, but this will make it much easier to actually enable eslint on test code. Thank you @berndsi 

Will wait for travis before merging.

test/features/buckets/key_value_bucket.js
```diff
@@ -1,4 +1,6 @@
-"use strict";
+'use strict';
+
+/* global describe, it, before */
```
Pchelolo: Actually, you can avoid doing that by placing a test-specific eslint config into the root of the test folder - I think it's much nicer like this. 

Example: https://github.com/wikimedia/preq/blob/master/test/.eslintrc.yml

berndsi:
Not sure why the test is failing. I just ran it twice locally, and it was fine.
Added the .eslintrc.yml file, as you recommended.

Pchelolo:
let's wait for it to run again, hopefully, they will fix themselves

Pchelolo:
ok, bien. travis kinda pass

Pchelolo:
Thank you @berndsi !

PR #1036, opened 2018-07-31
Update readme: node 6+ is required

berndsi:


Pchelolo:
Since we require node 6+ let's also add that info into the package.json

berndsi:
@Pchelolo Where in package.json?
I already see there:
```
   "engines": {
    "node": ">=6"
  },
  "deploy": {
    "node": "6.11.1",
    "target": "debian",
    "dependencies": {
      "_all": []
    }
  }
```
Not sure what else is missing.

Pchelolo:
Oh. Sorry about that..

PR #1037, opened 2018-07-31
Mathoid test failure

berndsi:
Changed the test code to be as it was probably intended, iterating over
all values in format. Before this simple change it was just testing the
last value in formats, `png`.
There is no `mml` in the following string:
`application/mathml+xml; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/MathML/1.0.0"`

Found by eslint warning "mutable variable is accessible inside closure".

### DIFF
--- a/test/features/mathoid.js
+++ b/test/features/mathoid.js
@@ -87,7 +87,7 @@ describe('Mathoid', function() {
     });
 
     for (let i = 0; i < formats.length; i++) {
-        var format = formats[i];
+        const format = formats[i];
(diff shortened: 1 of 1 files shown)


berndsi:
So, this would fix the test as it was intended but now it fails because of an issue it found.

Pchelolo:
Not a mathoid expert myself, need help from @d00rman @physikerwelt

d00rman:
Wow, nice find @berndsi ! The test now fails because `mml` is not a mime type, but `mathml` is, sigh. Fixed in #1040 hence closing this one.

PR #1038, opened 2018-07-31
Simplify eslint config for tests

berndsi:
The same can be accomplished by setting mocha to true.


Pchelolo:
Neat! We need to do the same in preq and mediawiki-title.

berndsi:
I actually found this in MCS. :) props to @niedzielski.

Pchelolo:
> Neat! We need to do the same in preq and mediawiki-title.

Just did that.

PR #1043, opened 2018-08-02
TEMP! Split PDF traffic

Pchelolo:
https://phabricator.wikimedia.org/T186748

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -48,6 +48,7 @@ default_project: &default_project
           # Cache PDF for 5 minutes since it's not purged
           cache_control: s-maxage=600, max-age=600
           uri: https://pdfrender-beta.wmflabs.org
+          new_uri: https://proton-beta.wmflabs.org
           secret: secret
           scheme: https
         transform:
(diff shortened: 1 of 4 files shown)


Pchelolo:
It's a no-op until a config change. We can make the config change in beta cluster only for now.

On a later stage we will expose the type and format parameters as well - as soon as we're sure we're switching to the new imlementation.

Pchelolo:
The tests pass locally... interesting

d00rman:
> We should also remember to make a grafana dashboard for Proton.

Created [T201158](https://phabricator.wikimedia.org/T201158) for that. Let's not merge this PR until that is completed.

d00rman:
@Pchelolo could we also have an undocumented query param that allows one to retrieve the PDF generated by Proton? I figure this would be very helpful for Reading to evaluate the visual output that it generates.

Pchelolo:
> could we also have an undocumented query param that allows one to retrieve the PDF generated by Proton?

Done, added `new_pdf` hidden query parameter.

d00rman:
> Done, added new_pdf hidden query parameter.

Obrigado @Pchelolo !

PR #1044, opened 2018-08-06
Remove jshint and jscs in favor of eslint

Pchelolo:


berndsi:
+1 
About the test failure. We had today (Monday) also a 504 result happening in the [periodic MCS test (diff test)](https://integration.wikimedia.org/ci/job/mobileapps-periodic-test/1972/). Coincidence?

d00rman:
That's probably MW in Beta failing to respond in both instances. This is known to happen from time to time.

PR #1045, opened 2018-08-08
Remove useless endpoints from non-wikipedia projects

Pchelolo:
Bug: https://phabricator.wikimedia.org/T201103


Pchelolo:
Config change https://gerrit.wikimedia.org/r/#/c/mediawiki/services/restbase/deploy/+/451523/

PR #1047, opened 2018-08-16
Added rvslots=main to the MW API revision request.

Pchelolo:
Bug: https://phabricator.wikimedia.org/T201974

### DIFF
--- a/sys/page_revisions.js
+++ b/sys/page_revisions.js
@@ -542,7 +542,7 @@ class PRS {
                 action: 'query',
                 prop: 'info|revisions',
                 continue: '',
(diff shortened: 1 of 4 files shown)


mdholloway:
Looks like there's a call that needs fixing in a test, too: https://github.com/wikimedia/restbase/blob/master/test/features/errors/invalid_request.js#L32-L42

Pchelolo:
Oh thank you, great find @mdholloway

mdholloway:
Actually, now that I look closer, why are you requesting contentmodel?  It doesn't appear to be used.

Pchelolo:
Hm... ok. let's redo this :)

PR #1048, opened 2018-08-20
PCS: expose pagelib CSS

berndsi:
http://localhost:7231/en.wikipedia.org/v1/data/css/mobile/pagelib

Bug: https://phabricator.wikimedia.org/T202105

### DIFF
--- a/v1/css.yaml
+++ b/v1/css.yaml
@@ -54,6 +54,49 @@ paths:
               content-type: /^text\/css; charset=utf-8/
             body: /.+/
 
+  /css/mobile/pagelib:
+    x-route-filters:
+      - path: lib/security_response_header_filter.js
+    get:
+      tags:
+        - Mobile
+      summary: Get styles from the wikimedia-page-library.
+      description: |
+        Gets common CSS mobile apps need to properly display pages using Page Content Service.
+
+        Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
+      produces:
+        - application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/CSS/1.0.0"
+        - application/problem+json
+      responses:
+        '200':
+          description: Success
+          headers:
+            ETag:
+              description: Different values indicate that the content has changed
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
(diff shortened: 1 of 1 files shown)


Pchelolo:
What prevents us from merging the 2 and make `{type}` a parameter with enum and 2 possible values `base` or `pagelib`?

berndsi:
> What prevents us from merging the 2 and make {type} a parameter with enum and 2 possible values base or pagelib?

Besides that we would probably need to do this for all three variations (base/pagelib/site), not much. There are some minor differences in the `summary` and the `x-amples.title` fields. I think the latter could be just two (or three) examples.

Pchelolo:
The latter can be just several examples - easy!

The former - we can just make a combined description with something like a bullet-point list with different variants.

I think it would be nicer then a giant copy-paste

berndsi:
Ok, will give it a shot.

Pchelolo:
👌

berndsi:
CI's still not happy but a bit better than before. Still, all seem unrelated issues. This time it's Node 6 with a PDF service 500.

Pchelolo:
I'm almost inclined to make the PDF tests just a warning. They're too annoying

v1/css.yaml
```diff
@@ -11,20 +11,37 @@ info:
     name: Apache licence, v2
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /css/mobile/base:
+  /css/mobile/{file}:
```
Pchelolo: Bikeshedding... Maybe 'type' or smth?

d00rman: +1, while this does get a file, in the context of the API `type` sounds more appropriate. Let's change that, otherwise LGTM.

berndsi: Ok, I could go with that. The other idea I had would be `bundle`. What do you think of that?

Pchelolo: > Ok, I could go with that. Another idea I had would be bundle. What do you think of that?

This is a totally bikeshedding session here, but I personally don't like `bundle` - the bundle is what's returned, and the parameter for the bundle is the `type` of the `bundle`.

berndsi: Ok, `type` it is then.

berndsi: Done

PR #1052, opened 2018-08-23
WIP: Add content negotiation filter

Pchelolo:
WIP WIP WIP!

1. To finish and test this we need some real conversion endpoint in Parsoid, even if it will just change the version number in the meta tag.
2. Accept header parsing is very basic, we need to think what features of the accept header will we support.
3. Vary: Accept is not being set, we also need to expand on https://gerrit.wikimedia.org/r/#/c/operations/puppet/+/281042/ - at least we need to replace the `patch` part of the accept to `.x` in order to avoid cache fragmentation on patch version updates.
4. The HTML endpoints can serve JSON in case section IDs are requested. It's questionable what to do in this case and whether someone actually uses the sections.

I will file some tasks for these questions.

cc @wikimedia/services @wikimedia/parsing


Pchelolo:
This is completely completed, was tested in Vagrant and seems to work OK. Next step is merging it and deploying to Labs to test VE some more. After - coordinated deploy together with @wikimedia/parsing as it's the first time we do anything like this and a lot of caution is needed.

lib/content_negotiation_filter.js
```diff
@@ -0,0 +1,100 @@
...
+        const storedVersion =
+            mwUtil.extractHTMLProfileVersion(htmlRes.headers['content-type']);
+        if (!storedVersion) {
+            // TODO: This can happen if sections are requested;
+            // Think what to do with that.
+            // If it's not the sections - something is very wrong on our side.
+            return htmlRes;
+        }
+
+        if (semver.satisfies(storedVersion, requestedVersion)) {
```
arlolra: This might be clearer as `eq` since the `requestedVersion` isn't a range.

But maybe you meant to use a partial or x-range (to ignore the patch)?

Pchelolo: We ignore the patch both in Varnish (by basically resetting the patch to 0) and in RESTBase

Pchelolo: https://github.com/wikimedia/restbase/pull/1069

PR #1056, opened 2018-08-31
Update and add metrics endpoints

jobar:
AQS have been updated to serve endpoints using page-title and
user-text instead of IDs. This patch updates the existing
endpoints that used ids and adds new endpoints by-user and
by-page.


nuria:
Looking at https://wikimedia.org/api/rest_v1/#/ seems like this has been deployed can @Pchelolo confirm?

PR #1059, opened 2018-09-07
Do not emit the dependency update event for non-latest rerenders.

Pchelolo:
Bug: [T203823](https://phabricator.wikimedia.org/T203823)
cc @wikimedia/services

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -138,19 +138,24 @@ function replaceSections(original, sectionsJson) {
     return `<body>${newBody}</body>`;
 }
 
-// HTML resource_change event emission
-function _dependenciesUpdate(hyper, req) {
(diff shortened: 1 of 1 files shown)


sys/parsoid.js
```diff
@@ -138,19 +138,24 @@ function replaceSections(original, sectionsJson) {
     return `<body>${newBody}</body>`;
 }
 
-// HTML resource_change event emission
-function _dependenciesUpdate(hyper, req) {
+/** HTML resource_change event emission
+ * hyper {HyperSwitch} the hyperswitch router object
+ * req {Object} the request
+ * [oldContent] {boolean} whether this is an older revision
```
d00rman: `s/oldContent/newContent/`

PR #1063, opened 2018-09-19
key_rev_value: Return the stored headers, if any

d00rman:


sys/key_rev_value.js
```diff
@@ -17,10 +17,9 @@ function returnRevision(req) {
     return (dbResult) => {
         if (dbResult.body && dbResult.body.items && dbResult.body.items.length) {
             const row = dbResult.body.items[0];
-            const headers = {
-                etag: mwUtil.makeETag(row.rev, row.tid),
-                'content-type': row.headers['content-type'] || 'application/octet-stream'
-            };
+            const headers = Object.assign({}, row.headers || {});
+            headers.etag = headers.etag || mwUtil.makeETag(row.rev, row.tid);
```
Pchelolo: This I'm not sure why is needed but tests fail without it

Pchelolo: We should investigate why

Pchelolo:
Since this one should fix lack of `content-language` for content served from the fallback storage we need a test.

Pchelolo:
Test failures are unrelated.

PR #1064, opened 2018-09-20
MCS: Remove tests that directly call the service

d00rman:
Over time we have accumulated numerous x-amples that bypass all RB features and directly call the Mobile Content Service. Since all of these are directly tested in MCS' x-amples, there is no need to do that in RB as well.

Nevertheless, the announcement x-ample has been left to assert that MCS is reachable to RB.


Pchelolo:
Travis is failing because of linting, so merging it is fine.

PR #1065, opened 2018-09-24
Fix up eslint linting for PDF and Lists

Pchelolo:
Accidentally broken the linting.


Pchelolo:
Will self-merge, it's obvious and I need to rebase #1052 on top of it.

PR #1068, opened 2018-09-25
Add more logging in metrics temporarily for content-negotiation

Pchelolo:
For the sake of getting more info on this first deployment let's log more.

cc @wikimedia/services @arlolra @subbuss


subbuss:
lgtm

PR #1069, opened 2018-09-25
Don't try upgrading if .minor version is satisfied ok

Pchelolo:
https://github.com/wikimedia/restbase/pull/1052#pullrequestreview-158677694

cc @wikimedia/services @arlolra

### DIFF
--- a/lib/content_negotiation_filter.js
+++ b/lib/content_negotiation_filter.js
@@ -60,6 +60,9 @@ module.exports = (hyper, req, next, options, specInfo) => {
             .then((newRes) => {
                 const newVersion =
                     mwUtil.extractHTMLProfileVersion(newRes.headers['content-type']);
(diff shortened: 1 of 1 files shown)


arlolra:
LGTM

arlolra:
Maybe deserves a test?

Pchelolo:
actually, just found another weirness. hold on

Pchelolo:
Updated. It does deserve a test indeed, but I'm having troubles creating one. These two issues are for content version upgrade, not downgrade, so we need somehow to obtain the previous version first to store it. Since the tests are running against deployed parsoid, we can only get the upgraded version, not the previous one. I guess I need to mock parsoid first to provide a fake lover version...

d00rman:
Needs a rebase.

d00rman:
Rebased, waiting on travis to merge

PR #1070, opened 2018-09-26
Update metrics top endpoints

jobar:
This patch is to match AQS patches related to T204707.
Wikistats-2 metrics top endpoint now follow the same
convention as top-pageviews, meaning they do not use
a date range anymore, but rather a single day or month.

### DIFF
--- a/v1/metrics.yaml
+++ b/v1/metrics.yaml
@@ -537,9 +537,6 @@ paths:
         You can filter by editor type (all-editor-types, anonymous, group-bot, name-bot, user)
         or page-type (all-page-types, content or non-content). You can choose between daily and
         monthly granularity as well.
-        The new pages value is computed as follow:
-          [number of created pages] - [number of deleted pages] + [number of restored pages]
-        for the chosen filters.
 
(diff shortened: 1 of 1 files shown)


Pchelolo:
Side note: I really want to resolve T204981 cause PRs like this are making me sick.

LGTM. But I will not merge it until tomorrow because of a huge deployment we're making with parisng and we might need to make some quick fixes. 

@jobar what's your timeline on this?

jobar:
@Pchelolo : It would be perfect if we could have this deploy by tomorrow EOD (we have other devs dependent on this one), but there it is not crucial. If you feel more comfortable waiting for early next week (no deploys on friday, right?), please do so :).
I also +1 resolving T204981. I'm a bit tired of duplicating all our endpoints.
Thanks !

Pchelolo:
This does not interfere with other stuff, so I will include it as soon as possible. But it might get undeployed and deployed back a couple of times.

If you could provide some ideas on T204981 that would be awesome. It's getting a bit overwhelming to do all of this without an automated way.

PR #1072, opened 2018-10-02
Remove old hack we reusing TID.

Pchelolo:
During the switch from the old storage to the new storage we've used
to use old storage instead of actually running Parsoid, so we've
needed to reuse the TID of the old content before storing the new
content. The hack was removed, but the reusing code remained. It's not
needed any more without the hack.

cc @wikimedia/services


d00rman:
Parsoid seems to be misconfigured in beta now ([T206107](https://phabricator.wikimedia.org/T206107)), so let's wait for that to be resolved before merging this, just to be on the safe side. Otherwise, LGTM.

d00rman:
Failures are not related to this change, merging.

PR #1073, opened 2018-10-10
Downgrade major-greater to minor-greater if no-cache is required

Pchelolo:
Issue: we have 1.7.0 in storage. Client requests 1.8.0, we rerender from scratch, Parsoid gives 2.0.0, it doesn't satisfy ^1.8.0. So we need to downgrade.

### DIFF
--- a/lib/content_negotiation_filter.js
+++ b/lib/content_negotiation_filter.js
@@ -6,8 +6,51 @@ const HTTPError = HyperSwitch.HTTPError;
 const URI = HyperSwitch.URI;
 const mwUtil = require('./mwUtil');
 
-module.exports = (hyper, req, next, options, specInfo) => {
+function downgrade(hyper, req, html, existingVersion, requestedVersion) {
     const rp = req.params;
+    const start = Date.now();
+    // The version is less then we expected - downgrade.
+    return hyper.post({
+        uri: new URI([rp.domain, 'sys', 'parsoid', 'transform', 'html', 'to', 'html']),
+        headers: {
+            'content-type': 'application/json',
+            'accept': req.headers.accept
+        },
+        body: {
+            original: {
+                html: {
+                    headers: html.headers,
+                    body: html.body.toString()
+                },
+                'data-parsoid': {
+                    body: { ids: {} }
+                }
+            }
+        }
+    })
+    .then((res) => {
+        hyper.metrics.endTiming('negotiation', start, 0.1);
+        // Parsoid transformation doesn't know about our caching policies
+        // or additional headers RESTBase sets, so merge headers from the original
+        // and important headers from the transformation.
+        const resHeaders = res.headers;
+        res.headers = html.headers;
+        res.headers.vary = resHeaders.vary;
+        res.headers['content-type'] = resHeaders['content-type'];
+        return res;
+    })
+    .catch((e) => {
+        hyper.logger.log('error/negotiation', {
+            message: 'Failed to transform content',
+            requested_version: requestedVersion,
+            existing_version: existingVersion,
+            error: e
+        });
+        throw e;
+    });
+}
+
+module.exports = (hyper, req, next, options, specInfo) => {
     if (!mwUtil.isHTMLRoute(specInfo.path)) {
         // The filter is  only supported on HTML routes,
         // but don't punish the client for our own mistake, just log an error.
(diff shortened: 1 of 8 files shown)


Pchelolo:
The tests fail because of an unrelated MCS bug.

lib/content_negotiation_filter.js
```diff
@@ -51,7 +94,8 @@ module.exports = (hyper, req, next, options, specInfo) => {
             return htmlRes;
         }
 
-        if (semver.gt(requestedVersion, storedVersion)) {
+        if (semver.gt(requestedVersion, storedVersion)
+                && !semver.satisfies(storedVersion)) {
```
arlolra: `!semver.satisfies(storedVersion)` needs a second argument

arlolra:
Needs a test, no?

While we're on the subject, did this test ever get done? https://github.com/wikimedia/restbase/pull/1069#issuecomment-424542885

lib/content_negotiation_filter.js
Pchelolo: Ouch, that's aremaining of my first idea how to fix that...

Pchelolo:
@arlolra since we can't deploy due to MW switchover, I will take time to implement the tests. Created a PR to get your opinion on the general approach.

arlolra:
The change here certainly solves the issue you've described.

Pchelolo:
> Needs a test, no? While we're on the subject, did this test ever get done? #1069 (comment)

Done for both this bug and the previous one.

berndsi:
Why does it matter for this if no-cache is required is required?

Pchelolo:
@bearND if you request Parsoid version greater then we have in storage we repeat the request with `no-cache` in order to upgrade. Then if Parsoid returns a major new version, we need to apply a downgrade to meet the client accept header.

Pchelolo:
Tests are failing due to https://phabricator.wikimedia.org/T206701

arlolra:
The tests LGTM

lib/content_negotiation_filter.js
```diff
@@ -61,6 +104,10 @@ module.exports = (hyper, req, next, options, specInfo) => {
                     mwUtil.extractHTMLProfileVersion(newRes.headers['content-type']);
                 if (semver.satisfies(newVersion, `^${requestedVersion}`)) {
                     return newRes;
+                } else if (semver.gt(newVersion, requestedVersion)) {
```
arlolra: Two lingering things from reading the above.

* On L68, the return `if  (!requestedVersion) {` means that it'll just return whatever is in storage, which is not quite the same thing as meaning it'll return the latest version.  Does that matter?

* Does the `no-cache` request on L101 store the response if the returned version is greater?

Pchelolo: > On L68, the return if (!requestedVersion) { means that it'll just return whatever is in storage, which is not quite the same thing as meaning it'll return the latest version. Does that matter?

If we did a `no-cache` for updating the version in that case it would be very prone to DDOS. If we hard-coded in RESTBase which version is currently the latest, that's really tight coupling. We're aiming to have all the clients provide the `Accept` header, so I consider lack of it a bug on the client side, so no guarantees for the buggy client.

> Does the no-cache request on L101 store the response if the returned version is greater?

It stores whatever Parsoid gives back. We aim to have latest parsoid version in storage, so why miss this opportunity to update

arlolra: > > Does that matter?
> If we did a no-cache for updating the version in that case it would be very prone to DDOS. 

Right, I was asking because our announcement tldr says,

> If omitted, they will always get the latest version of the HTML, regardless of any breaking changes it may contain.

which is incorrect.

> It stores whatever Parsoid gives back

Guess you need to be careful about not setting `Accept` when making parse requests then!!


Final question, what's the benefit of blanking the patch level (`requestedVersion = requestedVersion.replace(/\d+$/, '0');`)?  Clients will most often be behind on the patch level and benefit from the caret semantics.  But maybe a client really wants to ensure a bug fix that came in a patch update?

Pchelolo: > Guess you need to be careful about not setting Accept when making parse requests then!!

We totally don't: https://github.com/wikimedia/restbase/blob/master/v1/content.yaml#L223-L227

> Final question, what's the benefit of blanking the patch level (requestedVersion = requestedVersion.replace(/\d+$/, '0');)? Clients will most often be behind on the patch level and benefit from the caret semantics. But maybe a client really wants to ensure a bug fix that came in a patch update?

We actually are doing the same thing in Varnish to avoid too drastic cache fragmentation, so I've added the same logic here in order to make it work predictably with/without Varnish in front of it. After https://phabricator.wikimedia.org/T170606 is resolved (soon) and we accumulate some data we can see whether the cache fragmentation really be a problem for the clients or not and maybe reconsider the decision to ignore the semver patch version

PR #1074, opened 2018-10-17
Spread requests beetween MCS nodes for onthisday

Pchelolo:
The `/feed/onthisday` endpoint is the one that's mostly timing out on deploys and I believe the timeouts are caused by MCS, not the response hydration portion. 

For the `/feed/featured` we make several individual requests for every portion of the feed, so we spread the load between several MCS nodes. Let's try doing the same for the `onthisday` endpoint.

cc @wikimedia/services

### DIFF
--- a/lib/base_feed.js
+++ b/lib/base_feed.js
@@ -19,7 +19,7 @@ class BaseFeed {
         this.options = options;
     }
 
-    _assembleResult(result, dateKey) {
+    _assembleResult(result, dateKey, req) {
         // assemble the final response to be returned
         return {
             status: 200,
@@ -29,7 +29,7 @@ class BaseFeed {
                 etag: `${dateKey}/${uuid.now().toString()}`,
                 'content-type': this.options.content_type
             },
-            body: this.constructBody(result)
+            body: this.constructBody(result, req)
         };
     }
 
@@ -41,7 +41,7 @@ class BaseFeed {
         throw new Error('Abstract. Must be overwritten');
     }
 
-    constructBody(result) {
+    constructBody(result, req) {
         throw new Error('Abstract. Must be overwritten');
     }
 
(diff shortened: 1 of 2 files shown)


Pchelolo:
For comparison of the perf:

Old:
```
ppchelko@restbase-dev1004:~$ ab -c 5 -n 100 http://localhost:7231/en.wikipedia.org/v1/feed/onthisday/all/01/01
This is ApacheBench, Version 2.3 <$Revision: 1604373 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient).....done


Server Software:        
Server Hostname:        localhost
Server Port:            7231

Document Path:          /en.wikipedia.org/v1/feed/onthisday/all/01/01
Document Length:        2495137 bytes

Concurrency Level:      5
Time taken for tests:   83.326 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      249622700 bytes
HTML transferred:       249513700 bytes
Requests per second:    1.20 [#/sec] (mean)
Time per request:       4166.309 [ms] (mean)
Time per request:       833.262 [ms] (mean, across all concurrent requests)
Transfer rate:          2925.52 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:  3436 4078 314.7   4028    5039
Waiting:     3434 4077 314.7   4026    5038
Total:       3436 4078 314.7   4028    5039

Percentage of the requests served within a certain time (ms)
  50%   4028
  66%   4172
  75%   4244
  80%   4314
  90%   4482
  95%   4692
  98%   4961
  99%   5039
 100%   5039 (longest request)
```

PR #1076, opened 2018-10-18
Travis: Use Node 10

d00rman:
We will be skipping Node 8 and moving to Node 10 in the near future, so
let's start testing with it.


d00rman:
Tests passing, trivial, so self-merging.

PR #1077, opened 2018-10-23
Removed accept-language duplicate

clarakosi:
Added a` varyDoesNotContainDuplicates` function in test/utils/assert.js that throws an error whenever there is a duplicate.

Added a test in `test/features/pagecontent/language_variants.js` that checks to see if there are language duplicates in vary.

Changed the `addVaryHeader` method in lib/mwUtil.js to check if a version of the vary exists before adding a new one.

Bug: [T207324](https://phabricator.wikimedia.org/T207324)

### DIFF
--- a/test/utils/assert.js
+++ b/test/utils/assert.js
@@ -200,7 +200,7 @@ function validateListHeader(headerList, options) {
         });
     }
 
-    if (options.allowDuplicates === false || !options.allowDuplicates) {
+    if (!options.allowDuplicates) {
         const filterDuplicates = headerArray.filter((header, index) => headerArray.indexOf(header) === index);
         if (filterDuplicates.length !== headerArray.length) {
             throw new assert.AssertionError({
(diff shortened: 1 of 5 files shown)


test/utils/assert.js
```diff
@@ -160,46 +160,53 @@ function checkString(result, expected, message) {
...
-        });
-    }
-    const varyAsList = res.headers.vary.split(',')
-    .map(header => header.trim().toLowerCase());
-    if (!varyAsList.includes(header.toLowerCase())) {
-        throw new assert.AssertionError({
-            message: `Vary header does not contain ${header}`
-        });
+
+    if (options.allowDuplicates === false || !options.allowDuplicates) {
```
Pchelolo: `false` is obviously falsy, so you don't need the first part of the condition.

Pchelolo:
One last nit inlined, otherwise LGTM

PR #1078, opened 2018-10-26
Added title normalization and tests for wiktionary 

clarakosi:
Modified title in lib/normalize_title_filter.js to check for term in wiktionary before defaulting to title

Added tests to check case sensitivty and namespaces in wiktionary

Added wiktionary url to test/utils/server.js

Modified v1/definition.yaml to register the filter

Bug: [T207904](https://phabricator.wikimedia.org/T207904)

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -341,7 +341,8 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
     const backString = Array.apply(null, { length: pathSuffixCount }).map(() => '../').join('');
     let location;
     if (!options.dropPathAfterTitle) {
-        const pathPatternAfterTitle = path.substring(path.indexOf('{title}') - 1);
+        let pathPatternAfterTitle = path.substring(path.indexOf(`{${titleParamName}`) - 1);
+        pathPatternAfterTitle = pathPatternAfterTitle.replace(/term/, 'title');
         location = backString
             + new URI(pathPatternAfterTitle, newReqParams, true).toString().substr(1)
             + mwUtil.getQueryString(req);
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -341,8 +341,7 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
     const backString = Array.apply(null, { length: pathSuffixCount }).map(() => '../').join('');
     let location;
     if (!options.dropPathAfterTitle) {
-        let pathPatternAfterTitle = path.substring(path.indexOf(`{${titleParamName}`) - 1);
-        pathPatternAfterTitle = pathPatternAfterTitle.replace(titleParamName, 'title');
+        const pathPatternAfterTitle = path.substring(path.indexOf(`{${titleParamName}`) - 1);
         location = backString
             + new URI(pathPatternAfterTitle, newReqParams, true).toString().substr(1)
             + mwUtil.getQueryString(req);
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -342,7 +342,7 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
     let location;
     if (!options.dropPathAfterTitle) {
         let pathPatternAfterTitle = path.substring(path.indexOf(`{${titleParamName}`) - 1);
-        pathPatternAfterTitle = pathPatternAfterTitle.replace(/term/, 'title');
+        pathPatternAfterTitle = pathPatternAfterTitle.replace(titleParamName, 'title');
         location = backString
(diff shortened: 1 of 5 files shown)


lib/normalize_title_filter.js
```diff
@@ -29,6 +29,7 @@ module.exports = (hyper, req, next, options, specInfo) => {
             }
         });
     }
+    rp.title = rp[options.title] ? rp[options.title] : rp.title;
```
Pchelolo: 1. rp[options.title] || rp.title is nicer
2. if options.title is undefined we'd be in trouble. so we need to go
(options.title && rp[options.title]) || rp.title the brackets are not really needed, but it's easier to read without thinking.

test/features/pagecontent/redirects.js
```diff
@@ -9,6 +9,57 @@ describe('redirects', () => {
     before(() => { return server.start(); });
 
     describe('', () => {
+        it('should direct to a lowercase version of a title in wiktionary', () => {
```
Pchelolo: REdirect

test/features/pagecontent/redirects.js
```diff
@@ -9,6 +9,57 @@ describe('redirects', () => {
...
+            })
+            .then((res) => {
+                assert.deepEqual(res.status, 200);
+                assert.deepEqual(res.headers['content-location'],
+                    'https://en.wiktionary.org/api/rest_v1/page/definition/cat');
+                assert.deepEqual(res.headers['cache-control'], 'test_purged_cache_control');
+            });
+        });
+
+        it('should direct to an uppercase version of a title in wiktionary', () => {
```
Pchelolo: Idem

lib/mwUtil.js
```diff
@@ -341,7 +341,8 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
     const backString = Array.apply(null, { length: pathSuffixCount }).map(() => '../').join('');
     let location;
     if (!options.dropPathAfterTitle) {
-        const pathPatternAfterTitle = path.substring(path.indexOf('{title}') - 1);
+        let pathPatternAfterTitle = path.substring(path.indexOf(`{${titleParamName}`) - 1);
+        pathPatternAfterTitle = pathPatternAfterTitle.replace(/term/, 'title');
```
Pchelolo: No way! This is super specific to the case that we have here right now. What if we're gonna have a different endpoint in the future where we decide to call the page title 'ugabugafoobarxxyyblombalabam' - we would need to replace that with 'title' as well.

This is a utility function, and it does take `titleParamName` as an option, so it has to work correctly  whatever the `titleParamName` was provided to it right? Now it will only work if it's 'title' or a 'term'

lib/normalize_title_filter.js
```diff
@@ -29,6 +29,7 @@ module.exports = (hyper, req, next, options, specInfo) => {
             }
         });
     }
+    rp.title = (options.title && rp[options.title]) || rp.title;
```
Pchelolo: Our overall goal here is to get rid of the hard-coded property name `title` right? So why are we assigning the request parameter `title`? We know that that the page name that we need to normalize will be located inside the `options.title` property right? 

Also, having the options property named title is a bit confusing isn't it? now we have so many titles in the code that it's hard to understand which one refers to what. Would it be more understandable if the options property be called 'titleParamName' here as well as in the utility function?

test/features/pagecontent/redirects.js
```diff
@@ -9,12 +9,25 @@ describe('redirects', () => {
...
+                });
+        });
+
         it('should redirect to a normalized version of a title', () => {
             return preq.get({
                 uri: `${server.config.bucketURL}/html/Main%20Page?test=mwAQ`,
                 followRedirect: false
             })
             .then((res) => {
+                console.log(res);
```
Pchelolo: Ups...

Pchelolo:
That's what we're talking about! but you forgot a console.log in the test. Otherwise LGTM

Pchelolo:
Ok, let's ait for travis

lib/mwUtil.js
```diff
@@ -341,7 +341,7 @@ mwUtil.createRelativeTitleRedirect = (path, req, options) => {
     const backString = Array.apply(null, { length: pathSuffixCount }).map(() => '../').join('');
     let location;
     if (!options.dropPathAfterTitle) {
-        const pathPatternAfterTitle = path.substring(path.indexOf('{title}') - 1);
+        const pathPatternAfterTitle = path.substring(path.indexOf(`{${titleParamName}`) - 1);
```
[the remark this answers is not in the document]
d00rman: Oh, oh. I stand corrected mister :)

PR #1084, opened 2018-12-04
Add 'morelike' recommendation public API specification

kodchi:
Bug: T201192

### DIFF
--- a/v1/recommend.yaml
+++ b/v1/recommend.yaml
@@ -56,6 +56,45 @@ paths:
             request:
               uri: '{{options.host}}/{domain}/v1/translation/articles/{from_lang}/{{default(seed_article,"")}}'
       x-monitor: false
+  /recommendation/article/morelike/translation/{seed_article}:
+    x-route-filters:
+      - path: ./lib/access_check_filter.js
+        options:
+          title: 'seed_article'
+    get:
+      tags:
+        - Recommendation
+      summary: Recommend missing articles
+      description: |
+        Recommends articles similar to the seed article but are missing
+        from the domain language Wikipedia.
+
+        Currently, this end point works only on eswiki, fawiki, and uzwiki.
+
+        Stability: [unstable](https://www.mediawiki.org/wiki/API_versioning#Unstable)
+      produces:
+        - application/json
+        - application/problem+json
+      parameters:
+        - name: seed_article
+          in: path
+          description: The article title used to search similar but missing articles
+          type: string
+          required: true
+      responses:
+        '200':
+          description: the prioritized list of Wikidata IDs recommended for creation as Wikipedia articles
+          schema:
+            $ref: '#/definitions/morelike_result'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_backend:
+            request:
+              uri: '{{options.host}}/{{domain}}/v1/article/morelike/translation/{{seed_article}}'
+      x-monitor: false
 definitions:
   recommendation_result:
     type: object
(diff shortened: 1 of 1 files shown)


v1/recommend.yaml
```diff
@@ -56,6 +56,39 @@ paths:
             request:
               uri: '{{options.host}}/{domain}/v1/translation/articles/{from_lang}/{{default(seed_article,"")}}'
       x-monitor: false
+  /recommendation/article/morelike/translation/{seed_article}
```
Pchelolo: Transferring the question from Gerrit:

"One thing that could be interesting to think about - is whether we want to normalize the titles of the 'seed_article' or whether you want to follow on-wiki redirects here, because we do have special middleware for that purpose, we would just need to enable it." 

The title normalization should not be important here since it's mostly used for cache purging and we do not do any active purges. 

Would following redirects for redirect pages be useful?

Pchelolo:
Also, seems like it's not valid YAML

v1/recommend.yaml
Pchelolo: Missing a colon in the end.

kodchi: Yes, following redirects sounds like a good idea. As for normalizing titles, I need the title as it appears on Wikipedia. For example, I need 'Mount Everest' rather than 'Mount_Everest'.

Pchelolo: Ok, for redirecting you'd need to enable the 'lib/access_check_filter' with `seed_article` as a title option name. I can follow up with this after we fix the PR

kodchi: OK, made the changes below.

kodchi:
BTW, the recommendation-api service still needs data in the database to work correctly. We should probably delay merging this until then. Unless this doesn't go out to production upon merge.

Pchelolo:
It does not go streight into production, but agreed - let's wait. Ping me here when all is ready on your end.

kodchi:
OK, will do. Thanks!

kodchi:
@Pchelolo the data has been imported. We can go ahead with this PR. Also, could you enable this on eswiki, fawiki, and uzwiki? Thanks!

Pchelolo:
@kodchi 

The tests are still failing - the YAML is inparseable.

Also, we don't have the capability to enable the APIs per individual wikis, so this will be visible everywhere your other endpoint is visible, which is pretty much everywhere except wikidata, wiktionary and wikimedia.org. Do you want both endpoints to be 'confined' to a smaller set? We can do by-project.

v1/recommend.yaml
```diff
@@ -56,6 +56,43 @@ paths:
             request:
               uri: '{{options.host}}/{domain}/v1/translation/articles/{from_lang}/{{default(seed_article,"")}}'
       x-monitor: false
+  /recommendation/article/morelike/translation/{seed_article}:
+    x-route-filters:
+      - path: ./lib/access_check_filter.js
+        options:
+          title: '{{seed_article}}'
```
Pchelolo: Should be title: 'seed_article' , it's not a template.

Pchelolo:
One more thing inlined.

v1/recommend.yaml
```diff
@@ -56,6 +56,45 @@ paths:
...
+          description: the prioritized list of Wikidata IDs recommended for creation as Wikipedia articles
+          schema:
+            $ref: '#/definitions/morelike_result'
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/problem'
+      x-request-handler:
+        - get_from_backend:
+          request:
```
Pchelolo: Oh... Indentation!

Pchelolo:
There's an unrelated test failure, this however looks good to me.

PR #1088, opened 2018-12-13
Feed: Ensure `result[0].body.tfa` exists

d00rman:
There have been ~1200 failed requests in the last 24h where `result[0].body.tfa` was undefined, which made RB fail when trying to set the thumbnail. Since the Android app expects these fields to be present, ensure they are defined.


d00rman:
cc @Pchelolo @mdholloway @berndsi

PR #1090, opened 2018-12-20
PDF: Switch to using Proton

d00rman:
We are sunsetting Electron and switchng fully to Proton. Therefore, adjust the code to use Proton only if `new_uri` is specified (and consequently get rid of `new_probability` and `new_pdf`). However, for the sake of backwards compatibility, keep the code using Electron around.

Bug: [T210651](https://phabricator.wikimedia.org/T210651)

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -47,11 +47,7 @@ default_project: &default_project
         pdf:
           # Cache PDF for 5 minutes since it's not purged
           cache_control: s-maxage=600, max-age=600
-          uri: https://pdfrender-beta.wmflabs.org
           new_uri: https://proton-beta.wmflabs.org
-          new_probability: 1
-          secret: secret
-          scheme: https
         transform:
           cx_host: https://cxserver-beta.wmflabs.org
         skip_updates: false
--- a/v1/pdf.js
+++ b/v1/pdf.js
@@ -22,23 +22,14 @@ module.exports = options => ({
     operations: {
         generatePDF: (hyper, req) => {
             const rp = req.params;
-            const newProbability = options.new_probability || 0;
             return hyper.get(new URI([rp.domain, 'sys', 'page_revisions', 'page', rp.title]))
             .then((latestRevision) => {
-                if (options.new_uri && (Math.random() < newProbability || req.query.new_pdf)) {
-                    const newResult = hyper.get(new URI(
+                if (options.new_uri) {
+                    return hyper.get(new URI(
                         `${options.new_uri}/${rp.domain}/v1/`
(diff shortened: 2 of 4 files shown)


v1/pdf.js
```diff
@@ -22,23 +22,13 @@ module.exports = options => ({
...
-                    .catch((e) => {
-                        hyper.logger.log('error/proton', e);
-                        if (req.query.new_pdf) {
-                            throw e;
-                        }
-                    });
-                    if (req.query.new_pdf) {
-                        return newResult;
-                    }
+                    ));
```
Pchelolo: we need to do more - change the public API to allow for changing the hardcoded `/a4/desktop`. Optional parameters with defaults should not break backwards compatibility.

v1/pdf.js
```diff
@@ -22,23 +22,13 @@ module.exports = options => ({
...
             return hyper.get(new URI([rp.domain, 'sys', 'page_revisions', 'page', rp.title]))
             .then((latestRevision) => {
-                if (options.new_uri && (Math.random() < newProbability || req.query.new_pdf)) {
-                    const newResult = hyper.get(new URI(
+                if (options.new_uri) {
+                    return hyper.get(new URI(
                         `${options.new_uri}/${rp.domain}/v1/`
                         + `pdf/${encodeURIComponent(rp.title)}/a4/desktop`
-                    ))
-                    .catch((e) => {
```
Pchelolo: I'd still keep logging

config.example.wikimedia.yaml
```diff
@@ -47,11 +47,7 @@ default_project: &default_project
         pdf:
           # Cache PDF for 5 minutes since it's not purged
           cache_control: s-maxage=600, max-age=600
-          uri: https://pdfrender-beta.wmflabs.org
           new_uri: https://proton-beta.wmflabs.org
```
Pchelolo: maybe let's rename new_uri to uri?

v1/pdf.js
d00rman: Oh, damn, forgot that part. Very good point. Will change.

v1/pdf.js
d00rman: Won't throwing the error here make RB log it automatically? We don't want to end up with duplicate log entries.

config.example.wikimedia.yaml
d00rman: I was thinking about it, but to keep it backwards compatible, I opted with leaving it as-is. Perhaps we could go with this now and revisit in the future?

d00rman:
@Pchelolo updated the URI

d00rman:
Superseded by #1126

PR #1093, opened 2019-01-23
Removed mocha-eslint and eslint-config-node-services dependencies.

hknustwmf:


.eslintrc.yml
```diff
@@ -1 +1,31 @@
-extends: 'eslint-config-node-services'
\ No newline at end of file
+extends: 
+  - 'eslint-config-wikimedia/server'
```
jdforrester: This should work as just 'wikimedia/server' unless something's broken.

Pchelolo: +1

.eslintrc.yml
```diff
@@ -1 +1,31 @@
...
+rules:
+  array-bracket-spacing:
+    - off
+  camelcase:
+    - error
+    - properties: never
+  computed-property-spacing:
+    - off
+  indent: 
+    - off
```
Pchelolo: Actually, why is it off? We do want to validate indent, just want to override tabs to spaces.

package.json
```diff
@@ -47,18 +48,15 @@
     "semver": "latest"
   },
   "devDependencies": {
-    "ajv": "^6.5.4",
+    "ajv": "^6.7.0",
```
Pchelolo: yet again, let's not update unrelated dependencies.

hknustwmf: I updated only dev dependencies to bring them up to date. Did not deem that dangerous. I'll roll them back.

.eslintrc.yml
hknustwmf: https://github.com/wikimedia/change-propagation/pull/302#pullrequestreview-194255471

.eslintrc.yml
hknustwmf: No, it works with 'wikimedia/server', Going to update that.

maintenance/generate-wikimedia-domain-config.js
```diff
@@ -31,11 +27,11 @@ preq.get({
         '*': []
     };
 
-    Object.keys(sm).forEach(function(k) {
+    Object.keys(sm).forEach(function (k) {
```
d00rman: Maybe turn this and the ones below into `() => {}`

.eslintrc.yml
d00rman: This is different from [the one in `rb-m-t-spec`](https://github.com/wikimedia/restbase-mod-table-spec/pull/53/files#diff-0378d5111ab996d5882f4712e9350604R13). While both have the same effect, we should probably use the same values to minimise confusion.

berndsi:
Indentation looks good to me.

The only thing I would change is 
"operator-linebreak": [ "error", "after" ],
to 
"operator-linebreak": [ "error", "before" ],
since it's easier to see an operator at the beginning of the line than at the end, so one quickly sees that the line is a continuation.

d00rman:
> since it's easier to see an operator at the beginning of the line than at the end, so one quickly sees that the line is a continuation.

I tangentially agree, but we've gone with the former everywhere else, so let's keep it consistent.

PR #1096, opened 2019-02-20
Temp: Log VE requests

d00rman:
As part of the Parsoid storage simplification effort, we need to have
insight into which headers are actually sent by VE. This commit adds a
temporary logging facility to VE requests as a route filter for
/page/html and /transform end points.

Bug: [T215956](https://phabricator.wikimedia.org/T215956)

### DIFF
--- a/lib/ve_logger_filter.js
+++ b/lib/ve_logger_filter.js
@@ -1,0 +1,19 @@
+'use strict';
+
+module.exports = (hyper, req, next, options, specInfo) => {
+
+    options.probability = options.probability || 1;
+    const ua = req.headers && req.headers['user-agent'] || '';
+    const aua = req.headers && req.headers['api-user-agent'] || '';
+
+    if (`${ua}|${aua}`.includes('VisualEditor') &&
+            Math.floor(1000 * Math.random()) <= Math.floor(1000 * options.probability)) {
+        hyper.logger.log('warn/visualeditor', {
+            msg: 'VisualEditor request',
+            req_headers: JSON.stringify(req.headers, null, 2)
+        });
+    }
+
+    return next(hyper, req);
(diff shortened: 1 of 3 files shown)


lib/ve_logger_filter.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+module.exports = (hyper, req, next, options, specInfo) => {
+
+    options.probability = options.probability || 1;
+
+    if (req.headers && req.headers['user-agent'] &&
+            req.headers['user-agent'].includes('VisualEditor') &&
```
Pchelolo: Better check for `Api-User-Agent`, browser will override the user-agent

lib/ve_logger_filter.js
```diff
@@ -0,0 +1,18 @@
...
+            Math.floor(1000 * Math.random()) <= Math.floor(1000 * options.probability)) {
+        hyper.logger.log('warn/visualeditor', {
+            msg: 'VisualEditor request',
+            req_headers: req.headers
```
Pchelolo: Let's json-stringify them.. Could be too much cardinality for logstash to handle

v1/transform.yaml
```diff
@@ -93,6 +93,9 @@ paths:
             $ref: '#/definitions/problem'
 
       x-route-filters:
+        - path: ./lib/ve_logger_filter.js
+          options:
+            probability: 0.25
```
Pchelolo: let's add it to html/to/wikitext as well, it's used in stashing.

lib/ve_logger_filter.js
d00rman: VE sets both, but yeah, can check that one too

PR #1098, opened 2019-03-13
Replace `json-stable-stringify` with `fast-json-stable-stringify`

clarakosi:
fast-json-stable-stringify is being used because it does not have any dependencies and has been shown to be faster than json-stable-stringify

Bug: [T210426](https://phabricator.wikimedia.org/T210426)


Pchelolo:
LGTM, let's wait for Travis

PR #1100, opened 2019-03-21
Update v1/ yaml to be swagger 2.0 compliant

clarakosi:
Modified the optional parameters with a set of paths with required parameters
to make code swagger 2.0 complaint.

Also, deleted trending.yaml & .js since it has been deprecated.

Bug: T218218


Pchelolo:
The PR LGTM. 

As for the errors you seem to have submitted the spec built from the test config. Instead we need to validate the example config. See https://github.com/wikimedia/restbase/blob/master/test/features/router/buildTree.js#L1 - there we are building and getting the spec from the example config.

Let's try to rerun the validator with the proper spec and see what happens. Some of the error messages do not seem to be correct and might be errors in the validator. We might wanna use some other swagger validator implementation. There's a bunch, for example 'swagger-tools' npm package, but we want to find the one that supports both swagger 2 and swagger 3.

Pchelolo:
I've checked some tools that exist for swagger validation and I think the swagger-cli npm package provides the most meaningful set of errors. It is also capable of validating openapi 3.0, so I believe that's what we should use

Pchelolo:
I've tested myself and the package only gives you one error per run (sucks). Here's a pointer on how to fix the first error: 

https://github.com/wikimedia/hyperswitch/blob/master/lib/router.js#L513

clarakosi:
Ok. I'll look into it this morning

clarakosi:
The remaining errors are in regard to `patternProperties` which is not supported and `oneOf` which is part of 3.0.

Still need to: create unit-tests + create PR for Hyperswitch

Pchelolo:
> The remaining errors are in regard to patternProperties which is not supported and oneOf which is part of 3.0.

Both are used only in the section editing API which is being deprecated, so feel free to remove those. Otherwise LGTM.

test/features/schema_tests.js
```diff
@@ -101,3 +102,30 @@ parallel('Responses should conform to the provided JSON schema of the response',
...
+                return preq.post({uri: `http://online.swagger.io/validator/debug`, body: swagger})
+                    .then((res) => {
+                        // if valid will return an empty object: {}
+                        if (Object.keys(res.body).length !== 0) {
+                            throw new Error(`Swagger specification is not valid: ${res.body}`)
+                        }
+                    })
+            })
+    });
+});
```
Pchelolo: This has to be rebased after #1101 is merged and the spec validity test has to be moved here https://github.com/wikimedia/restbase/pull/1101/files#diff-ff16a5ee31f1814844151cee8c98d8aaR38

test/features/schema_tests.js
```diff
@@ -3,6 +3,7 @@
 // mocha defines to avoid JSHint breakage
 /* global it, before */
 
+const ServiceRunner = require('service-runner');
```
Pchelolo: Why?

test/features/schema_tests.js
```diff
@@ -34,9 +35,16 @@ parallel('Responses should conform to the provided JSON schema of the response',
...
         return preq.get({ uri: `${server.config.baseURL()}/?spec` })
-        .then((res) => {
-            // TODO: validate the openapi spec!
-        });
+            .then((res) =>  {
+                return preq.post({uri: `http://online.swagger.io/validator/debug`,
+                    body: JSON.stringify(res.body)})
+                    .then((res) => {
+                        // if valid will return an empty object: {}
+                        if (Object.keys(res.body).length !== 0) {
```
Pchelolo: Event better: assert.deepEquals({}, res.body, 'Spec must have no validation errors')

test/features/schema_tests.js
```diff
@@ -34,9 +35,16 @@ parallel('Responses should conform to the provided JSON schema of the response',
 
     it('should expose valid OpenAPI spec', () => {
         return preq.get({ uri: `${server.config.baseURL()}/?spec` })
-        .then((res) => {
-            // TODO: validate the openapi spec!
-        });
+            .then((res) =>  {
+                return preq.post({uri: `http://online.swagger.io/validator/debug`,
```
Pchelolo: I think it's better to install a dev dependency to swagger-cli and use it to validate. Lett dependencies on internet and external service.

clarakosi: One of my concerns with swagger-cli is that it uses swagger-parser which is using a deprecated dependency: https://github.com/APIDevTools/swagger-parser/pull/109

test/features/schema_tests.js
clarakosi: From the previous version. Will remove.

test/features/schema_tests.js
Pchelolo: hm. ok. Let's use the online one for now then and file a ticket to keep an eye on the swagger-cli thing.

Pchelolo:
Waiting for travis and merging.

v1/graphoid.yaml
```diff
@@ -34,7 +34,7 @@ paths:
         '200':
           description: The PNG render of the requested graph.
           schema:
-            type: blob
+            type: string
```
d00rman: PNGs are binary blobs, not strings, so we need `type: file` here (mental note: we'll need to change this to be v3-compliant)

Pchelolo:
Once @d00rman comment is fixed, let's not wait for travis. I've run tests locally to and it's all good. Travis is on vacation.

v1/graphoid.yaml
clarakosi: Done

PR #1102, opened 2019-03-28
Provide test drone configuration

Pchelolo:
WIP


Pchelolo:
Not satisfactory. Seems like the build timeout is 5 min that's not enough

PR #1103, opened 2019-04-01
Add storage RB config.

Pchelolo:
Bug: [T220855](https://phabricator.wikimedia.org/T220855)

### DIFF
--- a/config.example.storage.wikimedia.yaml
+++ b/config.example.storage.wikimedia.yaml
@@ -1,7 +1,4 @@
 # RESTBase wikimedia example config
-
-# Load some project templates. These are referenced / shared between domains
-# in the root_spec further down.
 default_project: &default_project
   x-modules:
     - spec:
@@ -10,32 +7,20 @@ default_project: &default_project
             x-modules:
               - spec:
                   paths:
-                    /parsoid_bucket:
-                      x-modules:
-                        - path: sys/multi_content_bucket.js
-                          options:
-                            grace_ttl: 1
-                            index_ttl: 864000
-                            renew_expiring: true
-                            delete_probability: 1
-                            table_name_prefix: parsoid_ng
-                            main_content_type:
-                              name: html
-                              value_type: blob
-                            dependent_content_types:
-                              - name: data-parsoid
-                                value_type: json
                     /key_value:
                       x-modules:
                         - path: sys/key_value.js
-                    /key_rev_value:
+                    /page_revisions:
+                      # We need to forward cookie to the API
+                      x-route-filters:
+                        - path: ./lib/mediawiki_auth_filter.js
                       x-modules:
-                        - path: sys/key_rev_value.js
+                        - path: sys/page_revisions.js
           /{api:sys}:
             x-modules:
               - spec:
                   paths:
-                    /table: &sys_table
+                    /table:
                       x-modules:
                         - path: sys/table.js
                           options:
@@ -47,32 +32,30 @@ default_project: &default_project
                               password: cassandra
                               defaultConsistency: one # or 'localQuorum' for production
                               storage_groups:
-                                - name: default.group.local
+                                - name: test.group.local
                                   domains: /./
                               dbname: test.db.sqlite3 # ignored in cassandra, but useful in SQLite testing
+                    /action:
+                      x-modules:
+                        - path: sys/action.js
+                          options:
+                            apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
+                            baseUriTemplate: "{{'https://{domain}/api/rest_v1'}}"
 
 
 # The root of the spec tree. Domains tend to share specs by referencing them
 # using YAML references.
 spec_root: &spec_root
   title: "The RESTBase Storage Service"
-  x-request-filters:
-    - path: lib/security_response_header_filter.js
+  x-sub-request-filters:
+    - type: default
+      name: http
+      options:
+        allow:
+          - pattern: /^https?:\/\/[a-zA-Z0-9\.]+\/w\/api\.php/
+            forward_headers: true
   paths:
     /{domain}: *default_project
-    # A robots.txt to make sure that the content isn't indexed.
-    /robots.txt:
-      get:
-        x-request-handler:
-          - static:
-              return:
-                status: 200
-                headers:
-                  content-type: text/plain
-                body: |
-                  User-agent: *
-                  Allow: /*/v1/?doc
-                  Disallow: /
 
 # Finally, a standard service-runner config.
 info:
@@ -102,12 +85,5 @@ logging:
 #  host: localhost
 #  port: 8125
 #  batch: true
-
-ratelimiter:
-  type: kademlia
-  # Cluster nodes
-  seeds:
-    # Port 3050 used by default
-    - 127.0.0.1
 
 num_workers: 0
--- a/config.example.storage.wikimedia.yaml
+++ b/config.example.storage.wikimedia.yaml
@@ -6,6 +6,31 @@ default_project: &default_project
   x-modules:
     - spec:
         paths:
+          /{api:v1}:
+            x-modules:
+              - spec:
+                  paths:
+                    /parsoid_bucket:
+                      x-modules:
+                        - path: sys/multi_content_bucket.js
+                          options:
+                            grace_ttl: 1
+                            index_ttl: 864000
+                            renew_expiring: true
+                            delete_probability: 1
+                            table_name_prefix: parsoid_ng
+                            main_content_type:
+                              name: html
+                              value_type: blob
+                            dependent_content_types:
+                              - name: data-parsoid
+                                value_type: json
+                    /key_value:
+                      x-modules:
+                        - path: sys/key_value.js
+                    /key_rev_value:
+                      x-modules:
+                        - path: sys/key_rev_value.js
           /{api:sys}:
             x-modules:
               - spec:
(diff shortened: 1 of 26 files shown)


d00rman:
Very early WIP, but yeah, this is the way to go. One thing we'll have to definitely change is using the `/sys` hierarchy. Since we will still need it in the front-end service, here we will probably just expose all of this via `/v1`. That is safe to do since the back-end service will not be exposed to the public. Oh and we need to figure out the names for these two services.

Pchelolo:
Sure sure, it's not ready for anything in any way, I've just started playing with this and decided to publish my playground early to have a venue for occasional bikesheds.

I actually like the idea of converting it to `/v1` - that would mean we wouldn't need to unrestrict the `/sys` that adds a layer of protection for the public RB instance.

Pchelolo:
heh.. it's tricky... the paths to the buckes are used internally under `/sys`...

d00rman:
> heh.. it's tricky... the paths to the buckes are used internally under `/sys`...

We could mount a simple proxy module in the config to redirect the requests to the back-end, something like:

```yaml
/{api:sys}:
  x-modules:
    - spec:
        paths:
          /page_revisions:
            x-modules:
              - path: sys/page_revisions.js
          /post_data: 
            x-modules:
              - path: sys/post_data.js
          /action:
            x-modules:
              - path: sys/action.js
                options: "{{options.action}}"
          /page_save:
            x-modules:
              - path: sys/page_save.js
          /parsoid:
            x-modules:
              - path: sys/parsoid.js
                options: &parsoid_options
                  parsoidHost: '{{options.parsoid.host}}'
                  grace_ttl: '{{default(options.parsoid.grace_ttl, 86400)}}'
          /{rest}:
            x-modules:
              - path: sys/proxy.js
                options:
                  backend_host: 127.0.0.1:7234
```

where `sys/proxy.js` would then send the request to the back-end service.

Pchelolo:
Basically, in the new version, I've switched summary to work in multi-instance configuration and switched one test suite to pass in the multi-configuration settings. (still need to push a bit of an update to hyperswitch), but I think this is basically a demonstration of how this whole thing will work.

The servers we start up for different CIs will be controlled by env variables.

Pchelolo:
So please have a look before I start working on it more

config.example.storage.wikimedia.yaml
```diff
@@ -0,0 +1,113 @@
...
+                            renew_expiring: true
+                            delete_probability: 1
+                            table_name_prefix: parsoid_ng
+                            main_content_type:
+                              name: html
+                              value_type: blob
+                            dependent_content_types:
+                              - name: data-parsoid
+                                value_type: json
+                    /key_value:
```
d00rman: I think we'll need here only `key_value` and `page_revisions` once the simplification is completed.

config.example.storage.wikimedia.yaml
```diff
@@ -0,0 +1,113 @@
...
+                  User-agent: *
+                  Allow: /*/v1/?doc
+                  Disallow: /
+
+# Finally, a standard service-runner config.
+info:
+  name: restbase
+
+services:
+  - name: restbase
```
d00rman: We probably want to chnge the name(s)

config.example.storage.wikimedia.yaml
```diff
@@ -0,0 +1,113 @@
...
+  streams:
+    - type: stdout
+
+#metrics:
+#  type: statsd
+#  host: localhost
+#  port: 8125
+#  batch: true
+
+ratelimiter:
```
d00rman: Would we need rate limiting for storage given that it's not going to be exposed?

config.example.storage.wikimedia.yaml
Pchelolo: Yes. But it's not complete yet, so I needed this to make it work. It will go away.

config.example.storage.wikimedia.yaml
Pchelolo: to something we didn't decide on yet.

config.example.storage.wikimedia.yaml
Pchelolo: indeed we don't

config.example.storage.wikimedia.yaml
```diff
@@ -0,0 +1,112 @@
...
+    - type: default
+      name: http
+      options:
+        allow:
+          - pattern: /^https?:\/\/[a-zA-Z0-9\.]+\/w\/api\.php/
+            forward_headers: true
+  paths:
+    /{domain}: *default_project
+    # A robots.txt to make sure that the content isn't indexed.
+    /robots.txt:
```
d00rman: `robots.txt` will be matched by `/{domain}` above, so it needs to be listed before.

config.example.storage.wikimedia.yaml
d00rman: Still not needed :P

sys/backend_proxy.js
```diff
@@ -0,0 +1,45 @@
...
+                req.params.path = `${modName}/${req.params.path}`;
+                return hyper.request({
+                    method: req.method,
+                    uri: backendURITemplate.expand({ request: req }).uri,
+                    headers: req.headers,
+                    body: req.body,
+                    query: req.query
+                })
+                .then((res) => {
+                    // TODO: Leave a comment why.
```
d00rman: Yes, please do :)

sys/backend_proxy.js
```diff
@@ -0,0 +1,45 @@
...
+                '/{+path}': {
+                    all: {
+                        operationId: 'proxy'
+                    }
+                }
+            }
+        },
+        operations: {
+            proxy: (hyper, req) => {
+                // Add the proxied module name to the path.
```
d00rman: Could you add a slightly-longer desc and an example for easier comprehension, please?

test/utils/run_tests.sh
```diff
@@ -22,30 +27,47 @@ runTest ( ) {
         rm -f test.db.sqlite3
     fi
 
-    if [ "$2" = "test" ]; then
-        "${mocha}"
-    elif [ "$2" = "coverage" ]; then
-        "${nyc}" --reporter=lcov node_modules/.bin/_mocha
+    if [ "x$3" = "x" ] || [ "$3" = "single_process" ]; then
```
d00rman: For future compatibility with WMF CI runs, it would be good for both `$2` and `$3` to be able to be set via `TEST_TARGET` and `TEST_MODE` (and for env vars to be preferred over the positional args).

config.example.wikimedia.yaml
```diff
@@ -6,6 +6,8 @@ default_project: &default_project
   x-modules:
     - path: projects/wmf_default.yaml
       options: &default_options
+        proxy:
```
d00rman: It would be good to also have an example config just for RESTRouter, since this file now contains a lot of unused config options.

Pchelolo: Only the `table` configuration is unused now. Do we want to create a copy-paste of a 200 lines long file just to make the example?

test/utils/run_tests.sh
Pchelolo: I was thinking about this, and I think we need to just remove the test runner script. Fallback to SQLite is default now via the default value of the RB_TEST_BACKEND env variable in the config. Locally running all configurations is unlikely to happen a lot, in Travis we can achieve running all the configs via env variables matrix. I think I'll just document the env variables and remove the run_tests.sh. What do you think?

config.example.storage.wikimedia.yaml
Pchelolo: Actually, it needs to be removed entirely. This is not going to be exposed outside.

d00rman: Let's not do that. Let's minimise the changes we make outside the split.

config.example.storage.wikimedia.yaml
d00rman: Heh, we've settled on restbase, so no need to change :)

config.example.wikimedia.yaml
d00rman: But most of the config is put in there, so I think it'll be clearer that way

Pchelolo: I'm not quite sure I follow regarding 'most of the config'

test/utils/run_tests.sh
Pchelolo: Doing that.

config.example.storage.wikimedia.yaml
Pchelolo: mmm? This is a completely new config, so I don't get how not removing this would minimize any of the changes. It's a copy-paste artifact.

config.example.wikimedia.yaml
d00rman: Euh, I meant most of the config vars that, say, come from puppet are declared here.

config.example.storage.wikimedia.yaml
d00rman: No, it's not. It's the front-end component of the current RB. And the current status quo is that RESTBase-the-router exposes it. So let's keep it for now.

test/utils/run_tests.sh
d00rman: I actually liked your earlier idea about removing this altogether and control stuff only with ENV VARS.

config.example.storage.wikimedia.yaml
Pchelolo: This is not the frontend component :) This is all complicated, but this is the example of how the backend component will look like when we finish up with the transition :)

d00rman: /me facepalms. Dang you are right. And it says nicely in the file name lol. Ok, yes, let's remove it.

config.example.wikimedia.yaml
d00rman: Nm...

package.json
```diff
@@ -37,6 +37,7 @@
   "dependencies": {
     "bluebird": "^3.5.5",
     "cassandra-uuid": "^0.1.0",
+    "colors": "^1.3.3",
```
Pchelolo: Ups...

sys/parsoid.js
```diff
@@ -77,7 +77,7 @@ function _dependenciesUpdate(hyper, req, newContent = true) {
     const rp = req.params;
     return mwUtil.getSiteInfo(hyper, req)
     .then((siteInfo) => {
-        const baseUri = siteInfo.baseUri.replace(/^https?:/, '');
+        const baseUri = siteInfo.baseUri.toString().replace(/^https?:/, '');
```
Pchelolo: Not needed anymore.

d00rman:
Very well done! Some more things I think should be done further, though:

- the projects shouldn't contain `/sys/table` and friends. We will use the projects in RESTRouter, so they shouldn't be loaded at all. As we discussed, in the interim we should have two versions of the projects: one for the whole vertical and one just for the front-end
- for completeness, it would be good imho to also have an example config for the front-end

d00rman:
Superseded by #1148

PR #1106, opened 2019-04-03
Simplify mcs storage

Pchelolo:
Bug: https://phabricator.wikimedia.org/T215960

### DIFF
--- a/sys/key_value.js
+++ b/sys/key_value.js
@@ -4,7 +4,6 @@
  * Key-value bucket handler
  */
 
-const uuid = require('cassandra-uuid').TimeUuid;
 const mwUtil = require('../lib/mwUtil');
 const HyperSwitch = require('hyperswitch');
 const stringify = require('fast-json-stable-stringify');
@@ -19,15 +18,9 @@ function returnRevision(req) {
     return (dbResult) => {
         if (dbResult.body && dbResult.body.items && dbResult.body.items.length) {
             const row = dbResult.body.items[0];
-            let headers = {
-                etag: row.headers.etag || mwUtil.makeETag('0', row.tid)
-            };
-            if (row.headers) {
-                headers = Object.assign(headers, row.headers);
-            }
             return {
                 status: 200,
-                headers,
+                headers: row.headers,
                 body: row.value
             };
         } else {
@@ -56,7 +49,7 @@ class KVBucket {
     }
 
     makeSchema(opts) {
-        const schemaVersionMajor = 5;
+        const schemaVersionMajor = 6;
 
         return {
             // Combine option & bucket version into a monotonically increasing
@@ -76,7 +69,6 @@ class KVBucket {
             },
             attributes: {
                 key: opts.keyType || 'string',
-                tid: 'timeuuid',
                 headers: 'json',
                 value: opts.valueType || 'blob'
             },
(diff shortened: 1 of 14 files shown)


sys/key_value.js
```diff
@@ -191,14 +167,15 @@ class KVBucket {
                         (!req.headers['content-type'] ||
                         req.headers['content-type'] === oldContent.headers['content-type'])) {
                     hyper.metrics.increment(`sys_kv_${req.params.bucket}.unchanged_rev_render`);
-                    return {
+                    throw new HTTPError({
```
d00rman: Why are we throwing it now? Isn't returning the response enough?

Pchelolo:
I believe I should've marked it as a WIP, it's yet again my playground.

> Why is tid management been completely removed from the key value bucket? How will its removal influence other parts that use key_value under the hood?

Storing a `TID` in a separate row makes no sense. What do we gain from it? We already have the TID in the etag.

I've refactored the key_value bucket quite substantially - we no longer assign a default etag and rely on the user to provide it. I am thinking to make externally provided etag mandatory - the user of the bucket should know better what exactly to set as a TID. Assigning some default etag (say, content hash) will just make it more error-prone I think.

> This PR completely replaces the old mobile_bucket without providing a fall-back. It could be good to have it for a while, since there are approx 40 req/s hitting mobile-sections.

This PR is what we will have eventually. I am thinking that it will be better to change how we do things - merge the PR that represents an eventual state first, then make a PR with fallbacks on top, so that later we can simply revert the second PR when we feel that it's time to remove the fallback. I think it's easier.

sys/key_value.js
Pchelolo: Should be enough, throwing is more explicit though.

d00rman:
> I've refactored the key_value bucket quite substantially - we no longer assign a default etag and rely on the user to provide it. I am thinking to make externally provided etag mandatory - the user of the bucket should know better what exactly to set as a TID. Assigning some default etag (say, content hash) will just make it more error-prone I think.

Hm, I do think the ETag should be mandatory, but the storage service should probably provide it, if the client doesn't. We can simply stick it in the header, so, yes, I agree there's no need to have a special colum for it. Or does it? For future uses, I can see cases for using `if-modified-since` or `if-match`, and there not having to extract it from the headers would be beneficial.

> This PR is what we will have eventually. I am thinking that it will be better to change how we do things - merge the PR that represents an eventual state first, then make a PR with fallbacks on top, so that later we can simply revert the second PR when we feel that it's time to remove the fallback. I think it's easier.

Sounds like a good idea, +1

Pchelolo:
> Or does it? For future uses, I can see cases for using if-modified-since or if-match, and there not having to extract it from the headers would be beneficial.

Only if we make use of LWT which does not seem to be planned for now. However, I would agree that maybe storing `tid` for if-modified-since and etag for if-match in separate columns might be beneficial as it's future-proof and does not require a lot of additional work. Will amend the PR.

Pchelolo:
One question we need to decide is what to use a default ETag. Ideally, we want the clients of the key_value bucket to provide the ETag for us and log a warning if it's not set. The most sencible default is some hind of a hash of the serialized values, however, serialization and computing the hash is not cheap for large values.

Given that it's a fallback that will never be used, I think using a hash of the serialized value is a solid solution. What do you think @d00rman ?

d00rman:
> However, I would agree that maybe storing `tid` for if-modified-since and etag for if-match in separate columns might be beneficial as it's future-proof and does not require a lot of additional work.

yup, exactly, +1, thnx!

> Given that it's a fallback that will never be used, I think using a hash of the serialized value is a solid solution.

This will hold true as long as RESTBase is the only direct user of the key value bucket, but that might change once we open up the storage service to others in the environment. How about having the ETag and tid be the same as a fallback? While not ideal, it prevents possible congestions in the future all the while having a unique ETag per key/value.

Pchelolo:
replaced by #1117

Pchelolo:
oh. I pushed 'comment' instead of 'close and comment'. Shame on me.

PR #1108, opened 2019-04-03
/edits/per-page endpoint throttling limit is now 5 reqs per sec

nuria:
Please see why we need to reduce requests ratios: https://phabricator.wikimedia.org/T219910


d00rman:
LGTM, but what about other end points? This particular one was hit this time, but perhaps others will be in the future? Should we lower them generally?

nuria:
We might need to do that eventually but doesn't seem we need to quite yet.  Issues are on edit data (backed by druid, not cassandra) and they are more pronounced if queried per per article, also true fix is this one: https://phabricator.wikimedia.org/T189623

PR #1109, opened 2019-04-04
/edits/per-page endpoint throttling limit is now 5

nuria:
Please see https://phabricator.wikimedia.org/T219910 for reasons as to why throttling is reduced


jobar:
LGTM, Thanks Nuria :)

PR #1111, opened 2019-04-08
Remove optional parameters in sys/

clarakosi:
Removed optional parameters from sys/ and test/test_module.yaml.

Bug: T218218

### DIFF
--- a/sys/key_rev_value.yaml
+++ b/sys/key_rev_value.yaml
@@ -8,13 +8,18 @@ paths:
     put:
       operationId: createBucket
 
+  /{bucket}/{key}: &bucket_key
+    get:
+      operationId: getRevision
+    put:
+      operationId: putRevision
+
   /{bucket}/{key}/:
     get:
       operationId: listRevisions
 
-  /{bucket}/{key}{/revision}{/tid}:
-    get:
-      operationId: getRevision
-  /{bucket}/{key}/{revision}{/tid}:
-    put:
-      operationId: putRevision
+  /{bucket}/{key}/{revision}:
+    <<: *bucket_key
+
+  /{bucket}/{key}/{revision}/{tid}:
+    <<: *bucket_key
--- a/sys/parsoid.yaml
+++ b/sys/parsoid.yaml
@@ -7,13 +7,24 @@ info:
     API through the `action` module.
 paths:
 
-  /pagebundle/{title}{/revision}{/tid}:
+  /pagebundle/{title}: &pagebundle_title
     get:
       summary: Retrieve a JSON bundle containing html and data-parsoid
       operationId: getPageBundle
     post:
       summary: Retrieve a JSON bundle containing html and data-parsoid
       operationId: getPageBundle
+
+  /pagebundle/{title}/{revision}:
+    <<: *pagebundle_title
+
+  /pagebundle/{title}/{revision}/{tid}:
+    <<: *pagebundle_title
+
+  /wikitext/{title}: &wikitext_title
+    get:
+      summary: Retrieve the wikitext for a title and revision.
+      operationId: getWikitext
 
   /wikitext/{title}/:
     get:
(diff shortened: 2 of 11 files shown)


sys/parsoid.yaml
```diff
@@ -26,11 +32,19 @@ paths:
           type: string
           required: false
 
-  /wikitext/{title}{/revision}{/tid}:
+  /wikitext/{title}/{revision}: &wikitext_title_revision
```
d00rman: `/wikitext/{title}` is missing

sys/parsoid.yaml
```diff
@@ -58,36 +73,66 @@ paths:
           type: string
           required: false
 
-  /data-parsoid/{title}{/revision}{/tid}:
+  /data-parsoid/{title}/{revision}: &data-parsoid_title_revision
```
d00rman: `/data-parsoid/{title}` is missing.

sys/parsoid.yaml
```diff
@@ -58,36 +73,66 @@ paths:
...
+  /transform/html/to/html: &transform_html_to_html
     post:
       summary: Update HTML, while optionally passing in title & revision.
       operationId: transformHtmlToHtml
 
-  /transform/html/to/wikitext{/title}{/revision}:
+  /transform/html/to/html/{title}:
+    <<: *transform_html_to_html
+
+  /transform/html/to/html/{title}/revision:
```
d00rman: `s/revision/{revision}`

test/test_module.yaml
```diff
@@ -10,8 +10,8 @@ paths:
             title: Wikimedia testing APIs
             x-is-api-root: true
           paths:
-            /service/test/{title}{/revision}:
-              get:
+            /service/test/{title}:
+              get: &service_test_title
```
d00rman: Better to be consistent and have the ref be for `/service/test/{title}` instead of `get`

v1/content.yaml
```diff
@@ -126,6 +126,14 @@ paths:
           description: The revision id
           type: integer
           required: true
+      x-request-handler:
+        - get_from_backend:
+            request:
+              uri: /{domain}/sys/page_revisions/page/{title}/{revision}
```
d00rman: Why are you replicating this here? swagger-router should know to use `{revision}` only if it's present, so the URI can be the same for both cases.

(This question goes for all replications in the files below)

clarakosi: > swagger-router should know to use {revision} only if it's present, so the URI can be the same for both cases

Tested it a bit and it doesn't seem to always be the case. Reverted it back to the original using optional params since it doesn't affect the final swagger spec.

clarakosi: Made 2 other commits testing the link in question and they both failed some tests. There was a at least one that didn't earlier today. Will try to recreate tomorrow.

d00rman: Hmmm this is a bit weird, it shouldn't be failing at all with `{/revision}` (as swagger-router knows not to attach the slash and the rev param if it's not there).

clarakosi: Maybe some miscommunication here. It does not fail with `{/revision}` that's why I reverted it after your comment. It fails with `/{revision}`

d00rman: We actually need to get rid of the optional syntax completely if we want to be fully v3-compliant, so yes, copy/pasta and duplication are needed here. Thnx @clarakosi for pointing out the obvious.

sys/key_rev_value.yaml
```diff
@@ -12,9 +12,11 @@ paths:
     get:
       operationId: listRevisions
 
-  /{bucket}/{key}{/revision}{/tid}:
+  /{bucket}/{key}/{revision}: &bucket_key_revision
```
Pchelolo: Where did the 'key' with no revision go?

sys/parsoid.yaml
```diff
@@ -58,36 +81,64 @@ paths:
           type: string
           required: false
 
-  /data-parsoid/{title}{/revision}{/tid}:
-    get:
-      summary: Retrieve the data-parsoid JSON for title & revision.
-      operationId: getDataParsoid
+  /data-parsoid/{title}/{revision}:
+    <<: *data-parsoid_title
+
```
Pchelolo: I do not think this is actually used ever, same as the one with no 'revision', at least it shouldn't be. But let's work on that as a part of parsoid storage simplification, not this one.

Pchelolo:
One comment inlined, otherwise LGTM

sys/key_rev_value.yaml
clarakosi: Done.

Pchelolo:
Nice PR number! :)

PR #1112, opened 2019-04-09
[WIP] Modify key_value bucket.

Pchelolo:
Open questions:
- How to provide headers like `vary` that we want to store and return back by which are not valid HTTP request headers.
- Where to do serialization/deserialization. If we make clients give us proper `content-type`, we will automatically parse the incoming data stream and will need to re-serialize it. So should we only support `application/octet-stream` as a content type and provide the actual content-type in `x-rv-content-type` header? if yes - we should return the data with the same header. Alternatively, we can add a patch to hyperswitch to NOT deserialize content and always store is as a Buffer and use proper client-provided content type. Doing so will basically make us reimplement kask here, which we did not want to do and might be to much work for an intermediate solution.
- What to use a default etag? a hash of the content buffer seems like an appropriate solution.
- How to prefix headers that need to be stored? Do header prefixing (that seem to be standard solution for existing http storage services (S3 for example) or put the content into an envelope containing headers and body? given that parsoid content is already an envelope containg multiple content parts, that might be anavoidable.

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -17,12 +17,13 @@ const mwUtil = {};
 /**
  * Create an etag value of the form
  * "<revision>/<tid>/<optional_suffix>"
- * @param   {Integer}   rev         page revision number
- * @param   {string}    tid         page render UUID
+ * @param   {Integer}   [rev]       page revision number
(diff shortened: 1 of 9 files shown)


Pchelolo:
Superceeded by #1116

PR #1116, opened 2019-04-14
Prepare key_value bucket for RESTBase split

Pchelolo:
This is based on #1115, needs to be rebased before review.

- Removed magic about etag handling
- Only store headers prefixed with `x-store-`
- Use new kv semantics everywhere
- removed some outdated tests


Pchelolo:
rebased after the merge of the base PR.

sys/key_value.js
```diff
@@ -102,58 +72,53 @@ class KVBucket {
...
-                    etag: req.headers && req.headers.etag || mwUtil.makeETag('0', tid)
-                }
-            };
+            return { status: 202 };
         }
 
+        const rp = req.params;
+        const tid = uuid.now().toString();
+
+        const headersToStore = {};
```
d00rman: `content-type` should be added straight away if present and then have it be overwritten if it's also present as `x-store-content-type`

sys/mathoid.js
```diff
@@ -114,10 +119,16 @@ class MathoidService {
                 }));
             }
             // construct the request object that will be emitted
+            Object.assign(completeBody[format].headers, {
+                'x-store-x-resource-location': hash
```
d00rman: `s/x-store-//` since `x-store-` is prepended later

test/features/specification/monitoring.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const parallel = require('mocha.parallel');
+const parallel = describe;//require('mocha.parallel');
```
d00rman: Euh?

Pchelolo:
> The only thing I'm wondering about it this direct change of key/val API that is visible to all of the modules using it. IIRC, we were discussing having a proxy module in the front-end RB instance.

I have started off with a proxy but then I thought - the endgame is to have no proxy here, and when we switch to using key_value over HTTP, we will want to switch one-by-one, thus either have 2 proxies or get the modules to look exactly as in this PR. Plus, there was some etag magic going on in the old key_value which I've removed, so I've decided proxy was no bueno.

> Hm, but how are we going to know which headers do need to be stored? Perhaps we should think about this a bit more.

Exactly. We would need to change the modules to provide only the headers they need to store. thus - change the modules.

test/features/specification/monitoring.js
Pchelolo: heh.

sys/mathoid.js
Pchelolo: indeed.

sys/key_value.js
Pchelolo: I actually am on the edge here. Why are we making an exception with the content-type again? I honestly don't quite see any benefit in having an exception for it, only confusion.

sys/mathoid.js
```diff
@@ -83,7 +84,11 @@ class MathoidService {
                 return P.join(
                     hyper.put({
                         uri: new URI([rp.domain, 'sys', 'key_value', 'mathoid_ng.check', hash]),
-                        headers: checkRes.headers,
+                        headers: {
```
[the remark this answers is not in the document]
d00rman: Also, since the headers are explicitly set on L79 and then returned, to minimise copy/pasta, let's convert them to `x-store` here in place rather than explicitly list them.

v1/definition.yaml
```diff
@@ -109,23 +109,25 @@ paths:
                 content-type: '{{storage.headers.content-type}}'
                 etag: '{{storage.headers.etag}}'
                 cache-control: '{{options.response_cache_control}}'
+                content-language: '{{storage.headers.content-language}}'
+                vary: '{{storage.headers.vary}}'
```
d00rman: Don't we want to store the headers too? L98 doesn't set them for storage.

Pchelolo: L98 is the GET request to the storage.

sys/mathoid.js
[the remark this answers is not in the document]
Pchelolo: done

Pchelolo:
ok. ok. Made it default to plain `content-type`.

d00rman:
LGTM, waiting for Travis...

PR #1121, opened 2019-04-17
Switch stashing to key_value with fallback

Pchelolo:
Step 1 of the process - only switches the stashing.

Bug: [T215956](https://phabricator.wikimedia.org/T215956)

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -201,7 +201,7 @@ class ParsoidService {
         };
     }
 
-    getStashBucketURI(rp, format, tid) {
+    getOldStashBucketURI(rp, format, tid) {
         const path = [rp.domain, 'sys', 'key_rev_value', `parsoid.stash.${format}-ng`, rp.title];
         if (rp.revision) {
             path.push(rp.revision);
@@ -210,6 +210,12 @@ class ParsoidService {
             }
         }
         return new URI(path);
+    }
+
+    getStashBucketURI(domain, title, revision, tid) {
+        return new URI([
+            domain, 'sys', 'key_value', 'parsoid-stash', `${title}:${revision}:${tid}`
+        ]);
     }
 
     getNGBucketURI(rp, format, tid) {
@@ -224,7 +230,7 @@ class ParsoidService {
     }
 
     getFallbackBucketURI(rp, format, tid) {
-        return this.getStashBucketURI(rp, format, tid);
+        return this.getOldStashBucketURI(rp, format, tid);
     }
 
     _getContentWithFallback(hyper, rp, format, tid) {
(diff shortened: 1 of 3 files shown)


sys/parsoid.js
```diff
@@ -224,7 +230,7 @@ class ParsoidService {
     }
 
     getFallbackBucketURI(rp, format, tid) {
-        return this.getStashBucketURI(rp, format, tid);
+        return this.getOldStashBucketURI(rp, format, tid);
```
d00rman: I think this is a relic from the transition to the complicated storage semantics. AFAICT, we don't need it any more, right?

sys/parsoid.js
```diff
@@ -281,26 +287,6 @@ class ParsoidService {
     generateAndSave(hyper, req, format, currentContentRes) {
         // Try to generate HTML on the fly by calling Parsoid
         const rp = req.params;
-        // Helper for retrieving original content from storage & posting it to
```
d00rman: This was commented out during the last transition, but we forgot to start using it again. It helps Parsoid making a quicker conversion for the same revid, so we should probably keep this and activate it after the current transition.

Either way, this change does not belong in this PR

sys/parsoid.js
```diff
@@ -510,23 +496,33 @@ class ParsoidService {
...
-
-        return P.props({
-            html: getStash('html'),
-            'data-parsoid': getStash('data-parsoid'),
-            wikitext: getStash('wikitext')
+        return hyper.get({
+            uri: this.getStashBucketURI(rp.domain, rp.title, rp.revision, tid)
         })
         .then((res) => {
+            res = JSON.parse(res.body.toString('utf8'));
```
d00rman: Let's be consistent and use `mwUtil.decodeBody()`

sys/parsoid.js
```diff
@@ -630,29 +626,18 @@ class ParsoidService {
...
-                hyper.put({
-                    uri: this.getStashBucketURI(rp, 'data-parsoid', tid),
-                    headers: original.body['data-parsoid'].headers,
-                    body: original.body['data-parsoid'].body
-                }),
-                hyper.put({
-                    uri: this.getStashBucketURI(rp, 'wikitext', tid),
+        return transformPromise.then((original) => hyper.put({
+            uri: this.getStashBucketURI(rp.domain, rp.title, rp.revision, tid),
+            body: Buffer.from(JSON.stringify({
```
d00rman: Also for the sake of consistency, perhaps we should create a `mwUtils.encodeBody()` function that encodes this, and use it in the mobile sections part too.

sys/parsoid.js
Pchelolo: Still need. this is for non-latest revision edits.

sys/parsoid.js
Pchelolo: We will not resurrect it ever afaik after talking to Subbu. I will remove it from this PR though, you're right.

sys/parsoid.js
Pchelolo: Ulala... let's not. Judging by the history of the whole 'decodeBody' magic will have to go. It's all about zipping up the content that we don't do anymore. Plus, after split this line will go away too.

sys/parsoid.js
Pchelolo: Idem

d00rman:
Alright! Let's try this stuff out :)

PR #1122, opened 2019-04-19
Update HTML content type profile to v2.1.0 in the spec

d00rman:
We have been serving v2.1.0 content for a while, but forgot apparently to update the spec to reflect that.

Bug: [T221432](https://phabricator.wikimedia.org/T221432)

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -189,7 +189,7 @@ paths:
           required: false
           type: string
       produces:
-        - text/html; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/HTML/1.7.0"
+        - text/html; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/HTML/2.1.0"
         - application/json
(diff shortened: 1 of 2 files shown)


Pchelolo:
this LGTM, but I'd rather not do it now. The ensure_content_type filter is enabled on these endpoints and given that we wanna transition parsoid storage to new model I'd rather not add extra complexity to it.

d00rman:
> this LGTM, but I'd rather not do it now. The ensure_content_type filter is enabled on these endpoints and given that we wanna transition parsoid storage to new model I'd rather not add extra complexity to it.

That's an excellent point. I agree, let's leave it for now and deploy it some other time - either in-between the stages or afterwards. In reality, though, all of the content that we do have should already be on version 2.1.0 so not much would change, but yes, let's minimise the surface of potential problems during the transition.

berndsi:
FWIW, MCS started requesting Parsoid 2.1.0 yesterday.

PR #1123, opened 2019-04-25
[Temp] Log all requests for Parsoid sections

d00rman:
We are sunsetting the section retrieval and editing portion of the REST API. The first step is to record the requests that do come in for sections.

Bug: [T216636](https://phabricator.wikimedia.org/T216636)

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -390,6 +390,7 @@ class ParsoidService {
     }
 
     getSections(hyper, req) {
+        hyper.logger.log('warn/parsoid/sections', 'getSections() called');
         const rp = req.params;
         const sections = req.query.sections.split(',').map((id) => id.trim());
         delete req.query.sections;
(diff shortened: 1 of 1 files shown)


berndsi:
@d00rman Where is this endpoint? I can't find it on https://en.wikipedia.org/api/rest_v1/#/Page%20content. I wonder if it might be useful in the future for section editing in the apps or in VE.

Pchelolo:
@berndsi requesting sections is done by a query parameter and the transform endpoint is under /transform

See conversation on the ticket regarding the usage of the endpoints.

d00rman:
> Where is this endpoint? I can't find it on https://en.wikipedia.org/api/rest_v1/#/Page%20content. I wonder if it might be useful in the future for section editing in the apps or in VE.

Sections can be retrieved through the `/page/html/{title}{/revision}{/tid}` end point by [supplying the `sections` query parameter](https://github.com/wikimedia/restbase/blob/2c804192c023cf601878e485e86af00a51559b43/v1/content.yaml#L164). However, it is important to note that here the definition of sections is different from what you might expect - sections here refer to paragraphs in the article. This feature was never really used so we want to remove it for now. That said, I agree that being able to retrieve/store sections can be very useful, but I think we'll need to come up with a way to truly provide sections (something a la `mobile-sections`).

PR #1125, opened 2019-04-26
Tests: Add CircleCI config

d00rman:
This is step one to adding CircleCI: we only run tests with SQLite. The next step is adding Cassandra.


d00rman:
Mathoid tests are failing, but they fail locally as well, so this is not related to CircleCI. I will investigate that failure separately.

d00rman:
Travis seems to be back to normal, closing this one

PR #1127, opened 2019-04-29
Changed Oracle JRE for OpenJDK JRE

hknustwmf:
to fix Travis builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ notifications:
 addons:
   apt:
     packages:
-    - oracle-java8-set-default
+    - openjdk-8-jre-headless
 
 env:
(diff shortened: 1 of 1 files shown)


d00rman:
For the Mathoid tests, it seems that there is some beta web proxy flakiness, as some tests succeed and some return 502, while they all pass for me locally. For Graphoid, it seems that the instance in beta is having problems.

d00rman:
We'll have to set up the services in beta again because of the move to k8s if we want to fix the failures here, cf. [T220235: Migrate Beta cluster services to use Kubernetes](https://phabricator.wikimedia.org/T220235)

Pchelolo:
Regardless of the failures of the services we test against, this PR fixes an issue for travis, thus merging.

PR #1128, opened 2019-04-30
Parsoid storage simplification, store old revision edits in new stash

Pchelolo:
Switching stash for editing older revisions and completely getting rid of using the old stashing buckets.

Technically, we don't need to stash all teh old revisions, only for the purpose of VE, but temporary it's ok.

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -219,7 +219,7 @@ class ParsoidService {
         ]);
     }
 
-    getNGBucketURI(rp, format, tid) {
+    getOldLatestBucketURI(rp, format, tid) {
         const path = [rp.domain, 'sys', 'parsoid_bucket', format, rp.title];
         if (rp.revision) {
             path.push(rp.revision);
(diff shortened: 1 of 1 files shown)


sys/parsoid.js
```diff
@@ -219,7 +219,13 @@ class ParsoidService {
         ]);
     }
 
-    getNGBucketURI(rp, format, tid) {
+    getLatestBucketURI(domain, title, revision, tid) {
+        return new URI([
+            domain, 'sys', 'key_value', 'parsoid', `${title}:${revision}:${tid}`
```
d00rman: If this is the latest bucket, why are we storing the title together with the revision and tid? it should just be the title

sys/parsoid.js
```diff
@@ -230,18 +236,28 @@ class ParsoidService {
...
+            uri: this.getOldLatestBucketURI(rp, format, tid)
+        })
+        .catch({ status: 404 }, (e) => {
+            if (rp.revision && tid) {
+                return hyper.get({
+                    uri: this.getStashBucketURI(rp.domain, rp.title, rp.revision, tid)
+                })
+                .then((res) => Object.assign(
+                    { status: 200 },
+                    JSON.parse(res.body.toString('utf8'))[format])
```
d00rman: It wouldn't be a bad idea to surround this by a try/catch perhaps?

sys/parsoid.js
```diff
@@ -274,15 +290,12 @@ class ParsoidService {
...
-            uri: this.getFallbackBucketURI(rp, 'data-parsoid', tid),
-            headers: parsoidResp.body['data-parsoid'].headers,
-            body: parsoidResp.body['data-parsoid'].body
-        })
-        .then(() => hyper.put({
-            uri: this.getFallbackBucketURI(rp, 'html', tid),
-            headers: parsoidResp.body.html.headers,
-            body: parsoidResp.body.html.body
-        }));
+            uri: this.getStashBucketURI(rp.domain, rp.title, rp.revision, tid),
```
d00rman: Let's add the headers here too

sys/parsoid.js
Pchelolo: This is wrong, you are right. THis is not even used in this PR. Shall remove.

sys/parsoid.js
Pchelolo: we are already inside a promise, what would we achieve with a try-catch?

sys/parsoid.js
Pchelolo: Indeed.

sys/parsoid.js
d00rman: A nice error msg. But this was just an observation, we can include it later

PR #1130, opened 2019-04-30
Parsoid simplification. Remove sections API

Pchelolo:
Bug: https://phabricator.wikimedia.org/T216636


Pchelolo:
We really need to fix beta mathoid and graphoid. CI not working is very annoying..

PR #1131, opened 2019-05-06
Parsoid simplification. Switch latest bucket and provide fallback.

Pchelolo:
TODO:
- More unit tests
- More documentation of what's going on
- A LOT more testing.
- Stere's quite a lot of code duplication, but that's a next step in the process.

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -125,16 +125,38 @@ class ParsoidService {
         };
     }
 
+    /**
+     * Get the URI of a bucket for the latest Parsoid content.
+     * @param {string} domain the domain name.
+     * @param {string} title the article title.
+     * @return {HyperSwitch.URI}
+     */
+    getLatestBucketURI(domain, title) {
+        return new URI([
+            domain, 'sys', 'key_value', 'parsoid', title
+        ]);
+    }
+
+    /**
+     * Get the URI of a bucket for stashing Parsoid content. Used both for stashing
+     * original HTML/Data-Parsoid for normal edits as well as for stashing transforms.
+     *
+     * @param {string} domain the domain name.
+     * @param {string} title the article title.
+     * @param {number} revision the revision of the article.
+     * @param {string} tid the TID of the content.
+     * @return {HyperSwitch.URI}
+     */
     getStashBucketURI(domain, title, revision, tid) {
         return new URI([
             domain, 'sys', 'key_value', 'parsoid-stash', `${title}:${revision}:${tid}`
         ]);
     }
 
-    getOldLatestBucketURI(rp, format, tid) {
-        const path = [rp.domain, 'sys', 'parsoid_bucket', format, rp.title];
-        if (rp.revision) {
-            path.push(rp.revision);
+    getOldLatestBucketURI(domain, format, title, revision, tid) {
+        const path = [domain, 'sys', 'parsoid_bucket', format, title];
+        if (revision) {
+            path.push(revision);
             if (tid) {
                 path.push(tid);
             }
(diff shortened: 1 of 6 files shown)


d00rman:
Needs a rebase first

Pchelolo:
Rebased. Twice.

test/features/mathoid.js
```diff
@@ -4,7 +4,7 @@ const assert = require('../utils/assert.js');
 const Server = require('../utils/server.js');
 const preq   = require('preq');
 
-describe('Mathoid', function() {
+describe.skip('Mathoid', function() {
```
d00rman: No need for this anymore, yay :)

sys/parsoid.js
```diff
@@ -125,109 +125,267 @@ class ParsoidService {
...
         const dataParsoidResponse = parsoidResp.body['data-parsoid'];
         const htmlResponse = parsoidResp.body.html;
-        return hyper.put({
-            uri: this.getStashBucketURI(rp.domain, rp.title, rp.revision, tid),
+        return hyper.get({ uri: this.getLatestBucketURI(domain, title) })
+        .then((existingRes) => {
+            const existingRev = mwUtil.parseETag(existingRes.headers.etag).rev;
+            const newRev = mwUtil.parseETag(parsoidResp.headers.etag).rev;
+            if (Number.parseInt(newRev, 10) >= Number.parseInt(existingRev, 10)) {
+                throw new HTTPError({ status: 404 });
```
d00rman: Using 404 is confusing to me. Perhaps 412 or similar would be better? Or a custom code, even better.

Also, this compares revisions. If they are the same, the code seems to assume that the content provided by Parsoid is newer than the one in storage, which in 99.9% of the cases will be true, but due to the async nature of updates and multiple DCs, there will be corner cases where the update will not happen. I think for this time around that is ok, but let's at least leave a comment to that effect.

sys/parsoid.js
```diff
@@ -125,109 +125,267 @@ class ParsoidService {
...
+                const resEtag = mwUtil.parseETag(res.headers.etag);
+                if (revision !== resEtag.rev) {
+                    throw new HTTPError({ status: 404 });
+                }
+                res.body = JSON.parse(res.body.toString('utf8'));
+                return res;
+            })
+            .catch({ status: 404 }, grabContentFromOldLatestBucket);
+        } else {
+            return hyper.get({ uri: this.getLatestBucketURI(domain, title) })
```
d00rman: Shouldn't the logic here be: since we have both the rev and tid, check the stash bucket first, then the latest and then finally the old parsoid_bucket?

sys/parsoid.js
```diff
@@ -236,72 +394,47 @@ class ParsoidService {
             const pageBundleUri = new URI([rp.domain, 'sys', 'parsoid', 'pagebundle',
```
d00rman: Perhaps this could be simplified so that the pagebundle method is called directly instead of going through the router?

sys/parsoid.js
```diff
@@ -236,72 +394,47 @@ class ParsoidService {
...
-                        newContent = false;
-                        // TODO: This will only need to be happening
-                        //  if we're requested by VE with a special flag
-                        return this.saveParsoidResultToFallback(hyper, req, tid, res);
-                    })
-                    .then(() => {
+                    let newContent = false;
+                    return this.saveParsoidResultToLatest(hyper, rp.domain, rp.title, res)
+                    .then((saveRes) => {
+                        if (saveRes.status === 201) {
```
d00rman: If it's not the latest, wouldn't the previous promise throw? I.e. how can we get here if the content is not the latest?

sys/parsoid.js
Pchelolo: There's a lot of simplification opportunities here but I don't want to touch anything not strictly required to be touched at this point.

sys/parsoid.js
Pchelolo: If it's not the latest, saveRes will be the previous latest and the status will be 200. If it was the latest and we saved - status will be 201.

PR #1140, opened 2019-05-23
Parsoid: Rate limit stashing requests

d00rman:
Stashing requests are expensive and take up storage, so make sure they are rate-limited. We check the rates only for external requests that explicitly request stashing. By default, at most 5 requests per second per client IP are allowed, but that can be adjusted by setting the `stash_ratelimit` in the module's configuration options.

Bug: [T224055](https://phabricator.wikimedia.org/T224055)

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -103,10 +103,7 @@ function compileReRenderBlacklist(blacklist) {
 
 class ParsoidService {
     constructor(options) {
-        this.options = options = options || {};
-        this.parsoidHost = options.parsoidHost;
-
-        this._blacklist = compileReRenderBlacklist(options.rerenderBlacklist);
+        this._initOpts(options);
 
         // Set up operations
         this.operations = {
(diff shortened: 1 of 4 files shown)


sys/parsoid.js
```diff
@@ -125,6 +122,51 @@ class ParsoidService {
         };
     }
 
+    _initOpts(opts) {
```
Pchelolo: Could use a default parameter

sys/parsoid.js
```diff
@@ -125,6 +122,51 @@ class ParsoidService {
...
+        if (!hyper.ratelimiter) {
+            return;
+        }
+        if (hyper._rootReq.headers['x-request-class'] !== 'external') {
+            return;
+        }
+        if (!((req.query && req.query.stash) || (req.body && req.body.stash))) {
+            return;
+        }
+        if (hyper._rootReq.headers['x-stash-rate-checked']) {
```
Pchelolo: What's the point of this? I don't think it would be called recursively and if it does, it's an error.

sys/parsoid.js
```diff
@@ -125,6 +122,51 @@ class ParsoidService {
...
+            return;
+        }
+        hyper._rootReq.headers['x-stash-rate-checked'] = true;
+        const key = `${hyper.config.service_name}.parsoid_stash.${req.params.domain}.` +
+            `${req.params.title || 'Transform'}.${req.params.revision || 0}|` +
```
Pchelolo: mmm... do we really wanna include title/revision in a key? This doesn't really limit anything, a malicious client can just iterate over all revisions of all pages

Pchelolo:
2 comments inlined. Plus need to update the docs.

sys/parsoid.js
d00rman: Good point, will remove.

sys/parsoid.js
d00rman: There are cases where parsoid calls itself via the URI, such as https://github.com/wikimedia/restbase/blob/de620a7f897a1f1e0840b009bb34293d11984b31/sys/parsoid.js#L577 (there are others too), so this ensures the rate is checked only once.

sys/parsoid.js
d00rman: Will do

sys/parsoid.js
Pchelolo: Yes and no. Actually that particular line you have shown is not calling itself, it's actually constructing an artificial request for code sharing. I wanna refactor it out. The only other case is calling the `pagebundle` endpoint, which I wanna get rid of too. So this is not needed.

d00rman: Ah you are correct. Ok, removing.

d00rman:
Ok, @Pchelolo addressed all the comments and added the rate limits to the spec docs.

Pchelolo:
Waiting gor travis

PR #1141, opened 2019-05-27
Build: Add the Blubber config file

d00rman:
This depends on the work being done in PR #1148.

Bug: [T223953](https://phabricator.wikimedia.org/T223953)

### DIFF
--- a/.pipeline/blubber.yaml
+++ b/.pipeline/blubber.yaml
@@ -1,0 +1,29 @@
+version: v3
+base: docker-registry.wikimedia.org/nodejs10-slim
+lives:
+  in: /srv/service
+runs:
+  environment: { APP_BASE_PATH: /srv/service }
+
+variants:
+  build:
+    base: docker-registry.wikimedia.org/nodejs10-devel
+    apt: { packages: [git, build-essential, python-pkgconfig] }
+    node: { requirements: [package.json] }
+    runs: { environment: { LINK: g++ } }
+  development:
+    includes: [build]
+    apt: { packages: [libsqlite3-0, ca-certificates] }
+    entrypoint: [node, server.js]
+  test:
+    includes: [build]
+    apt: { packages: [libsqlite3-0, ca-certificates] }
+    entrypoint: [npm, test]
+    runs: { environment: { TEST_TARGET: sqlite, TEST_MODE: fefs } }
+  prep:
+    includes: [build]
+    node: { env: production }
+  production:
+    copies: prep
+    node: { env: production }
+    entrypoint: [node, server.js]


Pchelolo:
I don't really know much about blubber, but looks good to me..

.pipeline/blubber.yaml
```diff
@@ -0,0 +1,29 @@
...
+base: docker-registry.wikimedia.org/nodejs-slim
+lives:
+  in: /srv/service
+runs:
+  environment: { APP_BASE_PATH: /srv/service }
+
+variants:
+  build:
+    base: docker-registry.wikimedia.org/nodejs-devel
+    apt: { packages: [git, python-pkgconfig, build-essential] }
```
akosiaris: Are there any native nodejs extensions in restbase? Otherwise I think build-essential and python-pkgconfig are not required

.pipeline/blubber.yaml
```diff
@@ -0,0 +1,29 @@
...
+    apt: { packages: [libsqlite3-0] }
+    entrypoint: [npm, test]
+    runs: { environment: { TEST_TARGET: sqlite, TEST_MODE: multi_process } }
+  prep:
+    includes: [build]
+    node: { env: production }
+  production:
+    copies: prep
+    node: { env: production }
+    apt: { packages: [ca-certificates] }
```
akosiaris: Does restbase reach out to any TLS enabled (and outside of wikimedia) endpoints ? If not, this isn't required.

.pipeline/blubber.yaml
d00rman: Only for prod stuff, the `heapdump` and `gc-stats`, but I don't know if that's already accounted for or not?

.pipeline/blubber.yaml
d00rman: Outside prod no, but isn't it the plan to have all connections in prod TLS'ed as well?

akosiaris: It is, but this package installs well-known x509 certs to be trusted by the system, not our own puppet CA (which is/will be used for this).

.pipeline/blubber.yaml
akosiaris: Good point, I 'll have to look into it

.pipeline/blubber.yaml
d00rman: Duh, good point! Will remove.

.pipeline/blubber.yaml
d00rman: It seems we do need them after all.

akosiaris: gc-stats seems to not require it, but heapdump does. So let's leave it as is.

akosiaris:
+1 from me

d00rman:
@akosiaris I changed the image to `nodejs10-slim` without a specific version tag. Can you confirm that's the proper one to use?

akosiaris:
Yeah, that's fine. We tend to upgrade those images every now and push `latest` for them so that all dependent images get the changes automatically on the next build.

PR #1144, opened 2019-06-02
[WIP] Split the logic between front-end, back-end and full-stack configs

d00rman:
This PR builds on top of #1103 . It creates the distinction between three types of RESTBase configs:

- front-end: the logic needed by the future front-end service instance, comprising public API end points, the business logic needed by them and proxying to the back-end storage service;
- back-end: the logic used by the future back-end storage service instance; and
- full-stack: the current RESTBase config with the addition of the `key_value` and `page_revisions` modules exposed to the public (this configuration will be needed during the transition period, and used for secondary configs such as Beta and MW-Vagrant)

TODO:
- adapt the tests to test for all three variants (fe + be, fe + fs, fs)


d00rman:
The actual diff from #1103 can be seen [here](https://github.com/Pchelolo/restbase/compare/split..d00rman:path-split).

Pchelolo:
oh gosh...

projects/wmf_wikidata.yaml
```diff
@@ -84,41 +84,12 @@ paths:
...
-                    grace_ttl: '{{default(options.parsoid.grace_ttl, 86400)}}'
-            /events:
-              x-modules:
-                - path: sys/events.js
-                  options: '{{merge({"skip_updates": options.skip_updates}, options.events)}}'
-        options: '{{options}}'
+      - path: projects/sys/business.js
+        options: '{{merge(options, {"exclude": "mobileapps"})}}'
+      - path: projects/sys/backend.js
+        options: '{{merge(options, {"exclude": "action"})}}'
```
Pchelolo: ???

d00rman:
> oh gosh...
> 
> I haven't looked super-closely, but here's a couple of thoughts:
> 
> 1. I don't think that FS RESTBase needs to expose key_value and page_revisions via `/v1` - this adds a lot of extra complexity. Instead we need to make a config option for hyperswitch to allow external access to `/sys`. This will be much simpler to implement and we will be able to not have any distinction between `v1` portion of the projects between FE and FS RESBase.

I don't like your trick. I think it's bad policy. I prefer to have a clear separation and have not so many work-arounds in the core piece. Which extra complexity are you exactly referring to? Note that FS will be needed for other things like vagrant and beta, which would mean a permanent hyperswitch "config option".

> 2. I think for the sake of simplifying everything, we need to merge config.test.yaml and config.example.wikimedia.yaml together into a single test config. Neither of these resemble what we're running in production and both of them are actually very close with the test config exposing some additional things for unit testing.

This is a TODO of the PR. I was thinking of having three configs (one for each type), which then we can tweak with env variables for tests.

> 3. In general, I think you got too far in avoiding copy pasta. All these module loaders and exclusions and yadayada makes it too complicated to reason about. I was thinking about something a bit simple. Say, we have a projects tructure like this:
> 
> ```
> /v1/
>  - default.yaml
>  - wikipedia.yaml
>  - ...
> /sys/
>  - business_default.yaml
>  - business_wikipedia.yaml
>  - ...
> - backend.yaml
> ```
> 
> Then we'd be able to mount the correct combinations within the config. So, for FS restbase we will mount both the business and backend project in the config under `{api:sys}`, for fronted config we will just mount the business sys config. More boilerplate but much much less magic.

Without copying around projects for different purposes you won't be able to achieve that. And you were against copying projects in the first place, so pick your poison.

Pchelolo:
> I don't like your trick. I think it's bad policy. I prefer to have a clear separation and have not so many work-arounds in the core piece. Which extra complexity are you exactly referring to? Note that FS will be needed for other things like vagrant and beta, which would mean a permanent hyperswitch "config option".

FS will indeed be needed for vagrant and beta, but the `/v1/key_value` and `/v1/page_revision` will only be needed for a transition period and will not be needed afterward. As soon as we complete the transfer the config option can be removed.

> This is a TODO of the PR. I was thinking of having three configs (one for each type), which then we can tweak with env variables for tests.

+1 to this, it's certainly the way forward.

> Without copying around projects for different purposes you won't be able to achieve that. And you were against copying projects in the first place, so pick your poison.

You will (ish). In my approach the amount of YAML code stays roughly the same, but it's split into more files. So, take `wmf_default.yaml` project. The spec for `/v1` (and only the spec) becomes `\v1/wmf_default.yaml`. The spec for `/sys` and only the spec (without `page_revisions`, `table` and `key_value` - what you have in `sys-business` now) becomes `/sts/wmf_default.yaml`. Then in the frontend config you just mount both specs under `/{api:v1}` and `/{api:sys}`. For full-stack RB under `/{sys}` you mount the sys project and add specs for table, page_revisions and key_value. the storage config can not use any projects at all and just use the config. 

So we're copy-pasting only 3 modules declarations.

d00rman:
> > I don't like your trick. I think it's bad policy. I prefer to have a clear separation and have not so many work-arounds in the core piece. Which extra complexity are you exactly referring to? Note that FS will be needed for other things like vagrant and beta, which would mean a permanent hyperswitch "config option".
> 
> FS will indeed be needed for vagrant and beta, but the `/v1/key_value` and `/v1/page_revision` will only be needed for a transition period and will not be needed afterward. As soon as we complete the transfer the config option can be removed.

This would mean that the API as seen by the FE would change, as FS and BE would not expose the same API. I'm personally not willing to trade the symmetry.

> You will (ish). In my approach the amount of YAML code stays roughly the same, but it's split into more files. So, take `wmf_default.yaml` project. The spec for `/v1` (and only the spec) becomes `\v1/wmf_default.yaml`. The spec for `/sys` and only the spec (without `page_revisions`, `table` and `key_value` - what you have in `sys-business` now) becomes `/sts/wmf_default.yaml`. Then in the frontend config you just mount both specs under `/{api:v1}` and `/{api:sys}`. For full-stack RB under `/{sys}` you mount the sys project and add specs for table, page_revisions and key_value. the storage config can not use any projects at all and just use the config.
> 
> So we're copy-pasting only 3 modules declarations.

We used to have it all in the config and moved away from it (into projects) because in doing so we could better control the compatibility between module code and module declarations. I'm not sure it's a good idea to going back to decoupling them, honestly.

Pchelolo:
> This would mean that the API as seen by the FE would change, as FS and BE would not expose the same API. I'm personally not willing to trade the symmetry.

I'm not sure I understand the benefit of having the same API in the transition period. On the contrary, just hiding `v1/key_value` from the docs is still a huge security concern. We would need to protect it so that only internal clients could use it, any malicious user starting to create buckets like crazy would crash cassandra pretty quickly.

> We used to have it all in the config and moved away from it (into projects) because in doing so we could better control the compatibility between module code and module declarations. I'm not sure it's a good idea to going back to decoupling them, honestly.

AFAIK we have created the concept of projects because we didn't have +2 on puppet and the config used to live in puppet. Now that this is not the case anymore we do not have the same problem. Not sure that compatibility was ever a concern (at least not in my mind)

d00rman:
> > This would mean that the API as seen by the FE would change, as FS and BE would not expose the same API. I'm personally not willing to trade the symmetry.
> 
> I'm not sure I understand the benefit of having the same API in the transition period. On the contrary, just hiding `v1/key_value` from the docs is still a huge security concern. We would need to protect it so that only internal clients could use it, any malicious user starting to create buckets like crazy would crash cassandra pretty quickly.

This is already done in this PR.

As for API stability, with it we don't need to modify the frontend once it's deployed. Less moving parts == more confidence == less stressful.

> 
> > We used to have it all in the config and moved away from it (into projects) because in doing so we could better control the compatibility between module code and module declarations. I'm not sure it's a good idea to going back to decoupling them, honestly.
> 
> AFAIK we have created the concept of projects because we didn't have +2 on puppet and the config used to live in puppet. Now that this is not the case anymore we do not have the same problem. Not sure that compatibility was ever a concern (at least not in my mind)

Even though these are YAML files, they are not config in my mind. They are an essential part of the service

d00rman:
Superseded by #1148

PR #1145, opened 2019-06-02
[DISCUSSION] Enable bucket creation caching in key_value

d00rman:
**DO NOT MERGE**: I'm creating this PR to discuss caching inside the `key_value` bucket.

Currently, we issue `createBucket` requests for each bucket in each domain. This means that we uselessly issue ~800 requests per table on each RB start-up, making it a major contributor to the start-up delay. This delay will be exacerbated once we split RESTBase in two services (cf. #1103 and #1144) because all these requests will be issued by the front-end service and directed to the back-end service, hence adding network latency.

In this PR, I'm proposing to explore the idea of caching the responses and return immediately when a particular bucket is known to exist. This is tracked in the `key_value` bucket by using the `storage_groups` config supplied to `sys/table`. If a `createBucket` request has already been issued for a particular combination of storage group and bucket name, the previous result is returned. Otherwise, the request is executed and its resulting status is stored in module's cache.

Note that once RB is split in two, this logic would need to reside in both the front-end and back-end service (for performance reasons).The pro of such an approach include increased performance during start-up, both in terms of querying the data store as well in minimising the network latency needed to complete the start-up procedure (after the split). The (obvious) contra in doing so is that we would need to replicate the storage groups configuration. However, IMHO, this is ok to do in this particular instance since (i) the storage groups definition is fairly stable and not trivial to change; and (ii) the performance benefits far outweigh the _conceptual breakage_ that doing so incurs.

The alternative is to put this logic in the respective storage modules (Cassandra, SQLite). Alas, the only pro here is the fact that the storage layer **is** the canonical place for storage groups definitions. Putting the caching there would bring us some performance benefits, but we would not be able to get rid of futile network requests after the split.

### DIFF
--- a/projects/wikimedia.org.yaml
+++ b/projects/wikimedia.org.yaml
@@ -95,6 +95,8 @@ paths:
             /key_value: &sys_key_value
               x-modules:
                 - path: sys/key_value.js
+                  options:
+                    groups: '{{options.table.storage_groups}}'
             /post_data: &sys_post_data
               x-modules:
                 - path: sys/post_data.js
--- a/projects/wmf_default.yaml
+++ b/projects/wmf_default.yaml
@@ -115,6 +115,8 @@ paths:
             /key_value: &sys_key_value
               x-modules:
                 - path: sys/key_value.js
+                  options:
+                    groups: '{{options.table.storage_groups}}'
             /page_revisions:
               x-modules:
                 - path: sys/page_revisions.js
--- a/projects/wmf_enwiki.yaml
+++ b/projects/wmf_enwiki.yaml
@@ -142,6 +142,8 @@ paths:
             /key_value: &sys_key_value
               x-modules:
                 - path: sys/key_value.js
+                  options:
+                    groups: '{{options.table.storage_groups}}'
             /page_revisions:
               x-modules:
                 - path: sys/page_revisions.js
--- a/projects/wmf_wikidata.yaml
+++ b/projects/wmf_wikidata.yaml
@@ -97,6 +97,8 @@ paths:
             /key_value: &sys_key_value
               x-modules:
                 - path: sys/key_value.js
+                  options:
+                    groups: '{{options.table.storage_groups}}'
             /page_revisions:
               x-modules:
                 - path: sys/page_revisions.js
--- a/projects/wmf_wikipedia.yaml
+++ b/projects/wmf_wikipedia.yaml
@@ -142,6 +142,8 @@ paths:
             /key_value: &sys_key_value
               x-modules:
                 - path: sys/key_value.js
+                  options:
+                    groups: '{{options.table.storage_groups}}'
             /page_revisions:
               x-modules:
                 - path: sys/page_revisions.js
--- a/projects/wmf_wikivoyage.yaml
+++ b/projects/wmf_wikivoyage.yaml
@@ -137,6 +137,8 @@ paths:
             /key_value: &sys_key_value
               x-modules:
                 - path: sys/key_value.js
+                  options:
+                    groups: '{{options.table.storage_groups}}'
             /page_revisions:
               x-modules:
                 - path: sys/page_revisions.js
--- a/projects/wmf_wiktionary.yaml
+++ b/projects/wmf_wiktionary.yaml
@@ -131,6 +131,8 @@ paths:
             /key_value: &sys_key_value
               x-modules:
                 - path: sys/key_value.js
+                  options:
+                    groups: '{{options.table.storage_groups}}'
             /page_revisions:
               x-modules:
                 - path: sys/page_revisions.js
--- a/sys/key_value.js
+++ b/sys/key_value.js
@@ -4,6 +4,7 @@
  * Key-value bucket handler
  */
 
+const P = require('bluebird');
 const uuid = require('cassandra-uuid').TimeUuid;
 const mwUtil = require('../lib/mwUtil');
 const HyperSwitch = require('hyperswitch');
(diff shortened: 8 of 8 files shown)


Pchelolo:
Alternative solution: eliminate createBucket calls allthogether.

We don't wanna expose the possibility of dynamic bucket creation and we already skip it in production because dynamic schema migrations are BAD and don't work. Maybe this is the time to finally solve the problem of dynamic schema creations once and for all. 

This particular solution I don't like cause it adds a bunch of code trying to workaround an issue we were trying to solve for quite some time. 

@eevans thoughts?

d00rman:
I agree, but I don't think it's a good idea too add yet another change to the already-tricky split. I think we should defer removing the calls to a later stage.

d00rman:
Clarification: the code in production does not create new keyspaces, but it checks that the ones that are needed by the code are there. This aspect should be kept, and for that this is still a valid idea/discussion as I don't think we should get rid of the check procedure on start-up.

eevans:
This is good timing, I've been thinking about exactly this (handling Cassandra schema/config) as part of [T220246](https://phabricator.wikimedia.org/T220246).  Any followup thoughts on this matter there is appreciated!

> ...does not create new keyspaces, but it checks that the ones that are needed by the code are there...I don't think we should get rid of the check procedure on start-up.

Can I play Devil's Advocate and ask why?  If schema management is something that happens entirely external to the app, what advantage does this check provide?  I assume we'd either except on startup because the check failed, or we'd except because a needed table, attribute, etc did not exist.

d00rman:
> This is good timing, I've been thinking about exactly this (handling Cassandra schema/config) as part of [T220246](https://phabricator.wikimedia.org/T220246). Any followup thoughts on this matter there is appreciated!

While closely related (and conflated in the current RB implementation), schema management and schema checking are two distinct mechanisms in my view. Here, I'd like to focus on the checking part :)

> Can I play Devil's Advocate and ask why? If schema management is something that happens entirely external to the app, what advantage does this check provide? I assume we'd either except on startup because the check failed, or we'd except because a needed table, attribute, etc did not exist.

I agree that schema management is something that should be done externally and is part of operational procedures. However, checking that everything is in place on start-up is a different thing. Up until now, RB has been checking the schemae are in place with the intent of creating them if they were not. I'm proposing to slightly change the semantics to mean _check the schemae are in place to ensure the proper functioning of the service_. Failing on start-up as soon as a discrepancy is found is a very good mechanism for detecting problems as soon as possible (as opposed to deploying a faulty config or version only for it to start failing later on).

Pchelolo:
> Failing on start-up as soon as a discrepancy is found is a very good mechanism for detecting problems as soon as possible (as opposed to deploying a faulty config or version only for it to start failing later on).

I agree 100% that we MUST not leave it for lazy evaluation, especially given how our deployments are done - having a canary not evaluating the operationability of the service makes close to 0 sense. However, we might think about these checks as operational procedure as well and implement them externally to the service. Several options on how that could be done would be
- Making whatever tool used for schema management have a checking mode and make the check externally
- Expand the set of `x-amples` (or whatever is coming to replace them) monitoring tests to cover all the keyspaces and run them as a part of a deploy or a k8s readiness check

> I agree that schema management is something that should be done externally and is part of operational procedures.

Having schema management external to the service and schema checking internally to the service will require duplication of the schema definition: once in some operational config, once in the service code. This is BAD.

d00rman:
> I agree 100% that we MUST not leave it for lazy evaluation, especially given how our deployments are done - having a canary not evaluating the operationability of the service makes close to 0 sense. However, we might think about these checks as operational procedure as well and implement them externally to the service. Several options on how that could be done would be
> 
> * Making whatever tool used for schema management have a checking mode and make the check externally
> * Expand the set of `x-amples` (or whatever is coming to replace them) monitoring tests to cover all the keyspaces and run them as a part of a deploy or a k8s readiness check

Given that the schema implicitly **is** part of the code (see my point below), I don't see much benefit in creating an external tool for doing a thing we are already doing.

> Having schema management external to the service and schema checking internally to the service will require duplication of the schema definition: once in some operational config, once in the service code. This is BAD.

I think this statement is somewhat misleading. The schema is implicitly already present in the code: a certain module expects a certain table to exist, and it expects the returned data to have certain columns which are of a certain type - this is what the schema **is**. Checking explicitly that all of these expectations are met on start-up is good behaviour and it seems that we all agree on keeping this aspect. The rabbit-hole we are now in, imho, is about the _detail_ of the schema definition to be checked. In that regard, RB is rather picky right now and perhaps we could relax it, but that's orthogonal to this discussion.

To sum up, since we already implicitly have the schema in the code, I don't see why we wouldn't check for it explicitly on start-up.

eevans:
> > This is good timing, I've been thinking about exactly this (handling Cassandra schema/config) as part of [T220246](https://phabricator.wikimedia.org/T220246). Any followup thoughts on this matter there is appreciated!
> 
> While closely related (and conflated in the current RB implementation), schema management and schema checking are two distinct mechanisms in my view. Here, I'd like to focus on the checking part :)

I think it's relevant because the scope of that ticket is (implicitly):  _How do we do things in a world where it's no longer in-application_?  I'm making (have been making) assumptions that applications can safely assume that the database schema is correct, and it sounds like you're saying otherwise.  We should reconcile this.

> > Can I play Devil's Advocate and ask why? If schema management is something that happens entirely external to the app, what advantage does this check provide? I assume we'd either except on startup because the check failed, or we'd except because a needed table, attribute, etc did not exist.
> 
> I agree that schema management is something that should be done externally and is part of operational procedures. However, checking that everything is in place on start-up is a different thing. Up until now, RB has been checking the schemae are in place with the intent of creating them if they were not. I'm proposing to slightly change the semantics to mean _check the schemae are in place to ensure the proper functioning of the service_. Failing on start-up as soon as a discrepancy is found is a very good mechanism for detecting problems as soon as possible (as opposed to deploying a faulty config or version only for it to start failing later on).

I don't disagree that failing fast is an important property, I'm just trying to understand why it requires complete schema validation.  I can't say I've ever been motivated to consider this in the past (Kask does not do it FWIW), and I can't recall ever seeing an example of it in the wild (it's not an industry standard approach).  I'm not suggesting that invalidates the idea, just that more explanation is needed. :)

Also: I think there'd be less to talk about if the code was already there (a vestige of the old schema management), and cost nothing to leave in as an additional check, but the cost is...well more than nothing, certainly.

eevans:
>  * Making whatever tool used for schema management have a checking mode and make the check externally

This begs the same question in my mind: Why?

Schema changes aren't something that should be done easily, or frequently.  In fact, I'm at pains to remember the last time we did anything other than create a new table, and migrate to it.  There are tons of things we treat as a given (and assume that errors would trickle up otherwise), why validate this?

>  * Expand the set of `x-amples` (or whatever is coming to replace them) monitoring tests to cover all the keyspaces and run them as a part of a deploy or a k8s readiness check

Now this sounds like a fantastic idea no matter what.  It tests everything, schema included.  A positive outcome from a synthetic transaction like this is the ultimate in determining readiness.

d00rman:
> I think it's relevant because the scope of that ticket is (implicitly): _How do we do things in a world where it's no longer in-application_? I'm making (have been making) assumptions that applications can safely assume that the database schema is correct, and it sounds like you're saying otherwise. We should reconcile this.

My assumption in this PR/discussion is that for the time being (i.e. the split), we don't want to get rid of this check. There are multiple reasons for it, but the most prominent (for me) is that the smaller number of changes/moving pieces we have during the split the better. Given that assumption, I want to optimise the checks so that they have the least amount of impact on start-up performance. Whether the checks should stay are a different conversation for me (which I'll address below).

> I don't disagree that failing fast is an important property, I'm just trying to understand why it requires complete schema validation. I can't say I've ever been motivated to consider this in the past (Kask does not do it FWIW), and I can't recall ever seeing an example of it in the wild (it's not an industry standard approach). I'm not suggesting that invalidates the idea, just that more explanation is needed. :)

I agree we might want to loosen the checking semantics. I'm not sure to what, though? Just require the table is there? Require the correct fields are present? Something else? :P 

However, I do think we should keep **a** form of checks. Failing fast and loud ASAP has proven invaluable in the past, and while things have stabilised in RESTBase (hopefully?), I do think this will keep our sanity (to a certain level).

One thing I worry about removing checking/creation are auxiliary installs: third-party and vagrant/beta (and/or their future replacements). Currently, while this mechanism is not good for production use, it's very useful in such environments, including local testing (and Travis). If we get rid of all of this, our maintenance burden increases substantially, which is a price I'm not sure it's worth paying.

> 
> Also: I think there'd be less to talk about if the code was already there (a vestige of the old schema management), and cost nothing to leave in as an additional check, but the cost is...well more than nothing, certainly.

I'm not sure I understand this point. We **do** have a checking and creation mechanism in the code already, haven't we?

d00rman:
@eevans @Pchelolo this sounds like a great topic to chat about at our off-site next week :)

Pchelolo:
Over a BBQ where @d00rman can eat a side dish for free

d00rman:
@eevans @Pchelolo I forgot what we concluded over that famous BBQ :P Are we going ahead with a way to cache/minimise these calls between RESTRouter and RESTBase or not?

PR #1147, opened 2019-06-03
Morelike: update recommend endpoint

kodchi:
Add a new return field, and change the existing field to match the
source.

Bug: T219505

### DIFF
--- a/v1/recommend.yaml
+++ b/v1/recommend.yaml
@@ -154,7 +154,10 @@ components:
           wikidata_id:
             type: string
             description: Wikidata ID for the item
-          score:
+          normalized_rank:
             type: number
-            description: Score of the recommendation. The higher the score, the more
-              important the recommendation is.
+            description: Normalized rank of the recommendation. The
+              higher the rank, the more important the recommendation is.
+          source_language:
+            type: string
+            description: Source language of the article
--- a/test/features/schema_tests.js
+++ b/test/features/schema_tests.js
@@ -108,5 +108,38 @@ parallel('Responses should conform to the provided JSON schema of the response',
             }
         });
     });
+
+    it('/data/recommendation/article/creation/translation/{from_lang} should conform to schema', () => {
+        return preq.get({ uri: `${server.config.baseURL()}/data/recommendation/article/creation/translation/uz?count=5` })
+        .then((res) => {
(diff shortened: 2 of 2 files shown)


v1/recommend.yaml
```diff
@@ -154,7 +154,10 @@ components:
...
             description: Wikidata ID for the item
-          score:
+          normalized_rank:
             type: number
-            description: Score of the recommendation. The higher the score, the more
-              important the recommendation is.
+            description: Normalized rank of the recommendation. The
+              higher the rank, the more important the recommendation is.
+          source_language:
+            type:string
```
Pchelolo: You need a space after `:` or it's not valid yaml.

Pchelolo:
It would be neat to add a schema test for recommendation API. See https://github.com/wikimedia/restbase/blob/master/test/features/schema_tests.js for inspiration

berndsi:
Changes like this should probably bump the version number.

kodchi:
> Changes like this should probably bump the version number.

Do you suggest a major version bump (because it's kind of a breaking change)? Since the API is marked as alpha quality, I was holding off on bumping the version as it may shoot up fast until we reach beta or stable, aka Chrome versioning.

berndsi:
Well, I commented before I looked up the version number. What I meant was the version number exposed in the response content-type header. But it looks like it's not set for this yet. (Compare the string after `profile` in https://en.wikipedia.org/api/rest_v1/#/Mobile/get_page_mobile_sections__title_.)

I guess you can ignore this for now until you add version numbers to your responses.

kodchi:
Some of the tests are failing because of https://phabricator.wikimedia.org/T215222. Can we still merge the pull request while we work to fix the above bug?

Pchelolo:
> Can we still merge the pull request while we work to fix the above bug?

No, this would not be good. It will make RESTBase test suite fail for other patches and having unrelated test failures makes it much less convenient to work.

We can merge it as soon as you deploy the updated fixed version of the recommendation-api.

kodchi:
OK, that makes sense.

d00rman:
@kodchi needs a rebase at least.

d00rman:
Closing due to inactivity

PR #1148, opened 2019-06-04
Introduce 3 running modes: fs, ftfs, ftbe

Pchelolo:


d00rman:
For our sanity, the diff since #1144 is [here](https://github.com/d00rman/restbase/compare/path-split..Pchelolo:split_3)

sys/backend_proxy.js
```diff
@@ -0,0 +1,84 @@
...
+    const backendURITemplate = new Template({
+        uri: `${options.backend_host_template}/{{path}}`
+    });
+    const usePathSegment = options.use_path_segment;
+    const blockExternalReqs = options.block_external_reqs;
+    return {
+        spec: {
+            paths: {
+                '/{+path}': {
+                    'x-hidden': true,
```
d00rman: Putting this here only hides `/{+path}` but not where it's mounted, so `/v1/key_value` will still show up in the spec.

config.frontend.test.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+# Finally, a standard service-runner config.
+info:
+  name: restbase
+
+services:
+  - name: restbase
+    module: hyperswitch
+    conf:
+      port: 7232
```
d00rman: 7232 is used by AQS, so let's use 7233 to keep our sanity (this is also most likely the port to be used in production)

config.fullstack.test.yaml
```diff
@@ -0,0 +1,251 @@
...
+                  <<: *default_options
+                  pageviews:
+                    host: https://wikimedia.org/api/rest_v1/metrics
+              - path: projects/proxy.yaml
+                options:
+                  backend_host_template: '{{"/{domain}/sys/legacy"}}'
+                  block_external_reqs: true
+          /{api:sys}:
+            x-modules:
+              - path: projects/proxy.yaml
```
d00rman: Why the double proy both here and in v1? We just need the one in v1 and then mount key_val and page_revs normally under sys/

sys/backend_proxy.js
d00rman: Since the HS PR will not be merged, this needs to be moved under `all`

config.fullstack.test.yaml
Pchelolo: This is one thing I wanna discuss.. It's because we've moved `JSON.parse(res.body.toString('utf8'))` into the proxy - so now modules don't support getting the Buffer from the backend anymore. Not sure what to do here.

sys/backend_proxy.js
Pchelolo: Done

config.frontend.test.yaml
```diff
@@ -0,0 +1,202 @@
...
+                options: *default_options
+              - path: test/test_module.yaml
+                options:
+                  events:
+                    uri: http://127.0.0.1:8085/v1/events
+                    topic: resource_change
+                    transcludes_topic: change-prop.transcludes.resource-change
+          /{api:sys}: *default_sys
+
+enwiki_project: &enwiki_project
```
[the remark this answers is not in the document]
Pchelolo: Hm... How did that happen?

config.frontend.test.yaml
Pchelolo: Done.

d00rman:
I think we should leave the projects under `projects/` for now. Literally all of the RB configs we have lying around in various places use them, so let's keep them for now (and add a comment on the top saying there are deprecated). This would allow us to merge the split, change the configs gradually and then finally remove the files altogether instead of having to do all at once.

berndsi:
Just out of curiosity. What do the three acronyms mean? fs, ftfs, ftbe

d00rman:
> Just out of curiosity. What do the three acronyms mean? fs, ftfs, ftbe

full-stack, front-end / full-stack and front-end / back-end. These are the combinations of services we will have after the split. ftbe is the final stage, while fefs is the intermediate step.

d00rman:
> > Just out of curiosity. What do the three acronyms mean? fs, ftfs, ftbe
> 
> full-stack, front-end / full-stack and front-end / back-end. These are the combinations of services we will have after the split. ftbe is the final stage, while fefs is the intermediate step.

For a "longer" explanation, cf. the description of #1144 .

config.frontend.test.yaml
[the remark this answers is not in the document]
Pchelolo: Done

sys/backend_proxy.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+const HyperSwitch = require('hyperswitch');
+const Template = HyperSwitch.Template;
+
+module.exports = (options) => {
+    options = options || {};
+    options.backend_host_template = options.backend_host_template || '/{domain}/sys';
```
d00rman: This should probably be `/{domain}/sys/legacy` now

sys/backend_proxy.js
```diff
@@ -0,0 +1,67 @@
...
+        spec: {
+            paths: {
+                '/{+path}': {
+                    'x-route-filters': options.block_external_reqs ?
+                        [{
+                            type: 'default',
+                            name: 'header_match',
+                            options: {
+                                whitelist: {
+                                    'x-client-ip': ['/^(?:::ffff:)?(?:10|127)\\./']
```
d00rman: Add `::1` since that's the IPv6 address for localhost.

sys/backend_proxy.js
```diff
@@ -0,0 +1,67 @@
...
+                })
+                .then((res) => {
+                    // We are slowly switching to storing data as binary blobs
+                    // in the storage component. When running multi-process,
+                    // the storage component will return the appropriate
+                    // content-type and deserializing the response will
+                    // happen internally in preq. However, when running
+                    // single process, preq is not involved, so for consistency
+                    // we need to manually deserialize the returned blob.
+                    if (res &&
```
d00rman: We also need to deserialise if the content type is `text/plain` as that's used in Mathoid for the indirection table

v1/summary_new.yaml
```diff
@@ -93,8 +93,10 @@ paths:
       x-setup-handler:
         # Set up a simple key-value bucket.
         - init:
-            method: put
+            method: 'put'
```
d00rman: Quotes are not needed

config.frontend.test.yaml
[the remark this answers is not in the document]
d00rman: We can get rid of this definition entirely since it's the same as `wp.org` and use that spec for `en.wp.org` on L152 further down.

config.frontend.test.yaml
```diff
@@ -0,0 +1,202 @@
...
+    module: hyperswitch
+    conf:
+      port: 7233
+      spec: *spec_root
+      salt: secret
+      default_page_size: 1
+      user_agent: RESTRouer-testsuite
+      ui_name: RESTBase
+      ui_url: https://www.mediawiki.org/wiki/RESTBase
+      ui_title: RESTBase docs
```
d00rman: REST API docs

config.frontend.test.yaml
```diff
@@ -0,0 +1,202 @@
...
+services:
+  - name: restrouter
+    module: hyperswitch
+    conf:
+      port: 7233
+      spec: *spec_root
+      salt: secret
+      default_page_size: 1
+      user_agent: RESTRouer-testsuite
+      ui_name: RESTBase
```
d00rman: REST API

config.fullstack.test.yaml
```diff
@@ -0,0 +1,248 @@
...
+                    uri: http://127.0.0.1:8085/v1/events
+                    topic: resource_change
+                    transcludes_topic: change-prop.transcludes.resource-change
+              - path: projects/proxy.yaml
+                options:
+                  backend_host_template: '{{"/{domain}/sys/legacy"}}'
+                  block_external_reqs: true
+          /{api:sys}: *default_sys
+
+enwiki_project: &enwiki_project
```
d00rman: Idem as for front-end, let's remove this and use only `wikipedia_project`

config.fullstack.test.yaml
```diff
@@ -0,0 +1,248 @@
...
+  x-sub-request-filters:
+    - type: default
+      name: http
+      options:
+        allow:
+          - pattern: /^https?:\/\/[a-zA-Z0-9\.]+\/w\/api\.php/
+            forward_headers: true
+          - pattern: /^https?:\/\/parsoid-beta.wmflabs.org.+/
+            forward_headers: true
+          # Need to forward cookie to backend RESTBase
```
d00rman: Euh, no we don't here. the FS *is* the back-end.

projects/proxy.yaml
```diff
@@ -0,0 +1,9 @@
+paths:
```
d00rman: Why is this file in the `projects` dir? It's not really a project...

sys/backend_proxy.js
d00rman: Actually, we can probably make it generally for all `text/*` content types. It makes sense.

d00rman:
And a minor nit coming from my OCD: `s/ft/fe` in the test type name (and the name of this PR :P)

sys/backend_proxy.js
Pchelolo: no.. we don't, mathoid uses 'string' cass type, so it's already decerialized.

config.fullstack.test.yaml
Pchelolo: No, can't do it, we add test_module to one of them and not another. For a reason of one is used to validate normal spec, one is used to do tests that use test_module

projects/proxy.yaml
Pchelolo: it's kinda a collection of modules which all the projects are..

config.frontend.test.yaml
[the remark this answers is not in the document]
Pchelolo: No. for testing we mount test_module to one and not to another.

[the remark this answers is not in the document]
d00rman: what about having a special test domain for this test module case? I'd like to keep the actual ones used in production as they are in production.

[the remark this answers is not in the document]
Pchelolo: With the tests we have that rely on the testing module that should work. Lemme try that.

d00rman:
Hm, it has just occurred to me that we'll also need an `fe` testing mode too, since we will need to be able to execute the tests for RESTRouter in the WMF CI pipeline. But we currently can't have that mode, as we don't have an FS RB to connect it to. This will need to be a follow-up once we deploy this PR everywhere.

config.frontend.test.yaml
```diff
@@ -0,0 +1,202 @@
...
+                    /mathoid:
+                      x-modules:
+                        - path: sys/mathoid.js
+                          options: *mathoid_options
+                    /post_data:
+                      x-modules:
+                        - path: sys/post_data.js
+                    /events:
+                      x-modules:
+                        - path: sys/events.js
```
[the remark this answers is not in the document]
Pchelolo: This is actually deliberate for testing. Will add a comment.

config.fullstack.test.yaml
```diff
@@ -0,0 +1,244 @@
...
+                    /mathoid:
+                      x-modules:
+                        - path: sys/mathoid.js
+                          options: *mathoid_options
+                    /post_data:
+                      x-modules:
+                        - path: sys/post_data.js
+                    /events:
+                      x-modules:
+                        - path: sys/events.js
```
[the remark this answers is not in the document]
Pchelolo: Again. deliberately not doing so.

sys/parsoid.js
```diff
@@ -204,7 +204,7 @@ class ParsoidService {
             uri: this.getStashBucketURI(domain, title, revision, tid)
         })
         .then((res) => {
-            res = JSON.parse(res.body.toString('utf8'));
+            res = res.body;
```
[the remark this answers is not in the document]
Pchelolo: meh.. Only hides errors elsewhere..

Pchelolo:
rebased and all comments addressed.

Pchelolo:
Woooohoooo!

PR #1149, opened 2019-06-04
Removed mwUtil.decodeBody

hknustwmf:
The buffer to string method is no longer needed since _getContentWithFallback provides the response object and the body is already decoded to string.

Bug: T221720


Pchelolo:
Good!

PR #1150, opened 2019-06-07
Mathoid: Move resource declarations into the sys/ module

d00rman:
The resource declarations for the images' buckets resided in
v1/mathoid.yaml, all of them creating the images' buckets in the global
domain. Alas, this causes RESTBase to issue check-and-create-bucket
requests for them for each domain it loads. Therefore, to optimise the
start-up process, move them to sys/mathoid.js, which is loaded only for
the global domain.


Pchelolo:
Makes sence.

Pchelolo:
Let's wait for travis

PR #1151, opened 2019-06-10
Rm backwards compatibility for mediawiki-basefields

mvolz:
Backwards compatibility for the format
mediawiki-basefields is no longer needed because
this is the native name of the format.

The ability to use the extra basefields parameter
was removed in citoid in change I7fc0004abe6

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "0.19.4",
+  "version": "0.19.5",
   "description": "REST storage and service dispatcher",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "0.19.4",
+  "version": "0.19.5",
```
d00rman: No version bump here, please

mvolz: done

PR #1152, opened 2019-06-11
Replace cassandra-uuid with uuid

hknustwmf:
Added uuidUtils module to provide some helper functions. Bumped the version number  to 0.19.5

Note: hyperswitch and restbase-mod-sqlite PRs for uuid need to be merged first for this to build.

Bug:  T223690

### DIFF
--- a/lib/uuidUtils.js
+++ b/lib/uuidUtils.js
@@ -1,0 +1,70 @@
+'use strict';
+
+const uuidRegEx = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/;
+
+const uuidUtils = {};
+
+uuidUtils.V1_EPOCH_OFFSET = 122192928000000000; //  100 nanosecond intervals since 15 October 1582
+
+/**
+ * Extract time in nanoseconds from UUID V1
+ *
+ * @param {string} uuidv1
+ * @return {number|null}
+ *
+ */
+uuidUtils.getTimeInNs = (uuidv1) => {
+    if (!(typeof uuidv1 === 'string' || uuidv1 instanceof String)) {
+        return undefined;
+    }
+
+    const uuidComponents = uuidv1.split('-');
+    return parseInt(
+        [
+            uuidComponents[2].substring(1),
+            uuidComponents[1],
+            uuidComponents[0]
+        ].join(''), 16);
+};
+
+/**
+ * Extract time in microseconds from UUID V1
+ *
+ * @param {string} uuidv1
+ * @return {number|undefined}
+ *
+ */
+uuidUtils.getTime = (uuidv1) => {
+    const timeInNs = uuidUtils.getTimeInNs(uuidv1);
+    // Convert ns intervals to ms since Jan 1 1970
+    return timeInNs ? Math.floor((timeInNs - uuidUtils.V1_EPOCH_OFFSET) / 10000) : undefined;
+};
+
+/**
+ * Extract date from UUID V1
+ *
+ * @param {string} uuidv1
+ * @return {Date|undefined}
+ *
+ */
+uuidUtils.getDate = (uuidv1) => {
+    const timeInMs = uuidUtils.getTime(uuidv1);
+    return timeInMs ? new Date(timeInMs) : undefined;
+};
+
+/**
+ * Validate UUID
+ *
+ * @param {string} uuid
+ * @return {boolean}
+ *
+ */
+uuidUtils.test = (uuid) => {
+    if (!(typeof uuid === 'string' || uuid instanceof String)) {
+        return false;
+    }
+
+    return uuidRegEx.test(uuid);
+};
+
+module.exports = uuidUtils;
--- a/package.json
+++ b/package.json
@@ -36,16 +36,16 @@
   "readme": "README.md",
   "dependencies": {
     "bluebird": "^3.5.5",
-    "cassandra-uuid": "^0.1.0",
     "content-type": "git+https://github.com/wikimedia/content-type#master",
     "entities": "^1.1.2",
     "fast-json-stable-stringify": "^2.0.0",
-    "hyperswitch": "^0.12.3",
+    "hyperswitch": "^0.12.4",
     "jsonwebtoken": "^8.5.1",
     "mediawiki-title": "^0.6.5",
-    "restbase-mod-table-cassandra": "^1.1.4",
+    "restbase-mod-table-cassandra": "^1.2.0",
     "semver": "latest",
-    "service-runner": "^2.7.0"
+    "service-runner": "^2.7.1",
+    "uuid": "^3.3.2"
   },
   "devDependencies": {
     "ajv": "^6.10.0",
@@ -63,7 +63,7 @@
     "nyc": "^14.1.1",
     "openapi-schema-validator": "^3.0.3",
     "preq": "^0.5.8",
-    "restbase-mod-table-sqlite": "^1.1.2"
+    "restbase-mod-table-sqlite": "^1.1.3"
   },
   "engines": {
     "node": ">=6"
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "0.19.4",
+  "version": "0.19.5",
```
d00rman: No version bump, here, please.

package.json
```diff
@@ -36,16 +36,16 @@
   "readme": "README.md",
   "dependencies": {
     "bluebird": "^3.5.5",
-    "cassandra-uuid": "^0.1.0",
     "content-type": "git+https://github.com/wikimedia/content-type#master",
     "entities": "^1.1.2",
     "fast-json-stable-stringify": "^2.0.0",
-    "hyperswitch": "^0.12.3",
+    "hyperswitch": "^0.12.4",
```
d00rman: This is not needed here

package.json
```diff
@@ -36,16 +36,16 @@
...
     "jsonwebtoken": "^8.5.1",
     "mediawiki-title": "^0.6.5",
-    "restbase-mod-table-cassandra": "^1.1.4",
+    "restbase-mod-table-cassandra": "^1.2.0",
```
d00rman: Idem

package.json
```diff
@@ -63,7 +63,7 @@
     "nyc": "^14.1.1",
     "openapi-schema-validator": "^3.0.3",
     "preq": "^0.5.8",
-    "restbase-mod-table-sqlite": "^1.1.2"
+    "restbase-mod-table-sqlite": "^1.1.3"
```
d00rman: Idem

sys/parsoid.js
```diff
@@ -59,7 +61,8 @@ function isModifiedSince(req, res) {
         if (req.headers['if-unmodified-since']) {
             const jobTime = Date.parse(req.headers['if-unmodified-since']);
             const revInfo = mwUtil.parseETag(res.headers.etag);
-            return revInfo && uuid.fromString(revInfo.tid).getDate() >= jobTime;
+
```
d00rman: No empty line

test/features/events/events.js
```diff
@@ -2,7 +2,7 @@
 
 const preq   = require('preq');
 const http   = require('http');
-const uuid   = require('cassandra-uuid').TimeUuid;
+const uuidTools   = require('../../../lib/uuidUtils');
```
d00rman: Consistency, please: `s/uuidTools/uuidUtils/`

lib/uuidUtils.js
```diff
@@ -0,0 +1,57 @@
...
+const uuidUtils = {};
+
+/**
+ * Extract time in nanoseconds from UUID V1
+ *
+ * @param {string} uuidv1
+ * @return {Integer|null}
+ *
+ */
+uuidUtils.getTime = (uuidv1) => {
```
d00rman: This will return the number of nanoseconds from 1582-10-15 but time is expressed in milliseconds elapsed since 1970-01-01

lib/uuidUtils.js
```diff
@@ -0,0 +1,57 @@
...
+ * @return {Date|null}
+ *
+ */
+uuidUtils.getDate = (uuidv1) => {
+    if (!(typeof uuidv1 === 'string' || uuidv1 instanceof String)) {
+        return null;
+    }
+
+    // Drop the ns for conversion
+    return new Date(Math.floor((uuidUtils.getTime(uuidv1) - uuidV1Offset) / 10000));
```
d00rman: Suffers from same the problem as `getTime`. Once that is fixed, here you'll need only `new Date(uuidUtils.getTime(uuidv1))`. You'll also need to check for the return value, as `new Date(undefined)` returns the current date.

lib/uuidUtils.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Extract time in nanoseconds from UUID V1
+ *
+ * @param {string} uuidv1
+ * @return {Date|null}
+ *
+ */
+uuidUtils.getDate = (uuidv1) => {
+    if (!(typeof uuidv1 === 'string' || uuidv1 instanceof String)) {
```
d00rman: Move this check to `getTime()` since this function uses that one. Also, better to return `undefined` rather than `null`

package.json
hknustwmf: 1.2.0 pulls in restbase-mod-table-spec 1.2.0 which uses uuid. Version 1.1.4 depends on restbase-mod-table-spec 1.1.0 which would install cassandra-uuid as a dependency. That would be taking a step backwards.

package.json
hknustwmf: This change involves code changes not just dev depency changes. Is there a document that outlines when version numbers should or should not be changed?

package.json
hknustwmf: I bumped the version of hyperswitch to 1.2.4 in the PR to indicate the change to uuid and code changes. Remaining at 0.12.3 would reintroduce cassandra-uuid.

package.json
hknustwmf: same as above. Reintroduces cassandra-uuid

lib/uuidUtils.js
hknustwmf: Added check to return statement

lib/uuidUtils.js
hknustwmf: Changed to undefined

lib/uuidUtils.js
hknustwmf: Added function to return ns and moved the ns to ms conversion to getTime()

test/features/events/events.js
hknustwmf: Changed the name

sys/parsoid.js
hknustwmf: Removed

package.json
d00rman: It's a bit confusing, I admit. The reason why the version shouldn't be bumped here is because we haven't changed RB's version in a while all the while introducing changes, so bumping the patch version here would be semantically wrong. On the other hand, we have the split coming up, and once that is merged we will create a new release, which will include this code.

package.json
d00rman: Good point. Meta comment: when you reference a different PR, please provide a link to it.

package.json
```diff
@@ -36,16 +36,16 @@
...
     "fast-json-stable-stringify": "^2.0.0",
-    "hyperswitch": "^0.12.3",
+    "hyperswitch": "^0.12.4",
     "jsonwebtoken": "^8.5.1",
     "mediawiki-title": "^0.6.5",
-    "restbase-mod-table-cassandra": "^1.1.4",
+    "restbase-mod-table-cassandra": "^1.2.0",
     "semver": "latest",
-    "service-runner": "^2.7.0"
+    "service-runner": "^2.7.0",
```
d00rman: Unrelated, but please bump to 2.7.1

package.json
hknustwmf: Rolled back.

package.json
hknustwmf: I assume this comment was in regards to the RB version number. I left all dependency version numbers the same except for service runner as requested

package.json
hknustwmf: Done

test/features/pagecontent/rerendering.js
```diff
@@ -81,6 +81,7 @@ describe('page re-rendering', function() {
             throw new Error('Expected a precondition failure');
         },
         (res) => {
+
```
d00rman: No empty line needed

package.json
hknustwmf: Will do going forward.

> Please add some tests for the UUID functionality. Some other comments in-lined.

Added tests

test/features/pagecontent/rerendering.js
hknustwmf: Removed

PR #1153, opened 2019-06-21
Add Suggested Edits endpoints

mdholloway:

The caption suggestion endpoints are supported only for Commons domains,
and the description suggestion endpoints are supported only for Wikidata
domains.  (This is enforced in the service configuration, and is not reflected here.)

Support for commons.wikimedia.org as a default project is added to
config.example.wikimedia.yaml for development and testing.

Support for recommendations endpoints is added to the wikidata.org
project in the example config.


mdholloway:
@mateusbs17 This is an example of adding a basic endpoint or set of endpoints to RESTBase with no provision for Cassandra storage.

mateusbs17:
I'm gonna try to answer these questions.

> [...] e.g. what is the meaning of `source` and `target` if the end point is accessed via `en.wp.org`? 
It will return `400 - Unsupported domain`

The [allowed domains](https://github.com/wikimedia/mediawiki-services-recommendation-api/blob/master/config.dev.yaml#L105) are:
```
description_allowed_domains:
        - www.wikidata.org
        - test.wikidata.org
        - wikidata.beta.wmflabs.org
      caption_allowed_domains:
        - commons.wikimedia.org
        - test-commons.wikimedia.org
        - commons.wikimedia.beta.wmflabs.org
```

> What is the source language for end points that only expect the target to be provided by the client?
I might be wrong, but the addition endpoints only need to know the target language which will get the new description/caption, the source doesn't matter since there is no manipulation of previously existing description/caption.

> Better yet, for end points that have both source and target, we should probably rework them so that the source is taken from the domain and then let the client specify the target (that's how it is currently done).
That's something I'm don't know how to answer, probably @joewalsh have a better idea about the service architecture.

I might be missing something, but these endpoints are supposed to work exclusively with commons and there is no need to do language differentiation per client since the users [need to have set multiple languages in the app](https://phabricator.wikimedia.org/T223131)

Pchelolo:
So, I guess we'd need a special project for commons now :(

mdholloway:
There's nothing requiring us to use the wikidata and commons domains per se.  We could just treat the domain as the target language, since there's a target language in every case.  The source could be a path parameter.  (Addition endpoints have no source language.)

IOW, they could look like:
- /{target}.wikipedia.org/api/rest_v1/data/recommendation/description/addition
- /{target}.wikipedia.org/api/rest_v1/data/recommendation/description/translation/from/{source}

The big drawback to using the wiki language subdomain as the language is that it means we can't specify a language variant as the target.  That would be unfortunate.

The same applies for the caption addition and translation endpoints, though Wikipedia domains would be a more awkward fit in that case, since the results are Commons file pages for files that aren't necessarily used on Wikipedias.

mdholloway:
The reason for preferring wikidata and commons for description and caption suggestions, respectively, is that those are the domains that the results most relate to.

d00rman:
@mdholloway @mateusbs17 FYI, we have merged #1148, so now you can rebase and start working on this.

mateusbs17:
@Pchelolo and @d00rman I created `v1/recommend-caption.yaml` and `v1/recommend-description.yaml` instead of changing `v1/recommend.yaml` to add the endpoints to the projects where they are allowed, let me know if you have any concerns.

config.frontend.test.yaml
```diff
@@ -98,6 +98,19 @@ wiktionary_project: &wiktionary_project
                 options: *default_options
           /{api:sys}: *default_sys
 
+commons_project: &commons_project
+  x-modules:
+    - spec:
+        paths:
+          /{api:v1}:
+            x-modules:
+              - path: projects/v1/commons.wmf.yaml
```
Pchelolo: The point I've had on the meeting about being able not to create a commons project. We can just use the default project and add the recommned-captions right here in the config

Pchelolo:
The `tests/features/specification/monitoring` test suite has a list of domains we try out for local `x-amples` testing. Now that we have endpoints specific to commons and to wikidata, we need to add those domains to the list.

mateusbs17:
I fixed the recommend.yaml but tests still fails. When accessing `http://localhost:7231/wikidata.org/v1/?spec`, it returns:
```
{
type: "https://mediawiki.org/wiki/HyperSwitch/errors/server_error",
title: "Site info fetch failed.",
method: "get",
detail: "SiteInfo is unavailable for wikidata.org",
uri: "/wikidata.org/v1/"
}
```

The last commit removes wikidata.org from the monitoring test just to make CI happy, any thoughts about that?

d00rman:
@mateusbs17 the domain name is `www.wikidata.org`, not `wikidata.org`, that's why the site info fetch fails.

mateusbs17:
> @mateusbs17 the domain name is `www.wikidata.org`, not `wikidata.org`, that's why the site info fetch fails.

Thanks, that was the problem indeed. Seems fixed, let me know if there is anything else.

mdholloway:
Looks like all comments are addressed.  Are we still on track to deploy this today?

Pchelolo:
patience, @mdholloway maybe yes, maybe no... we haven't tried again deploying RB yet..

mdholloway:
OK, thanks for the update, @Pchelolo.  @mateusbs17 please keep Joe up to date on what's happening with this.

config.fullstack.test.yaml
```diff
@@ -135,6 +135,32 @@ wiktionary_project: &wiktionary_project
...
+        paths:
+          /{api:v1}:
+            x-modules:
+              - path: projects/v1/default.wmf.yaml
+                options: *default_options
+              - path: v1/recommend-caption.yaml
+                options: '{{options.recommendation}}'
+              - path: projects/proxy.yaml
+                options: *proxy_options
+          /{api:sys}: *default_sys
```
Pchelolo: This portion actually needs to be transferred to the production config. Want me to do it or want me to show you how it's done @mateusbs17 ?

Pchelolo: Also, we need to update `config.fronted.wikimedia.yaml` similarly.

Pchelolo: Went ahead and made a prod config change for you https://gerrit.wikimedia.org/r/c/mediawiki/services/restbase/deploy/+/519500

v1/recommend-caption.yaml
```diff
@@ -0,0 +1,143 @@
...
+          description: Error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/problem'
+      x-request-handler:
+        - get_from_backend:
+            request:
+              uri: '{{options.host}}/{{domain}}/v1/caption/addition/{target}'
+      x-amples:
```
Pchelolo: Needs x-monitor: true as well

v1/recommend-caption.yaml
```diff
@@ -0,0 +1,143 @@
...
+          description: Error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/problem'
+      x-request-handler:
+        - get_from_backend:
+            request:
+              uri: '{{options.host}}/{{domain}}/v1/caption/translation/from/{source}/to/{target}'
+      x-amples:
```
Pchelolo: Needs x-monitor: true as well

v1/recommend-description.yaml
```diff
@@ -0,0 +1,155 @@
...
+          description: Error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/problem'
+      x-request-handler:
+        - get_from_backend:
+            request:
+              uri: '{{options.host}}/{{domain}}/v1/description/addition/{target}'
+      x-amples:
```
Pchelolo: Needs x-monitor: true as well

v1/recommend-description.yaml
```diff
@@ -0,0 +1,155 @@
...
+          description: Error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/problem'
+      x-request-handler:
+        - get_from_backend:
+            request:
+              uri: '{{options.host}}/{{domain}}/v1/description/translation/from/{source}/to/{target}'
+      x-amples:
```
Pchelolo: Needs x-monitor: true as well.

Actually. while you're at it, could you amend our tests to fail if there IS x-amples and no x-monitoring. Right about here: https://github.com/wikimedia/restbase/blob/master/test/features/specification/monitoring.js#L43

d00rman: `x-monitor` is actually not needed if it's value it's supposed to be `true`. The default is `x-monitor: true` in service-checker

Pchelolo: olala... that's the shitty part of having the checker reimplemented in node for testing. out tests need x-monitor

mateusbs17:
After adding x-monitor true it was possible to see that the x-amples tests don't work properly, the reason I found is related with the wmflabs host passed as options.
- https://recommendation-api-beta.wmflabs.org/www.wikidata.org/v1/description/translation/from/en/to/zh breaks because labs is not able to access wikidata
- https://app-editor-tasks.wmflabs.org/www.wikidata.org/v1/description/translation/from/en/to/zh works fine

Even though I added the options to both wikidata and commons, some tests still gets 404.

config.frontend.test.yaml
```diff
@@ -98,6 +100,30 @@ wiktionary_project: &wiktionary_project
...
+  x-modules:
+    - spec:
+        paths:
+          /{api:v1}:
+            x-modules:
+              - path: projects/v1/default.wmf.yaml
+                options: *default_options
+              - path: projects/proxy.yaml
+                options: *proxy_options
+          /{api:sys}: *default_sys
```
Pchelolo: How's this definition different from the default_project? I think you forgot to bring in recommendation endpoints here.

config.fullstack.test.yaml
Pchelolo: Same question.

projects/v1/default.wmf.yaml
```diff
@@ -96,3 +96,5 @@ paths:
       - path: v1/javascript.yaml
         options:
           host: '{{options.mobileapps.host}}'
+      - path: v1/recommend-caption.yaml
+        options: '{{options.appeditortasks}}'
```
Pchelolo: Mmmm... Haven't we decided that this will only be exposed on commons?

config.frontend.test.yaml
thesocialdev: Adding the recommend-caption endpoints in the config file seems is breaking on tests, it seems that the host for appeditortask is ignored.

Pchelolo: It shouldn't be if you pass the options in there

thesocialdev: My attempt were:
```
commons_project: &commons_project
  x-modules:
    - spec:
        paths:
          /{api:v1}:
            x-modules:
                options: *default_options
              - spec:
                  paths:
                    /data:
                      x-modules:
                        - path: v1/recommend-caption.yaml
                          options: '{{options.appeditortasks}}'
              - path: projects/proxy.yaml
                options: *proxy_options
          /{api:sys}: *default_sys
```
The routes are fine when accessing localhost:7231/commons.wikimedia.org/?spec, but when accessing the endpoint the results got 404 status, That's why I moved to the default project.

Pchelolo: Should work pretty much exactly how it does for the wikidata descriptions thing.

thesocialdev: I thought the same, but got different results when testing, am I missing something here?

Pchelolo: you can
a) put the /data right into the parts in the recommend-caption
b) Do this

```
              - spec:
                  paths:
                    /data:
                      x-modules:
                        - path: v1/recommend-caption.yaml
                          options: '{{options.appeditortasks}}'
                 options: '{{options}}'
```

You're passing `options` on every module load and here you have 2.

Pchelolo:
As for `x-amples`... hm... Would the labs instance support beta commons and beta wikidata? You can expose those under beta domains and not specify domain in the test explicitly - that should work

projects/v1/default.wmf.yaml
thesocialdev: Done.

config.frontend.test.yaml
thesocialdev: Done.

config.frontend.test.yaml
```diff
@@ -29,6 +29,8 @@ default_project: &default_project
                     host: https://citoid-beta.wmflabs.org
                   recommendation:
                     host: https://recommendation-api-beta.wmflabs.org
+                  appeditortasks:
+                    host: https://app-editor-tasks.wmflabs.org
```
Pchelolo: What is this domain pointing too? it's a different instance of recommendation api service or what is this?

thesocialdev:
> As for `x-amples`... hm... Would the labs instance support beta commons and beta wikidata? You can expose those under beta domains and not specify domain in the test explicitly - that should work

That's something @mdholloway would know more about. I can't find the documented thing (e-mail/wiki page) that explains this labs setup.

config.frontend.test.yaml
thesocialdev: That's the labs instance @mdholloway was working on. I think this instance has a special setup for the endpoints to work with [Extension:WikimediaEditorTasks](https://www.mediawiki.org/wiki/Extension:WikimediaEditorTasks), commons and wikidata.

Pchelolo: Hm. but where would we point production too? To recommendation-api LVS? I think we need a bit more info from @mdholloway here.

thesocialdev: @Pchelolo see, https://github.com/wikimedia/restbase/pull/1153#issuecomment-507436692

thesocialdev: > Hm. but where would we point production too? To recommendation-api LVS? I think we need a bit more info from @mdholloway here.

Yes, the deployment followed the regular recommendation-api target machines. The limitations of the beta-cluster that made us create a new instance to perform the actions the Apps team needed for testing.

Pchelolo: I'm not entirely sure why the normal `recommendation-api` service can not handle these recommendations.. Since the normal recommendation-api.beta.wmflabs.org is as well completely under our control, why can't we make that one work and not create additional installations of the same thing?

Pchelolo: Perhas cause the recommendation service installed in beta doesn't seem to work. Then we need to fix it and not create a separate installation.

mdholloway: This should indeed point to the Beta Cluster. `app-editor-tasks` was just for prototyping and testing, particularly before the code was merged onto recommendation-api master.

mdholloway: (Pointing to `app-editor-tasks` in this PR rather than BC like the others was probably just an error on my part.)

projects/v1/default.wmf.yaml
mdholloway: > Mmmm... Haven't we decided that this will only be exposed on commons?

Er, that's news to me. Was that discussed in the sync meeting last week?  The description suggestion endpoints don't make sense on Commons.

config.frontend.test.yaml
mdholloway: Ah, OK, I see what's going on here.  Yeah, we definitely don't want to introduce an `appeditortasks` service to RB.  I'll update the PR.

mdholloway:
I'm not able to force-push my updates, so I opened a new PR at #1159.

v1/recommend-description.yaml
mdholloway: Which config file should we now be using for ordinary testing?  These now pass with `x-monitor: true` when using `config.fullstack.test.yaml` but only for beta cluster domains.  But won't this put beta cluster domains on the Swagger doc page, even in production?

Also, I noticed that most endpoints have `x-monitor: false` in RESTBase.  That's why I did likewise in the new PR before seeing this comment.

projects/v1/default.wmf.yaml
mdholloway: Never mind, misunderstood the comment ;)

Pchelolo:
ok, now we have 2 of these and also something clearly went wrong with git history here... we definitely can't merge this with broken history.

mdholloway:
Sorry for the mess.  Cleaned up and rebased in #1160.

@thesocialdev You're not configured as a reviewer in this repo, so I can't add you, but your reviews are also welcome.

PR #1156, opened 2019-06-26
Expose new talk endpoint

thesocialdev:
Expose new talk endpoint in the wikipedia projects only, no cassandra
storage is needed, only regular Varnish layer. The endpoints are:

  /{domain}/v1/page/talk/{title}
  /{domain}/v1/page/talk/{title}/{revision}

Both return JSON with structured talk page.

Bug: T225733

### DIFF
--- a/projects/v1/wikipedia.wmf.yaml
+++ b/projects/v1/wikipedia.wmf.yaml
@@ -90,6 +90,10 @@ paths:
       - path: v1/random.yaml
         options: '{{merge({"random_cache_control": "s-maxage=2, max-age=1"},
                       options.mobileapps)}}'
+      - path: v1/talk.yaml
+        options:
+          response_cache_control: '{{options.purged_cache_control}}'
+          host: '{{options.mobileapps.host}}'
       - path: v1/pdf.js
         options: '{{options.pdf}}'
       - path: v1/common_schemas.yaml # Doesn't really matter where to mount it.
--- a/projects/v1/wikipedia.wmf.yaml
+++ b/projects/v1/wikipedia.wmf.yaml
@@ -92,7 +92,6 @@ paths:
                       options.mobileapps)}}'
       - path: v1/talk.yaml
         options:
-          response_cache_control: '{{options.purged_cache_control}}'
           host: '{{options.mobileapps.host}}'
       - path: v1/pdf.js
         options: '{{options.pdf}}'
--- a/test/features/specification/monitoring.js
+++ b/test/features/specification/monitoring.js
@@ -26,13 +26,22 @@ function constructTestCase(title, path, method, request, response) {
     };
 }
 
+function filterPath(paths, pathStr, method) {
+    const p = paths[pathStr][method];
+    if (p['x-amples'] && typeof p['x-monitor'] === 'undefined') {
+        throw new Error (`${'A Method that contains examples should declare x-monitor: true.'
+        + ' Path: '}${pathStr} Method: ${method}`)
+    }
+    return p['x-monitor'];
+}
+
 function constructTests(spec, options, server) {
     const paths = spec.paths;
     const ret = [];
     Object.keys(paths).forEach((pathStr) => {
         if (!pathStr) { return; }
         Object.keys(paths[pathStr]).filter((method) => {
-            return paths[pathStr][method]['x-monitor'];
+            return filterPath(paths, pathStr, method);
         })
         .forEach((method) => {
             const p = paths[pathStr][method];
--- a/v1/talk.yaml
+++ b/v1/talk.yaml
@@ -1,0 +1,140 @@
+openapi: 3.0.1
+info:
+  version: 0.1.0
+  title: MediaWiki Structured Talk page API
+  description: API for retrieving structured representation of the talk page
+  termsOfService: https://www.mediawiki.org/wiki/REST_API#Terms_and_conditions
+  contact:
+    name: Reading Infrastructure
+    url: https://www.mediawiki.org/wiki/Wikimedia_Reading_Infrastructure_team
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /talk/{title}:
+    get: &talk_title_revision_get_spec
+      tags:
+        - Talk pages
+      summary: Get structured talk page contents
+      description: Gets structured talk page contents for the provided title.
+      parameters:
+        - name: title
+          in: path
+          description: 'Page title. Use underscores instead of spaces. Example: `Main_Page`.'
+          required: true
+          schema:
+            type: string
+        - name: redirect
+          in: query
+          description: |
+            Requests for [redirect pages](https://www.mediawiki.org/wiki/Help:Redirects) return HTTP 302 with a redirect target in `Location` header and content in the body.
+            To get a 200 response instead, supply `false` to the `redirect` parameter.
+          schema:
+            type: boolean
+      responses:
+        200:
+          description: structured talk page JSON.
+          headers:
+            ETag:
+              description: |
+                Syntax: "{revision}/{tid}". Example: "701384379/154d7bca-c264-11e5-8c2f-1b51b33b59fc"
+              schema:
+                type: string
+          content:
+            application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/Talk/0.1.0":
+              schema:
+                type: string
+        301:
+          description: |
+            A permanent redirect is returned if the supplied article title was not in the normalized form.
+            To avoid this kind of redirect, you can use the [mediawiki-title](https://github.com/wikimedia/mediawiki-title) library to perform
+            title normalization client-side.
+
+            Beware that redirected pre-flighted cross-origin requests (such as those sending custom request headers like `Api-User-Agent`)
+            will fail in most current browsers [due to a spec bug](https://github.com/whatwg/fetch/issues/204).
+        302:
+          description: |
+            The page is a [redirect page](https://www.mediawiki.org/wiki/Help:Redirects).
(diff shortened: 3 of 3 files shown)


projects/v1/wikipedia.wmf.yaml
```diff
@@ -90,6 +90,10 @@ paths:
       - path: v1/random.yaml
         options: '{{merge({"random_cache_control": "s-maxage=2, max-age=1"},
                       options.mobileapps)}}'
+      - path: v1/talk.yaml
+        options:
+          response_cache_control: '{{options.purged_cache_control}}'
```
Pchelolo: But it's not purged.. The endpoints we purge are cached quite extensively, so this will draw varnish responses very much out of date. I'm not really sure what the correct value would be for talk pages, it's more a product decision.

v1/talk.yaml
```diff
@@ -0,0 +1,140 @@
...
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/problem'
+      x-request-handler:
+        - get_from_pcs:
+            request:
+              method: get
+              headers:
+                cache-control: '{{cache-control}}'
```
Pchelolo: Since we're not storing it, propagating cache-control is not needed

Pchelolo: Does it support lang variants? If so you need to propagate accept-language.

v1/talk.yaml
```diff
@@ -0,0 +1,140 @@
...
+      x-amples:
+        - title: Get structured talk page for enwiki Salt article
+          request:
+            params:
+              domain: en.wikipedia.org
+              title: Salt
+          response:
+            status: 200
+            headers:
+              content-type: application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/Talk/0.1.0"
```
Pchelolo: It's better to specify a regex here allowing version changes. If you specify specific version RESTBase checks will fail as soon as MCS updates a version.

v1/talk.yaml
```diff
@@ -0,0 +1,140 @@
...
+              vary: /.+/
+              etag: /^"[^/"]+/[^/"]+"$/
+              access-control-allow-origin: '*'
+              access-control-allow-headers: accept, x-requested-with, content-type
+              content-security-policy: default-src
+              x-content-security-policy: default-src
+              x-frame-options: SAMEORIGIN
```
Pchelolo: No need to verify access-contol headers either.

v1/talk.yaml
```diff
@@ -0,0 +1,140 @@
...
+        - title: Get structured talk page for enwiki Salt article
+          request:
+            params:
+              domain: en.wikipedia.org
+              title: Salt
+          response:
+            status: 200
+            headers:
+              content-type: application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/Talk/0.1.0"
+              vary: /.+/
```
Pchelolo: I'm not sure we need to verify Vary.. it's quite an internal thingy.

v1/talk.yaml
```diff
@@ -0,0 +1,140 @@
...
+  title: MediaWiki Structured Talk page API
+  description: API for retrieving structured representation of the talk page
+  termsOfService: https://www.mediawiki.org/wiki/REST_API#Terms_and_conditions
+  contact:
+    name: Reading Infrastructure
+    url: https://www.mediawiki.org/wiki/Wikimedia_Reading_Infrastructure_team
+  license:
+    name: Apache licence, v2
+    url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
```
Pchelolo: These probably need access_check_filter for redirect support? Or it doesn't?

projects/v1/wikipedia.wmf.yaml
berndsi: Yes, we still need purging. Not sure how long we should cache this. Probably don't need to keep around very long. I suspect this to be low traffic, at least initially, since initially this will be used by the iOS app to display a user talk pages.

v1/talk.yaml
berndsi: Yes, we should support language variants here as well.

v1/talk.yaml
berndsi: Yes.

projects/v1/wikipedia.wmf.yaml
Pchelolo: For the purging we need to mangle change-prop then. Could you create a ticket so that we don't forget. I can make a rule.

berndsi: Here's a strawman proposal: How about 5 minutes?

v1/talk.yaml
mateusbs17: Done.

v1/talk.yaml
mateusbs17: Done.

v1/talk.yaml
mateusbs17: Couldn't find any example about how to do it, could you please explain to it to me?

v1/talk.yaml
mateusbs17: Done.

v1/talk.yaml
Pchelolo: https://github.com/wikimedia/restbase/blob/master/v1/transform.yaml#L337

projects/v1/wikipedia.wmf.yaml
berndsi: > For the purging we need to mangle change-prop then. Could you create a ticket so that we don't forget. I can make a rule.

https://phabricator.wikimedia.org/T226669

mateusbs17: > [...] Could you create a ticket so that we don't forget. [...]
https://phabricator.wikimedia.org/T226670

v1/talk.yaml
mateusbs17: Done.

v1/talk.yaml
mateusbs17: @berndsi what do you think?

berndsi: Since I can't find `Vary` mentioned in mobileapps.yaml and mobile-html.yaml I agree with Petr that it's not needed.

mateusbs17: Done.

v1/talk.yaml
Pchelolo: I mean, all the access-control headers. CSP, x-CPS, x-frame-options - nothing

v1/talk.yaml
```diff
@@ -0,0 +1,143 @@
...
+            request:
+              method: get
+              headers:
+                accept-language: '{{accept-language}}'
+              uri: '{{options.host}}/{domain}/v1/page/talk/{title}/{revision}'
+            return:
+              status: 200
+              headers: '{{ get_from_pcs.headers }}'
+              body: '{{get_from_pcs.body}}'
+      x-amples:
```
Pchelolo: You also need `x-monitor: true` to enable it.

v1/talk.yaml
```diff
@@ -0,0 +1,143 @@
...
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/problem'
+      x-request-handler:
+        - get_from_pcs:
+            request:
+              method: get
+              headers:
+                accept-language: '{{accept-language}}'
+              uri: '{{options.host}}/{domain}/v1/page/talk/{title}/{revision}'
```
Pchelolo: This ain't gonna work. 'revision' will be undefined here and you'd end up with a trailing `/`. Luckily, you can do `{title}{/revision}` here and it's gonna fix it.

Pchelolo: not addressed.

v1/talk.yaml
berndsi: Done

v1/talk.yaml
berndsi: Done

v1/talk.yaml
```diff
@@ -0,0 +1,141 @@
...
+          required: true
+          schema:
+            type: integer
+        - name: redirect
+          in: query
+          description: |
+            Requests for [redirect pages](https://www.mediawiki.org/wiki/Help:Redirects) return HTTP 302 with a redirect target in `Location` header and content in the body.
+            To get a 200 response instead, supply `false` to the `redirect` parameter.
+          schema:
+            type: boolean
```
Pchelolo: this will ingerit the whole talk_title_revision_get_spec thus the x-amples too, so it will fail. we need to x-monitor "false" this one.

v1/talk.yaml
berndsi: Should we drop the revision altogether for this path? We'd have it in the other one still.

v1/talk.yaml
mateusbs17: Done.

v1/talk.yaml
mateusbs17: Done.

test/features/specification/monitoring.js
```diff
@@ -26,13 +26,22 @@ function constructTestCase(title, path, method, request, response) {
     };
 }
 
+function filterPath(paths, pathStr, method) {
+    const p = paths[pathStr][method];
+    if (p['x-amples'] && typeof p['x-monitor'] === 'undefined') {
+        throw new Error (`${'A Method that contains examples should declare x-monitor: true.'
+        + ' Path: '}${pathStr} Method: ${method}`)
```
Pchelolo: There's something really weird with the quotes here :)

Pchelolo:
Ok, travis was failing due to some instability on apache.org site, rerunning tests. Otherwise  LGTM

joewalsh:
@Pchelolo any rough idea of when this will be deployed?

PR #1157, opened 2019-06-26
Stop separating mobileapps route definitions to v1 and sys

Pchelolo:
https://phabricator.wikimedia.org/T226579


projects/dev.yaml
```diff
@@ -22,7 +22,7 @@ paths:
                       response_cache_control: '{{options.purged_cache_control}}'
                       cx_host: '{{options.transform.cx_host}}'
                 - path: v1/common_schemas.yaml # Doesn't really matter where to mount it.
-                - path: v1/mobileapps.yaml
+                - path: v1/mobileapps.js
```
d00rman: Since we are still keeping around the old projects, all of them should have the same change as `dev.yaml`

projects/wmf_enwiki.yaml
```diff
@@ -65,7 +65,7 @@ paths:
                   options:
                     response_cache_control: '{{options.purged_cache_control}}'
                     cx_host: '{{options.transform.cx_host}}'
-                - path: v1/mobileapps.yaml
+                - path: v1/mobileapps.js
```
d00rman: The /sys hierarchy should be removed too (in all old-style projects)

Pchelolo: done

PR #1158, opened 2019-07-02
Proton: Support additional options

thesocialdev:
Support format and device type parameters in the RESTBase public PDF
rendering API

Bug: T226677

### DIFF
--- a/v1/pdf.yaml
+++ b/v1/pdf.yaml
@@ -12,8 +12,8 @@ info:
     name: Apache licence, v2
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /pdf/{title}/{format}/{type}:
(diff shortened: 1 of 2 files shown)


v1/pdf.yaml
```diff
@@ -12,7 +12,7 @@ info:
     name: Apache licence, v2
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /pdf/{title}:
+  /pdf/{title}/{format}/{type}:
```
Pchelolo: We cant really do this, this is not backwards compatible so we will break all the existing clients. Both format and type should be optional, this this needs to be split into 3 separate routes. Also, need to document the defaults for the optional parameters.

PR #1159, opened 2019-07-02
Add suggested edits endpoints, take 2

mdholloway:

The caption suggestion endpoints are supported only for Commons domains,
and the description suggestion endpoints are supported only for Wikidata
domains.  (This is enforced in the service configuration, and is not reflected here.)

Support for commons.wikimedia.org as a default project is added to
config.example.wikimedia.yaml for development and testing.

Support for recommendations endpoints is added to the wikidata.org
project in the example config.

(Note: This is a continuation of #1153 after rewriting some history in a way Git doesn't necessarily appreciate.)


config.frontend.test.yaml
```diff
@@ -98,6 +98,32 @@ wiktionary_project: &wiktionary_project
...
+commons_project: &commons_project
+  x-modules:
+    - spec:
+        paths:
+          /{api:v1}:
+            x-modules:
+              - path: projects/v1/default.wmf.yaml
+                options: *default_options
+              - path: v1/recommend-caption.yaml
+                options: '{{options.recommendation}}'
```
thesocialdev: This doesn't work because options is 'undefined' when declared like this. The right thing to do is done in the other PR https://github.com/wikimedia/restbase/pull/1153/files#diff-b91bbd009bed6f8dfe87647993d47521R119

mdholloway: All right, I'll try picking that one back up

PR #1160, opened 2019-07-03
Add Suggested Edits endpoints to Recommendations

mdholloway:

The caption suggestion endpoints are supported for Commons domains,
and the description suggestion endpoints are supported for Wikidata domains.

### DIFF
--- a/config.frontend.test.yaml
+++ b/config.frontend.test.yaml
@@ -186,6 +186,9 @@ spec_root: &spec_root
 
     # labs, used for most tests
     /{domain:en.wikipedia.beta.wmflabs.org}: *wikipedia_project
+    /{domain:commons.wikimedia.beta.wmflabs.org}: *commons_project
+    /{domain:wikidata.beta.wmflabs.org}: *wikidata_project
+
     # Serbian wiki in beta for language variants tests
     /{domain:sr.wikipedia.beta.wmflabs.org}: *wikipedia_project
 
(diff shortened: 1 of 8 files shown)


Pchelolo:
Ok, so you gave up on the x-ample tests :) 

Lemme merge this and try to experiment with that a little..

Pchelolo:
After travis is passing

mdholloway:
Yeah, they pass with Beta Cluster domains but not with prod domains (which makes sense given that that's what the test configs point to).  So we can re-add those for BC domains and set monitoring to true if you want.

Pchelolo:
> Yeah, they pass with Beta Cluster domains but not with prod domains (which makes sense given that that's what the test configs point to). So we can re-add those for BC domains and set monitoring to true if you want.

I actually think it might be a bit harder, since operations-software-service-checker script is involved too. I think I'll just merge this once Travis is done and followup

Pchelolo:
Actually, I need to file a followup ticket here with these x-amples tests. The thing is that in production only endpoints from `en.wikipedia.org` endpoints are checked and all the rest are ignored. Given that we have a non-zero number of endpoints that are not on enwiki, we need some enhancements on how we execute the checks in prod. 

For now, could you please write a couple of tests in just plain javascript for these endpoints and add a comment that "TODO convert it to x-amples once we support running x-amples for multiple domains.

mdholloway:
OK, added tests. Not sure what's up with Travis but it seemed stalled earlier.

PR #1161, opened 2019-07-03
Save SQLite DB file to temp location 

Pchelolo:
For testing, we run in SQLite mode in docker and so can't create the test file in current directory because of permissions. Instead, let's create the DB file in OS temp location.

cc @wikimedia/services @akosiaris

### DIFF
--- a/.pipeline/blubber.yaml
+++ b/.pipeline/blubber.yaml
@@ -19,7 +19,10 @@ variants:
     includes: [build]
     apt: { packages: [libsqlite3-0, ca-certificates] }
     entrypoint: [npm, test]
-    runs: { environment: { TEST_TARGET: sqlite, TEST_MODE: fefs } }
+    runs:
+      environment:
+        TEST_TARGET: sqlite
+        TEST_MODE: fefs
   prep:
     includes: [build]
     node: { env: production }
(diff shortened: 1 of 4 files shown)


eevans:
Insofar as I understand what `insecurely: true` is doing (`COPY [., .]`), this seems reasonable (though I'm not sure I actually understand/appreciate the insecurity).

One alternative would be to create the sqlite database as a unique temp file (the JS equivalent of something like `mktemp -t sqlite.XXXXXXXXXX`), log the resulting filename in case locating it for debugging is required, and leave it to the OS to clean them up.  This might result in more robust tests anyway (for example, ensuring each test run gets a fresh database, keeping the working directory free of detritus, etc)

But to be clear, I'm personally OK with the PR as submitted.  :+1:

akosiaris:
> Insofar as I understand what insecurely: true is doing (COPY [., .]), this seems reasonable (though I'm not sure I actually understand/appreciate the insecurity).

insecurely means that it skips the step separating the owner of the files from the execution user. That is, instead of switching running the commands under the `runuser` the commands will be run using the `somebody` user. Useful while still changing the image. But it could have been named better I think.

LGTM.

Pchelolo:
Thanks @eevans and @akosiaris 

I have changed the premise of the PR to create DB file in the temp directory.

eevans:
Nice; I like it! :+1:

PR #1162, opened 2019-07-08
Ignore cache-control on store requests

Pchelolo:
A fallout from split - we've accidentally started to respect 'cache-control: no-cache' header on write in case `if-none-hash-match` header is provided as well.

Bug: https://phabricator.wikimedia.org/T226983


Pchelolo:
Trivial and UBN, self-merging.

eevans:
LGTM?

PR #1164, opened 2019-07-09
Expose fronted RB on 7231 and backend RB on 7233

Pchelolo:
Bug: https://phabricator.wikimedia.org/T223953#5290864

### DIFF
--- a/config.fullstack.test.yaml
+++ b/config.fullstack.test.yaml
@@ -253,8 +253,9 @@ info:
 services:
   - name: restbase
     module: hyperswitch
-    conf:
-      port: 7231
+    conf: &rb_conf
+      # In production we expose both 7231 and 7233 ports from fullstack RB.
+      port: 7233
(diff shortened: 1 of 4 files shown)


config.fullstack.test.yaml
```diff
@@ -253,8 +253,9 @@ info:
 services:
   - name: restbase
     module: hyperswitch
-    conf:
-      port: 7231
+    conf: &rb_conf
```
d00rman: In reality, we need two services in this config; one on port 7231 and the other 7233. They ought to have the same config.

Pchelolo: See comment: 

>  In production we expose both 7231 and 7233 ports from fullstack RB.

I can't do it in here because of fsfe test configuration. 2 processes would use the same port in that case.

d00rman: Ah, missed it! Thnx!

PR #1165, opened 2019-07-10
Add wikibase format to spec

mvolz:
Done in the service here: https://gerrit.wikimedia.org/r/#/c/mediawiki/services/citoid/+/513975/
And deployed last month. 

Bug: T208213

### DIFF
--- a/v1/citoid.yaml
+++ b/v1/citoid.yaml
@@ -32,6 +32,7 @@ paths:
           - `zotero`: format used by [Zotero](https://www.zotero.org/).
           - `bibtex`: format used in conjunction with LaTeX documents.
             See [bibtex.org](http://www.bibtex.org/).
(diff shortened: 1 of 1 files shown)


mvolz:
Any idea why this isn't building?

clarakosi:
This appears to be an npm/Travis issue. Adding @Pchelolo as a reviewer and will look more into it next week.

Pchelolo:
Ok. So myself and @clarakosi have discovered why the build was failing. Totally unrelated to this PR. Long story short, certain npm versions contained a bug that made npm publish the `.git` directory in the repo, which didn't break `npm install` in a clean repo, but if you try to `npm install` some more packages into a folder with already installed packages it broke. We have fixed it, but travis is still failing, which means it might be caching something or something? It's not quite clear to me.

The PR looks good, but in the purpose of scientific discovery, I will close it and open a new one from the same branch and fork.

PR #1166, opened 2019-07-17
Switch event destination to eventgate.

Pchelolo:
Depends on a config change https://gerrit.wikimedia.org/r/c/mediawiki/services/restbase/deploy/+/524058

Bug: T226522


d00rman:
Since this commit came after the release of v1, we should have kept compat with the old-style events.

PR #1170, opened 2019-07-24
Only expose unique values for the operationId.

Pchelolo:
We have to only expose unique operationIds because duplicates break swagger-ui autogenerated docs.


eevans:
LGTM :+1: (Travis tests notwithstanding)

PR #1179, opened 2019-08-23
Mechanism for storing and retrieving Parsoid HTML from JS and PHP

hknustwmf:
It may look like a lot of changes, but there are actually only a few. The key pieces are parsoid.js and the new test variant_config.js. The transform.js test battery is now executed twice -- once for each variant. The tests themselves did not change except for the headers.

### DIFF
--- a/config.fullstack.test.yaml
+++ b/config.fullstack.test.yaml
@@ -8,6 +8,7 @@ default_project: &default_project
                 options: &default_options
                   parsoid:
                     host: https://parsoid-beta.wmflabs.org
+                    host_php: https://parsoid-beta.wmflabs.org
                     grace_ttl: 1000000
                   action:
                     apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
--- a/projects/sys/default.wmf.yaml
+++ b/projects/sys/default.wmf.yaml
@@ -11,9 +11,13 @@ paths:
       - path: sys/page_save.js
   /parsoid:
     x-modules:
-      - path: sys/parsoid.js
-        options:
+      - path: sys/parsoid_proxy.js
+  /parsoidjs:
+    x-modules:
+      - path: sys/parsoid.js 
+        options: &parsoidopts
           parsoidHost: '{{options.parsoid.host}}'
+          bucketName: 'parsoid'
           response_cache_control: '{{options.purged_cache_control}}'
           grace_ttl: '{{default(options.parsoid.grace_ttl, 86400)}}'
           # A list of pages that we don't currently want to re-render on
@@ -116,6 +120,13 @@ paths:
               - '/Commons:Undeletion_requests\//'
               - '/Commons:WikiProject_Aviation\/recent_uploads\//'
               - '/^(?:User|Benutzer):/'
+  /parsoidphp:
+    x-modules:
+      - path: sys/parsoid.js
+        options:
+          <<: *parsoidopts
+          parsoidHost: '{{options.parsoid.host_php}}'
+          bucketName: 'parsoidphp'
   /events:
     x-modules:
       - path: sys/events.js
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -54,7 +54,7 @@ function extractTidMeta(html) {
  *  in `if-unmodified-since` header of the request
  * @param  {Object} req the request
  * @param  {Object} res the response
- * @return {boolean}    true if content has beed modified
+ * @return {boolean}    true if content has been modified
  */
 function isModifiedSince(req, res) {
     try {
@@ -125,10 +125,11 @@ class ParsoidService {
     _initOpts(opts = {}) {
         this.options = opts;
         this.parsoidHost = opts.parsoidHost;
+        this.bucketName = opts.bucketName || 'parsoid';
         this.options.stash_ratelimit = opts.stash_ratelimit || 5;
         this.options.grace_ttl = opts.grace_ttl || 86400;
         this._blacklist = compileReRenderBlacklist(opts.rerenderBlacklist);
-        if (!opts.parsoidHost) {
+        if (!this.parsoidHost) {
             throw new Error('Parsoid module: the option parsoidHost must be provided!');
         }
     }
@@ -170,7 +171,7 @@ class ParsoidService {
      */
     getLatestBucketURI(domain, title) {
         return new URI([
-            domain, 'sys', 'key_value', 'parsoid', title
+            domain, 'sys', 'key_value', this.bucketName, title
         ]);
     }
 
@@ -186,7 +187,7 @@ class ParsoidService {
      */
     getStashBucketURI(domain, title, revision, tid) {
         return new URI([
-            domain, 'sys', 'key_value', 'parsoid-stash', `${title}:${revision}:${tid}`
+            domain, 'sys', 'key_value', `${this.bucketName}-stash`, `${title}:${revision}:${tid}`
         ]);
     }
 
@@ -799,7 +800,7 @@ module.exports = (options) => {
         // Dynamic resource dependencies, specific to implementation
         resources: [
             {
-                uri: '/{domain}/sys/key_value/parsoid',
+                uri: `/{domain}/sys/key_value/${ps.bucketName}`,
                 headers: {
                     'content-type': 'application/json'
                 },
@@ -808,7 +809,7 @@ module.exports = (options) => {
                 }
             },
             {
-                uri: '/{domain}/sys/key_value/parsoid-stash',
+                uri: `/{domain}/sys/key_value/${ps.bucketName}-stash`,
                 headers: {
                     'content-type': 'application/json'
                 },
--- a/sys/parsoid_proxy.js
+++ b/sys/parsoid_proxy.js
@@ -1,0 +1,48 @@
+'use strict';
+
+/*
+ * Simple proxy to route requests to the client-requested
+ * Parsoid variant (JS or PHP) during the transition
+ * period. Parsoid/JS is being phased out and replaced by
+ * Parsoid/PHP.
+ */
+
+const HyperSwitch = require('hyperswitch');
+const URI = HyperSwitch.URI;
+const mwUtil = require('../lib/mwUtil');
+const VARIANT_HDR_NAME = 'x-parsoid-variant';
+
+module.exports = () => {
+    return {
+        spec: {
+            paths: {
+                '/{+path}': {
+                    all: {
+                        operationId: 'proxy_parsoid_variant'
+                    }
+                }
+            }
+        },
+        operations: {
+            proxy_parsoid_variant: (hyper, req) => {
+                const rootReqHeaders = hyper._rootReq.headers || {};
+                rootReqHeaders[VARIANT_HDR_NAME] = rootReqHeaders[VARIANT_HDR_NAME] || 'JS';
+                const isPhpVariant = /PHP/i.test(rootReqHeaders[VARIANT_HDR_NAME]);
+
+                return hyper.request({
+                    method: req.method,
+                    uri: new URI(req.uri.toString().replace('/parsoid/',
+                        isPhpVariant ? '/parsoidphp/' : '/parsoidjs/')),
+                    headers: req.headers,
+                    body: req.body,
+                    query: req.query
+                })
+                .tap((res) => {
+                    res.headers = res.headers || {};
+                    res.headers[VARIANT_HDR_NAME] = rootReqHeaders[VARIANT_HDR_NAME];
+                    mwUtil.addVaryHeader(res, VARIANT_HDR_NAME);
+                });
+            }
+        }
+    };
+};
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -52,8 +52,10 @@ services:
                     # XXX Check the base RESTBase URI
                     baseUriTemplate: "{{'http://{domain}:7231/{domain}/v1'}}"
                   parsoid:
-                    # XXX Check Parsoid URL!
+                    # XXX Check Parsoid/JS URL!
                     host: http://localhost:8142
+                    # XXX Check Parsoid/PHP URL!
+                    host_php: http://localhost:8142
                   table:
                     backend: sqlite
                     dbname: db.sqlite3
--- a/config.frontend.test.yaml
+++ b/config.frontend.test.yaml
@@ -9,6 +9,7 @@ default_project: &default_project
                 options: &default_options
                   parsoid:
                     host: https://parsoid-beta.wmflabs.org
+                    host_php: https://parsoid-beta.wmflabs.org
                     grace_ttl: 1000000
                   action:
                     apiUriTemplate: "{{'https://{domain}/w/api.php'}}"
--- a/test/features/parsoid/php_variant.js
+++ b/test/features/parsoid/php_variant.js
@@ -1,0 +1,71 @@
+'use strict';
+
+const assert = require('../../utils/assert.js');
+const Server = require('../../utils/server.js');
+const preq   = require('preq');
+
+const testPage = {
+    title: 'User:Pchelolo%2fRestbase_Test',
+    revision: '275854',
+    wikitext: '<div id=bar>Selser test'
+    // html is fetched dynamically
+};
+
+describe('php parsoid variant', function() {
+    this.timeout(20000);
+    let contentTypes;
+    const server = new Server();
+    before(() => {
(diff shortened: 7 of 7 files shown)


Pchelolo:
I haven't looked in any of the details, but I strongly disagree with the implementation approach taken here.
 
Touching parsoid.js in this manner was never intended and doing so creates so much more complexity... Here's what needs to be done in my opinion:

1. We need to mount `sys/parsoid.js` in project twice under 2 separate paths with 2 separate options - one for JS and one for PHP
2. Instead of mounting `/v1/parsoid.yaml` we need to write a little JS proxy module that would look at the cookie and decide which parsoid version (PHP vs JS) to use, call it, and add a Vary header in the end - that way all the code related to this transition will be confined to a single place and not spread around the codebase. This change will be removed after full transition anyway, so we want it to be as self-confined as possible
3. I don't understand the purpose of one more copy of the config. What we need to do is to have tests running for JS parsoid-only, for PHP parsoid-only and maybe one or two tests verifying that actual switching works. For the purpose of the first two, we can add support in preq to set additional headers based on an environment variable and add one more dimention in the travis testing matrix. For the second - we just write a couple of new tests in a separate suite.

d00rman:
> Touching parsoid.js in this manner was never intended and doing so creates so much more complexity... Here's what needs to be done in my opinion:

I agree. Too many places in `parsoid.js` have been touched just for the transition period. I agree with @Pchelolo on the course of action outlined below. Some comments follow.

> 
> 1. We need to mount `sys/parsoid.js` in project twice under 2 separate paths with 2 separate options - one for JS and one for PHP
> 2. Instead of mounting `/v1/parsoid.yaml` we need to write a little JS proxy module that would look at the cookie and decide which parsoid version (PHP vs JS) to use, call it, and add a Vary header in the end - that way all the code related to this transition will be confined to a single place and not spread around the codebase. This change will be removed after full transition anyway, so we want it to be as self-confined as possible

I agree this is the way to go, but there are some caveats. These could be mounted onto `/sys/parsoid/` and `/sys/parsoidphp/`. In the mount options, then, one would get the Parsoid/JS host supplied for `parsoidHost`, while the other would get the Parsoid/PHP host. Additionally, each would need an extra option passed to it, `bucket_name` (or similar) that instructs it which buckets to use (the values are `parsoid` and `parsoidphp` respectively).

> 3. I don't understand the purpose of one more copy of the config. What we need to do is to have tests running for JS parsoid-only, for PHP parsoid-only and maybe one or two tests verifying that actual switching works. For the purpose of the first two, we can add support in preq to set additional headers based on an environment variable and add one more dimention in the travis testing matrix. For the second - we just write a couple of new tests in a separate suite.

I think we can tackle the tests after the first two points have been completed (but still in this PR).

hknustwmf:
> Needs to be reworked as per above discussion.

I added the proxy and a couple basic tests.

projects/sys/default.wmf.yaml
```diff
@@ -10,10 +10,124 @@ paths:
...
+    x-modules:
+      - path: sys/parsoid_proxy.js
+        options:
+          parsoidPhpHost: '{{options.parsoid.host_php}}'
+  /parsoidjs:
     x-modules:
       - path: sys/parsoid.js
         options:
           parsoidHost: '{{options.parsoid.host}}'
+          bucketName: 'parsoid'
```
d00rman: Instead of copy/pasting the options, it's better to use YAML references. For `/parsoidjs`, add an anchor to options:

```yaml
options: &parsoidopts
```

And then reuse them below for `/parsoidphp` back-filling the differences:

```yaml
options:
  <<: *parsoidopts
  bucketName: parsoidphp
  parsoidHost: '{{options.parsoid_php_host}}'
```

config.fullstack.test.yaml
```diff
@@ -6,8 +6,10 @@ default_project: &default_project
             x-modules:
               - path: projects/v1/default.wmf.yaml
                 options: &default_options
+                  block_external_reqs: true
```
d00rman: This shouldn't be here

sys/parsoid.js
```diff
@@ -816,6 +817,25 @@ module.exports = (options) => {
                     valueType: 'blob',
                     default_time_to_live: options.grace_ttl
                 }
+            },
+            {
```
d00rman: It would be better if the parsoid module exported only one set of resources - those needed by it. By having both js and php exporting both variants we double the number of requests to Cassandra, and that's not good.

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+/*
+ * Simple proxy to route requests to the client-requested
+ * Parsoid variant (JS or PHP) during the transistion
```
d00rman: `s/transistion/transition/`

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+ * persiod. Parsoid/JS is being phased out and replaced by
```
d00rman: `s/persiod/period/`

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+
+module.exports = (options) => {
+    options = options || {};
+    const phpHost = options.parsoidPhpHost;
+    const VARIANT_HDR_NAME = 'x-parsoid-variant';
+    return {
+        spec: {
+            paths: {
+                '/{+path}': {
+                    'x-route-filters': options.block_external_reqs ?
```
d00rman: Not needed

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+                            type: 'default',
+                            name: 'header_match',
+                            options: {
+                                whitelist: {
+                                    'x-client-ip': ['/^(?:::1)|(?:::ffff:)?(?:10|127)\\./']
+                                }
+                            }
+                        }] : [],
+                    all: {
+                        operationId: 'proxy',
```
d00rman: You need to change the operation id, as that one already exists

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+                        'x-monitor': false,
+                        'x-hidden': true
```
d00rman: These two `x-*` stanzas above are not needed as `/sys` routes are not visible nor monitored.

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+            }
+        },
+        operations: {
+            proxy: (hyper, req) => {
+                const rootReqHeaders = hyper._rootReq.headers;
+                if (!rootReqHeaders[VARIANT_HDR_NAME]) {
+                    rootReqHeaders[VARIANT_HDR_NAME] = 'JS';
+                }
+                const isPhpVariant = /PHP/i.test(rootReqHeaders[VARIANT_HDR_NAME]);
+                if (isPhpVariant && !phpHost) {
```
d00rman: This is superfluous here because RB will fail to load if both modules don't get their respective host option.

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+const HyperSwitch = require('hyperswitch');
+const URI = HyperSwitch.URI;
+const HTTPError = HyperSwitch.HTTPError;
+const mwUtil = require('../lib/mwUtil');
+
+module.exports = (options) => {
+    options = options || {};
+    const phpHost = options.parsoidPhpHost;
```
d00rman: Not needed.

projects/sys/default.wmf.yaml
```diff
@@ -10,10 +10,124 @@ paths:
     x-modules:
       - path: sys/page_save.js
   /parsoid:
+    x-modules:
+      - path: sys/parsoid_proxy.js
+        options:
```
d00rman: These options are not needed.

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+                        operationId: 'proxy',
+                        'x-monitor': false,
+                        'x-hidden': true
+                    }
+                }
+            }
+        },
+        operations: {
+            proxy: (hyper, req) => {
+                const rootReqHeaders = hyper._rootReq.headers;
```
d00rman: Add: `|| {}` to avoid unlikely cases when the headers are not defined.

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+                    method: req.method,
+                    uri: newUri,
+                    headers: req.headers,
+                    body: req.body,
+                    query: req.query,
+                    params: req.params
+                })
+                .then((res) => {
+                    if (res && res.headers) {
+                        if (res.headers['content-type'] && res.headers['content-type'].startsWith('application/json') && Buffer.isBuffer(res.body)) {
```
d00rman: No, you shouldn't mess with the body, leave it as-is. Remove this if-block.

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+                return hyper.request({
+                    method: req.method,
+                    uri: newUri,
+                    headers: req.headers,
+                    body: req.body,
+                    query: req.query,
+                    params: req.params
+                })
+                .then((res) => {
+                    if (res && res.headers) {
```
d00rman: You should only check if `res` is defined here. If `res.headers` is not defined, it should be added.

projects/sys/default.wmf.yaml
hknustwmf: Much better :)

config.fullstack.test.yaml
hknustwmf: Removed

sys/parsoid.js
hknustwmf: That's what I had initially but it was throwing exceptions in the tests (due to a bad reference). Fixed.

sys/parsoid_proxy.js
hknustwmf: Oh, boy! VS Code does not come with a spell checker. Adding one. :)

sys/parsoid_proxy.js
hknustwmf: Removed

sys/parsoid_proxy.js
hknustwmf: Added

sys/parsoid_proxy.js
hknustwmf: Removed

sys/parsoid_proxy.js
hknustwmf: Added

projects/sys/default.wmf.yaml
hknustwmf: Removed

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,54 @@
...
+                    }
+                }
+            }
+        },
+        operations: {
+            proxy_parsoid_variant: (hyper, req) => {
+                const rootReqHeaders = hyper._rootReq.headers || {};
+                if (!rootReqHeaders[VARIANT_HDR_NAME]) {
+                    rootReqHeaders[VARIANT_HDR_NAME] = 'JS';
+                }
```
Pchelolo: Using `||` is shorter and nicer then a full if.

hknustwmf: Agreed. Changed it.

projects/example.yaml
```diff
@@ -52,11 +52,226 @@ paths:
               x-modules:
                 - path: sys/page_save.js
             /parsoid:
+              x-modules:
```
Pchelolo: In light of #1180 you can keep this file intact.

projects/example.yaml
```diff
@@ -52,11 +52,226 @@ paths:
...
               x-modules:
                 - path: sys/parsoid.js
                   options:
                     parsoidHost: '{{options.parsoid.host}}'
                     response_cache_control: '{{options.purged_cache_control}}'
                     grace_ttl: '{{default(options.parsoid.grace_ttl, 86400)}}'
+                    # A list of pages that we don't currently want to re-render on
+                    # each edit. Most of these are huge bot-edited pages, which are
+                    # rarely viewed in any case.
+                    rerenderBlacklist:
```
Pchelolo: Why did you copy-paste this here?

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,52 @@
...
+                const isPhpVariant = /PHP/i.test(rootReqHeaders[VARIANT_HDR_NAME]);
+
+                return hyper.request({
+                    method: req.method,
+                    uri: new URI(req.uri.toString().replace('/parsoid/',
+                        isPhpVariant ? '/parsoidphp/' : '/parsoidjs/')),
+                    headers: req.headers,
+                    body: req.body,
+                    query: req.query,
+                    params: req.params
```
Pchelolo: I don't think you need to pass through 'params', but needs testing.

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,52 @@
...
+                })
+                .then((res) => {
```
Pchelolo: 1. can be replaced with `tap`
2. if 'res' is falsy, we're in trouble... Maybe no need to check for it to be truthy

sys/parsoid_proxy.js
```diff
@@ -0,0 +1,52 @@
...
+                    if (res) {
+                        res.headers = res.headers || {};
+                        res.headers[VARIANT_HDR_NAME] = rootReqHeaders[VARIANT_HDR_NAME];
```
Pchelolo: I'm not quite sure this is a right behavior that wouldn't accidentally mask issues. We `assume` we've routed the request to correct Parsoid, but who knows? Why would we set the response header for parsed variant here??

Pchelolo:
A couple of issues left, but you're close. 

One thing I'm wondering about is testing. With the current patch we will be testing PHP parsoid compatibility with exactly 1 test. We need more. We need to have an env variable that would set a parsoid variant header in preq for all requests, (I believe we'd need to support this in preq, but maybe `request` already supports something like that. Needs to be checked). Then we need to add 2 versions of this variable to the test matrix in travis.yaml - that would allow us to run all the tests with both parsoid versions and be sure we're compatible with both.

sys/parsoid_proxy.js
hknustwmf: It's in the ticket: "Furthermore, RESTBase must include this header in the response, ..."  https://phabricator.wikimedia.org/T230791

projects/example.yaml
hknustwmf: Not intentionally. Removed.

sys/parsoid_proxy.js
hknustwmf: You are correct. Not needed.

projects/example.yaml
hknustwmf: Reversed the changes.

sys/parsoid_proxy.js
hknustwmf: 1. Replace with tap. 2. Removed

sys/parsoid_proxy.js
hknustwmf: Ideally, the variant header would be added by the Parsoid instances instead of Restbase

Pchelolo: Yeah, that's what I meant. Or, the content-type would be different. Would you mind filing a ticket, something like 'Parsoid JS/PHP should indicate which Parsoid implementation the page was rendered with'?

Pchelolo:
1. This needs to be manually rebased.
2. Now that we have parsoid in beta accessible, we can pick up the work here again. See T231569 for reference regarding parsoid in beta cluster.

There's a bunch of things to be done to be able to run CI and local testing agains Parsoid-PHP. We can't access production wikis on beta PHP Parsoid, see T231933 so we need to convert all the tests to only talk to beta wikis. I think we can work on that in separate PRs.

hknustwmf:
Based on the discussion with Giuseppe yesterday, let's also think/about create tasks to remove external dependencies from integration tests.

Pchelolo:
> Based on the discussion with Giuseppe yesterday, let's also think about/create tasks to remove external dependencies from integration tests.

Creating tasks makes sense, but actually blocking this on them is a bit too ambitious in my opinion.

hknustwmf:
It was meant a suggestion not a blocker

d00rman:
Superseeded by #1207

PR #1180, opened 2019-08-30
Simplify config.example.yaml to avoif using projects

Pchelolo:
I think having an example config not using the projects is simpler for small installs and developers.


Pchelolo:
I'm quite indifferent :) Your proposal makes it a couple lines longer, but I guess it shows more how combining different sub-trees works... mmmm... yeah. I'll do it.

PR #1181, opened 2019-09-03
MCS: Stop using storage for wiktionary definitions

d00rman:
Bug: [T231361](https://phabricator.wikimedia.org/T231361)

### DIFF
--- a/v1/definition.yaml
+++ b/v1/definition.yaml
@@ -86,61 +86,11 @@ paths:
             application/problem+json:
               schema:
                 $ref: '#/components/schemas/problem'
-      x-setup-handler:
-        # Set up a simple key-value bucket.
-        - init:
-            method: put
-            uri: /{domain}/sys/key_value/term.definition-ng
-            headers:
-              content-type: application/json
-            body:
-              valueType: json
       x-request-handler:
-        - storage:
-            request:
-              method: get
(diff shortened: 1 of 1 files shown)


v1/definition.yaml
```diff
@@ -150,35 +100,7 @@ paths:
                 content-type: '{{extract.headers.content-type}}'
                 vary: '{{extract.headers.vary}}'
               body: '{{extract.body}}'
-        - emit_change_event:
```
Pchelolo: The `response_cache_control` here will still be pretty long. We need to either return some kind of purging or make much smaller cache max-age

d00rman: In [the CP config patch](https://gerrit.wikimedia.org/r/534122) we keep the purging side of things, so this will not be an issue.

Pchelolo: No. If you don't emit the resource-change for this route, no purge will be emitted.

d00rman: Indeed, I'll add a purge event to the CP config, that ought to fix it.

d00rman: @Pchelolo updated the CP config patch to reflect what we discussed here

Pchelolo:
+1, waiting for travis

PR #1184, opened 2019-09-18
Update appservice.wmflabs.org references to mobileapps.wmflabs.org

mdholloway:
Move dev/testing traffice away from the legacy appservice labs instance,
which is being replaced by mobileapps.wmflabs.org (a web proxy for
mobileapps.mobile.eqiad.wmflabs).

Bug: https://phabricator.wikimedia.org/T225837


d00rman:
LGTM, we should be able to move on this early next week.

PR #1185, opened 2019-09-19
Parsoid: Remove the TID suffix, if present

d00rman:
As per RfC7232, ETags can be suffixed depending on the content type sent. Normally, the suffix should not be present in the If-Match header sent by the client. Alas, in certain circumstances (yet to be exactly pinned down) VE sends us an If-Match header that is suffixed by `-df` (presumably indicating deflate-compressed content, which is not the case), causing RESTBase to throw 404s for content that is in reality available in storage.

This has been happening for 15 days now with at least 2000 occurrences per day, which is problematic since users are unable to save their modifications because of it. To work around this problem, RESTBase now strips the suffix and complains only by logging the event.

Bug: [T230272](https://phabricator.wikimedia.org/T230272)


d00rman:
Self-merging as it's a UBN

PR #1196, opened 2019-10-07
Add .pipeline/config.yaml with publish stage

rakugama:
Adds a basic pipeline configuration with a publish stage for development
images to be used in local-charts (and eventually elsewhere, probably).

Bug: T234580

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 restbase
 coverage
 config.yaml
+!.pipeline/config.yaml
 node_modules
 npm-debug.log
 package-lock.json
--- a/.pipeline/config.yaml
+++ b/.pipeline/config.yaml
@@ -1,0 +1,9 @@
+pipelines:
+  publish:
+    blubberfile: blubber.yaml
+    stages:
+      - name: dev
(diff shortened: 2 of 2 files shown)


d00rman:
@rakugama we already [have a trigger in `integration/config`](https://github.com/wikimedia/integration-config/blob/29d3ff36039c2b57a723b3bdeaa84fe93895e159/zuul/layout.yaml#L7151) for building images in production. Since this would be essentially the same image, is this config needed? Also, if this PR is merged, would that invalidate building the image used in production by the RESTRouter chart?

rakugama:
@d00rman It shouldn't invalidate the building of the production image, but I'm having trouble figuring out how/when the production image is being published since merges happen on github instead of gerrit and looking at the registry it doesn't seem like a publish is happening after every merge (is that correct?). We'd like to publish a dev image for every merge to master. We could discuss during the meeting on Thurday if that would be easier.

rakugama:
Closing due to existing trigger for image publishing and because it's not feasible to publish an image for every merge with the current setup.

PR #1200, opened 2019-10-09
Parsoid: Use the ETag's revid to stash the content under

d00rman:
Clients may want to stash the contents of a document regardless of
whether they provide the revid in the URI or not. If they do, then the
contents being stashed needs to match that revid (ensured by
`generateAndSave`), but if they don't then we have to use the revid
present in the ETag. They either have to match or we must use the one in
the ETag as the client has not supplied it. Therefore, always use the
one found in the response's ETag.

Furthermore, ensure the returned ETag is the one we stash the data under.
We use the HTML's response's ETag to stash the data, but return the
overall response's ETag. While they are equal in most cases, ensure that
the proper ETag is returned without counting on the magic of them being
equal.

Also, in case we don't have an ETag to stash and return, error out as
it's better to fail in that case than create a new one since that means
that either we have incomplete data or that Parsoid did not return a
proper response (or both).

Bug: [T234928](https://phabricator.wikimedia.org/T234928)

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -218,14 +218,14 @@ class ParsoidService {
         const rp = req.params;
         const dataParsoidResponse = parsoidResp.body['data-parsoid'];
         const htmlResponse = parsoidResp.body.html;
-        const tid = mwUtil.parseETag(parsoidResp.headers.etag).tid;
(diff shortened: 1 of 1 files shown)


d00rman:
Self-merging as I've tested this locally extensively and want to deploy ASAP as this potentially fixes the problems we are seeing in production.

PR #1206, opened 2019-10-11
PCS: Do not use storage for /page/metadata; release v1.1.4

d00rman:
We will not be pre-generating content for it any more, so bypass the
storage and always hit PCS directly.

Bug: [T235173](https://phabricator.wikimedia.org/T235173)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "restbase",
-  "version": "1.1.3",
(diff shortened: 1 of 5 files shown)


d00rman:
Published as [v1.1.4](https://github.com/wikimedia/restbase/tree/v1.1.4)

PR #1207, opened 2019-10-13
Parsoid: Add the variant proxy

d00rman:
This PR enables RESTBase to communicate with both Parsoid/JS and Parsoid/PHP and use separate storage for each of them. It introduces three `sys` modules: `parsoid-js`, `parsoid-php` and `parsoid`, which handle the JS variant, PHP variant and the *main* Parsoid module acting as the proxy, respectively.

The common Parsoid business logic has been extracted into `lib/parsoid.js`, while `sys/parsoid-js` and `sys/parsoid-php` only deal with variant-specific details: issuing requests to Parsoid and determining the buckets to use when storing/retrieving content. This code extraction and reshuffling was done for the following reasons:
- to have a clean separation between Parsoid/JS and Parsoid/PHP;
- to allow for their interchangeability, so that in case of problems it is easy to revert to having only one of them active;
- to have a *clean* code situation once the transition has been completed; and
- to avoid introducing any new permanent config parameters that are not used during the transition (i.e. users can choose to use one or the other without needing to adjust the config).

The proxy allows directing requests to either variant. It loads both variants' modules internally and uses their operations to complete requests. It is designed in such a way so as to allow an easy transition between fully using JS to fully using PHP with no config changes. When first introduced, its defaults emulate the JS-only scenario. Once the switch is fully achieved, then simply changing `sys/parsoid.js` for `sys/parsoid-php.js` in `projects/sys/default.wmf.yaml` with no config change results in having a fully-functional Parsoid/PHP module. The proxy can. thus, function properly with only one of variant modules loaded and configured.

In order to support the transition period, the proxy has three modes of operation: single, mirror and split. In single mode, only one variant is used, defined by the `default_variant` configuration value, defaulting to `js`. This allows us to start using the proxy with no config changes. In the final stages of the transition (before we remove the proxy), it can be changed to `php` to only use the PHP variant. The mirror mode is used to asynchronously mirror traffic to the PHP variant. Requests are issued to both variants, but only the JS one is returned. The amount of traffic to be mirrored can be tuned with the `percentage` configuration parameter. The important caveat here is that only requests for `/page/{format}` end points are mirrored - we cannot do so reliably for transforms since they rely on stashed content, which is likely not to be available for the PHP variant. Furthermore, when the proxy is configured in mirror mode, dependency update events are emitted only for the JS variant, so as to avoid duplicates. Finally, the split mode is used to split the traffic between the two variants based on the request domain. If one of the patterns given in the `pattern` configuration parameter is matched, then the variant not defined in `default_variant` is used, otherwise the default one is used. This mode supports the second stage of the transition, where JS will be authoritative for the majority of domains, while we will be slowly moving projects one by one (or group by group) over to using Parsoid/PHP. Because of this gradual switch, there is also a built-in fall-back for transform end points in this mode: if a request for the non-default variant 404s, then we retry the request using the default variant in order to cover the clients that are in the middle of an editing session during the switch.

Apart from these modes, the proxy also supports clients directly telling it which variant to use. If the incoming request has the `PARSOID_VARIANT` cookie or the `X-Parsoid-Variant` header set, then the request is sent directly to that variant regardless of the proxy's mode. When deciding where to send the request, the proxy gives precedence to the header in case both are set.

Bug: [T230791](https://phabricator.wikimedia.org/T230791)


d00rman:
Pro tip: looking at the diff of the whole PR is a bit confusing because of the code move, but skipping the first commit and looking at the diff of the other commits paints the real picture. The first commit only moves the ParsoidService class to lib/parsoid.js without any changes to it.

d00rman:
Tests are failing due to [T235375](https://phabricator.wikimedia.org/T235375)

d00rman:
> Tests are failing due to [T235375](https://phabricator.wikimedia.org/T235375)

This has now morphed into [T235478](https://phabricator.wikimedia.org/T235478)

d00rman:
> Oh gosh holy cow this is not a nice change to review :)

Hehe, yeah, sorry about that @Pchelolo . But! You can simply look at [this diff](https://github.com/wikimedia/restbase/pull/1207/files/69ae00d572b0997582a8423726f74eab4a98125e..36d1d4ace75caae776157b75952295fa69513c31) which only holds the actual code changes... Kind of :P

d00rman:
> > Oh gosh holy cow this is not a nice change to review :)
> 
> Hehe, yeah, sorry about that @Pchelolo . But! You can simply look at [this diff](https://github.com/wikimedia/restbase/pull/1207/files/69ae00d572b0997582a8423726f74eab4a98125e..36d1d4ace75caae776157b75952295fa69513c31) which only holds the actual code changes... Kind of :P

Oops, sorry @Pchelolo wrong link. [This diff](https://github.com/wikimedia/restbase/pull/1207/files/bfa9fadae42ddf8303ba09613dfbea96db2a9420..36d1d4ace75caae776157b75952295fa69513c31) is authoritative :)

lib/parsoid.js
```diff
@@ -104,12 +104,15 @@ class ParsoidService {
...
-        this.options.grace_ttl = opts.grace_ttl || 86400;
         this._blacklist = compileReRenderBlacklist(opts.rerenderBlacklist);
-        if (!opts.parsoidHost) {
-            throw new Error('Parsoid module: the option parsoidHost must be provided!');
+        if (!this.parsoidUri) {
+            throw new Error('Parsoid module: the option host must be provided!');
+        }
+        // remove the trailing slash, if any
+        if (this.parsoidUri.slice(-1) === '/') {
+            this.parsoidUri = this.parsoidUri.slice(0, -1);
```
Pchelolo: Why?

lib/parsoid.js
```diff
@@ -142,34 +145,6 @@ class ParsoidService {
         }
     }
 
-    /**
-     * Get the URI of a bucket for the latest Parsoid content.
-     * @param {string} domain the domain name.
-     * @param {string} title the article title.
-     * @return {HyperSwitch.URI}
-     */
-    getLatestBucketURI(domain, title) {
```
Pchelolo: Let's leave these in and throw some AssertionError or smth

lib/parsoid.js
```diff
@@ -142,34 +145,6 @@ class ParsoidService {
...
-     * Get the URI of a bucket for stashing Parsoid content. Used both for stashing
-     * original HTML/Data-Parsoid for normal edits as well as for stashing transforms.
-     *
-     * @param {string} domain the domain name.
-     * @param {string} title the article title.
-     * @param {number} revision the revision of the article.
-     * @param {string} tid the TID of the content.
-     * @return {HyperSwitch.URI}
-     */
-    getStashBucketURI(domain, title, revision, tid) {
```
Pchelolo: Idem

lib/parsoid.js
```diff
@@ -395,7 +368,10 @@ class ParsoidService {
                     }
                 })
                 .then(() => {
-                    const dependencyUpdate = _dependenciesUpdate(hyper, req, newContent);
+                    let dependencyUpdate = P.resolve();
+                    if (!this.options.skip_updates) {
```
Pchelolo: This is unrelated I believe, but huge kudos for finding this :)

lib/parsoid.js
d00rman: Because, as the comment two lines above says, we want to get rid of the last slash in case it is provided since we count on it not being there in `_getParsoidReq()`.

d00rman:
@Pchelolo added the empty methods to `lib/parsoid.js` including `_getParsoidReq()`. I used a generic `Error`, though, because `AssertionError` is not available in Node 6.

d00rman:
So the new diff to look at is [here](https://github.com/wikimedia/restbase/pull/1207/files/bfa9fadae42ddf8303ba09613dfbea96db2a9420..14ea99577b7a061bdd51fa08c0b595c183b4aa99) :)

Pchelolo:
LGTM. We need to wait till the varnish issue is resolved. 

We can bike shed about the proxy complexity, but since it's temporary, I'm happy with whatever you're happy with.

d00rman:
> LGTM. We need to wait till the varnish issue is resolved.

Agreed.

> We can bike shed about the proxy complexity, but since it's temporary, I'm happy with whatever you're happy with.

Yup, I concur here too. But I went for a more complex approach on purpose so as to make all of the transitions easier with less moving pieces during the actual process of transition.

d00rman:
Ported over the fix for #1208 as well.

PR #1208, opened 2019-10-15
Parsoid: Stash: Honour our own ETag when retrieving the stash

d00rman:
If a client is switching between wikitext and HTML and in one of the two
transforms they fail to provide the revision in the URI, trust that the
ETag is correct and try to retrieve the stashed content using the
information in the ETag. As a fall-back, though, try with the
client-supplied revision as well.

Bug: [T235465](https://phabricator.wikimedia.org/T235465)

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -596,8 +596,24 @@ class ParsoidService {
             contentPromise = P.resolve(undefined);
         } else {
             if (etag && etag.suffix === 'stash' && from === 'html' && to === 'wikitext') {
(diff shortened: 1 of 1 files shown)


Pchelolo:
Something seemingly unrelated is breaking the tests!

d00rman:
> Something seemingly unrelated is breaking the tests!

Random Cassandra failures. I've encountered these a number of times over the last period. Only the last set of tests is failing with Cassandra not responding. Restarted them. Locally all is good, ofc.

PR #1211, opened 2019-10-30
Only mirror no-cache requests to Parsoid-PHP and not store results.

Pchelolo:
Bug: https://phabricator.wikimedia.org/T236838

cc @subbuss


Pchelolo:
Ok, travis just doesn't work today...

d00rman:
I think both of these changes (mirror only no-cache reqs, do not store) should have been config options for the proxy, since eventually we want to send everything to parsoid/php and store the result as well

PR #1212, opened 2019-10-30
Parsoid-PHP: Don't pass through all the request headers.

Pchelolo:
Bug: T236837

### DIFF
--- a/sys/parsoid-php.js
+++ b/sys/parsoid-php.js
@@ -21,7 +21,7 @@ class ParsoidPHP extends Parsoid {
     _getParsoidReq(req, path, headers, body) {
         return {
             uri: new URI(`${this.parsoidUri}/${req.params.domain}/v3/${path}`),
-            headers: Object.assign({ host: req.params.domain }, headers || req.headers),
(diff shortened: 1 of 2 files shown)


sys/parsoid-php.js
```diff
@@ -21,7 +21,7 @@ class ParsoidPHP extends Parsoid {
     _getParsoidReq(req, path, headers, body) {
         return {
             uri: new URI(`${this.parsoidUri}/${req.params.domain}/v3/${path}`),
-            headers: Object.assign({ host: req.params.domain }, headers || req.headers),
+            headers: Object.assign({ host: req.params.domain }, headers || {}),
```
d00rman: Euh not sure this is entirely correct. If one wants to avoid sending any headers, they should simply pass in `{}` for the headers params, don't they?

Pchelolo: > Euh not sure this is entirely correct. If one wants to avoid sending any headers, they should simply pass in `{}` for the headers params, don't they?

Yes, true, but there's 0 reason to pass through all the request headers, so I did it this way.

PR #1213, opened 2019-11-06
Parsoid: Use fall-back for 421s too

d00rman:
Parsoid/PHP will emit 421s if we provide it with data-parsoid belonging
to Parsoid/JS. While this should never happen, ensure that the fall-back
is in place to keep our sanity.

Bug: [T237563](https://phabricator.wikimedia.org/T237563)

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -157,7 +157,11 @@ class ParsoidProxy {
             res.headers = res.headers || {};
             res.headers['x-parsoid-variant'] = variant;
             return P.resolve(res);
-        }).catch({ status: 404 }, (e) => {
+        }).catch({ status: 404 }, { status: 421 }, (e) => {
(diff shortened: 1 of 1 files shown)


sys/parsoid.js
```diff
@@ -157,7 +157,7 @@ class ParsoidProxy {
             res.headers = res.headers || {};
             res.headers['x-parsoid-variant'] = variant;
             return P.resolve(res);
-        }).catch({ status: 404 }, (e) => {
+        }).catch({ status: 404 }, { status: 421 }, (e) => {
```
Pchelolo: Should we log a warning in case of 421? that way some critical problem woundln go unnoticed.

d00rman: @Pchelolo {{done}}

PR #1214, opened 2019-11-08
Expose pcs endpoints for js and css

joewalsh:
Exposes new `pcs` endpoints, removes now unused `pagelib_body_start` and `pagelib_body_end` endpoints

https://phabricator.wikimedia.org/T237745

### DIFF
--- a/v1/css.yaml
+++ b/v1/css.yaml
@@ -28,6 +28,9 @@ paths:
         The first two are the same regardless of what domain is used.
         For these we suggest meta.wikimedia.org.
 
+        You can still pass pagelib for type, but this is a legacy version of the CSS for 
+        existing app clients.
+
         Stability: [experimental](https://www.mediawiki.org/wiki/API_versioning#Experimental)
       parameters:
         - name: type
--- a/v1/css.yaml
+++ b/v1/css.yaml
@@ -22,8 +22,8 @@ paths:
         Gets common CSS mobile apps need to properly display pages using Page Content Service.
         In most cases all of the types are needed (preferably in this order):
         * base (Common mobile CSS from ResourceLoader)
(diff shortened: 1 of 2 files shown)


v1/css.yaml
```diff
@@ -38,6 +38,7 @@ paths:
             enum:
               - base
               - pagelib
+              - pcs
```
d00rman: You are removing `pagelib` from the description, but not from the enum here. It must be one or the other :)

joewalsh: Fixed - `pagelib` is deprecated but needs to remain available for app clients that still point to that endpoint. I added some clarifying comments in the documentation

joewalsh:
> Also, are we sure nothing is using the end points being removed (in the apps) ?

Yes - the `page_start` and `page_end` endpoints were only in use by `mobile-html` but that changed and the start and end code is now in `pcs`.

The `pagelib` endpoints are still in use by the Android app shuold remain available until the Android app switches to mobile-html and enough users upgrade.

v1/javascript.yaml
```diff
@@ -97,52 +63,7 @@ paths:
...
-          content:
-            application/problem+json:
-              schema:
-                $ref: '#/components/schemas/problem'
-      x-request-handler:
-        - get_from_pcs:
-            request:
-              method: get
-              uri: '{{options.host}}/{domain}/v1/data/javascript/mobile/pagelib_body_end'
+              uri: '{{options.host}}/{domain}/v1/data/javascript/mobile/pagelib'
```
d00rman: `s/pagelib/{type}/`

joewalsh: Fixed

d00rman:
Should be deployed in the next few days.

PR #1219, opened 2019-11-25
Remove duplicate filter definitions

d00rman:
When hyperswitch loads a route, it automatically assigns it the filters
loaded for its hierarchical parent. This can cause problems in cases
where the filter is not idempotent, as is the case of the language
variant filter which performs an extra action (calling Parsoid for
conversion); we then end up chaining the action twice, each using the
result of the previous one.

Bug: [T234266](https://phabricator.wikimedia.org/T234266)

### DIFF
--- a/v1/content.yaml
+++ b/v1/content.yaml
@@ -394,19 +394,7 @@ paths:
       x-monitor: false
 
   /html/{title}/{revision}:
-    x-route-filters: &html_title_revision_filters
-      - path: ./lib/access_check_filter.js
-        options:
-          redirect_cache_control: '{{options.response_cache_control}}'
-          attach_body_to_redirect: true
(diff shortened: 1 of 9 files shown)


Pchelolo:
Oh holy cow. We were too smart at some point implementing the filters feature...

d00rman:
> Oh holy cow. We were too smart at some point implementing the filters feature...

Inheritance is actually a desirable feature given the hierarchy of the paths. The problem is that we didn't take that into account when we were removing paths with optional params and that came around and bit us. The only two filters that are "dangerous" are lang variants and content negotiation, the rest are fine being duplicates functionally.

Note that this list could be further streamlined to increase performance, but I want to get the bug out of the way now and don't want to change too many things just to prune filters.

PR #1229, opened 2019-12-05
Fix accept-language header for /page/related

thesocialdev:
1) Forward headers to summary endpoint to fetch proper displaytitle for
language variant
2) Add test using zh beta
3) Stop is deleting accept-language header and not forwarding it (language_variants_filter.js#12)

Bug: T231609


lib/language_variants_filter.js
```diff
@@ -7,6 +7,8 @@ const mwUtil = require('./mwUtil');
 module.exports = (hyper, req, next, options, specInfo) => {
     const rp = req.params;
     const acceptLanguage = req.headers && req.headers['accept-language'];
+    // TODO: investigate why the following block is not allowing related to forward
+    // accept-language to summary request
     if (!acceptLanguage || mwUtil.isNoCacheRequest(req)) {
```
thesocialdev: /page/related endpoint is performing multiple requests to /page/summary, all of them pass through this filter. During debug, I could identify that, for some reason, if one of the requests hits isNoCacheRequest as true and delete accept-language header, all of the requests will be forwarded without the accept-language header.

If change line 12 with: `if (!acceptLanguage) {` the new added `related.js` test will pass (all tests pass).

Also, my patch assumes there is no other way to forward accept-language header since I'm not acquainted with restbase functionality.

Any thoughts @Pchelolo and @d00rman cc/ @mdholloway

Pchelolo: 1. There's a bug here. We need to clone the request and deleted the headers from the clone. no-cache requests are used to rerender the stored content, and we only store default variant. So, we don't want to rerender a non-default variant. 
2. For your case of 'related' there shouldn't be no-cache requests - we don't store anything. Cache-control is removed by Varnish, so this header is under our internal control, thus there shouldn't be any no-cache requests for related.

Pchelolo:
One more thing - you probably would want to add `accept-language` to the related spec.

v1/related.js
```diff
@@ -45,7 +46,7 @@ class Related {
             delete res.body.items;
 
             // Step 2: Hydrate response as always.
-            return mwUtil.hydrateResponse(res, (uri) => mwUtil.fetchSummary(hyper, uri));
+            return mwUtil.hydrateResponse(res, (uri) => mwUtil.fetchSummary(hyper, uri, rh));
```
Pchelolo: You probably would want to add `vary` and `content-language` here. Conditionally on `mwUtil.shouldConvertAcceptLanguage` or based on one of the summary responses.

Pchelolo: Also, just to make sure, clone rh and delete 'cache-control' header.

lib/language_variants_filter.js
```diff
@@ -86,8 +88,20 @@ module.exports = (hyper, req, next, options, specInfo) => {
                 // TODO: Eventually we want to think of a better way to support
                 // other content pure-fetching, but right now let's use a magic
                 // header.
-                req.headers['cache-control'] = 'no-cache,no-store';
-                return next(hyper, req);
+                mReq.headers['cache-control'] = 'no-cache,no-store';
+                
+                //
+                return next(hyper, mReq)
+                .then((res) => mwUtil.canConvertLangVariant(hyper, mReq, revision.page_language)
```
d00rman: This is not needed. If `shouldConvert` returns `true`, then so will `canConvert`. Since at this point in the code we are sure `shouldConvert` returned `true`, no need to check this again.

lib/mwUtil.js
```diff
@@ -412,13 +412,19 @@ mwUtil.isSelfRedirect = (req, location) => {
...
  */
-mwUtil.fetchSummary = (hyper, uri) => {
-    return hyper.get({ uri })
+mwUtil.fetchSummary = (mRes, hyper, uri, headers = undefined) => {
+    return hyper.get({ uri, headers })
     .then((res) => {
+        // Assign content-language and vary header based on one of summary responses
+        if (mRes.headers && !mRes.headers['content-language']) {
+            mRes.headers['content-language'] = res.headers && res.headers['content-language']
+            mwUtil.addVaryHeader(res, 'accept-language');
```
d00rman: Why are you modifying both `res` and `mRes` headers?

thesocialdev: Nice catch, thanks.

lib/language_variants_filter.js
thesocialdev: Done.

v1/related.js
thesocialdev: Done.

lib/language_variants_filter.js
thesocialdev: Done.

thesocialdev:
I believe tests flaked with 502 HTTP errors. I'm missing something or rechecking CI is needed?

Pchelolo:
> I believe tests flaked with 502 HTTP errors. Am I missing something or rechecking CI is needed?

Restarted a build. Let's seee...

thesocialdev:
This PR needs another review after rebasing to make sure it complies with recent parsoid cleanup.

thesocialdev:
recheck?

lib/mwUtil.js
```diff
@@ -636,4 +644,18 @@ mwUtil.addVaryHeader = (res, vary) => {
...
+        return;
+    }
+    return {
+        method: req.method,
+        uri: req.uri,
+        headers: Object.assign({}, req.headers),
+        query: req.query,
+        body: req.body,
+        params: req.params
+    };
```
Pchelolo: If we're cloning - let's clone, deep clone. Or for performance we can provide an optional parameter with an array that would specify which portions of the request to clone

test/features/related.js
```diff
@@ -0,0 +1,29 @@
...
+const Server = require('../utils/server.js');
+const preq   = require('preq');
+
+describe('Page Related', () => {
+
+    const server = new Server();
+    before(() => server.start());
+    after(() => server.stop());
+
+    it('retrieve correct displaytitle for language variants', () => {
```
Pchelolo: You can put it into /pagecontent/languagevariants test suite.

lib/mwUtil.js
```diff
@@ -636,4 +644,18 @@ mwUtil.addVaryHeader = (res, vary) => {
     }
 };
 
+mwUtil.cloneRequest = (req) => {
+    if (!req) {
+        return;
```
Pchelolo: let's return the `req` so that we don't convert 'null' to 'undefined'. Wouldn't matter probably, just a nice style.

lib/mwUtil.js
```diff
@@ -412,13 +412,21 @@ mwUtil.isSelfRedirect = (req, location) => {
...
-    return hyper.get({ uri })
+mwUtil.fetchSummary = (hyper, uri, headers, pRes) => {
+    return hyper.get({ uri, headers })
     .then((res) => {
+        // Assign content-language and vary header to parent response
+        // based on one of summary responses
+        if (pRes && pRes.headers && !pRes.headers['content-language']) {
+            pRes.headers['content-language'] = res.headers && res.headers['content-language'];
+            mwUtil.addVaryHeader(pRes, 'accept-language');
+        }
```
Pchelolo: Maybe we can return an object [summary, content-language] instead of assigning the header here. I think it would be less magical.

Pchelolo:
RE-reviewd. LGTM, just some little suggestions.

mdholloway:
Ping!  Is this good to merge?

mdholloway:
Cool, thanks.  Would you mind pinging me next time you deploy restbase so we can test and close out the ticket?

Pchelolo:
sure

PR #1232, opened 2019-12-12
service-runner 2.8.0 and hyperswitch to 0.14.0 and metrics api implementation

shdubsh:
automated builds will fail until service-runner and hyperswitch are released


Pchelolo:
Superseded by #1282

PR #1234, opened 2019-12-18
Disable all parsoid-php vs parsoid-js special cases

Pchelolo:
Once this is deployed there will be no special cases and all the traffic will go into the selected variant (PHP) with no mirroring. 

After this, once we're very confident no revert will be needed, I will clean up the whole splitting code and config.

cc @wikimedia/parsing

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -125,31 +125,6 @@ class ParsoidProxy {
         return ret;
     }
 
-    _getStickyVariant(hyper, req) {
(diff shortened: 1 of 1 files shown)


Pchelolo:
> LGTM! What's the current value of `this.splitRegex` here?

It's a catch-all `/.*/`.. it's a bit inverse now, as default_variant is JS but we invert everything :) I'll clean all this up after we're done with transition

PR #1236, opened 2020-01-06
T241756: Parsoid cleanup

clarakosi:


config.frontend.test.yaml
```diff
@@ -11,8 +11,6 @@ default_project: &default_project
                     host: https://parsoid-beta.wmflabs.org
                     grace_ttl: 1000000
                     php_host: https://en.wikipedia.beta.wmflabs.org/w/rest.php
```
Pchelolo: let's remove `php_host` and just use `host`? we don't need 2 of these any longer.

config.fullstack.test.yaml
```diff
@@ -10,8 +10,6 @@ default_project: &default_project
                     host: https://parsoid-beta.wmflabs.org
                     grace_ttl: 1000000
                     php_host: https://en.wikipedia.beta.wmflabs.org/w/rest.php
```
Pchelolo: idem: let's remove `php_host` and just use `host`.

Pchelolo:
LGTM. Now we also need a patch for the deploy repo https://gerrit.wikimedia.org/r/admin/projects/mediawiki/services/restbase/deploy since we've renamed a config variable. Ping me on IRC if it's not clear what's going on in that repo.

PR #1238, opened 2020-02-03
Update the random title endpoint to use wikifeeds

mdholloway:
The wikifeeds service has been updated to serve the /page/random/title
endpoint, which is consumed by the apps for the Explore feed. This
updates RESTBase to obtain it from wikifeeds. After this change is live,
the random page code will be removed from mobileapps.


Pchelolo:
LGTM. Please rebase on #1240 to make the travis pass

PR #1239, opened 2020-02-06
Add mobile-html-offline-resources endpoint

mdholloway:
Bug: https://phabricator.wikimedia.org/T244175


Pchelolo:
Hm.. it seems like PCS /v1/transform/html/to/mobile-html/{title} has changed somehow and now it's not really transforming parsoid html but returns html for the provided {title}

PR #1240, opened 2020-02-06
Fix broken test for transform/wikitext/to/mobile-html

mdholloway:
This test was broken by a change to mobileapps to more aggressively
strip empty sections in mobile-html. Some section body text is added to
the input to fix it.


Pchelolo:
Seems good. lemme wait for travis

PR #1241, opened 2020-02-06
Create GitHub actions

Pchelolo:
Testing GH actions for CI


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,39 @@
...
+on: [push,pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [6.x, 10.x]
```
jdforrester: Should probably initiate testing on 12.x at some point.

jdforrester:
This is quite the set of tests!

Pchelolo:
> This is quite the set of tests!

hehe, it's mostly a version/configuration matrix.. If only I could make it work...

btw, I'm impressed with GH actions, I've spent no more then 15 minutes on it and it's almost at a working state. Very nice.

.github/workflows/nodejs.yml
```diff
@@ -1,12 +1,16 @@
 name: Node.js CI
 
-on: [push,pull_request]
+on:
```
jdforrester: You can simplify this down to just `on: [push,pull_request]` unless non-master patches happen a lot?

Pchelolo: In that case it runs both 'push' and 'pull_request' if the development branch is in upstream and not in the fork as it is the case with this one. We usually don't do it, this time it happened cause I've used GitHub UI to create the actions file, but being explicit wouldn't hurt imho

Pchelolo:
Hmmm... I don't see a way to only rerun failed jobs.. Travis had that. But otherwise I think it's neat, we can live with it.

PR #1242, opened 2020-03-06
Add i18n endpoint

joewalsh:
Exposes the i18n endpoint in mobileapps service https://phabricator.wikimedia.org/T247096

### DIFF
--- a/projects/v1/default.wmf.yaml
+++ b/projects/v1/default.wmf.yaml
@@ -96,3 +96,6 @@ paths:
       - path: v1/javascript.yaml
         options:
           host: '{{options.mobileapps.host}}'
+      - path: v1/i18n.yaml
+        options:
+          host: '{{options.mobileapps.host}}'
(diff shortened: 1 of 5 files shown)


mdholloway:
The PR looks good as best I can tell.  That said, at this point I am unable to test it in RESTBase locally.  My local installation appears up to date, but 60 tests fail on `npm test` and none of the provided config files appear to provide the expected routes.

joewalsh:
This is ready for review as the endpoint was deployed in mobileapps

Pchelolo:
I think the test failures have something to do with https://phabricator.wikimedia.org/T246854

Pchelolo:
This looks like it will work, but given that parsoid in beta is broken -> tests are broken, I would be much more comfortable if we waited for that to be fixed. However, if it's blocking some other development, please say so and I will merge/deploy it with some manual testing.

joewalsh:
Hi @Pchelolo - thanks for the code reviews and update on the broken tests. This and #1244 are blocking further development and release of the apps. Our timeline is somewhat flexible - so if [T246854](https://phabricator.wikimedia.org/T246854) is fixed in time for these changes to go live this week or next, it works for us, otherwise it'd be better to go forward with manual testing if possible.

PR #1244, opened 2020-03-10
Add ensure_content_type filter to mobile-html and mobile-html-offline-resources

joewalsh:
Added the `ensure_content_type` filter to `mobile-html` and `mobile-html-offline-resources`. It looks like the spec format had changed, so I updated the filter itself to work with the new format. Let me know if I missed something or if those changes were unnecessary.

https://phabricator.wikimedia.org/T247295

### DIFF
--- a/lib/ensure_content_type.js
+++ b/lib/ensure_content_type.js
@@ -82,7 +82,7 @@ module.exports = (hyper, req, next, options, specInfo) => {
     const responseCodes = responses && Object.keys(responses);
     let defaultContentType;
     const validContentTypes = new Set();
-    responseCodes.forEach(code => {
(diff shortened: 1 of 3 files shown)


lib/ensure_content_type.js
```diff
@@ -78,18 +78,34 @@ function checkContentType(hyper, req, next, expectedContentType, responsePromise
 }
 
 module.exports = (hyper, req, next, options, specInfo) => {
-    const produces = specInfo.spec.produces;
```
Pchelolo: Omg, this is something that has been missed entirely when updating to openapi 3. Thank you for noticing and fixing it.

PR #1245, opened 2020-03-13
Use parsoid-external-ci-access.beta.wmflabs.org for testing

cscott:
Bug: T247589

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -44,4 +44,5 @@ jobs:
         export PATH=${PATH}:../apache-cassandra-${{ matrix.cassandra-version }}/bin
         cassandra
     - run: npm install
-    - run: npm run lint && npm run coverage -- ${{ matrix.test-target }} ${{ matrix.test-mode }}
+    - run: npm run lint
+    - run: SAVE_PAGE_NAME="Save_test%2F${{ matrix.node-version }}-${{ matrix.cassandra-version }}-${{ matrix.test-target }}-${{ matrix.test-mode }}" npm run coverage -- ${{ matrix.test-target }} ${{ matrix.test-mode }}
--- a/config.frontend.test.yaml
+++ b/config.frontend.test.yaml
@@ -8,7 +8,7 @@ default_project: &default_project
               - path: projects/v1/default.wmf.yaml
                 options: &default_options
                   parsoid:
-                    host: https://en.wikipedia.beta.wmflabs.org/w/rest.php
+                    host: http://parsoid-external-ci-access.beta.wmflabs.org/w/rest.php
                     grace_ttl: 1000000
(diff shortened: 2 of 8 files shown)


cscott:
Two failing tests:
'item requests > should request page lints. no revision:' is failing for some reason, even though this works fine from CLI:
```
curl -d '{}' -X POST -H 'Content-Type: application/json' -H 'Host: en.wikipedia.beta.wmflabs.org' http://parsoid-external-ci-access.beta.wmflabs.org/w/rest.php/en.wikipedia.beta.wmflabs.org/v3/transform/wikitext/to/lint/User%3APchelolo%2FLintTest

curl -d '{}' -X POST -H 'Content-Type: application/json' -H 'Host: en.wikipedia.beta.wmflabs.org' http://parsoid-external-ci-access.beta.wmflabs.org/w/rest.php/en.wikipedia.beta.wmflabs.org/v3/transform/wikitext/to/lint/User%3APchelolo%2FLintTest/409437
```
It seems to be the redirect which is causing problems, since the redirect goes to:
https://en.wikipedia.beta.wmflabs.org/w/rest.php/en.wikipedia.beta.wmflabs.org/v3/transform/wikitext/to/lint/User%3APchelolo%2FLintTest/409437
not `parsoid-external-ci-access.beta.wmflabs.org`

Second failure is 'router - security > should forward cookies on request to parsoid'.
That gets an assertion error "Mocks not satisfied".  Probably the mock needs to be updated?

cscott:
Ok, this should fix the tests.

It's *possible* the issue with `parsoid.js:getLintErrors()` is a parsoid bug with how the redirect is generated.

There are some transient failures which seem to be caused by github parallelizing the tests which create and mutate https://en.wikipedia.beta.wmflabs.org/wiki/Save_test

cscott:
The "prevent concurrent execution" thing doesn't actually seem to work.  The page name should include `node-version`, `cassandra-version`, `test-target`, and `test-mode` to prevent conflicts, instead of trying to keep the test runs separate.

cscott:
OK, this is good now.

Pchelolo:
Hm.. thank you for setting up the parsoid-ci-access, however with the test fixes - I'm not sure they're correct. For example the page_save test that you're setting the page via an env variable now is actually completely mocked and should be not calling real services (we were getting complaints for creating gazillion pages on beta).

I think I should just take it on from here and try to poke some other approaches to fixing those tests, there's probably just something hardcoded somewhere, `nock` is not a very user-friendly library to debug..

Pchelolo:
Superseded by #1248 - mostly same code, thank you @cscott, just changed the save_api to actually use the mocks - we have switched that off cause parsoid and MW was sharing the domain and we couldn't use nock properly, now they don't so we can return to using the mocks.

PR #1246, opened 2020-03-16
Deprecate PCS /page/media and /page/metadata endpoints

mdholloway:
These endpoints are unused by Wikimedia clients and should be dropped in the near future.


Pchelolo:
Given that they were marked as experimental - no need to deprecate. just delete.

mdholloway:
See #1249.

PR #1247, opened 2020-03-16
Deprecate PCS /page/references endpoint

mdholloway:
This content has been incorporated into mobile-html and the separate
endpoint is no longer needed.


Pchelolo:
Is it used by any app versions? It was marked as experimental, so if there's no known usages - just drop it.

mdholloway:
See #1249.

PR #1249, opened 2020-03-17
Remove PCS media, metadata, and references endpoints

mdholloway:
These endpoints are marked experimental, were never used by the apps, and will never be.


Pchelolo:
LGTM, but first a couple of things need to happen:
- References to these should be removed from change-prop sources
- References to these should be removed from change-prop deploy repo
- Task should be created to drop Cassandra keyspaces for the references endpoint

Pchelolo:
And one test is failing. needs fixing ofc

mdholloway:
Thanks Petr.  I'll do those follow-ups in the morning.  I assume the Graphoid spec test faliing on some of these CI jobs is unrelated.

Pchelolo:
>  I assume the Graphoid spec test faliing on some of these CI jobs is unrelated.

Ye, ignore that one. I'm actually thinking to remove it - it's unreliable and we're probably gonna get rid of graphoid anyway

mdholloway:
I made the updates in change-propagation and change-propagation/deploy and added you to both as a reviewer.

The change-propagation code repo has some issues.  There are four tests failing on master over some kind of schema validation issue.  Also, CI insta-fails on master because the blubberfile uses v3, which is obsolete.  I updated the Blubberfile as part of my patch but haven't worked out what's going on with the test failures yet.

PR #1252, opened 2020-04-01
Bump mobile-html and mobile-html-offline-resources filter version to 1.2.1

joewalsh:
https://phabricator.wikimedia.org/T249150

### DIFF
--- a/v1/pcs/mobile-html-offline-resources.yaml
+++ b/v1/pcs/mobile-html-offline-resources.yaml
@@ -31,7 +31,7 @@ paths:
         200:
           description: links to scripts and styles to accompany the mobile-html of the page for offline consumption
           content:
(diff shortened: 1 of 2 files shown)


Pchelolo:
LGTM. When do you need this to be deployed?

joewalsh:
@Pchelolo later this week or early next week would be ideal. Same for #1251

Pchelolo:
Was mcs/pcs deployed already so I can proceed and deploy this any time?

joewalsh:
Yes, mcs/pcs was deployed with the new version

Pchelolo:
When do you need this deployed? I can do tomorrow probably

joewalsh:
@Pchelolo could you merge #1251 before deploying? That's the bigger blocker for us. Today would be great, early next week (before Thursday) is fine too.

PR #1253, opened 2020-04-03
Update eslint config

edg2s:


edg2s:
Failures appear unrelated to the patch.

package.json
```diff
@@ -51,10 +51,9 @@
     "ajv": "^6.10.2",
     "bunyan": "^1.8.12",
     "coveralls": "^3.0.5",
-    "eslint": "^5.16.0",
-    "eslint-config-wikimedia": "^0.13.1",
-    "eslint-plugin-jsdoc": "^15.8.0",
-    "eslint-plugin-json": "^1.4.0",
+    "eslint": "^6.8.0",
```
Pchelolo, line 54: Unfortunately we can't do that yet. We still need to be able to run this on node6

Pchelolo:
We unfortunately can't do this yet, we still need to be able to run this on node 6. eslint is not a prod dependency, but we want to be able to run tests on node 6 as well.

edg2s:
We could disable linting tests on 6, they only need to be checked in one environment...

Pchelolo:
Superseded by #1299

PR #1255, opened 2020-04-08
For mobile-html always use the configured CSP

berndsi:
Check first for mobile_html instead of accepting backend CSP.
Also allow overrides from config file.

### DIFF
--- a/lib/security_response_header_filter.js
+++ b/lib/security_response_header_filter.js
@@ -88,17 +88,21 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
 
         let csp;
         if (isHtmlSvgContent(res)) {
-            // Let backend services override CSP headers for HTML / SVG
-            // XXX: Re-consider this policy
-            if (rh['content-security-policy']) {
-                csp = rh['content-security-policy'];
-            } else if (options.mobile_html) {
+            if (options.mobile_html) {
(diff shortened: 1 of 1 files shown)


lib/security_response_header_filter.js
```diff
@@ -88,17 +88,21 @@ module.exports = function addCSPHeaders(hyper, req, next, options) {
...
-            } else if (options.mobile_html) {
+            if (options.mobile_html) {
                 // TODO: This is a copy-paste from MCS. T229016
                 // In future we would want MCS to manage it's storage,
                 // but currently we can not store special CSP as changing them would require
                 // truncating all the stored content. So we have no option other then
                 // setting them explicitly in RESTBase code.
-                csp = MOBILE_HTML_CSP;
+                if (hyper.config.mobile_html_csp) {
+                    csp = hyper.config.mobile_html_csp;
```
Pchelolo, line 98: it would be better to pass these in the options of the filter itself.

mdholloway:
Is there a Phab task for this?

berndsi:
> Is there a Phab task for this?

I can't find one, and don't remember why I put this in the commit message. I thought it was more convenient for testing things locally, so one could have the CSP specified in config.yaml. Not sure if it's needed, so I'm closing for now. We can reopen if it becomes an issue again.

PR #1259, opened 2020-05-01
Fix 'morelike' recommendation API response

kodchi:
Fixes https://phabricator.wikimedia.org/T251587.


kodchi:
Why are these unrelated tests failing?

PR #1260, opened 2020-05-05
Update MCS/PCS endpoint stability designations

mdholloway:
* Bump css, javascript, and media-list to unstable
* Add experimental designation to mobile-html-offline-resources and talk, which had no stability designation

Phab: https://phabricator.wikimedia.org/T250505

### DIFF
--- a/v1/css.yaml
+++ b/v1/css.yaml
@@ -25,7 +25,7 @@ paths:
         * site (Site-specific, mobile CSS from ResourceLoader, as defined in MediaWiki\:Mobile.css)
         * pcs (CSS for the Page Content Service)
 
-        The first two are the same regardless of what domain is used.
+        The `base` and `pcs` responses are the same regardless of what domain is used.
         For these we suggest meta.wikimedia.org.
 
         You can still pass pagelib for type, but this is a legacy version of the CSS for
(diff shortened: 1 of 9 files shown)


berndsi:
@mdholloway I think we could keep the talk endpoint as experimental. This one seems like that could change significantly or become obsolete with the talk page consultation project also happening.

mdholloway:
> @mdholloway I think we could keep the talk endpoint as experimental. This one seems like that could change significantly or become obsolete with the talk page consultation project also happening.

Done

mdholloway:
Removed the mobile-sections deprecations following further discussions about ongoing mobile-sections maintainership.

v1/css.yaml
```diff
@@ -28,10 +28,12 @@ paths:
         The first two are the same regardless of what domain is used.
```
berndsi, line 28: Probably should update this, too. Site is in "the first two". It probably meant to say the first and the third, but probably better to be more explicit here.

mdholloway: Updated this text.

mdholloway:
Good catch spotting that.  Should be fixed now.  For the record, for a tweak like that I don't mind if you just push a fix onto the PR.

PR #1270, opened 2020-06-30
mobile-html-offline-resources: make 'title' parameter optional

mdholloway:
The mobile-html-offline-resources endpoint accepts a required title and
optional revision and tid path parameters despite not using any of them.
This is undesirable from a caching perspective because it will result
in the same content being cached for many different URLs per wiki.

This change makes the title parameter optional so that we can encourage
clients to stop sending title parameters to improve the cache hit rate,
while continuing to support existing clients that send it.

This change depends on a corresponding change to mobileapps itself.
(https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/608730)

Bug: https://phabricator.wikimedia.org/T254490

### DIFF
--- a/v1/pcs/mobile-html-offline-resources.yaml
+++ b/v1/pcs/mobile-html-offline-resources.yaml
@@ -11,7 +11,7 @@ info:
     name: Apache licence, v2
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /mobile-html-offline-resources/{title}:
+  /mobile-html-offline-resources:
     x-route-filters:
       - path: lib/ensure_content_type.js
     get: &mobile-html-offline-resources_title_revision_get_spec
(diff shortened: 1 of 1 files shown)


mdholloway:
Note: CI will fail until the mobileapps change is merged.

mdholloway:
Pinging @thesocialdev as well for review/comment

mdholloway:
Per in-meeting discussion. The title param may be used in the future for lazy-loaded references.

berndsi:
Maybe we could add something in Varnish or RB to all point to the same URI?

mdholloway:
We considered that, but decided for maintainability's sake it wasn't desirable to include bits of routing logic in a different repo; also, I doubt that it would be welcomed by the RESTBase or Varnish maintainers.

PR #1272, opened 2020-07-30
Prevent RESTBase CORS redirect resolve from hitting frontend caches

cicalese:
Bug: T259054

### DIFF
--- a/lib/content_location_filter.js
+++ b/lib/content_location_filter.js
@@ -3,31 +3,27 @@
 const HyperSwitch = require('hyperswitch');
 const URI = HyperSwitch.URI;
 const mwUtil = require('./mwUtil');
-const path = require('path');
 const P = require('bluebird');
 
 function isAbsoluteRedirect(location) {
     return /^https?:/.test(location);
 }
 
-function resolveRelativeRedirect(hyper, req, res) {
-    return mwUtil.getSiteInfo(hyper, req)
-    .then((siteInfo) => {
-        const pathArr = req.uri.path.slice(2).map(encodeURIComponent);
-        pathArr.push('..');
-        pathArr.push(res.headers.location);
-        return `${siteInfo.baseUri}/${path.join.apply(path, pathArr)}`;
-    });
+function resolveRelativeRedirect(req, res) {
+    const pathArr = req.uri.path.slice(0, -1).map(encodeURIComponent);
+    pathArr.unshift('');
+    pathArr.push(res.headers.location);
+    return pathArr.join('/');
 }
 
 function resolvedRedirectResponse(hyper, req, res) {
     let getContentURI;
     if (isAbsoluteRedirect(res.headers.location)) {
-        getContentURI = P.resolve(res.headers.location);
+        getContentURI = res.headers.location;
     } else {
-        getContentURI = resolveRelativeRedirect(hyper, req, res);
+        getContentURI = resolveRelativeRedirect(req, res);
     }
-    return getContentURI.then((contentURI) => hyper.request({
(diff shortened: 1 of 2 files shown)


lib/content_location_filter.js
```diff
@@ -3,31 +3,27 @@
...
     let getContentURI;
     if (isAbsoluteRedirect(res.headers.location)) {
-        getContentURI = P.resolve(res.headers.location);
+        getContentURI = res.headers.location;
     } else {
-        getContentURI = resolveRelativeRedirect(hyper, req, res);
+        getContentURI = resolveRelativeRedirect(req, res);
     }
-    return getContentURI.then((contentURI) => hyper.request({
+    return P.resolve(getContentURI).then((contentURI) => hyper.request({
```
Pchelolo, line 26: You don't need to do P.resolve anymore, hyper.request returns a promise, so you can  simplify this to
```
return hyper.request({
   method: req.method,
    uri: new URI(contentURI)
})...
```

Pchelolo:
Mmmm... So the cyclic redirects test failure is interesting. So the test is setup to catch problems when you have a redirect cycle (page A -> page B -> page A)... Theoretically this should not happen on wiki, but it sometimes does.

So before we were going over the actual network, so we were using a `request` library that follows maximum of 10 redirects, and when the limit's exceeded, it throws 504. Now we are not actually going over the network - restbase detects that the request is local (starts with /) and doesn't perform a network request, but just calls a handler internally. So, we're not going via `request` library, and there's no automatic redirect handling - we need to implement it ourselves. Something like this should work:

```
function resolvedRedirectResponse(hyper, req, res, redirectNum = 0) {
 blablabla
 return hyper.request({ blablabla })
 .then(res => {
  if (res.status >=300) {
   if (redirectNum > 10) {
   throw new HTTPError({
   status: 504
   ...
   }
   }
   return resolvedRedirectResponse(hyper, req, res, redirectNum++)
   }
  }
  ... blablabla
 });
}
```

lib/content_location_filter.js
```diff
@@ -29,6 +29,22 @@ function resolvedRedirectResponse(hyper, req, res) {
         // We don't pass the `origin` header further, so possible
         // redirect looping will be handled by normal redirect limiting in `preq`
     })
+    .then((res) => {
+        if (res.status >= 300) {
+            if (redirectNum > 10) {
+                throw new HTTPError({
+                    status: 504,
+                    body: {
+                        type: 'redirect loop',
```
Pchelolo, line 38: we set the 'type' ti an identifier of the http error, so it should be gateway_timeout for 504

lib/content_location_filter.js
```diff
@@ -29,6 +29,22 @@ function resolvedRedirectResponse(hyper, req, res) {
...
+                        detail: 'Exceeded maxRedirects'
```
Pchelolo, line 39: maxRedirects written like this suggests there's some config parameter with this name. just write 'max redirects'

Pchelolo:
Test failutre seem unrelated.

Additionally, please add an assertion to the test verifying there's no 'age' header

lib/content_location_filter.js
cicalese: I have it written that way, since that's what's in the test. Is there a reason that the test was written that way?

cicalese: Regardless, I fixed as you mentioned, including in the test.

Pchelolo: yeah, fix the test :) this error was originally produced by 'request' library where there's indeed a maxRedirects setting

PR #1279, opened 2020-09-24
Remove the code for feed summary hydration and feed storage support

Pchelolo:
Bug: T263133

### DIFF
--- a/v1/feed.js
+++ b/v1/feed.js
@@ -2,7 +2,6 @@
 
 const P = require('bluebird');
 const HyperSwitch = require('hyperswitch');
-const URI = HyperSwitch.URI;
 const Template = HyperSwitch.Template;
 const mwUtil = require('../lib/mwUtil');
 const BaseFeed = require('../lib/base_feed');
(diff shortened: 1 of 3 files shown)


v1/feed.js
```diff
@@ -53,73 +52,6 @@ const PARTS_URIS = {
...
-                    toViewsForDate.views += filteredViews[0].views;
-                }
-            });
-            return orig;
-        };
-        if (res.body.tfa && res.body.tfa.$merge && res.body.tfa.$merge.length) {
-            const summaryURI = res.body.tfa.$merge[0];
-            const title = decodeURIComponent(
-                summaryURI.substr(summaryURI.lastIndexOf('/') + 1));
-            const highQualityThumbRequest = hyper.get({
```
mdholloway, line 75: Blagh.  I forgot we had all of this random business logic intermingled with summary hydration.  Why was this happening in RESTBase?!

Pchelolo: I moved it over. summary gives too small thumbnails for apps to use as featured images

mdholloway: Is the removeDuplicateTitles/updateForDupes stuff accounted for as well?  IIRC, this was needed because we were getting raw most-read titles from the AQS response, but RESTBase was resolving them internally, so that deduplication was needed.

What I'm afraid is happening now is that redirect titles are being discarded instead of resolved (since we're not resolving them yet in the service, and they won't have a summary found in RB). This means that our view_history counts will be slightly off.  I don't think this is a blocker for going forward, just something to note and fix up later.

Pchelolo: Hm... This is an excellent point. 

We actually are resolving all redirects in the feeds now, remember the 30 seconds mean latency problem. So here we end up getting duplicate summaries and are removing them correctly, but you're right, I didn't more the removeDuplocates code into wikifeeds... I guess I should do it now before this is merged.

mdholloway:
Confirmed that this fixes language variant support.

v1/feed.js
Pchelolo: Ok, https://gerrit.wikimedia.org/r/c/mediawiki/services/wikifeeds/+/630684 moves the deduce logic into wikifeeds.

mdholloway (approved):

PR #1280, opened 2020-09-25
Redirect to internal RESTBase url for internal requests.

Pchelolo:
Bug: T263798


Pchelolo:
Given the dramatic effect it's having, I'm going to self-merge and deploy it.

PR #1281, opened 2020-11-03
Replace preq with api-testing REST client

kevinbazira:
Bug: T265846

I have been able to replace `preq` with the `api-testing` REST `client` within 11 out of 13 tests.

The two tests that are failing when I switch them to `client.post` are;

1. The `html2wt with scrub_wikitext` test fails with error 400 whenever the `scrub_wikitext: 1` parameter is added.

2. The `supports stashing content` test fails at the second `client.post`.

Could there be something I am missing that is causing these two tests to fail?

### DIFF
--- a/.api-testing.config.json
+++ b/.api-testing.config.json
@@ -1,0 +1,4 @@
+{
+    "base_uri": "",
+    "main_page": "Main_Page"
+}
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
   },
   "devDependencies": {
     "ajv": "^6.10.2",
+    "api-testing": "^1.3.0",
     "bunyan": "^1.8.12",
     "coveralls": "^3.1.0",
     "eslint": "^5.16.0",
--- a/test/features/parsoid/transform.js
+++ b/test/features/parsoid/transform.js
@@ -3,6 +3,7 @@
 const assert = require('../../utils/assert.js');
 const Server = require('../../utils/server.js');
 const preq   = require('preq');
+const { REST } = require('api-testing');
 
 const testPage = {
     title: 'User:Pchelolo%2fRestbase_Test',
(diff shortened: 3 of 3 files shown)


test/features/parsoid/transform.js
```diff
@@ -15,123 +16,108 @@ describe('transform api', function() {
     this.timeout(20000);
     let contentTypes;
     const server = new Server();
+    const client = new REST('');
```
Pchelolo, line 19: Nothing to do here for you, but it would've been sweet if we could inject the config into the clients @clarakosi

test/features/parsoid/transform.js
```diff
@@ -3,6 +3,7 @@
 const assert = require('../../utils/assert.js');
```
Pchelolo, line 3: Can you try reimporting 'assert' from api-testing too and use chai asserts instead of this?

e.g. const { REST, assert } = require('api-testing') and then your 'assert' will be a reexport of [chai library](https://www.chaijs.com/api/assert/). Then you can change some of the asserts to use the lib, which is more powerful

Pchelolo:
> Could there be something I am missing that is causing these two tests to fail?

Can you add the changes you made for these tests into the PR too? It's hard to debug not knowing what you've done.

Pchelolo:
> The html2wt with scrub_wikitext test fails with error 400 whenever the scrub_wikitext: 1 parameter is added.

I've tried to change this one locally and it works for me.

> The supports stashing content test fails at the second client.post.

So, this one requires you to set a `if-match` header and I haven't found a way to do it in api-testing. @clarakosi writes in the api-testing library under request jsdoc:

>  The request has not been sent when this method returns, and can still be modified like a superagent request.

So I've attempted to just use `.set` there, but to no success - since the 'api-testing.request' method is async, it returns a promise, and that JSDoc is incorrect. Thus, the api-testing library needs fixing - we need a way to set arbitrary headers to requests, at least in REST. You can work with @clarakosi on that.

test/features/parsoid/transform.js
kevinbazira: Thanks @Pchelolo, I have reimported `assert` from api-testing and used chai asserts as advised.

kevinbazira:
> Can you add the changes you made for these tests into the PR too? It's hard to debug not knowing what you've done.

The two tests that are failing have been added into the PR.

Hope we can find a way to make them work with  the `api-testing` REST `client`.

test/features/parsoid/transform.js
```diff
@@ -15,197 +16,166 @@ describe('transform api', function() {
...
-                headers: {
-                    'if-match': etag
-                },
-                body: {
-                    html: res.body.replace('>ABCDEF<', '>FECDBA<')
-                }
-            });
+            return client.post(
+                `${server.config.baseURL('en.wikipedia.beta.wmflabs.org')}/transform/html/to/wikitext/${testPage.title}/${testPage.revision}`,
+                { 'if-match': etag, html: res.text.replace('>ABCDEF<', '>FECDBA<') }
```
Pchelolo, line 165: Yeah, ofc it's failing, 'if-match' is a header, not a part of the body. 

Unfortunately it seems like there's no way to set custom headers in api-testing, lemme chat with @clarakosi and decide how to add that capability.

test/features/parsoid/transform.js
```diff
@@ -15,197 +16,166 @@ describe('transform api', function() {
...
-        return preq.post({
-            uri: `${server.config.baseURL('en.wikipedia.beta.wmflabs.org')}/transform/html/to/wikitext`,
-            body: {
-                html: '<h2></h2>',
-                scrub_wikitext: 1
-            }
-        })
+        return client.post(
+            `${server.config.baseURL('en.wikipedia.beta.wmflabs.org')}/transform/html/to/wikitext`,
+            { html: '<h2></h2>', scrub_wikitext: 1 }
```
Pchelolo, line 128: Change 1 to 'true' - this parameter should be a boolean, and preq was automatically doing a conversion while api-testing doesn't.

test/features/parsoid/transform.js
```diff
@@ -15,197 +16,166 @@ describe('transform api', function() {
...
+        )
         .then((res) => {
             assert.deepEqual(res.status, 200);
             const pattern = /Selser test/;
-            if (!pattern.test(res.body)) {
+            if (!pattern.test(res.text)) {
                 throw new Error(`Expected pattern in response: ${pattern
                 }\nSaw: ${JSON.stringify(res, null, 2)}`);
             }
-            assert.contentType(res, contentTypes.wikitext);
```
Pchelolo, line 169: This can be upstreamed to api-testing. Seems like a good general-purpose assertion.

test/features/parsoid/transform.js
kevinbazira: Thanks @Pchelolo, I have fixed the `html2wt with scrub_wikitext` test as per your advice.

test/features/parsoid/transform.js
clarakosi: Done! @kevinbazira There is now an assert.contentType in the assert library: https://github.com/wikimedia/mediawiki-tools-api-testing/blob/master/lib/assert.js#L55

test/features/parsoid/transform.js
clarakosi: I've updated the HTTP methods to now accept an object of headers as the third parameter. Please update this to pass the `if-match` headers

https://github.com/wikimedia/mediawiki-tools-api-testing/blob/master/lib/REST.js#L103

package.json
```diff
@@ -49,6 +49,7 @@
   },
   "devDependencies": {
     "ajv": "^6.10.2",
+    "api-testing": "^1.3.0",
```
clarakosi, line 52: The new changes were released as api-testing 1.4.0.

test/features/parsoid/transform.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
-const assert = require('../../utils/assert.js');
 const Server = require('../../utils/server.js');
 const preq   = require('preq');
```
clarakosi, line 4: I don't think you are using this anymore, right?

clarakosi (commented):
I've updated the api-testing package to address some of the issues you were having. You can find the new changes in api-testing 1.4.0

package.json
kevinbazira: Thanks @clarakosi, `api-testing` has been updated from 1.3.0 to 1.4.0.

test/features/parsoid/transform.js
kevinbazira: Nope, `preq` has now been totally removed from `transform.js`.

test/features/parsoid/transform.js
kevinbazira: Thanks @clarakosi and @Pchelolo, the `if-match` header has been added and the `supports stashing content` test is now passing.

test/features/parsoid/transform.js
kevinbazira: Thanks @clarakosi and @Pchelolo, all tests are now using the `assert.contentTypes`.

kevinbazira:
@clarakosi and @Pchelolo, all instances of `preq` have been replaced with the `api-testing` REST `client` in `transform.js` and all 13 tests are passing. Thanks to both of you!

Pchelolo:
Ok, this is great, I consider the experiment a success. Thank you for this, the work here is done.

However, I will not merge this just yet since we still need to have node 6 running in CI, and api-testing can't be used under node 6.

test/features/parsoid/transform.js
```diff
@@ -15,197 +15,167 @@ describe('transform api', function() {
...
-        })
+        return client.post(
+            `${server.config.baseURL('en.wikipedia.beta.wmflabs.org')}/transform/wikitext/to/html/${testPage.title}`,
+            { wikitext: '== Heading ==' }
+        )
         .then((res) => {
             assert.deepEqual(res.status, 200);
-            assert.contentType(res, contentTypes.html);
+            assert.contentTypes(res, mwUtil.constructRegex([contentTypes.html]));
             const pattern = /<h2.*>Heading<\/h2>/;
```
Pchelolo, line 41: can you replace all pattern matchings with assert.match?

test/features/parsoid/transform.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
-const assert = require('../../utils/assert.js');
 const Server = require('../../utils/server.js');
-const preq   = require('preq');
+const { REST, assert } = require('api-testing');
```
Pchelolo, line 4: Given that we still need to run CI with node 6, which doesn't support async/await, you can surround this with try-catch and just return and skip this whole test suite if api-testing is not loadable.

test/features/parsoid/transform.js
clarakosi: So the current workaround for the JSON file would be to declare an environment variable for the base URI. Probably easiest to do it in the test command in package.json but that would complicate things when trying to run a single file. Will look into why we did it this way next week

.api-testing.config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "base_uri": "",
+    "main_page": "Main_Page"
```
clarakosi, line 3: You don't need the main_page info

clarakosi (commented):
Well done! 🎉  One more nit below

test/features/parsoid/transform.js
```diff
@@ -15,197 +15,167 @@ describe('transform api', function() {
...
-            }
-        })
+        return client.post(
+            `${server.config.baseURL('en.wikipedia.beta.wmflabs.org')}/transform/wikitext/to/html/${testPage.title}`,
+            { wikitext: '== Heading ==' }
+        )
         .then((res) => {
             assert.deepEqual(res.status, 200);
-            assert.contentType(res, contentTypes.html);
+            assert.contentTypes(res, mwUtil.constructRegex([contentTypes.html]));
```
clarakosi, line 40: I published a new version that makes contentType singular here. Can you update the package to 1.4.1 and update this assert accordingly?

kevinbazira: Thanks @clarakosi, `api-testing` has been updated from 1.4.0 to 1.4.1 and I was to use the singular `contentType` but @Pchelolo [advised](https://github.com/wikimedia/restbase/pull/1281#discussion_r518781424) that we replace all pattern matchings with `assert.match`.

.api-testing.config.json
kevinbazira: The "Main_Page" info has been removed from `.api-testing.config.json` file as per your advice.

test/features/parsoid/transform.js
kevinbazira: Thanks @Pchelolo all pattern matchings are now using `assert.match` as per your advice.

test/features/parsoid/transform.js
kevinbazira: A try-catch block has been added as per your advice.

package.json
```diff
@@ -49,6 +49,7 @@
   },
   "devDependencies": {
     "ajv": "^6.10.2",
+    "api-testing": "^1.4.1",
```
clarakosi, line 52: Ok, last comment I promise. Can you bump this 1.4.2? I found a bug yesterday and released a new version

kevinbazira: No worries @clarakosi, the comments can keep coming. `api-testing` has been updated from 1.4.1 to 1.4.2.

PR #1284, opened 2020-11-17
GH Actions: use cassadnra action

Pchelolo:


Pchelolo:
Doesn't seem to work :(

PR #1286, opened 2020-12-15
Add metrics to config.example.yaml

cscott:
Ensure that the standalone restbase example configuration doesn't throw
"No such metrics client: 'undefined'" errors on startup.

Change-Id: I4dfbe33a0f25ef28f877c41290d433d5bcf50984


Pchelolo:
This should be fixed in service-runner. I'll ask @clarakosi to look into this.

PR #1287, opened 2021-02-12
Remove references to graphoid.

Pchelolo:
Bug: T242855


akosiaris (approved):
Errors from CI look unrelated to the change in my eyes and not sure if they were broken before. But LGTM

PR #1292, opened 2021-05-14
Add redirect filter for domain redirects

clarakosi:
The filter adds a domain level redirect similar to the one found
in apache for mediawiki to redirect restbase APIs.

Bug: T279588

### DIFF
--- a/config.fullstack.test.yaml
+++ b/config.fullstack.test.yaml
@@ -197,6 +197,9 @@ spec_root: &spec_root
   title: "The RESTBase root"
   # Some more general RESTBase info
   x-request-filters:
+    - path: lib/redirect_filter.js
+      options:
+        ar.wikipedia.beta.wmflabs.org: en.wikipedia.beta.wmflabs.org
     - path: lib/security_response_header_filter.js
 
   x-sub-request-filters:
--- a/lib/redirect_filter.js
+++ b/lib/redirect_filter.js
@@ -1,0 +1,20 @@
+'use strict';
+
+const redirectLocation = (redirectTarget, uri) => {
+    const uriArray = uri.split('/');
+    uriArray[1] = redirectTarget;
+    return uriArray.join('/');
+};
+
+module.exports = (hyper, req, next, options) => {
+    const redirectSource = req.params.domain || `${req.uri}`.split('/')[1];
+    if (redirectSource in options) {
+        return {
+            status: 301,
+            headers: {
+                location: redirectLocation(options[redirectSource], `${req.uri}`)
+            }
+        };
+    }
+    return next(hyper, req);
+};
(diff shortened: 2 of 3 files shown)


config.fullstack.test.yaml
```diff
@@ -197,6 +197,9 @@ spec_root: &spec_root
   title: "The RESTBase root"
   # Some more general RESTBase info
   x-request-filters:
+    - path: lib/redirect_filter.js
+      options:
+        ar.wikipedia.beta.wmflabs.org: en.wikipedia.beta.wmflabs.org
```
Pchelolo, line 202: Would probably be nice to wrap this into a 'domain_redirects' property - just in case we ever want to add some other type of redirects.

lib/redirect_filter.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const redirectLocation = (redirectTarget, uri) => {
+    const uriArray = uri.split('/');
+    uriArray[1] = redirectTarget;
+    return uriArray.join('/');
+};
+
+module.exports = (hyper, req, next, options) => {
+    const redirectSource = req.params.domain || `${req.uri}`.split('/')[1];
```
Pchelolo, line 10: If there's no domain, don't do anything.

lib/redirect_filter.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const redirectLocation = (redirectTarget, uri) => {
```
Pchelolo, line 3: This ain't gonna work... The problem is that we will receive the uri as internal to the cluster, e.g. `http://restbase.svc.eqiad.wmnet/en.wikipedia.org` etc. 

So your location will also become internal, and the client that receives this redirect will have no idea what to do.

So we need to generate absolute redirect for external requests, and relative redirects for internal requests. normalize_title_filter.js L73-96 does this properly. Now that we have a need for to generate absolute redirects in 2 places, you can extract some code from normalizeTitleFilter into mwutil and use it here too. not sure what the API should be, something like mwUtil.createAbsoluteRedirect

test/features/redirect_filter.js
```diff
@@ -0,0 +1,17 @@
...
+    before(() => server.start());
+    after(() => server.stop());
+
+    it('should redirect the request to the set redirect target', () => {
+        return preq.get({ uri: `${server.config.baseURL('ar.wikipedia.beta.wmflabs.org')}/page/html/Main_Page` })
+            .then((res) => {
+                assert.deepEqual(res.status, 200);
+                assert.checkString(res.headers['content-location'], /en.wikipedia.beta.wmflabs.org/);
+            });
+    });
```
Pchelolo, line 16: Neet a test for internal and external requests.

PR #1294, opened 2021-08-27
Bump mediawiki-title to v0.7.4

arlolra:
Should fix a redirect loop where the core Title implementation allows
soft-hyphens (U+00AD) but mediawiki-title was stripping them.  This
mismatch happened because a change only got reverted in one of the
implementations.

See https://github.com/wikimedia/mediawiki-title/pull/51

Bug: T284734

/cc @Pchelolo @subbuss

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "fast-json-stable-stringify": "^2.1.0",
     "hyperswitch": "^0.14.0",
     "jsonwebtoken": "^8.5.1",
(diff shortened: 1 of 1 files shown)


kelson42:
No sure if this PR breaks things... or if the whole CI is broken...

arlolra:
> No sure if this PR breaks things... or if the whole CI is broken...

I get the same 6 failures on the main branch

PR #1295, opened 2021-09-08
Resolve redirects for page/related endpoint

VadimKovalenkoSNF:
Implement private _relatedPageRedirect() method for resolving article redirects

BUG: T289575

### DIFF
--- a/v1/related.js
+++ b/v1/related.js
@@ -12,14 +12,14 @@ class Related {
     }
 
     _relatedPageRedirect(hyper, domain, title) {
-        const apiURI = new URI(`https://${domain}/w/api.php`);
+        const apiURI = new URI([domain, 'sys', 'action', 'rawquery']);
 
         return hyper.get({
             uri: apiURI,
             headers: {
                 'content-type': 'application/json'
             },
-            query: {
+            body: {
                 action: 'query',
                 format: 'json',
                 titles: title,
--- a/v1/related.js
+++ b/v1/related.js
@@ -11,59 +11,89 @@ class Related {
         this._options = options;
     }
 
+    _relatedPageRedirect(hyper, domain, title) {
+        const params = `action=query&format=json&titles=${title}&redirects=1&converttitles=1`;
+        const apiURL = `https://${domain}/w/api.php?${params}`;
+        return hyper.get({
+            uri: apiURL,
+            headers: {
+                'content-type': 'application/json'
+            }
+        }).then((res) => {
+            return res.body;
+        }).catch((e) => {
+            hyper.logger.log('error/related', {
+                message: 'Failed to fetch related pages',
+                error: e
+            });
+            throw e;
+        });
+    }
+
     getPages(hyper, req) {
         const rp = req.params;
-        const rh = Object.assign({}, req.headers);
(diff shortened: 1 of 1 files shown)


v1/related.js
```diff
@@ -11,59 +11,89 @@ class Related {
         this._options = options;
     }
 
+    _relatedPageRedirect(hyper, domain, title) {
+        const params = `action=query&format=json&titles=${title}&redirects=1&converttitles=1`;
+        const apiURL = `https://${domain}/w/api.php?${params}`;
```
clarakosi, line 16: Instead of building the URL here, you should be able to use the action API wrapper here. Example: https://github.com/wikimedia/restbase/blob/master/v1/lists.js#L142-L154

v1/related.js
```diff
@@ -11,59 +11,89 @@ class Related {
...
-                gsrsearch: `morelike:${rp.title}`,
-                gsrnamespace: 0,
-                gsrwhat: 'text',
-                gsrinfo: '',
-                gsrprop: 'redirecttitle',
-                gsrlimit: 20
+        return this._relatedPageRedirect(hyper, rp.domain, rp.title).then((res) => {
+
+            // Return the right title to redirect if it has another language option
+            if (res.query.converted) {
```
clarakosi, line 39: This will cause an error if `res` is empty. I'd suggest additional checks here and below to see if `res` and `res.query` exist. Similar to line 90 below

clarakosi (commented):
Thanks for the PR! I left a couple of small comments below

v1/related.js
VadimKovalenkoSNF: Done.

v1/related.js
VadimKovalenkoSNF: Done.

v1/related.js
```diff
@@ -11,59 +11,98 @@ class Related {
...
 
-    getPages(hyper, req) {
-        const rp = req.params;
-        const rh = Object.assign({}, req.headers);
-        // we don't store /page/related no need for cache-control
-        if (rh['cache-control']) {
-            delete rh['cache-control'];
-        }
+    _relatedPageRedirect(hyper, domain, title) {
+        const apiURI = new URI(`https://${domain}/w/api.php`);
```
clarakosi, line 15: This still doesn't use the Action API wrapper but rather calls this URI directly which then makes an external call to the MW instance instead of a local call. This will also miss out on error handling provided by the Action API wrapper since the Action API returns 200 on errors.

To use the Action API wrapper you'd need to modify URI to something like:
```suggestion
        const apiURI = new URI([domain, 'sys', 'action', 'rawquery']);
```

clarakosi (changes requested):

Pchelolo:
please rebase it on current master, it has CI fixed.

v1/related.js
VadimKovalenkoSNF: Thank you for the suggestion, pls check recent changes.

VadimKovalenkoSNF:
> please rebase it on current master, it has CI fixed.

Done, pls check.

clarakosi (approved):

PR #1297, opened 2021-11-29
Add content purge filter

Pchelolo:
This filter can be added in case ALL content in a certain
time range needs to be purged.

config.fullstack.yaml contains an example of how to add a filter,
this needs to be done in the production config in the deploy repo.

Change-Id: I076160215421d57cd708edeb31c8fa16215997c5


johngian:
Re-opening this in case we need to use it for PCS switchover to core page html

PR #1299, opened 2022-02-07
Update dependencies

Pchelolo:
DO NOT MERGE YET!!!

Still preserving compatibility with node 10

### DIFF
--- a/v1/common_schemas.yaml
+++ b/v1/common_schemas.yaml
@@ -108,20 +108,16 @@ components:
         lang:
           type: string
           description: The page language code
-          example: en
         dir:
           type: string
           description: The page language direction code
-          example: ltr
         timestamp:
           type: string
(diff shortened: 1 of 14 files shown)


v1/common_schemas.yaml
```diff
@@ -108,20 +108,16 @@ components:
         lang:
           type: string
           description: The page language code
-          example: en
```
clarakosi, line 111: Is there a reason for getting rid of the examples?

clarakosi (approved):

v1/common_schemas.yaml
Pchelolo: ya. 'example' is not a supported key in json-schema, `examples` is. and in openapi `example` is the supported key, while `examples` is not. newer version of json validator package we use fails when it sees `example` and openapi-spec validator fails if it sees `examples`. So, I've decided to just remove them instead of messing with this incompatibility.

clarakosi: Ohh ok. Makes sense

PR #1305, opened 2022-09-02
Parsoid: Empty the list of pages that we don't want to re-render

MatmaRex:
Perf improvements and resource limit improvements made it unnecessary,
and it hasn't been updated in years.

The list is a problem because we want to parse some of these pages now.
See https://phabricator.wikimedia.org/T316914


thesocialdev (approved):
LGTM. I will give an opportunity to others to chime in before merging it.

kemayo:
Seems like nobody wants to chime in -- want to go ahead and merge?

thesocialdev (approved):

thesocialdev:
I don't have merge rights, @johngian will look at it.

johngian:
Looks good to me.

johngian:
Looks good to me. On top of that this config is picked up by the tests so restbase initialization should work fine.

MatmaRex:
Thanks for the reviews. It looks like the change is not deployed to Wikimedia wikis yet. Is that automated like the deployment train, or what would it take to make that happen?

MatmaRex:
(It was deployed by @johngian)

Thank you both for the reviews!

PR #1306, opened 2022-09-20
tests: Fix CI errors

johngian:
* Fix mathoid endpoint references for testing
* Fix missing titles and revisions from beta
* Fix displaytitle assertion for language variants


johngian:
@thesocialdev @brightbyte It looks like CI is fixed with this PR.

PR #1307, opened 2022-09-22
pcs sunset: Emit metrics for PCS latency

johngian:
* Reduce max parallelism to avoid 500 errors from beta cluster

Change-Id: Iae4bc53742e3aef938078c8d2706537553e7de9b


johngian:
@brightbyte I added  a few more timing metrics for PCS specifically.

PR #1310, opened 2022-10-18
pcs: Fix timing metric

johngian:
Change-Id: Ife4a089ba341761608575aa2f2462822ffd5e905


johngian:
Fixes a bug on the way we send PCS timings, instead of send the timestamp send the time difference.

johngian:
@thesocialdev Can you take a look?

thesocialdev (approved):

PR #1311, opened 2022-12-20
Remove pngs from math endpoint

physikerwelt:
Bug: T311620
Change-Id: I7dc8be1900d2d7b502f7e66e7f3207404a8b3dec

### DIFF
--- a/sys/mathoid.js
+++ b/sys/mathoid.js
@@ -6,7 +6,7 @@ const URI = HyperSwitch.URI;
 const HTTPError = HyperSwitch.HTTPError;
 const mwUtil = require('../lib/mwUtil');
 
-const FORMATS = ['mml', 'svg', 'png'];
+const FORMATS = ['mml', 'svg'];
 
 function prefixHeaders(headers, prefix = 'x-store-') {
     const prefixedHeaders = {};
@@ -141,12 +141,6 @@ class MathoidService {
                 }, prefixHeaders(completeBody[format].headers)),
                 body: completeBody[format].body
             };
-            if (format === 'png' && reqObj.body && reqObj.body.type === 'Buffer') {
-                // for png, we need to convert the encoded data manually
-                // because we are receiving it wrapped inside a JSON
-                reqObj.body = Buffer.from(reqObj.body.data);
-                completeBody[format].body = reqObj.body;
-            }
             // store the emit Promise
             reqs[idx] = hyper.put(reqObj);
         }
@@ -289,15 +283,6 @@ module.exports = (options) => {
                 body: {
                     keyType: 'string',
                     valueType: 'string'
-                }
-            }, {
-                uri: '/{domain}/sys/key_value/mathoid_ng.png',
-                headers: {
-                    'content-type': 'application/json'
-                },
-                body: {
-                    keyType: 'string',
-                    valueType: 'blob'
                 }
             }
         ]
--- a/test/features/mathoid.js
+++ b/test/features/mathoid.js
@@ -9,8 +9,8 @@ describe('Mathoid', function() {
     const server = new Server();
     const f = 'c^2 = a^2 + b^2';
     const nf = 'c^{2}=a^{2}+b^{2}';
-    const formats = ['mml', 'svg', 'png'];
-    const formats_regex = [/mathml/, /svg/, /png/];
+    const formats = ['mml', 'svg'];
+    const formats_regex = [/mathml/, /svg/];
     let hash;
 
     before(() => server.start());
--- a/v1/mathoid.yaml
+++ b/v1/mathoid.yaml
@@ -153,7 +153,6 @@ paths:
             enum:
               - svg
               - mml
-              - png
         - name: hash
           in: path
           description: The hash string of the previous POST data
@@ -171,10 +170,6 @@ paths:
             application/mathml+xml:
               schema:
                 type: string
-            image/png:
-              schema:
-                type: string
-                format: binary
         404:
           description: Unknown format or hash ID
           content:


physikerwelt:
cf. https://gerrit.wikimedia.org/r/c/mediawiki/services/restbase/+/869763

johngian:
Codewise it looks OK. Meanwhile i am trying to fix the tests so CI can pass.

physikerwelt:
@johngian cool. Can you also deploy to beta?

johngian:
Please rebase to latest changes.

physikerwelt:
@johngian thank you for fixing the test. Now the tests pass.

brightbyte:
Do I understand correctly that this will remove support for the https://wikimedia.org/api/rest_v1/media/math/render/png endpoint? I am seeing 64k requests per hour there... So something is apparently using it? https://w.wiki/6PEA

physikerwelt:
I can't see the stats. Maybe the requests come from non-WMF wikis. The change was announced in technews (https://en.wikipedia.org/wiki/Wikipedia:Village_pump_(technical)/Archive_201#Tech_News:_2022-48) and discussed here https://phabricator.wikimedia.org/T311620#8400577. People running math extensions before https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Math/+/858661 (approx 1.40.0-wmf.12) still would trigger requests from people using their websites.

brightbyte:
Looks like you are right - according to the Referrer header, nearly all of them come from Fandom.

mszabo-wikia:
Hi all,

We've expedited the work to disable PNG rendering support and hope to do it as soon as we determine whether there's a need for any user communication.

We've also reprioritized the work to (in the medium/long term) move off the WMF infrastructure for Math rendering.

Thank you for your patience and understanding.

physikerwelt:
Math development is mostly driven by the community. You can join us here https://meta.wikimedia.org/wiki/Wikimedia_Community_User_Group_Math

mszabo-wikia:
We just deployed a change to disable PNG rendering for math & chem formulas, so hopefully this change should be unblocked momentarily.

physikerwelt:
@mszabo-wikia thank you. @brightbyte can you double check please?

brightbyte:
Requests are down by 85%. Still seeing about two requests per second, though (down from 15). Let's see how it looks tomorrow.

physikerwelt:
@brightbyte do you think it would make sense to backport the change to 1_39 REL (I guess not everyone might be willing to upgrade to 1_40 REL)

brightbyte:
> @brightbyte do you think it would make sense to backport the change to 1_39 REL (I guess not everyone might be willing to upgrade to 1_40 REL)

Are we even doing releases of RESTbase? I'm not seeing release branches or anything like that.

physikerwelt:
@brightbyte The requests are initiated by the Math extension. And these extensions follow the regular release cycle. My suggestion would be https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Math/+/898882

brightbyte:
Looks like the hits originating from Fandom are pretty much gone now. 
We are seeing about one req/sec coming from paradoxplaza.com wikis, see https://wikis.paradoxplaza.com/.

physikerwelt:
I can redo the removal of PNG rendering mode on 1.35 (there are too many other code changes between 1.35 and 1.39). 
Also note that there are two reasons of why PNG images are being requested. For one the browser feels that it can not serve SVG images and asks for PNG.



However, if we run node <=10 in restbase anyhow, I don't see any reason to hurry. It depends on the urgency to remove node 10. I was under the impression that this would be very urgent.

brightbyte:
Can you make a wiki page explaining the removal of PNG support, with instructions of how to fix it on 3rd party wikis?

We could then make requests for PNG redirect to a single PNG file that points to that wiki page.

brightbyte:
> However, if we run node <=10 in restbase anyhow, I don't see any reason to hurry. It depends on the urgency to remove node 10. I was under the impression that this would be very urgent.

That I don't know.

johngian:
How is the node version of restbase related to the mathoid service?

physikerwelt:
> Can you make a wiki page explaining the removal of PNG support, with instructions of how to fix it on 3rd party wikis?

Yes



> How is the node version of restbase related to the mathoid service?

We must remove png support from mathoid as the node librsvg binding does not work on node versions > 10. If we keep services running on node 10 anyhow, I think it does not matter if there is one or two services running with node 10.

johngian:
I think there are a few things that are not necessarily coupled:

* Does the fact that MW 1.35 has support until end of 2023 mean that also extensions have the same support? From the math extension it looks like the requirement is `>=1.40`.
*  Since the node upgrade is inevitable sooner or later I think it makes sense to communicate with the consumers left using png for the upcoming deprecation.
* The main issue that I see is [that the endpoint ](https://en.wikipedia.org/api/rest_v1/#/Math/get_media_math_render__format___hash_) is considered stable so a signature change shouldn't happen without further communications and deprecation. 

@physikerwelt Other than the technical aspect (disabling png on RESTBase and updating the backing service) to me it makes sense to:
* Announce to a mailing list with a broader audience about the upcoming deprecation (mostly for external consumers).
* Update the swagger docs on the endpoint
* When we end up with an amount of reqs that we feel comfortable disable png on restbase.

I think since this discussion has little to do with actual restbase, lets move it to phabricator for better visibility.

brightbyte:
> Does the fact that MW 1.35 has support until end of 2023 mean that also extensions have the same support? From the math extension it looks like the requirement is >=1.40.

My thinking is that the version of Math that was released along with MW 1.35 should not break as long as 1.35 is supported. The fact that the current version of Math requires MW 1.40 isn't relevant for that. Websites that run MW 1.35 with Math are relying on the service, and should be able to as long as 1.35 is supported.

jdforrester:
> > Does the fact that MW 1.35 has support until end of 2023 mean that also extensions have the same support? From the math extension it looks like the requirement is >=1.40.
> 
> My thinking is that the version of Math that was released along with MW 1.35 should not break as long as 1.35 is supported. 

Math wasn't a supported release until 1.38 when we started bundling it in the tarball. I don't think you should do any extra work to try to support 1.35 in staff time.

physikerwelt:
I am happy to spend a few hours, if you tell me what fits the overall strategy.

physikerwelt:
As discussed with @brightbyte via mail, https://github.com/wikimedia/restbase/pull/1324 is the alternative approach to redirect from png to SVG images.

PR #1317, opened 2023-03-22
mcs: Update endpoint stability information

johngian:
Related phabricator patch: https://phabricator.wikimedia.org/T328036

### DIFF
--- a/test/features/mobileapps.js
+++ b/test/features/mobileapps.js
@@ -18,6 +18,7 @@ describe('Mobile Content Service', () => {
         })
         .then((res) => {
             assert.deepEqual(res.status, 200);
+            assert.deepEqual(res.headers.sunset, 'Sat, 01 Jul 2023 00:00:00 GMT');
             assert.deepEqual(/^application\/json/.test(res.headers['content-type']), true);
             assert.deepEqual(!!res.headers.etag, true);
             assert.deepEqual(!!res.body.lead, true);
@@ -31,6 +32,7 @@ describe('Mobile Content Service', () => {
         })
         .then((res) => {
             assert.deepEqual(res.status, 200);
+            assert.deepEqual(res.headers.sunset, 'Sat, 01 Jul 2023 00:00:00 GMT');
             assert.deepEqual(/^application\/json/.test(res.headers['content-type']), true);
             assert.deepEqual(!!res.headers.etag, true);
         });
@@ -42,6 +44,7 @@ describe('Mobile Content Service', () => {
         })
         .then((res) => {
             assert.deepEqual(res.status, 200);
+            assert.deepEqual(res.headers.sunset, 'Sat, 01 Jul 2023 00:00:00 GMT');
             assert.deepEqual(/^application\/json/.test(res.headers['content-type']), true);
             assert.deepEqual(!!res.headers.etag, true);
         });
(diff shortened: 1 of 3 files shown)


johngian:
TODO: Link to the announcement

thesocialdev (approved):
LGTM. Waiting for the link.

test/features/mobileapps.js
```diff
@@ -18,6 +18,7 @@ describe('Mobile Content Service', () => {
         })
         .then((res) => {
             assert.deepEqual(res.status, 200);
+            assert.deepEqual(res.headers.deprecation, 'Sat, 01 Jul 2023 00:00:00 GMT');
```
dr0ptp4kt, line 21: Should this and the other be `sunset` instead of `deprecation`? I think @brightbyte was suggesting following the IETF draft in https://www.rfc-editor.org/rfc/rfc8594.html#section-3 (if `deprecation` is just a magical alias for `sunset`, please advise). No strong feeling here, just wanted to make sure I understood correctly!

v1/mobileapps.js
```diff
@@ -37,6 +43,7 @@ class MobileApps {
             return this._fetchFromMCS(hyper, req);
         })
         .catch({ status: 404 }, () => this._fetchFromMCSAndStore(hyper, req))
+        .tap(this._injectDeprecationHeaders.bind(this))
```
dr0ptp4kt, line 46: Is this symbol available in another patch? I think by convention `res.headers` is intended to fairly explicitly align with real HTTP header values, although there's nothing wrong with this `_injectDeprecationHeaders` value observing on that header and applying any other aspects it thinks are germane (e.g., for internal enrichment of headers above and beyond `Sunset:` or `Deprecation:` as it were).

dr0ptp4kt (commented):
Question/observation on `Sunset:` and header enrichment.

test/features/mobileapps.js
johngian: I think you are right, updating the patch.

v1/mobileapps.js
johngian: I am not sure if that's what you mean but here is the definition:
https://github.com/wikimedia/restbase/pull/1317/files#diff-6805c4ed202dfbf57927df5bdad305acce64cd6a04110293c3952017be860b82R22

test/features/mobileapps.js
brightbyte: The Sunset header is appropriate when we have set a data for decomissioning. 
There is also a draft for a Deprecation header, which can be used earlier in the life cycle: https://greenbytes.de/tech/webdav/draft-ietf-httpapi-deprecation-header-latest.html

brightbyte:
I think something somewhere shoudl reference [https://www.rfc-editor.org/rfc/rfc8594.html](https://www.rfc-editor.org/rfc/rfc8594.html). Probably in a comment in the source code.

test/features/mobileapps.js
johngian: This is the spec Ι confused with the one you mentioned before @brightbyte

v1/mobileapps.js
dr0ptp4kt: I glossed right over it, sorry!

PR #1319, opened 2023-04-05
pcs: Customize cache control

johngian:
* This patch allows PCS to have customized cache control headers so we can test optimizations only for the specific endpoint


thesocialdev (approved):
LGTM.

PR #1321, opened 2023-05-17
Remove page edit API

brightbyte:
This removed the POST handlers for page content under v1/page/html and v1/page/wikitext.

Bug: T336851
Change-Id: I204ac7ff26be9b83f26a6368bd454fd1584776b8


johngian:
I just tried running the tests against master and they are passing.

brightbyte:
Perhaps page_save was used by the tests somehow to set things up...

PR #1323, opened 2023-05-23
Optimize the performance of parsoid regular expressions

sbassett29:
* Sets length ranges for a few regular expressions that may
  result in polynomial-time performance in worst-case
  scenarios.  A max range length of 128 seems reasonable
  based upon a limited analysis of available patterns.

Bug: T337274


johngian:
I think tests failures are transient from beta, master is also failing.

sbassett29:
> I think tests failures are transient from beta, master is also failing.

I would assume so, given the fact that the tests are passing everywhere else.  Are we still that concerned about supporting Node 10/sqlite for restbase?

johngian:
I think its OK to merge. Either way though we need to fix the tests before deploying.

PR #1326, opened 2023-06-22
Remove public endpoints for editing pages

brightbyte:
This removed the POST handlers for page content under v1/page/html and v1/page/wikitext.

Internal code for editing pages will be removed in the next patch.

Bug: T336851
Change-Id: I0b6af8dd1c6944c755e1df0df79bb1bdb017305d


brightbyte:
ok, this passes. Let's see what happens if we also remove sys/page_save.js

brightbyte:
I think this is ready to be merged.

johngian (approved):
No strong opinions on the comments, just being a bit more defensive because I don't know if there are other references of the error codes in RESTBase. Other than that it looks good.

johngian:
Lets squash to a single commit before merging.

brightbyte:
> Lets squash to a single commit before merging.

Not sure how to do this in the github workflow... I though you would do that while merging, using the "squash and merge" function...

PR #1329, opened 2023-07-17
Use HTTPS to access parsoid backend

brightbyte:
The Mocha tests should use HTTPS to access the parsoid-external-ci-access.beta.wmflabs.org host. Using plain HTTP may result in POST requests being rejected with "Insecure request forbidden, use HTTPS instead.".

Note that parsoid-external-ci-access.beta.wmflabs.org doesn't have a valid TLS certificate. Setting NODE_TLS_REJECT_UNAUTHORIZED=0 can be used to disable certificate validation.

Change-Id: Ib8f4af45687e13e44f77409c697779c2d8dcfdc3

### DIFF
--- a/config.frontend.test.yaml
+++ b/config.frontend.test.yaml
@@ -8,7 +8,8 @@ default_project: &default_project
               - path: projects/v1/default.wmf.yaml
                 options: &default_options
                   parsoid:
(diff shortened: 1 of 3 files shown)


brightbyte:
Obsolete

PR #1330, opened 2023-08-24
Disable storage

brightbyte:
Allow storage of parsoid output to be disabled based on the domain parameter


config.fullstack.test.yaml
```diff
@@ -7,8 +7,9 @@ default_project: &default_project
               - path: projects/v1/default.wmf.yaml
                 options: &default_options
                   parsoid:
-                    host: http://parsoid-external-ci-access.beta.wmflabs.org/w/rest.php
-                    grace_ttl: 1000000
+                    host: https://parsoid-external-ci-access.beta.wmflabs.org/w/rest.php
+                    # disabled_storage: true
+                    grace_ttl: 1000007
```
brightbyte, line 12: These changes should probably not be in this patch.
Though I can't test without https...

projects/sys/default.wmf.yaml
```diff
@@ -14,6 +14,7 @@ paths:
       - path: sys/parsoid.js
         options:
           host: '{{options.parsoid.host}}'
+          disabled_storage: '{{options.parsoid.disabled_storage}}'
```
brightbyte, line 17: is this actually needed? not sure...

test/features/pagecontent/language_variants.js
```diff
@@ -158,7 +158,7 @@ describe('Language variants', function() {
             assert.deepEqual(res.headers['content-language'], 'de');
             assert.deepEqual(res.headers['x-restbase-sunset'] || null, 'true');
             assert.checkString(res.headers.etag, /^"\d+\/[a-f0-9-]+"$/);
-            assert.deepEqual('Das ist eine testseite', res.body.extract);
+            assert.deepEqual('Das ist eine testseite.', res.body.extract);
```
brightbyte, line 161: The content of the page on beta was actually changed! Took me forever to figure that out...

https://de.wikipedia.beta.wmflabs.org/w/index.php?title=RESTBase_Testing_Page&diff=prev&oldid=30418

test/features/pagecontent/rerendering.js
brightbyte, line 1: should not be in this patch

test/features/pagecontent/pagecontent.js
brightbyte, line 1: should not be in this patch

sys/parsoid.js
```diff
@@ -519,47 +544,63 @@ class ParsoidService {
...
-                        status: 500,
-                        body: {
-                            title: 'no_etag',
-                            description: 'No ETag has been provided in the response'
-                        }
-                    });
+                    if (disabledStorage) {
+                        // Generate an ETag, for consistency
+                        const uuid = uuidv1();
+                        res.headers.etag = mwUtil.makeETag(res.headers['content-revision-id'] || '0', uuid);
```
johngian, line 588: I think here we are missing some DOM manipulation that't happening on `generateAndSave` to pass the tid to html meta.

johngian: https://github.com/wikimedia/restbase/blob/6a8e93390e59552e01d296e0563c9ef3ae1867bb/sys/parsoid.js#L444

johngian (changes requested):

johngian:
Do we have a phab ticket for reference?

brightbyte:
> Do we have a phab ticket for reference?

https://phabricator.wikimedia.org/T344945

brightbyte:
> It might also be useful to add a header for the fact that parsoid storage is disabled on responses and add a test to assert that for a test wiki from the local env setup.

Something like `x-restbase-cache-disabled`?

sys/parsoid.js
```diff
@@ -391,10 +403,23 @@ class ParsoidService {
...
-            `page/pagebundle/${encodeURIComponent(rp.title)}/${rp.revision}`
-        ));
+            path
+        );
+
+        return hyper.get(parsoidReq)
+          .then((resp) => {
+              return resp;
+          }, (error) => {
+              throw error;
```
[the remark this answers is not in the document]
brightbyte: I don't remember exactly, but I often do this kind of thing when I want to be able to set a break point.

johngian:
> > It might also be useful to add a header for the fact that parsoid storage is disabled on responses and add a test to assert that for a test wiki from the local env setup.
> 
> Something like `x-restbase-cache-disabled`?

Yeah, so we know what kind of requests we are getting (in case we need to debug from the client side) and also to be able to test this.

brightbyte:
> > Something like `x-restbase-cache-disabled`?
> 
> Yeah, so we know what kind of requests we are getting (in case we need to debug from the client side) and also to be able to test this.

I implemented it in the latest version.

brightbyte:
For the record, I don't understand why the "should render & re-render independent revisions" test is now failing with storage enabled. The behavior for that case shouldn't have changed at all. Did I actually break something, or is it an unrelated issue?...

johngian (approved):

PR #1331, opened 2023-09-14
Disable announcements healthcheck

nosmo:
This check is no longer needed as we have moved to rest-gateway for this endpoint.


johngian:
CI failures were unrelated to the patch. +1 on the actual change.

PR #1333, opened 2023-10-11
Fix CI for disabled storage

johngian:
Add a new beta domain in config just for testing the disabled parsoid storage feature. This way we also test the feature flag configured by domain.


johngian:
@brightbyte wanna take a look?

PR #1336, opened 2023-11-09
Use GET when fetching related pages

brightbyte:
Could it be this simple?

Bug: T347034


jdforrester (approved):
If it works, it works.

johngian:
Just tested it locally. It looks like GET request returns results.

PR #1337, opened 2023-11-13
Make parsoid transform follow disable_storage setting.

brightbyte:
If disable_storage is set, don't try to load the original HTML from cache for transform operations.

Bug: T350219
Change-Id: I33c3baae4f0579121b89f6d913852d6c118ebd56

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -936,12 +936,21 @@ class ParsoidService {
 
     _getOriginalContent(hyper, req, revision, tid) {
         const rp = req.params;
-        return this._getContentWithFallback(hyper, rp.domain, rp.title, revision, tid)
-            .then((res) => {
-                res = res.body;
-                res.revid = revision;
-                return res;
-            });
+
+        const disabledStorage = this._isStorageDisabled(req.params.domain);
+        let contentReq;
+
+        if (disabledStorage) {
+            contentReq = this._getPageBundleFromParsoid(hyper, req);
+        } else {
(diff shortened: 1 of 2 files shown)


thesocialdev (approved):
LGTM

jdforrester (approved):

sys/parsoid.js
```diff
@@ -936,12 +936,21 @@ class ParsoidService {
...
-                res = res.body;
-                res.revid = revision;
-                return res;
-            });
+
+        const disabledStorage = this._isStorageDisabled(req.params.domain);
+        let contentReq;
+
+        if (disabledStorage) {
+            contentReq = this._getPageBundleFromParsoid(hyper, req);
```
subbuss, line 944: Can you confirm that this will fetch a stashed version from core? Otherwise, this could lead to dirty diffs? How are you testing this?

brightbyte: Disabling storage also disabeld stashing, since RESTbase uses the storage mechanism to implement the stash. It doesn't use the Stash in MW.

But that should be a non-issue, since VE doesn't use this API anymore. As far as I know, nothing uses stashing in RESTbase. I have a PR up for removing it entirely.

johngian:
Other than a nit comment, lets squash and merge.

brightbyte:
> Other than a nit comment, lets squash and merge.

Which comment?

brightbyte (commented):
thank you!

PR #1339, opened 2023-11-29
Use beta for tests

brightbyte:
Mocha: Target the normal beta instances

Replace the use of http://parsoid-external-ci-access.beta.wmflabs.org/
with https://en.wikipedia.beta.wmflabs.org/.

Bug: T350353


johngian:
Something is wrong with the beta mathoid endpoint.
Maybe you can try this instead:
```
host: https://mathoid-beta-svc.wmcloud.org
```

PR #1341, opened 2024-01-25
Add content purge filter

johngian:
This filter can be added in case ALL content in a certain
time range needs to be purged.

config.fullstack.yaml contains an example of how to add a filter,
this needs to be done in the production config in the deploy repo.

Change-Id: I076160215421d57cd708edeb31c8fa16215997c5


johngian:
I pinned the jsonc-parser dependency only for dev-dependencies so CI can work. (This is the last known version that works with node 12)

johngian:
Just tested it locally. Other than some minor changes on the configuration for start/end time (initial example mentioned unix timestamps but it expected iso8601 strings) it worked OK.

PR #1343, opened 2024-03-07
Add an empty body to workaround parsoid failing requests

johngian:
Bug: T359509

### DIFF
--- a/sys/parsoid.js
+++ b/sys/parsoid.js
@@ -415,6 +415,9 @@ class ParsoidService {
             path
         );
 
+        // HACK: T359509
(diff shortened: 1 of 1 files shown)


sys/parsoid.js
```diff
@@ -415,6 +415,9 @@ class ParsoidService {
             path
         );
 
+        // HACK: T359509
+        parsoidReq.body = '{}';
+
         return hyper.get(parsoidReq)
```
brightbyte, line 421: This seems super strange - POST requests requrie a valid body, but GET request should require *no* body.

PR #1345, opened 2024-06-04
Parsoid: set s-maxage to 60 seconds

brightbyte:
This sets s-maxage for Parsoid to 60 seconds. If we can get away with this, we can disable active purging (and don't need to implement active purging for the equivalent core endpoints).

Bug: T365630
Change-Id: Ifad202d8f0df5bc383151aa534091e9e16795ffa


johngian:
Ideally this should have been a config entry managed from scap deployment repo but since we are almost ready to deprecate i think its fine to have it hardcoded.

johngian:
Some tests fail because of the change of the cache header.

brightbyte:
Tests for some old versions of node fail because restbase uses the ?? operator which isn't available until node 14 or so.

Can we stop running tests for node versions oder than 14, and start running tests for node 18 and 20?

brightbyte:
The remaining error is:

```
  1) Language variants
       should request mobile-sections with no variant and store it:

      AssertionError [ERR_ASSERTION]: Expected: sr; Result: undefined
      + expected - actual

      -false
      +true
```

This test passes for me. Also, it uses an unrelated endpoint (mobile-html).

Not sure how to procede... @johngian do you have an idea what might be going on here?

johngian:
@brightbyte CI should be passing now

PR #1346, opened 2024-07-02
Redirect docs for metrics endpoints to new docs site

apaskulin:
As part of the project to replace the RESTBase `/metrics` endpoints (AQS 1.0) with a new set of services (AQS 2.0), we've created a new documentation site for the `/metrics` endpoints that uses OpenAPI specs generated by the new AQS 2.0 services.

My goal is to redirect users viewing the old docs (https://wikimedia.org/api/rest_v1/#/) to the new docs (https://doc.wikimedia.org/analytics-api), without breaking RESTBase or making the other portions of its docs unusable. 

Ideally, I'd like to be able to keep as much of the old structure as possible (the sections headings and endpoint titles) and replace the content with a link to the new docs site. Or, we could just remove the `/metrics` endpoints from being listed on https://wikimedia.org/api/rest_v1/#/. But I wasn't able to find a way to do either of these options without breaking RESTBase's CI. Is there an easy way to do something like this without breaking things?

This PR adds a note to the intro section of https://wikimedia.org/api/rest_v1/#/ with a link to the new docs site.

Bug: https://phabricator.wikimedia.org/T368973

### DIFF
--- a/projects/v1/wikimedia.wmf.yaml
+++ b/projects/v1/wikimedia.wmf.yaml
@@ -35,6 +35,13 @@ info:
         change policy, according to
         [our API version policy](https://www.mediawiki.org/wiki/API_versioning).
       - Endpoint specific usage limits.
+      
+      ### Metrics endpoints
+
+      For documentation for `/metrics` endpoints, including pageviews, unique
+      devices, edited pages, editors, edits, registered users, bytes difference,
+      and mediarequests data, see the
+      [Wikimedia Analytics API documentation](https://doc.wikimedia.org/analytics-api).
 
   termsOfService: https://wikimediafoundation.org/wiki/Terms_of_Use
   contact:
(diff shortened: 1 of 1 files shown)


apaskulin:
@nosmo would you be able to help with this?

nosmo:
This change lgtm at the base. Unfortunately I don't know enough about the internals to comment on the solutions beyond this but it'd be good to remove the metrics paths at least so there's less confusion. Does OpenAPI accept absolute urls for things like the `x-modules` `path` parameter?

nosmo (approved):

apaskulin:
That's a good question. If I understand correctly, x-modules are a custom setup and not something standard to OpenAPI, so I'm not sure what the limitations are. If it works for you, let's merge this as-is, and I'll open a separate PR that tries to remove the metrics paths.

apaskulin:
Also I don't have merge access, so you'll need to merge for me. Thanks!

PR #1347, opened 2024-07-11
pcs: Use if-unmodified-since to avoid unnecessary rendering

johngian:
Replicate what Parsoid on RESTBase is doing to avoid unnecessary rendering from template changes.

* From what it looks from prod metrics we do get changeprop requests that the job date is for content that has already been superseded (old events)
* Currently parsoid is doing all the heavylifting that checks the `if-unmodified-since` and if the events are stale doesn't propagate more resource change events
* By disabling pregeneration on parsoid and replacing the pregeneration trigger for PCS/summary with MW resource changes we end up with the same problem (and increased load).
* This patch is replicating the logic from parsoid to PCS/Summary endpoints

### DIFF
--- a/lib/mwUtil.js
+++ b/lib/mwUtil.js
@@ -182,6 +182,33 @@ mwUtil.isNoCacheRequest = (req) => req.headers && /no-cache/i.test(req.headers['
  */
 mwUtil.isNoStoreRequest = (req) => req.headers && /no-store/i.test(req.headers['cache-control']);
 
+/**
+ * Checks whether the request is an 'if-unmodified-since' request
+ * @param   {Object}    req
+ * @return  {boolean}
+ */
+mwUtil.isUnmodifiedSinceRequest = (req) => Boolean(req.headers && req.headers['if-unmodified-since']);
+
+/**
+ * Checks whether the response has been modified since the timestamp
+ * in `if-unmodified-since` header of the request
+ * @param  {Object} req the request
+ * @param  {Object} res the response
+ * @return {boolean}    true if content has beed modified
+ */
+mwUtil.isUnmodifiedSince = (req, res) => {
+    try {
+        if (req.headers['if-unmodified-since']) {
+            const jobTime = Date.parse(req.headers['if-unmodified-since']);
+            const revInfo = mwUtil.parseETag(res.headers.etag);
+            return revInfo && uuidUtils.getDate(revInfo.tid) < jobTime;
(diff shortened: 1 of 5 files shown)


johngian:
I think now with the fix for summary endpoint this is ready for review @brightbyte

lib/mwUtil.js
```diff
@@ -182,6 +182,33 @@ mwUtil.isNoCacheRequest = (req) => req.headers && /no-cache/i.test(req.headers['
...
+mwUtil.isUnmodifiedSinceRequest = (req) => Boolean(req.headers && req.headers['if-unmodified-since']);
+
+/**
+ * Checks whether the response has been modified since the timestamp
+ * in `if-unmodified-since` header of the request
+ * @param  {Object} req the request
+ * @param  {Object} res the response
+ * @return {boolean}    true if content has beed modified
+ */
+mwUtil.isModifiedSince = (req, res) => {
```
brightbyte, line 199: It's a bit surprising that a method called isModifiedSince() will ignored the if-modified-since header, and instead look at the is-unmodified-since header.

I suggest to either turn the logic around and make this isUnmodifiedSince, or make it check both conditions and rename to something like meetsTimestampConditions() or some such.

sys/key_value.js
```diff
@@ -75,11 +78,22 @@ class KVBucket {
...
             if (dbResult.body && dbResult.body.items && dbResult.body.items.length) {
                 const row = dbResult.body.items[0];
-                return {
+                const result = {
                     status: 200,
                     headers: row.headers,
                     body: row.value
                 };
+
+                if (isNoCacheRequest && isUnmodifiedSinceRequest) {
```
brightbyte, line 87: IF we are not looking at the information in row to decide whether we should throw, why can't we do this in line 63 already?

johngian: Because we need to fetch data from cassandra to compare the result timestamp with the value from the header. If the request is just a no-cache request we avoid the fetching part and just return 404 so the logic on each endpoint fetches and stores the latest (as its already implemented).

lib/mwUtil.js
johngian: I renamed the function just to avoid implementing more functionality.

sys/key_value.js
johngian: Also typo. Fixed now

thesocialdev (approved):

PR #1348, opened 2024-07-15
Remove metrics endpoints

apaskulin:
Because the `/metrics` endpoints now provided through AQS 2.0 and the REST gateway, they no longer need to be included in the RESTBase docs at https://wikimedia.org/api/rest_v1/#/

This PR attempts to remove the `/metrics` endpoints from the docs by removing the paths from the specs, configuration, and tests. Note that this PR is risky since it includes major changes, and I wasn't able to test it locally to ensure that it works correctly. Feel free to close if it's too risky to merge.


apaskulin:
@nosmo, Looks like this passed CI, but I wasn't able to test it locally to confirm that it correctly removes the metrics endpoints.

nosmo (approved):
Looks good to me, but I'll loop someone from the team working on restbase stuff to give the final say

PR #1350, opened 2024-08-29
Set up user-agent for PCS stored requests

johngian:
Endpoints affected:
* summary
* mobile-html
* media-list

Top level config is not maintained when custom headers are
defined on hyper requests. Also top level config is not available
on template expansion, thus we need to redefine the user-agent in
multiple places.

Bug: https://phabricator.wikimedia.org/T319365


johngian:
Skipping a failed test because it looks unrelated to the patch.

PR #1352, opened 2025-04-18
enable testing against nodejs 18

eevans:
Bug: [T357791](https://phabricator.wikimedia.org/T357791)


jdforrester:
We should (at least) drop Node 10 testing and support, given we're running in 12 in prod. But not for this patch. :-)

PR #1354, opened 2025-08-13
Update config.yaml

PrayashM1:
Updated config for build


PrayashM1:
comitting changes for config

mvolz:
I assume you probably meant to do this on a private fork? 

Also you've exposed your password publicly so you should cycle it for security reasons.

